PR #24, opened 2019-03-21
cmd/loop: add new keyword argument for amt

alexbosworth:
In this commit there is a new keyword argument for amt to add an alternative to the positional argument for amt

### DIFF
--- a/cmd/loop/loopout.go
+++ b/cmd/loop/loopout.go
@@ -31,24 +31,33 @@ var loopOutCommand = cli.Command{
 				"should be sent to, if let blank the funds " +
 				"will go to lnd's wallet",
 		},
+		cli.Uint64Flag{
+			Name:   "amt",
+			Usage:  "the amount in satoshis to loop out",
+		},
 	},
 	Action: loopOut,
 }
 
 func loopOut(ctx *cli.Context) error {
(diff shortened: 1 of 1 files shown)


cmd/loop/loopout.go
```diff
@@ -31,24 +31,33 @@ var loopOutCommand = cli.Command{
...
-	}
-
 	args := ctx.Args()
 
-	amt, err := parseAmt(args[0])
+	var amtStr string
+	if ctx.IsSet("amt") {
+	   amtStr = ctx.String("amt")
+	} else if ctx.NArg() > 0 {
+	   amtStr = args[0]
```
joostjager: 8 space indent

cmd/loop/loopout.go
```diff
@@ -31,24 +31,33 @@ var loopOutCommand = cli.Command{
...
-		return nil
-	}
-
 	args := ctx.Args()
 
-	amt, err := parseAmt(args[0])
+	var amtStr string
+	if ctx.IsSet("amt") {
+	   amtStr = ctx.String("amt")
+	} else if ctx.NArg() > 0 {
```
joostjager: Usual standard is to do this with a switch:
```
switch {
case ctx.IsSet("amt"):
       ....
<etc>
}
```

PR #25, opened 2019-03-22
cmd/loopd: point to testnet loop server for testnet mode

### LINKED ISSUE #20: loop out failes on testnet
When trying to use loop on testnet (loopd started with `network=testnet` option), the loop binary fails with this error:

```
pat@ubuntu:~/.lnd/data/chain/bitcoin/testnet$ loop out 1000000
Max swap fees for 1000000 loop out: 54972
CONTINUE SWAP? (y/n), expand fee detail (x): x

Max on-chain fee:                 411
Max off-chain swap routing fee:   50010
Max off-chain prepay routing fee: 422
Max swap fee:                     4129
Max no show penalty:              8258
CONTINUE SWAP? (y/n): y
[loop] rpc error: code = Unknown desc = invoice not for current active network 'testnet3'
```

Environment:
- loop version 0.0.1
- Linux ubuntu-s-2vcpu-2gb-fra1-01 4.15.0-38-generic #41-Ubuntu SMP Wed Oct 10 10:59:38 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Roasbeef:
Fixes #20.

### DIFF
--- a/cmd/loopd/daemon.go
+++ b/cmd/loopd/daemon.go
@@ -26,6 +26,12 @@ func daemon(config *config) error {
 	}
 	defer lnd.Close()
 
+	// If the user is targeting the testnet network, then we'll point
+	// towards the testnet swap server rather than the mainnet endpoint.
+	if config.Network == "testnet" {
+		config.SwapServer = testnetServer
+	}
+
 	swapClient, cleanup, err := getClient(
 		config.Network, config.SwapServer, config.Insecure, &lnd.LndServices,
 	)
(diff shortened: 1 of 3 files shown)


cmd/loopd/daemon.go
```diff
@@ -26,6 +26,12 @@ func daemon(config *config) error {
 	}
 	defer lnd.Close()
 
+	// If the user is targetting the testnet network, then we'll point
```
alexbosworth:
```suggestion
	// If the user is targeting the testnet network, then we'll point
```

PR #30, opened 2019-03-23
multi: add persistent logging 

### LINKED ISSUE #5: loopd logs aren't persistent

Roasbeef:
In PR, we add persistent logic to the loopd daemon. Users can now designate a default logging directory, as well as log rotation options. Additionally, we've added a new `--debuglevel` command which allows users to crank up the logging for particular sub-systems on start up.

Fixes #5.

### DIFF
--- a/cmd/loopd/config.go
+++ b/cmd/loopd/config.go
@@ -1,5 +1,25 @@
 package main
 
+import (
+	"path/filepath"
+
+	"github.com/btcsuite/btcutil"
+)
+
+var (
+	loopDirBase = btcutil.AppDataDir("loop", false)
+
+	defaultConfigFilename = "loopd.conf"
+
+	defaultLogLevel    = "info"
+	defaultLogDirname  = "logs"
+	defaultLogFilename = "loopd.log"
+	defaultLogDir      = filepath.Join(loopDirBase, defaultLogDirname)
+
+	defaultMaxLogFiles    = 3
+	defaultMaxLogFileSize = 10
+)
+
 type lndConfig struct {
 	Host         string `long:"host" description:"lnd instance rpc address"`
 	MacaroonPath string `long:"macaroonpath" description:"Path to lnd macaroon"`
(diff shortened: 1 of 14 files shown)


cmd/loopd/config.go
```diff
@@ -1,5 +1,25 @@
...
+
+var (
+	loopDirBase = btcutil.AppDataDir("loop", false)
+
+	defaultConfigFilename = "loopd.conf"
+
+	defaultLogLevel    = "info"
+	defaultLogDirname  = "logs"
+	defaultLogFilename = "loopd.log"
+	defaultLogDir      = filepath.Join(loopDirBase, defaultLogDirname)
```
joostjager: This creates the network directory level in two locations. There is `~/.loop/logs/regtest` for the logs and also `~/.loop/regtest` for the db. I think it is better to keep all files for a network underneath the same dir.

cmd/loopd/daemon.go
```diff
@@ -62,7 +62,7 @@ func daemon(config *config) error {
 	looprpc.RegisterSwapClientServer(grpcServer, &server)
 
 	// Next, start the gRPC server listening for HTTP/2 connections.
-	logger.Infof("Starting gRPC listener")
+	loopdLog.Infof("Starting gRPC listener")
```
joostjager: In other packages we use as a variable name just `log`. I see that we also need to instantiate loggers for other packages in `loopd`, but that is just instantiate, pass in and done. I would also just use `log` here.

cmd/loopd/log.go
```diff
@@ -1,24 +1,207 @@
...
+		if !validLogLevel(logLevel) {
+			str := "The specified debug level [%v] is invalid"
+			return fmt.Errorf(str, logLevel)
+		}
+
+		setLogLevel(subsysID, logLevel)
+	}
+
+	return nil
+}
```
joostjager: Instead of copy/pasting all this, can't we move this code into the lnd build package and reuse it here? We are already depending on the build package, so it won't be a new dependency.

joostjager:
Build fails because tests don't build

guggero:
Replaced by #99

PR #31, opened 2019-03-25
multi: always supply chain params when decoding addresses

joostjager:
The commit https://github.com/btcsuite/btcutil/commit/3ac1210f4b38284536002486ef8a21ac54c14497 in btcutil changed the address decoding behaviour. This surfaced that we were not passing in the chain params everywhere, causing panics.


go.sum
```diff
@@ -115,10 +119,14 @@ github.com/lightninglabs/neutrino v0.0.0-20190313035638-e1ad4c33fb18/go.mod h1:v
 github.com/lightninglabs/neutrino v0.0.0-20190313220016-a6cc2a8361d5/go.mod h1:NXpKUhcszEdn7v7ze5vsnSHI1KX6HqFdNRu0eqhkJzk=
 github.com/lightninglabs/neutrino v0.0.0-20190314214430-643615b8c132 h1:yaKXc3D/qeEy2wv+bIQwmrq+kThdbWg0p2ggea3RAsg=
 github.com/lightninglabs/neutrino v0.0.0-20190314214430-643615b8c132/go.mod h1:NXpKUhcszEdn7v7ze5vsnSHI1KX6HqFdNRu0eqhkJzk=
+github.com/lightninglabs/neutrino v0.0.0-20190321023416-6dac90b98052 h1:YVi5x7xCumfIVtEejy3zBAej+NqY9EONy/fRim2+azQ=
```
Roasbeef: I think if you run `go mod tidy`, then these duplicates will go away.

go.mod
```diff
@@ -3,14 +3,13 @@ module github.com/lightninglabs/loop
...
-	github.com/btcsuite/btcutil v0.0.0-20190207003914-4c204d697803
+	github.com/btcsuite/btcutil v0.0.0-20190316010144-3ac1210f4b38
 	github.com/coreos/bbolt v1.3.2
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/golang/protobuf v1.3.1
 	github.com/grpc-ecosystem/grpc-gateway v1.8.5
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/lightninglabs/neutrino v0.0.0-20190314214430-643615b8c132 // indirect
-	github.com/lightningnetwork/lnd v0.0.0-20190314214430-b4a1024ac74fc576c65c8074288a5ffaf6bd1ec4
+	github.com/lightningnetwork/lnd v0.5.1-beta.0.20190322220528-6ad8be25e1aa
```
Roasbeef: Should't this be 0.5.2? Not sure how `go mod` handles internal release/tag branching.

joostjager: This is what it adds if I specify the commit `6ad8be25e1aa`

go.sum
joostjager: Ran `go mod tidy` and this is the delta that remains.

PR #34, opened 2019-04-01
Loop In

joostjager:
This PR adds loop in functionality to the Lightning Loop client.

Loop in allows a lightning channel to be 'refilled' with on-chain coins. It allows a channel to remain open and offers an alternative over closing and reopening the channel.

# High level flow

1. User generates a secret preimage.
2. User publishes an on-chain htlc tied to the hash of the secret preimage.
3. Server picks up confirmation of the on-chain htlc.
4. Server sends an off-chain lightning payment to the user.
5. User settles the payment by revealing the secret preimage.
6. Server learns the preimage and uses this to sweep the on-chain htlc.

# Usage

Similar to Loop Out. A new command has been added to the `loop` command line tool:

`loop in <amt>`

Also the `--external` flag can be supplied. In that case step 2 of the flow above is skipped and the user is given the opportunity to publish the htlc manually. An example use of this is to withdraw from an exchange into a lightning channel by looping in.

There is also the `--channel` flag that allows selection of the channel to loop in to. This functionality is not implemented yet.

# Testnet only

Loop in is currently only supported on testnet. There are still several changes required to `lnd` for it to be used on mainnet.


loopin.go
```diff
@@ -0,0 +1,585 @@
...
+		// sweeping the htlc after paying the invoice.
+		err := s.lnd.Invoices.CancelInvoice(globalCtx, s.hash)
+		if err != nil && err != channeldb.ErrInvoiceAlreadySettled {
+			return err
+		}
+
+		// TODO: Add miner fee of timeout tx to swap cost balance.
+	}
+
+	// Wait for a final state of the swap invoice. It should either be
```
Roasbeef: I think we should move this to be _above_ waiting for the HTLC spend, or be concurrent even. I tried another Loop in on testnet now and realized that things are shown as pending until the HTLC is swept on chain. However, from the PoV of the client, the Loop In process is complete once its invoice is settled, as it now has more outbound capacity.

Roasbeef:
The README should also be updated to instruct users to build with the `invoicerpc` tag as well.

loopin.go
joostjager: Everywhere in loop swaps are only marked complete when all information needed for accounting is there. For loop in on the client, this means the htlc sweep tx needs to be confirmed and the swap invoice either canceled or settled. Also after a restart, a swap that is completed from the user PoV, still needs to be resumed to complete the accounting info.

An intermediate state that the swap transitions to when the invoice is paid could be a solution here.

joostjager: Added a commit that adds this intermediate state.

joostjager:
@Roasbeef ptal

Roasbeef:
Panic on `loop monitor` w/ the latest attempting a Loop In:
```
oroutine 68 [running]:
main.(*swapClientServer).marshallSwap(0xc0001c2290, 0xc00024f530, 0xc0006fe420, 0xb51840, 0xc000343dc0)
        /root/loop/cmd/loopd/swapclient_server.go:122 +0x134
main.(*swapClientServer).Monitor.func1(0x6d5a1c0, 0xed42e0803, 0x14cc500, 0xb680a30070e9430c, 0xaa124c441abd7ade, 0xc887a080307fdaae, 0xd00a71501b250270, 0x102, 0x511c5b616b2f239f, 0x7ef9a4a62c974031, ...)
        /root/loop/cmd/loopd/swapclient_server.go:134 +0x4a
main.(*swapClientServer).Monitor(0xc0001c2290, 0xc000343d20, 0xef9960, 0xc000174a30, 0x0, 0x0)
        /root/loop/cmd/loopd/swapclient_server.go:197 +0x877
github.com/lightninglabs/loop/looprpc._SwapClient_Monitor_Handler(0xc3b860, 0xc0001c2290, 0xef84c0, 0xc0001049c0, 0x14eaa10, 0xc000102c00)
        /root/loop/looprpc/client.pb.go:975 +0x109
google.golang.org/grpc.(*Server).processStreamingRPC(0xc000200180, 0xefc2a0, 0xc00024b800, 0xc000102c00, 0xc0001bf0b0, 0x14b49e0, 0x0, 0x0, 0x0)
        /root/gocode/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:1175 +0xb12
google.golang.org/grpc.(*Server).handleStream(0xc000200180, 0xefc2a0, 0xc00024b800, 0xc000102c00, 0x0)
        /root/gocode/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:1254 +0xcbe
google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc000332030, 0xc000200180, 0xefc2a0, 0xc00024b800, 0xc000102c00)
        /root/gocode/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:690 +0x9f
created by google.golang.org/grpc.(*Server).serveStreams.func1
        /root/gocode/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:688 +0xa1
```

Roasbeef:
Nvm, it was due to me having issued a swap using the old DB format.

loopin.go
```diff
@@ -582,16 +601,18 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
...
+	err := s.store.UpdateLoopIn(s.hash, s.lastUpdateTime, s.state)
 	if err != nil {
 		return err
 	}
 
 	// Send out swap update
 	return s.sendUpdate(ctx)
 }
+
+// setState updates the swap state and last update timestamp.
```
Roasbeef: I think we should also roll in `persistState` into this method as well. Otherwise, I see a few instances above, where we modify the swap state, but don't persist it to the database.

Roasbeef: Scratch that, I see it's persisted for the first thing that causes it to break out of the loop.

cmd/loopd/view.go
```diff
@@ -28,37 +30,40 @@ func view(config *config) error {
...
+	if err := viewIn(swapClient, chainParams); err != nil {
 		return err
 	}
 
-	if len(swaps) == 0 {
-		fmt.Printf("No swaps\n")
+	return nil
+}
+
+func viewOut(swapClient *loop.Client, chainParams *chaincfg.Params) error {
```
Roasbeef: I think we should instead make this an command for the `loop` binary, but it isn't blocking and can be done in a follow up.

PR #45, opened 2019-04-17
lndclient+cmd/loopd: use unique macaroon per sub-server to avoid users having to delete admin.macaroon

Roasbeef:
In this PR, we add a new struct, the `macaroonPouch`. This struct bundles all the macaroons we need for each sub-server. Along the way, we also export the set of default* paths for lnd, and add a new set of variables that store the default file names of each of the macaroons for each sub-server.

The `WithMacaroonAuth` method on the `serializedMacaroon` type will allow us to attach a macaroon for each call rather than attaching it at the connection level. This slight change will allow us to use multiple macaroons over a single RPC connection, rather than a single global macaroon.

Before this PR, we used the `admin.macaroon` for everything which meant that users on older versions of `lnd` may not have all the capabilities on their existing macaroon that was added for some of the sub-servers. After this commit, users will no longer need to do this.

### DIFF
--- a/cmd/loopd/daemon.go
+++ b/cmd/loopd/daemon.go
@@ -34,7 +34,8 @@ func daemon(config *config) error {
 
 	// Create an instance of the loop client library.
 	swapClient, cleanup, err := getClient(
-		config.Network, config.SwapServer, config.Insecure, &lnd.LndServices,
+		config.Network, config.SwapServer, config.Insecure,
+		&lnd.LndServices,
 	)
 	if err != nil {
 		return err
--- a/lndclient/lnd_services.go
+++ b/lndclient/lnd_services.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"path/filepath"
 	"time"
 
@@ -12,11 +11,8 @@ import (
 	"github.com/btcsuite/btcutil"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightningnetwork/lnd/lncfg"
-	"github.com/lightningnetwork/lnd/macaroons"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
-
-	macaroon "gopkg.in/macaroon.v2"
 )
 
 var rpcTimeout = 30 * time.Second
@@ -30,6 +26,8 @@ type LndServices struct {
 	Invoices      InvoicesClient
 
 	ChainParams *chaincfg.Params
+
+	macaroons *macaroonPouch
 }
 
 // GrpcLndServices constitutes a set of required RPC services.
@@ -40,13 +38,25 @@ type GrpcLndServices struct {
 }
 
 // NewLndServices creates a set of required RPC services.
-func NewLndServices(lndAddress string, application string,
-	network string, macPath, tlsPath string) (
-	*GrpcLndServices, error) {
+func NewLndServices(lndAddress, application, network, macaroonDir,
+	tlsPath string) (*GrpcLndServices, error) {
+
+	// If the macaroon directory isn't set, then we can't proceed as we
+	// need then to obtain the macaroons for all sub-servers.
+	if macaroonDir == "" {
+		return nil, fmt.Errorf("macarooon dir must be set")
+	}
+
+	// Now that we've ensured our macaroon directory is set properly, we
+	// can retrieve our full macaroon pouch from the directory.
+	macaroons, err := newMacaroonPouch(macaroonDir)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain macaroons: %v", err)
+	}
 
 	// Setup connection with lnd
 	logger.Infof("Creating lnd connection to %v", lndAddress)
-	conn, err := getClientConn(lndAddress, network, macPath, tlsPath)
+	conn, err := getClientConn(lndAddress, network, tlsPath)
 	if err != nil {
 		return nil, err
 	}
@@ -101,6 +111,7 @@ func NewLndServices(lndAddress string, application string,
 			Signer:        signerClient,
 			Invoices:      invoicesClient,
 			ChainParams:   chainParams,
+			macaroons:     macaroons,
 		},
 		cleanup: cleanup,
 	}
@@ -135,7 +146,7 @@ var (
 	defaultSignerFilename            = "signer.macaroon"
 )
 
-func getClientConn(address string, network string, macPath, tlsPath string) (
+func getClientConn(address string, network string, tlsPath string) (
 	*grpc.ClientConn, error) {
 
 	// Load the specified TLS certificate and build transport credentials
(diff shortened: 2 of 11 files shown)


lndclient/lnd_services.go
```diff
@@ -40,13 +38,25 @@ type GrpcLndServices struct {
...
-func NewLndServices(lndAddress string, application string,
-	network string, macPath, tlsPath string) (
-	*GrpcLndServices, error) {
+func NewLndServices(lndAddress, application, network, macaroonDir,
+	tlsPath string) (*GrpcLndServices, error) {
+
+	// If the macaroon directory isn't set, then we can't proceed as we
+	// need then to obtain the macaroons for all sub-servers.
+	if macaroonDir == "" {
+		return nil, fmt.Errorf("macarooon dir must be set")
```
joostjager: s/macarooon/macaroon

lndclient/lnd_services.go
```diff
@@ -40,13 +38,25 @@ type GrpcLndServices struct {
...
 // NewLndServices creates a set of required RPC services.
-func NewLndServices(lndAddress string, application string,
-	network string, macPath, tlsPath string) (
-	*GrpcLndServices, error) {
+func NewLndServices(lndAddress, application, network, macaroonDir,
+	tlsPath string) (*GrpcLndServices, error) {
+
+	// If the macaroon directory isn't set, then we can't proceed as we
+	// need then to obtain the macaroons for all sub-servers.
+	if macaroonDir == "" {
```
joostjager: Do we still want to allow running without macaroons? This was possible before.

cmd/loopd/daemon.go
```diff
@@ -7,34 +7,71 @@ import (
...
+	// Based on the network, we may need to swap our items like the
+	// swapserver or the default location of macaroons if a path isn't set.
+	switch config.Network {
+	case "testnet":
+		// If the user is targeting the testnet network, then we'll
+		// point towards the testnet swap server rather than the
+		// mainnet endpoint.
+		config.SwapServer = testnetServer
+
+		if config.Lnd.MacaroonPath == "" {
```
joostjager: Rename config option to `MacaroonDir`? It isn't the option that it used to be. If we want to keep using `MacaroonPath`, the error further down referencing "macaroon dir" should be updated.

cmd/loopd/daemon.go
```diff
@@ -7,34 +7,71 @@ import (
...
+			config.Lnd.MacaroonPath = filepath.Join(
+				lndclient.DefaultLndDir,
+				lndclient.DefaultDataDir,
+				lndclient.DefaultChainSubDir, "bitcoin",
+				"simnet",
+			)
+		}
+	}
+
```
joostjager: This default dir logic isn't applied when the `view` command is executed.

joostjager: And also not when starting the server. Probably better to move this to a location where it is reusable.

lndclient/macaroon_pouch.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return serializedMacaroon(hex.EncodeToString(macBytes)), nil
+}
+
+// WithMacaroonAuth modifies the passed context to include the macaroon KV
+// metadata of the target macaroon. This method can be used to add the macaroon
+// at call time, rather than when the connection to the gRPC server is created.
+func (s serializedMacaroon) WithMacaroonAuth(ctx context.Context) context.Context {
+	return metadata.AppendToOutgoingContext(ctx, "macaroon", string(s))
+}
```
joostjager: What is the impact of this in terms of network traffic and performance (is the macaroon validated in `lnd` for each call?) compared to opening one tcp connection per sub server and using connection level macaroons?

Roasbeef: It's the exact same. There's a special header that's used to tack on arbitrary "metadata" which are KV pairs: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests (search for "metadata").

lndclient/lnd_services.go
Roasbeef: IMO no, as we don't really recommend running `lnd` itself without macaroons. We may want to add a way to specify the raw hex macaroons over the command line though. Adding this line helped me find some bugs in the prior implementation, and for example will make things fail if users run on `regtest`.

cmd/loopd/daemon.go
Roasbeef: Renamed!

lndclient/lnd_services.go
Roasbeef: Actually ended up reverting this in order to consolidate the default macaroon directory code for re-use.

Roasbeef:
PTAL @joostjager!

cmd/loopd/daemon.go
Roasbeef: Moved it to reside inside of `lndclient.NewLndServices`. As a result of this change, the move to make some default dir params public is no longer needed, so I removed it.

willcl-ark:
I would like to know more about this particular wording:
> will allow us to attach a macaroon *for each call* rather than attaching it at the connection level.

I hope this doesn't (and sure it probably doesn't!) mean that macaroons are regenerated for each RPC
call, as in my experience gRPC connection stubs would need to be regenerated for each call and this is expensive/slow

lndclient/macaroon_pouch.go
joostjager: That's nice. Then the chosen solution is definitely the best by having just a single tcp connection.

lndclient/lnd_services.go
```diff
@@ -82,10 +87,12 @@ func NewLndServices(lndAddress, application, network, macaroonDir,
 		)
 	}
 
-	notifierClient := newChainNotifierClient(conn)
-	signerClient := newSignerClient(conn)
-	walletKitClient := newWalletKitClient(conn)
-	invoicesClient := newInvoicesClient(conn)
+	// With the network check passed, we'll now initialize the rest of the
+	// sub-sever connections, giving each of them their specific macaroon.
```
joostjager: s/sever/server

lndclient/lnd_services.go
```diff
@@ -40,13 +38,47 @@ type GrpcLndServices struct {
...
+			macaroonDir = filepath.Join(
+				defaultLndDir, defaultDataDir,
+				defaultChainSubDir, "bitcoin", "mainnet",
+			)
+
+		case "simnet":
+			macaroonDir = filepath.Join(
+				defaultLndDir, defaultDataDir,
+				defaultChainSubDir, "bitcoin", "simnet",
+			)
```
joostjager: regtest missing

lndclient/macaroon_pouch.go
```diff
@@ -0,0 +1,98 @@
...
+
+// macaroonPouch holds the set of macaroons we need to interact with lnd for
+// Loop. Each sub-server has its own macaroon, and for the remaining temporary
+// calls that directly hit lnd, we'll use the admin macaroon.
+type macaroonPouch struct {
+	// invoiceMac is the macaroon for the invoice sub-server.
```
joostjager: s/invoice/Invoices

Roasbeef:
> I hope this doesn't (and sure it probably doesn't!) mean that macaroons are regenerated for each RPC call

No they're all held in memory and attached to each call. On the HTTP/2 level, everything is the same as it's just a special header KV passed along.

Roasbeef:
Latest comments addressed!

PR #46, opened 2019-04-18
Introduced docker integration

### LINKED ISSUE #43: 'x509: certificate signed by unknown authority' when running in Docker container
I'm trying to create a Dockerfile for loop/loopd. The Dockerfile itself is quite similar to the LND Dockerfile, but I'm having issues actually running loop.

I have the container running on my testnet server, connecting to my testnet LND node. This is what it outputs:

```
2019-04-15 16:17:31.277 [INF] LOOPD: Version: 0.1.0-alpha commit=
2019-04-15 16:17:31.278 [INF] LNDCLIENT: Creating lnd connection to lnd:10009
2019-04-15 16:17:31.281 [INF] LNDCLIENT: Connected to lnd
2019-04-15 16:17:31.520 [INF] LNDCLIENT: Using network testnet
2019-04-15 16:17:31.523 [INF] LOOPD: Starting gRPC listener
2019-04-15 16:17:31.526 [INF] LOOPD: Starting REST proxy listener
2019-04-15 16:17:31.534 [INF] LOOPD: Starting swap client
2019-04-15 16:17:31.534 [INF] LOOPD: Waiting for updates
2019-04-15 16:17:31.535 [INF] LOOPD: RPC server listening on [::]:11010
2019-04-15 16:17:31.535 [INF] LOOPD: REST proxy listening on 127.0.0.1:8081
2019-04-15 16:17:31.602 [INF] CLIENT: Connected to lnd node OpinionatedGeek Test ⚡ with pubkey 02639adf914ec9dfd8649c47de1d232de4a23b0719b1d8bd2bef9beb32f948018e
2019-04-15 16:17:31.602 [INF] CLIENT: Wait for first block ntfn
2019-04-15 16:17:31.606 [INF] CLIENT: Starting event loop at height 1489187
```
It looks to me like `loopd` is connecting to `lnd` properly.

But when I run `loop` in the container running `loopd`, I get the following error:
```
[loop] rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority"
```
I'm not even clear which component is the source of the error.
* It looks to be `loop` connecting to `loopd`, but I didn't think that involved TLS at all. And even if it did use TLS, `loop` and `loopd` are both running in the same container.
* `loopd` and `lnd` are on different containers so I can see how TLS problems could occur, but that connection seems to be working (as shown in the logs)

Is this a problem anyone else has come across?

Here's the current version of the Dockerfile:
[Dockerfile.txt](https://github.com/lightninglabs/loop/files/3081234/Dockerfile.txt)

OpinionatedGeek:
Introduced docker integration:
* Added Dockerfile to build loopd and loop with no dependencies.
* Added instructions to build and run docker image.

There's no MAINTAINER tag in the Dockerfile (yet). I wasn't sure who to put. Much of the Dockerfile contents come from @Roasbeef's LND Dockerfile though.

I don't know if the tone of the instructions is OK. Any thoughts?

Fixes #43.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -32,13 +32,14 @@ Docker is very flexible so you can use that information however you choose. This
 
 One way of running `loopd` is 
 ```
-docker run --rm -it --name loopd -v $HOME:/root loop:latest loopd --network=testnet --lnd.host <my-lnd-ip-address>:10009
+docker run --rm -it --name loopd -v $HOME/.lnd:/root/.lnd -v $HOME/.loop:/root/.loop loop:latest loopd --network=testnet --lnd.host <my-lnd-ip-address>:10009
 ```
 
 Things to note from this docker command:
 * You can stop the server with Control-C, and it'll clean up the associated stopped container automatically.
 * The name of the running container is 'loopd' (which you may need to know to run the `loop` command).
-* Your home directory is mapped into the container as the container user's home directory. `loopd` will look for your tls.cert and macaroon in the default locations in ~/.lnd, and will store its own data in ~/.loop. If mapping your home directory isn't appropriate for your case you can map whatever directories you choose and override where `loopd` looks for files using additional command-line parameters.
+* The '.lnd' directory in your home directory is mapped into the container, and `loopd` will look for your tls.cert and macaroon in the default locations. If this isn't appropriate for your case you can map whatever directories you choose and override where `loopd` looks for them using additional command-line parameters.
+* The '.loop' directory in your home directory is mapped into the container, and `loopd` will use that directory to store some state.
 * You probably need to specify your LND server host and port explicitly, since by default `loopd` looks for it on localhost and there is no LND server on localhost within the container.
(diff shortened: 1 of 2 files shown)


alexbosworth:
looks good to me, although I'm not yet a Docker user

DOCKER.md
```diff
@@ -0,0 +1,82 @@
...
+* The binary `loop`, at `/go/bin/loop`
+
+Docker is very flexible so you can use that information however you choose. This guide isn't meant to be prescriptive.
+
+
+### Example: Running loopd
+
+One way of running `loopd` is 
+```
+docker run --rm -it --name loopd -v $HOME/.lnd:/root/.lnd loop:latest loopd --network=testnet --lnd.host <my-lnd-ip-address>:10009
```
tdickman: Probably worth mapping the `~/.loopd` directory as well since it stores the state of active loops.

OpinionatedGeek: Excellent point! Thanks for that. I've updated that line to map ~/.loop, as well as added a note about why.

PR #59, opened 2019-05-24
lndclient/basic_client: pass full macaroon path, not just directory.

valentinewallace:
This is useful for lndmon's usage of loop's lndclient package so we can restrict lndmon to `readonly.macaroon` rather than `admin.macaroon`.

### DIFF
--- a/lndclient/basic_client.go
+++ b/lndclient/basic_client.go
@@ -13,11 +13,48 @@ import (
 	macaroon "gopkg.in/macaroon.v2"
 )
 
+// BasicClientOption is a functional option argument that allows adding arbitrary
+// lnd basic client configuration overrides, without forcing existing users of
+// NewBasicClient to update their invocation. These are always processed in
+// order, with later options overriding earlier ones.
+type BasicClientOption func(*basicClientOptions)
+
+// basicClientOptions is a set of options that can configure the lnd client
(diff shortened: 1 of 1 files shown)


lndclient/basic_client.go
```diff
@@ -17,7 +17,7 @@ import (
 // "basic" as it uses a global macaroon (by default the admin macaroon) for the
 // entire connection, and falls back to expected defaults if the arguments
 // aren't provided.
-func NewBasicClient(lndHost, tlsPath, macDir, network string) (lnrpc.LightningClient, error) {
+func NewBasicClient(lndHost, tlsPath, macPath, network string) (lnrpc.LightningClient, error) {
```
Roasbeef: I think we can avoid changing the set of arguments too much by using a functional argument here. This path will also mean that callers won't need to update their set of arguments they want the read only macaroon (admin being the default for now, though we should revisit, least authority principle n all).

valentinewallace: switched it to take in a `readonly` boolean param.

valentinewallace: ok updated to use functional optional args.

Roasbeef:
Ping!

valentinewallace:
> Ping!

Updated!

PR #60, opened 2019-06-06
lndclient: add routerrpc

joostjager:


lndclient/router_client.go
```diff
@@ -0,0 +1,161 @@
...
+	TrackPayment(ctx context.Context, hash lntypes.Hash) (
+		chan PaymentStatus, chan error, error)
+}
+
+// PaymentStatus describe the state of a payment.
+type PaymentStatus struct {
+	State    routerrpc.PaymentState
+	Preimage lntypes.Preimage
+	Fee      lnwire.MilliSatoshi
+	Route    []uint64
```
Roasbeef: Perhaps we should just use `router.Route` here in order to give the caller all the information available from the response of the call. Although, perhaps we only need the last few channel IDs in our particular use case.

lndclient/router_client.go
```diff
@@ -0,0 +1,161 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/status"
+
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lntypes"
+)
+
+// RouterClient exposes payment functionality.
+type RouterClient interface {
+	SendPayment(ctx context.Context, invoice string,
```
Roasbeef: Missing `godoc` comments here and below.

lndclient/router_client.go
```diff
@@ -0,0 +1,161 @@
...
+	}()
+
+	return statusChan, errorChan, nil
+}
+
+func unmarshallPaymentStatus(rpcStatus *routerrpc.PaymentStatus) (
+	*PaymentStatus, error) {
+
+	status := PaymentStatus{
+		State: rpcStatus.State,
```
Roasbeef: What about the returned errors?

lndclient/router_client.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return &routerClient{
+		client:       routerrpc.NewRouterClient(conn),
+		routerKitMac: routerKitMac,
+	}
+}
+
+func (r *routerClient) SendPayment(ctx context.Context, invoice string,
+	maxFee btcutil.Amount, maxCltv *int32, outgoingChannel *uint64,
+	timeout time.Duration) (chan PaymentStatus, chan error, error) {
```
Roasbeef: If we have the structs like `TrackPaymentMessage` below, why not use them here instead of unrolling all the fields into individual arguments?

lndclient/router_client.go
joostjager: Failed payment are just another payment state. Other errors like "payment not found" are already handled above as a stream err.

lndclient/router_client.go
joostjager: Can't use `TrackPaymentMessage` because it is in the test package, but created a new struct for this.

lndclient/router_client.go
joostjager: Added

lndclient/router_client.go
joostjager: Converted to `route.Route`

joostjager:
@Roasbeef  ptal

PR #65, opened 2019-06-26
multi: expose confirmation target for loop out HTLC sweeps

wpaulino:
In this PR, we allow clients to specify the confirmation target for their HTLC sweeps within Loop Out swaps. We provide this as an alternative to allowing clients to specify a fixed fee rate since Loop already relies on confirmation targets everywhere.

We restrict this confirmation target by introducing a delta from the on-chain HTLC's expiration, which serves as the maximum confirmation target allowed. Since it's possible that the sweep doesn't confirm within the intended target in the event of an increasing demand for block space, once the delta from the HTLC's expiration is reached, we'll begin to use the default sweep confirmation target which should provide a timely confirmation. This is needed because if the preimage has already been revealed, then we need to make sure we sweep the HTLC before the server does.

Depends on #63.


cmd/loop/loopout.go
```diff
@@ -75,8 +82,10 @@ func loopOut(ctx *cli.Context) error {
 	}
 	defer cleanup()
 
+	sweepConfTarget := int32(ctx.Uint64("conf-target"))
 	quoteReq := &looprpc.QuoteRequest{
-		Amt: int64(amt),
+		Amt:        int64(amt),
+		ConfTarget: sweepConfTarget,
```
Roasbeef: We should likely modify this command to be: `loop out quote` or `loop quote out` for symmetric consistency. Also the case that right now, there's no way to get a quote for a loop in quote on the command line.

cmd/loop/loopout.go
```diff
@@ -36,6 +36,13 @@ var loopOutCommand = cli.Command{
 			Name:  "amt",
 			Usage: "the amount in satoshis to loop out",
 		},
+		cli.Uint64Flag{
+			Name: "conf-target",
```
Roasbeef: Why no underscore?

cmd/loop/quote.go
```diff
@@ -12,12 +12,25 @@ var quoteCommand = cli.Command{
 	Usage:       "get a quote for the cost of a swap",
 	ArgsUsage:   "amt",
 	Description: "Allows to determine the cost of a swap up front",
-	Action:      quote,
+	Flags: []cli.Flag{
+		cli.Uint64Flag{
+			Name: "conf-target",
```
Roasbeef: Same comment here re underscore.

looprpc/client.proto
```diff
@@ -127,6 +127,12 @@ message LoopOutRequest {
     lowest routing fee for the swap payment to the server.
     */
     uint64 loop_out_channel = 8;
+
+    /**
+    The number of blocks from the on-chain HTLC's confirmation height that it
+    should be swept within.
+    */
+    int32 sweep_conf_target = 9;
```
joostjager: Having both `max_miner_fee` and `sweep_conf_target` somehow suggests that we have not yet settled on the best set of parameters to control the chain fees. But I think it is something we can find out later.

There is also the following case to be aware of: if we get really close to the expiry, we may actually want to spend an amount on chain fees up to the full swap amount, because we are also risking losing the full amount if we don't make it in time.

cmd/loop/quote.go
```diff
@@ -12,12 +12,25 @@ var quoteCommand = cli.Command{
...
+			Usage: "the confirmation target that should be used " +
+				"either for the sweep of the on-chain HTLC " +
+				"broadcast by the swap server in the case of " +
+				"a Loop Out, or for the confirmation of the " +
+				"on-chain HTLC broadcast by the swap client " +
+				"in the case of a Loop In",
```
joostjager: Description may be confusing because there is no way yet to control loop in chain fees.

cmd/loop/loopout.go
wpaulino: Seems like a good candidate for a follow-up PR?

cmd/loop/loopout.go
wpaulino: Hyphens are usually the standard for flags containing multiple words, but we seem to have different patterns (no separator, underscore, hyphen) across all of our CLI flags within `lnd`, `lncli` and `loop`.

cmd/loopd/swapclient_server.go
```diff
@@ -283,9 +294,13 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 
 	logger.Infof("Loop in quote request received")
 
+	confTarget, err := validateConfTarget(req.ConfTarget, defaultConfTarget)
```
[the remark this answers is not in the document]
wpaulino: Done.

cmd/loop/quote.go
```diff
@@ -12,12 +12,25 @@ var quoteCommand = cli.Command{
...
+			Value: 6,
```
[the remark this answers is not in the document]
wpaulino: Since the quote command will be used for all types of swaps (in and out) I opted to not use that constant here.

looprpc/client.proto
wpaulino: Yeah I'm not sure how useful `max_miner_fee` is anymore given the risk once the preimage is revealed. Seems like we'd want a better strategy at that point, e.g., positive yield?

cmd/loop/quote.go
wpaulino: Removed loop in portion.

cmd/loop/loopout.go
Roasbeef: Yeah

cmd/loop/loopout.go
Roasbeef: Most of the `lncli` flags use undercores. Only for `lnd` flags do we sometimes use a hyphen.

cmd/loop/quote.go
Roasbeef: Current "inherited" style from `lncli` is underscore delimiters.

wpaulino: Done.

PR #70, opened 2019-07-27
 loop+loopout: validate hash of swap invoice

bjornoj:
Right now, the loop client trusts that the loop server responds with a correct swap invoice.  This makes it possible for a loop server to respond with a regular(non-hold) invoice whose hash is generated by the server, and the client would still pay it.

To prevent this from happening, this PR makes the client check that the hash of the swap invoice is the same as the hash it generated. If the hashes are not equal, the client returns an error and aborts the swap.

### DIFF
--- a/loopout.go
+++ b/loopout.go
@@ -86,12 +86,12 @@ func newLoopOutSwap(globalCtx context.Context, cfg *swapConfig,
 		return nil, fmt.Errorf("cannot initiate swap: %v", err)
 	}
 
-	err = validateLoopOutContract(cfg.lnd, currentHeight, request, swapResp)
+	err = validateLoopOutContract(cfg.lnd, currentHeight, request, swapHash, swapResp)
 	if err != nil {
 		return nil, err
 	}
 
-	// Instantie a struct that contains all required data to start the swap.
+	// Instantiate a struct that contains all required data to start the swap.
 	initiationTime := time.Now()
 
 	contract := loopdb.LoopOutContract{
@@ -660,21 +660,26 @@ func (s *loopOutSwap) sweep(ctx context.Context,
 // validateLoopOutContract validates the contract parameters against our
 // request.
 func validateLoopOutContract(lnd *lndclient.LndServices,
-	height int32,
-	request *OutRequest,
+	height int32, request *OutRequest, swapHash lntypes.Hash,
 	response *newLoopOutResponse) error {
 
 	// Check invoice amounts.
 	chainParams := lnd.ChainParams
 
-	swapInvoiceAmt, err := swap.GetInvoiceAmt(
+	swapInvoiceHash, swapInvoiceAmt, err := swap.DecodeInvoice(
 		chainParams, response.swapInvoice,
 	)
 	if err != nil {
 		return err
 	}
 
-	prepayInvoiceAmt, err := swap.GetInvoiceAmt(
+	if swapInvoiceHash != swapHash {
+		return fmt.Errorf(
+			"cannot initiate swap, swap invoice hash %v not equal generated swap hash %v",
+			swapInvoiceHash, swapHash)
+	}
+
+	_, prepayInvoiceAmt, err := swap.DecodeInvoice(
 		chainParams, response.prepayInvoice,
 	)
 	if err != nil {
--- a/swap/fees.go
+++ b/swap/fees.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/zpay32"
 )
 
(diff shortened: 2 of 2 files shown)


loopout.go
```diff
@@ -86,12 +86,25 @@ func newLoopOutSwap(globalCtx context.Context, cfg *swapConfig,
...
+		cfg.lnd.ChainParams, swapResp.swapInvoice,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if swapInvoiceHash != swapHash {
+		return nil, fmt.Errorf(
+			"cannot initiate swap, swap invoice hash %s not equal generated swap hash %s",
+			swapInvoiceHash.String(), swapHash.String())
```
joostjager: If you use `%v`, the `String()` call is not necessary.

loopout.go
```diff
@@ -86,12 +86,25 @@ func newLoopOutSwap(globalCtx context.Context, cfg *swapConfig,
...
 	}
 
+	swapInvoiceHash, err := swap.GetInvoiceHash(
+		cfg.lnd.ChainParams, swapResp.swapInvoice,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if swapInvoiceHash != swapHash {
```
joostjager: The purpose of the `validateLoopOutContract` function is to do this kind of validation. You can move this code into that function too. Added benefit is that the invoice only needs to be decoded once.

loopout.go
bjornoj: Fixed!

loopout.go
bjornoj: Fixed in the latest commit. Moved the logic inside validateLoopOutContract and replaced GetInvoiceAmt and GetInvoiceHash with a more generic DecodeInvoice.

Question: I imagine DecodeInvoice already exists in lnd somewhere. Is there a point in duplicating the logic in loop?

loopout.go
```diff
@@ -662,19 +662,26 @@ func (s *loopOutSwap) sweep(ctx context.Context,
 func validateLoopOutContract(lnd *lndclient.LndServices,
 	height int32,
 	request *OutRequest,
+	swapHash lntypes.Hash,
 	response *newLoopOutResponse) error {
```
joostjager: nit: could have fewer newlines (more args on a line)

loopout.go
joostjager: I am not aware of such a wrapper function in `lnd`. There the call is typically made to `zpay32` directly.

swap/fees.go
```diff
@@ -27,38 +27,24 @@ func FeeRateAsPercentage(feeRate int64) float64 {
 	return float64(feeRate) / (FeeRateTotalParts / 100)
 }
 
-// GetInvoiceAmt gets the invoice amount. It requires an amount to be
-// specified.
-func GetInvoiceAmt(params *chaincfg.Params,
-	payReq string) (btcutil.Amount, error) {
+// DecodeInvoice gets the hash and the amount of an invoice.
+// It requires an amount to be specified.
+func DecodeInvoice(params *chaincfg.Params,
```
halseth: Squash this commit with the first one.

loopout.go
```diff
@@ -675,19 +662,26 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
-	swapInvoiceAmt, err := swap.GetInvoiceAmt(
+	swapInvoiceHash, swapInvoiceAmt, err := swap.DecodeInvoice(
 		chainParams, response.swapInvoice,
 	)
 	if err != nil {
 		return err
 	}
 
-	prepayInvoiceAmt, err := swap.GetInvoiceAmt(
+	if swapInvoiceHash != swapHash {
```
halseth: This commit can be squashed with the second one.

loopout.go
bjornoj: nit fixed

swap/fees.go
halseth: So the comits would be:
1. add  the `DecodeInvoice`method
2. Add the hash check within `validateLoopOutContract`

bjornoj: Attempt 2. Is the current commit more like it?

halseth: @bjornoj  Actually, can you just squash all changes in the PR into one commit? Thanks! 😀

bjornoj: Of course! Done

halseth: Awesome, looks good!

bjornoj: Would it be better to omit all commit messages other than the first one? ie the commit message for the PR would just be this:
```
loop+loopout: validate hash of swap invoice
    
    This commit fixes a possible exploit by the loop server, where
    - in a loop out - the server could claim money off-chain, without
    publishing an on-chain swap htlc.
    
    The server could do this by responding with a regular invoice, whose hash
    is different than the hash in the NewLoopOutSwap request. To prevent
    the exploit, we validate that the hash of the swap invoice is equal to the
    hash the client generated.
```

halseth: Yes, sounds good!

bjornoj: There we go, all good now

PR #75, opened 2019-08-06
SwapPublicationDeadline for LoopOut contracts

halseth:
## SwapPublicationDeadline option
The SwapPublicationDeadline option  is added for the client to signal to the swap server that it is prepared to wait a given amount of time before the swap is published on-chain. This let the server take this into account when doing calculating the swap fee, potentially batching swaps together or waiting for less congested fee periods, reducing the swap fee for the clients.

By default the CLI will set a maximum swap deadline of 30 minutes, while setting the `--fast` flag will request an immediate swap.


batcher/batcher.go
```diff
@@ -0,0 +1,266 @@
...
+	// txPublishTicker represents the time between each time we'll check
+	// whether it is time to batch and send the current set of send
+	// requests.
+	s.cfg.TxPublishTicker.Resume()
+	defer s.cfg.TxPublishTicker.Stop()
+
+	for {
+
+		select {
+		// A new send is being requested. Add it to rou set of requests
```
joostjager: rou?

batcher/batcher.go
```diff
@@ -0,0 +1,266 @@
...
+		// and start the ticker.
+		case req := <-s.sendOutputChan:
+			// Add the new request to our slice, and sort it.
+			sendRequests = append(sendRequests, req)
+			sort.Slice(sendRequests, func(i, j int) bool {
+				return sendRequests[i].cutoff.Before(
+					sendRequests[j].cutoff,
+				)
+
+			})
```
joostjager: newlines

batcher/batcher.go
```diff
@@ -0,0 +1,266 @@
...
+			var txOuts []*wire.TxOut
+			for i := range sendRequests {
+				// If we reach the maximum batch size, we'll
+				// break here. The rest will be sent in the
+				// next batch.
+				if uint32(i) == s.cfg.MaxBatchSize {
+					break
+				}
+
+				txOuts = append(txOuts, sendRequests[i].txOut)
```
joostjager: Isn't this nicer with a slice operation? First decide the number of items (min(maxbatchsize, n)) and then slice in one operation?

batcher/batcher.go
```diff
@@ -0,0 +1,266 @@
...
+	defer s.cfg.TxPublishTicker.Stop()
+
+	for {
+
+		select {
+		// A new send is being requested. Add it to rou set of requests
+		// and start the ticker.
+		case req := <-s.sendOutputChan:
+			// Add the new request to our slice, and sort it.
+			sendRequests = append(sendRequests, req)
```
joostjager: Instead of letting this grow beyond max size, shouldn't we publish right away when we've added the max'th item?

loopdb/loopout.go
```diff
@@ -155,6 +160,17 @@ func deserializeLoopOutContract(value []byte, chainParams *chaincfg.Params) (
...
+	err = binary.Read(r, byteOrder, &contract.MaxSwapWaitTimeSec)
+	switch {
+	// Field not available for old versions.
+	case err == io.ErrUnexpectedEOF:
+		fallthrough
+	case err == io.EOF:
+		return &contract, nil
+	case err != nil:
+		return nil, err
+	}
```
joostjager: There is a migration mechanism already in place to which this migration can be added. It is slightly more work, but prevents this legacy code from staying around forever.

looprpc/server.proto
```diff
@@ -22,6 +22,10 @@ message ServerLoopOutRequest {
     bytes swap_hash = 2;
 
     uint64 amt = 3;
+
+    uint64 max_swap_wait_time_sec = 4;
```
joostjager: Isn't it better to use an absolute time here? That would be better observable.

cmd/loopd/swapclient_server.go
```diff
@@ -75,6 +75,7 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 		MaxSwapRoutingFee:   btcutil.Amount(in.MaxSwapRoutingFee),
 		MaxSwapFee:          btcutil.Amount(in.MaxSwapFee),
 		SweepConfTarget:     sweepConfTarget,
+		MaxSwapWaitTimeSec:  in.MaxSwapWaitTimeSec,
```
joostjager: I wouldn't mind grouping some changes together in a single commit, but that's my personal preference.

batcher/batcher.go
```diff
@@ -0,0 +1,266 @@
...
+				if uint32(i) == s.cfg.MaxBatchSize {
+					break
+				}
+
+				txOuts = append(txOuts, sendRequests[i].txOut)
+			}
+
+			// Create and publish a transaction paying to the
+			// gathered outputs.
+			tx, err := s.publishTxOuts(ctx, txOuts)
```
joostjager: One thing to keep in mind here, is that we change the swap state to `Published` right before the actual publish call. We need to persist the state first, to be absolutely sure that we won't ever publish twice after a restart.

The batcher adds a potentially large timespan between offering the `TxOut` and the actual publish. Any restart within that span would cause the swap not to progress anymore. The publish hasn't happened yet and will also never happen anymore.

Maybe an extra signal back to the caller is required here to keep the time window between the state change and the actual publication of the tx as small as possible.

looprpc/server.proto
```diff
@@ -74,4 +78,4 @@ message ServerLoopInQuoteResponse {
     uint64 min_swap_amount = 4;
     uint64 max_swap_amount = 5;
     int32 cltv_delta = 6;
-}
\ No newline at end of file
+}
```
joostjager: `max_swap_wait_time_sec` needs to be added to the quote call too, to be able to see the difference in fee

looprpc/server.proto
joostjager: On the `cli` it could still be a relative time as a convenience.

batcher/batcher_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}, wallet)
+
+	runCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go b.Run(runCtx)
+
+	// Keep track of the time as we go, so we can be sure to always
+	// increase it on every tick.
+	now := time.Now()
```
joostjager: non-deterministic

halseth:
PR changed to only consist of client changes.

loopdb/loopout.go
```diff
@@ -45,6 +45,10 @@ type LoopOutContract struct {
 	// MaxPrepayRoutingFee is the maximum off-chain fee in msat that may be
 	// paid for the prepayment to the server.
 	MaxPrepayRoutingFee btcutil.Amount
+
+	// MaxSwapWaitTimeUnix is the maximum time in unix seconds the client
+	// allows the server to wait before publishing the on chain swap.
+	MaxSwapWaitTimeSec int64
```
guggero: Could we name this `SwapDeadline` or at least `MaxSwapWaitTimestamp`? As far as I understood it, it no longer contains the relative number of seconds to wait but an absolute timestamp at this point. So it's a bit confusing that it still has the same name as in the gRPC message.

looprpc/server.proto
```diff
@@ -51,6 +56,9 @@ message ServerLoopOutQuote {
     uint64 max_swap_amount = 6;
 
     int32 cltv_delta = 7;
+
+    /// The unix time in seconds we want the swap to be initiated by.
+    int64 max_swap_wait_time_unix = 8;
```
joostjager: This is part of the request?

Maybe we should also now add `swap_amt` to the request and return `swap_fee` in the response. We can then deprecate `swap_fee_base` and `swap_fee_rate`, as a preparation for more advanced pricing models.

interface.go
```diff
@@ -65,6 +65,10 @@ type OutRequest struct {
 	// LoopOutChannel optionally specifies the short channel id of the
 	// channel to loop out.
 	LoopOutChannel *uint64
+
+	// MaxSwapWaitTimeSec can be set by the client to allow the server
+	// delaying publication of the swap HTLC to save on chain fees.
+	MaxSwapWaitTimeSec uint64
```
joostjager: I think this should be `time.Time`?

loopout.go
```diff
@@ -77,10 +77,19 @@ func newLoopOutSwap(globalCtx context.Context, cfg *swapConfig,
...
+	initiationTime := time.Now()
+	logger.Infof("Initiating swap request at height %v, time %v",
+		currentHeight, initiationTime)
+
+	// We'll take our current time and add our max swap time to it.  This
+	// will be given to the server for it to use as the latest swap
+	// publication time.
+	maxSwapTimestamp := initiationTime.Add(
+		time.Duration(request.MaxSwapWaitTimeSec) * time.Second,
+	)
```
joostjager: I think that it may be better to perform this calculation in the cli or rpc caller, to keep the flexibility to pass in an absolute time.

looprpc/client.proto
```diff
@@ -133,6 +133,14 @@ message LoopOutRequest {
...
     */
     int32 sweep_conf_target = 9;
+
+    /**
+    The maximum time (in seconds) we allow the server to wait before publishing
+    the HTLC on chain. Setting this to a larger value will give the server the
+    opportunity to batch multiple swaps together, and wait for low-fee periods
+    before publishing the HTLC, potentially resulting in a lower swap fee.
+    */
+    uint64 max_swap_wait_time_sec = 10;
```
joostjager: As mentioned before, convert to absolute (unix) timestamp to support absolute deadlines.

cmd/loop/loopout.go
```diff
@@ -113,6 +120,7 @@ func loopOut(ctx *cli.Context) error {
 		MaxSwapRoutingFee:   int64(*limits.maxSwapRoutingFee),
 		LoopOutChannel:      unchargeChannel,
 		SweepConfTarget:     sweepConfTarget,
+		MaxSwapWaitTimeSec:  ctx.Uint64("max_swap_wait_time"),
```
joostjager: This is where we'd convert the relative deadline from the cmd line to an absolute time.

guggero:
This looks good!
I'm still testing it but haven't found any bugs now.

The main thing I tripped over was the naming of the variable. It wasn't always clear when it is a relative number of seconds and when a timestamp.
@joostjager's suggestion to only accept absolute timestamps, even on gRPC would solve that. So I would go for that too.

loopdb/loopout.go
halseth: I liked the "deadline" idea, so converted all these to `SwapPublicationDeadline`, and using `time.Time`: https://github.com/lightninglabs/loop/pull/75/commits/1cff25ac6aed4fee758741d6fef8c46b4f017cce

loopdb/migration_02_max_swap_time.go
```diff
@@ -0,0 +1,158 @@
...
+
+		// Read the old format....
+		contract, err := deserializeLoopOutContractV01(
+			contractBytes, chainParams,
+		)
+		if err != nil {
+			return err
+		}
+
+		contract.MaxSwapWaitTimeSec = 0
```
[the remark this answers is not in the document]
halseth: Using initiation time 👍

looprpc/server.proto
```diff
@@ -22,6 +22,11 @@ message ServerLoopOutRequest {
     bytes swap_hash = 2;
 
     uint64 amt = 3;
+
+    /// The unix time in seconds we want the swap to be initiated by.
+    int64 max_swap_wait_time_unix = 4;
```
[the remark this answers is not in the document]
halseth: Changed name, and added more spcific comment.

looprpc/server.proto
halseth: can be done in later PR, but yeah that sounds reasonable.

loopout.go
halseth: Done!

halseth:
> This looks good!
> I'm still testing it but haven't found any bugs until now.
> 
> The main thing I tripped over was the naming of the variable. It wasn't always clear when it is a relative number of seconds and when a timestamp.
> @joostjager's suggestion to only accept absolute timestamps, even on gRPC would solve that. So I would go for that too.

Changed name to `SwapPublicationDeadline`, let me know if that makes more sense! 😀

looprpc/server.proto
joostjager: Ok, but `max_swap_wait_time_unix` should still be part of the quote request? Otherwise, what is returned here by the server?

halseth: `swap_publication_deadline` is in the quote request, if that is what you mean?

joostjager: it is currently in `ServerLoopOutQuote`, but I thought it should be added to `ServerLoopOutQuoteRequest`?

halseth: oh, yeah that's a mistake 🤦‍♂

halseth: Can even just consider removing it for now? As it is not used anywhere currently

guggero:
Needs rebase :)

halseth:
Rebased. 

Also changed the serialization format of the swap publication deadline to be the same as creation time:  https://github.com/lightninglabs/loop/pull/75/commits/604262096b1918bbe956412f7c7b157d5583edae
This let us simplify the migration significantly: https://github.com/lightninglabs/loop/pull/75/commits/50c0101683b8bff83d3f13a594f9e80d21a467d2

loopdb/migration_02_swap_publication_deadline.go
```diff
@@ -134,33 +33,26 @@ func migrateSwapPublicationDeadline(tx *bbolt.Tx, chainParams *chaincfg.Params)
...
 			return errors.New("contract not found")
 		}
 
-		// Read the old format....
-		contract, err := deserializeLoopOutContractV01(
-			contractBytes, chainParams,
-		)
-		if err != nil {
+		// Write the current contract serialization into a buffer.
+		b := &bytes.Buffer{}
```
guggero: Can't this just be `b := bytes.NewBuffer(contractBytes)`? Should be ok since we only read from `contractBytes` later. Or is this considered bad practice?

loopdb/migration_02_swap_publication_deadline.go
```diff
@@ -134,33 +33,26 @@ func migrateSwapPublicationDeadline(tx *bbolt.Tx, chainParams *chaincfg.Params)
...
-		contract.SwapPublicationDeadline = time.Unix(
-			contract.InitiationTime.Unix(), 0,
-		)
-
-		// ...write the new format.
-		bs, err := serializeLoopOutContract(contract)
-		if err != nil {
+		// We migrate to the new format by copying the first 8 bytes
+		// (the creation time) to the end (the swap deadline)
+		var swapDeadline [8]byte
```
guggero: Nice that we can solve it that way and remove all that other migration code!

cmd/loopd/swapclient_server.go
```diff
@@ -68,14 +69,15 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
...
-		SweepConfTarget:     sweepConfTarget,
+		Amount:                  btcutil.Amount(in.Amt),
+		DestAddr:                sweepAddr,
+		MaxMinerFee:             btcutil.Amount(in.MaxMinerFee),
+		MaxPrepayAmount:         btcutil.Amount(in.MaxPrepayAmt),
+		MaxPrepayRoutingFee:     btcutil.Amount(in.MaxPrepayRoutingFee),
+		MaxSwapRoutingFee:       btcutil.Amount(in.MaxSwapRoutingFee),
+		MaxSwapFee:              btcutil.Amount(in.MaxSwapFee),
+		SweepConfTarget:         sweepConfTarget,
+		SwapPublicationDeadline: time.Unix(int64(in.SwapPublicationDeadline), 0),
```
guggero: nit: line too long

loopdb/migration_02_swap_publication_deadline.go
halseth: I'm not sure tbh... The docs say
```
NewBuffer is intended to prepare a Buffer to read existing data. It can also be used to set the initial size of the internal buffer for writing. To do that, buf should have the desired capacity but a length of zero.
```
maybe safest to avoid it?

cmd/loopd/swapclient_server.go
halseth: Fixed: 4c4ac6d

halseth:
@alexbosworth `--fast` flag added. Is it a good idea though, to change the default behavior from fast to slow?

alexbosworth:
I wouldn't put it too slow for that reason, like no more than a slow block could reasonably take.

The question is really about expectations:

1. You should already have an expectation that a swap is going to take a little while to complete due to the preimage sweep phase
2. You should already have an expectation that confirmation is a part of the flow, which can be hours
3. You could also say that the default expectation is to get an economically priced swap, with the batching strategy delayed swap execution is pretty much necessary in order to make swaps economic, unless there is huge volume

The missing feature I see here is a status indicator on the swap. We don't have any way to query the status of active swaps and the UI leans on the Blockchain to carry the messaging of swap progress, when it really should be out of band

PR #74, opened 2019-08-03
update api for loop in

callmekurisu:
### Description
Updating API for loop in (^_^)

### Steps to Test
GET http://localhost:8081/v1/loop/in/terms

GET http://localhost:8081/v1/loop/in/quote/250001

POST http://localhost:8081/v1/loop/in
content-type: application/json

{
  "amt": "",
  "loop_in_channel": "",
  "max_miner_fee": "",
  "max_swap_fee": "",
  "external_htlc: true
}

### Screenshots
image

image

PR #86, opened 2019-09-20
loopd: don't fail external loop in on fee estimation failure

### LINKED ISSUE #84: loop in external doesn't work if there's no onchain funds
I'm trying to do a `loop in --amt 250000 --external` and I got this response
```
[loop] rpc error: code = Unknown desc = insufficient funds available to construct transaction
```
I have more than 50 channels with plenty inbound capacity, at that moment I had 13451 satoshis on-chain wallet balance, so I sent to my node 1000000 onchain btc and after that I could run the same command, I understand that loop in always is looking for that `--amt` in the node wallet to make the swap and is ignoring the `--external` flag.

guggero:
In the `LoopInQuote` function of `client.go` an on-chain fee estimation for paying the HTLC is done by calling `EstimateFeeToP2WSH` on lnd.
This fails in case the lnd wallet has no UTXOs left to construct an estimation transaction with.
For an internal loop in, this can be seen as a hard failure and the swap can't continue.
For an external loop however, it doesn't matter that the lnd wallet has no funds. So we just warn about not being able to estimate the on-chain fee.

Fixes #84 

I'm not too happy about sending back an error as a string. The two operations (getting a swap fee quote from the swap server and doing the on-chain fee estimation in lnd) would ideally be separate calls so the client could handle the distinct errors. But this would be a bigger refactor. Open to suggestions here.

Note: The large diff in the `looprpc` package is mainly due to #74 which was probably generated with a wrong version of `github.com/grpc-ecosystem/grpc-gateway`. It also included some manual changes in the `client.swagger.json` that are overwritten when generating again. This can also go into a separate PR if it's too big for this bugfix.


looprpc/client.proto
```diff
@@ -372,4 +372,11 @@ message QuoteResponse {
     An estimate of the on-chain fee that needs to be paid to sweep the HTLC.
     */
     int64 miner_fee = 3;
+    
+    /**
+    If an error happens during on-chain fee estimation in lnd, this contains
+    the error. The client should decide if it's ok to continue (i.e. external
+    loop in) or not.
+    */
+    string miner_fee_estimation_err = 4;
```
Roasbeef: Can't we just return a regular error here using the existing gRPC error codes?

Roasbeef: IIRC, the original issue failed just as a result of an attempted swap execution, or is it that the issue also bled over into the quote call?

client.go
```diff
@@ -470,17 +470,26 @@ func (s *Client) LoopInQuote(ctx context.Context,
...
+	// are doing an external loop in anyway. It is not OK for an internal
+	// loop in though, because this will fail certainly. So we let the
+	// client handle this by forwarding the error to it.
 	minerFee, err := s.lndServices.Client.EstimateFeeToP2WSH(
 		ctx, request.Amount, request.HtlcConfTarget,
 	)
+	var errString string
 	if err != nil {
-		return nil, err
+		// Return the error as string so the client can decide to
```
Roasbeef: Can't we instead return a wrapped error here?

cmd/loop/loopin.go
```diff
@@ -67,6 +67,21 @@ func loopIn(ctx *cli.Context) error {
 	}
 
 	limits := getInLimits(amt, quote)
+	external := ctx.Bool("external")
+	
+	// There could have been an error in the on-chain fee estimation if
```
Roasbeef: Following from my last comment, we can use `Is/As` here instead and gate the warning or exit based on that.

looprpc/client.proto
```diff
@@ -372,4 +372,11 @@ message QuoteResponse {
     An estimate of the on-chain fee that needs to be paid to sweep the HTLC.
     */
     int64 miner_fee = 3;
+    
+    /**
+    If an error happens during on-chain fee estimation in lnd, this contains
+    the error. The client should decide if it's ok to continue (i.e. external
+    loop in) or not.
```
joostjager: I am wondering what the user expects in a quote response when the intention is to use an external address. A zero miner fee? Or a happy case fee (1 inputs, 1 htlc out, 1 change out)?

How about passing in an `external` bool to the quote request to skip the lnd wallet fee estimation? It makes it all a bit more transparent. Returning an error sometimes for an external funded swap quote may be confusing.

guggero: Good idea! I like that much more than passing an error back.
I wasn't sure if it was OK to just skip the fee estimation completely on an external swap in.

guggero:
I did look into returning an error in the response message directly. I haven't found anything useful, it looks like there is no error primitive in protobuf corresponding to a golang error. We could have used numeric error codes. Or send some additional data in the GRPC error body and then parse that in the client (to still get the swap fee that was calculated).

But I like Joost's suggestion much more to just skip on-chain fee estimation completely for an external loop in.

PR #88, opened 2019-09-26
server proto: add amt to ServerLoopOutQuoteRequest

halseth:
In order to move to a dynamic fee model for Loop Outs, we add a amount
field to the ServerLoopOutQuoteRequest. The idea is that the server will
return a total fee estimate based on the information given in the
QuoteRequest. For now this is only the swap amount, but this can be
extended in the future.

Given the amount, the server will give a total fee estimate, int the
SwapFeeBase fee. The total fee is kept here for backwards compatibility.


looprpc/server.proto
```diff
@@ -74,4 +78,4 @@ message ServerLoopInQuoteResponse {
     uint64 min_swap_amount = 4;
     uint64 max_swap_amount = 5;
```
joostjager: I think we should already now add a `LoopOutTerms` call to the server to expose the static terms (at least `min_swap_amount` and `max_swap_amount`). Otherwise someone willing to find out the min/max range needs to guess an amount within that range first to get a valid quote response.

halseth: Added

client.go
```diff
@@ -353,11 +353,14 @@ func (s *Client) LoopOutQuote(ctx context.Context,
...
-	logger.Infof("Offchain swap destination: %x", terms.SwapPaymentDest)
+	quote, err := s.Server.GetLoopOutQuote(ctx, request.Amount)
+	if err != nil {
+		return nil, err
+	}
 
-	swapFee := swap.CalcFee(
-		request.Amount, terms.SwapFeeBase, terms.SwapFeeRate,
-	)
+	logger.Infof("Offchain swap destination: %x", quote.SwapPaymentDest)
```
joostjager: Why add this?

client.go
```diff
@@ -381,7 +384,7 @@ func (s *Client) LoopOutQuote(ctx context.Context,
 	return &LoopOutQuote{
 		SwapFee:      swapFee,
```
joostjager: `swapFee` helper variable, while `quote.PrepayAmount` is copied into the response directly.

looprpc/client.proto
```diff
@@ -339,7 +339,7 @@ message TermsResponse {
     /**
     On-chain cltv expiry delta
     */
-    int32 cltv_delta = 7;
+    int32 cltv_delta = 7 [deprecated=true];
```
joostjager: We don't set those values anymore, so we could just as well remove them in the proto completely and only leave a `reserved`?

client.go
```diff
@@ -381,7 +384,7 @@ func (s *Client) LoopOutQuote(ctx context.Context,
...
 		MinerFee:     minerFee,
-		PrepayAmount: btcutil.Amount(terms.PrepayAmt),
+		PrepayAmount: btcutil.Amount(quote.PrepayAmount),
```
joostjager: Set `Dest` and `CltvDelta`?

cmd/loopd/swapclient_server.go
```diff
@@ -266,10 +267,13 @@ func (s *swapClientServer) LoopOutQuote(ctx context.Context,
...
 
+	dest := hex.EncodeToString(quote.SwapPaymentDest[:])
 	return &looprpc.QuoteResponse{
-		MinerFee:  int64(quote.MinerFee),
-		PrepayAmt: int64(quote.PrepayAmount),
-		SwapFee:   int64(quote.SwapFee),
+		MinerFee:        int64(quote.MinerFee),
+		PrepayAmt:       int64(quote.PrepayAmount),
+		SwapFee:         int64(quote.SwapFee),
+		SwapPaymentDest: dest,
```
joostjager: Why is this actually a string field?

looprpc/client.proto
halseth: That would break users of the protos, but maybe that's better than having them keep using values not set anymore?

client.go
halseth: It was there from before, just on the terms instead of quote.

cmd/loopd/swapclient_server.go
halseth: It was from before, do you think it makes more sense as raw bytes?

client.go
halseth: Fixed: 0714370

cmd/loopd/swapclient_server.go
joostjager: Now that we move it, I think there is an opportunity to change. Idk why it was string based, but in general I don't think we should do hex strings. The only reason I know of is for parameters that show up in rest urls.

looprpc/client.proto
halseth: Did change to reserved

cmd/loopd/swapclient_server.go
halseth: Changed to `bytes`

client.go
```diff
@@ -465,10 +470,13 @@ func (s *Client) LoopInQuote(ctx context.Context,
 		return nil, ErrSwapAmountTooHigh
 	}
 
+	quote, err := s.Server.GetLoopInQuote(ctx, request.Amount)
+	if err != nil {
+		return nil, err
+	}
+
 	// Calculate swap fee.
```
guggero: nit: maybe remove the comment or change it since there's nothing calculated any more.

halseth: Done

PR #89, opened 2019-09-27
loopout: compare delta from htlc expiry correctly

wpaulino:
This addresses an issue where using a sweep confirmation target greater than the default would result in most cases not revealing the preimage due to the default confirmation target yielding a higher fee than the max miner fee backed by the confirmation target provided.

### DIFF
--- a/loopout_test.go
+++ b/loopout_test.go
@@ -6,6 +6,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
 	"github.com/lightninglabs/loop/lndclient"
 	"github.com/lightninglabs/loop/loopdb"
 	"github.com/lightninglabs/loop/sweep"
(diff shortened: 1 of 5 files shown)


alexbosworth:
Should it have had an associated test?

joostjager:
Change LGTM. But indeed, at least one sanity checking test for this would be useful.

wpaulino:
Test added.

loopout_test.go
```diff
@@ -99,3 +100,115 @@ func TestLateHtlcPublish(t *testing.T) {
...
+
+	ctx.NotifyConf(htlcTx)
+
+	// The client should then register for a spend of the HTLC and attempt
+	// to sweep it using the custom confirmation target.
+	ctx.AssertRegisterSpendNtfn(swap.htlc.PkScript)
+
+	expiryChan <- time.Now()
+
+	ctx.AssertConfTargetQueried(testRequest.SweepConfTarget)
```
joostjager: This assert seems to rely a lot on the internals of the swap implementation. I'd think that it is much more robust to inspect the output value of the tx. This is the swap amount minus the miner fee.

loopout_test.go
```diff
@@ -99,3 +100,115 @@ func TestLateHtlcPublish(t *testing.T) {
...
+		t.Fatalf("expected state %v, got %v",
+			loopdb.StatePreimageRevealed, status.State)
+	}
+
+	sweepTx := ctx.ReceiveTx()
+	if sweepTx.TxIn[0].PreviousOutPoint.Hash != htlcTx.TxHash() {
+		t.Fatalf("expected sweep tx to spend %v, got %v",
+			htlcTx.TxHash(), sweepTx.TxIn[0].PreviousOutPoint)
+	}
+
```
joostjager: Generate some more blocks here and assert that the tx is rbf'ed with the default conf target?

wpaulino: Done.

loopout_test.go
```diff
@@ -99,3 +101,158 @@ func TestLateHtlcPublish(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Set up the required dependencies to execute the swap.
+	sweeper := &sweep.Sweeper{Lnd: &lnd.LndServices}
+	blockEpochChan := make(chan interface{})
+	statusChan := make(chan SwapInfo)
+	expiryChan := make(chan time.Time)
```
joostjager: Creating test context like `loopInTestContext` for loop in would be good to reduce code duplication and maintenance cost in the future. Maybe it is too much for a 1-character PR though. Otherwise you could add a TODO in both of these tests so that the next dev might do it.

loopout_test.go
```diff
@@ -99,3 +101,158 @@ func TestLateHtlcPublish(t *testing.T) {
...
+				htlcTx.TxHash(), sweepTx.TxIn[0].PreviousOutPoint)
+		}
+
+		sweepTxFee := btcutil.Amount(
+			htlcTx.TxOut[0].Value - sweepTx.TxOut[0].Value,
+		)
+		customFee, err := sweeper.GetSweepFee(
+			context.Background(), swap.htlc.AddSuccessToEstimator,
+			swap.DestAddr, expConfTarget,
+		)
```
joostjager: It is tricky to use the code under test in the assertion. Bugs may go unnoticed. Can't this assertion be done by calculating the published fee rate based on the tx weight?

wpaulino: Wouldn't I need to do the same thing `GetSweepFee` does to retrieve the weight?

joostjager: I thought using `SerializeSize` and `SerializeSizeStripped` on the `MsgTx`

wpaulino: Done.

loopout_test.go
```diff
@@ -99,3 +102,163 @@ func TestLateHtlcPublish(t *testing.T) {
...
+		weight := blockchain.GetTransactionWeight(btcutil.NewTx(sweepTx))
+		feeRate, err := ctx.Lnd.WalletKit.EstimateFee(
+			context.Background(), expConfTarget,
+		)
+		if err != nil {
+			t.Fatalf("unable to retrieve fee estimate: %v", err)
+		}
+		minFee := feeRate.FeeForWeight(weight)
+
+		if minFee > fee {
```
joostjager: Why not equal here?

loopout_test.go
joostjager: Do you want to add the TODO?

wpaulino: Done.

loopout_test.go
joostjager: Sorry, missed that. Maybe you could add some margin, to at least verify that it actually publishes with the lower fee rate.

wpaulino: Wouldn't adding a margin be the same as getting the weight estimate as done in `GetSweepFee`? That's why I opted for it initially, as we really care about the confirmation target being used for the sake of the test, not the fee itself.

PR #93, opened 2019-10-04
loop+cmd: extract types into common module

guggero:
If we also want to use the types defined in loop on the server, we currently need to pull in the root module with all its dependencies.
By extracting the types into their own module, we avoid pulling in dependent sub-modules like `looprpc` for example when using the types in the server code.

### DIFF
--- a/cmd/loop/loopin.go
+++ b/cmd/loop/loopin.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcutil"
-	"github.com/lightninglabs/loop"
 	"github.com/lightninglabs/loop/looprpc"
+	"github.com/lightninglabs/loop/swap"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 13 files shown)


looptypes/interface.go
```diff
@@ -1,4 +1,4 @@
-package loop
+package looptypes
 
 import (
 	"time"
```
joostjager: It looks like there is more moved to the new package than necessary. For example `OutRequest`.

cmd/loop/loopin.go
```diff
@@ -69,7 +69,7 @@ func loopIn(ctx *cli.Context) error {
 	}
 
 	limits := getInLimits(amt, quote)
-	err = displayLimits(loop.TypeIn, amt, limits, external)
+	err = displayLimits(looptypes.TypeIn, amt, limits, external)
```
joostjager: If possible we make sure that all commits build. Seeing this change, the previous commit probably doesn't.

looptypes/interface.go
guggero: So you would suggest to only move the types into this new module that are currently used by the server? This feels a bit strange to me to have these core types in multiple packages.

joostjager:
Build failed

guggero:
Oops, fixed it.

guggero:
Yes, this should have been obvious, sorry. It does make most sense that way.
The linter urged me to not call the logger `swap.SwapLog` so I renamed it to `swap.PrefixLog` which should more accurately describe what it does.

PR #94, opened 2019-10-07
Add linter to makefile, lint on Travis build

guggero:
Use the same configuration for linting as in lnd. The command `make lint` can now be run and linting is also run on Travis build.

I fixed the most obvious errors then locked the linter down to only report new issues, added after c859f0c1b87e490ec1d7dd7aa15eea0dbf8c3ef6.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,43 @@
+run:
+  # timeout for analysis
+  deadline: 4m
+
+  # Linting uses a lot of memory. Keep it under control by only running a single
+  # worker.
+  concurrency: 1
+
+linters-settings:
+  gofmt:
+    # simplify code: gofmt with `-s` option, true by default
+    simplify: true
+
+linters:
+  enable-all: true
+  disable:
+  # Global variables are used in many places throughout the code base.
+  - gochecknoglobals
+
+  # Some lines are over 80 characters on purpose and we don't want to make them
+  # even longer by marking them as 'nolint'.
+  - lll
+
+  # We don't care (enough) about misaligned structs to lint that.
+  - maligned
+
+  # We have long functions, especially in tests. Moving or renaming those would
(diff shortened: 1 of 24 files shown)


lndclient/invoices_client.go
```diff
@@ -52,10 +52,10 @@ func (s *invoicesClient) WaitForFinished() {
 func (s *invoicesClient) SettleInvoice(ctx context.Context,
 	preimage lntypes.Preimage) error {
 
-	rpcCtx, cancel := context.WithTimeout(ctx, rpcTimeout)
+	_, cancel := context.WithTimeout(ctx, rpcTimeout)
 	defer cancel()
```
joostjager: This now looks useless. I think this timeout context needs to be passed into the `WithMacaroonAuth` call.

guggero: Right, fixed it.

guggero:
I noticed there's a `.golangci.yml` in lnd that has some additional settings. I added that file for loop too but removed the build tag and mobile items that aren't needed here.

halseth:
file conflict?

guggero:
Fixed

guggero:
Fixed Travis too

.golangci.yml
```diff
@@ -0,0 +1,46 @@
...
+  deadline: 4m
+
+  # Linting uses a lot of memory. Keep it under control by only running a single
+  # worker.
+  concurrency: 1
+
+linters-settings:
+  govet:
+    # Don't report about shadowed variables
+    check-shadowing: false
```
halseth: check  if we can remove.

guggero: I removed it and ran it for the whole git history. Didn't mention any new issues so we can use the configuration without this exception.

guggero:
I added some more small fixes for things that came into master after this PR was opened.

PR #99, opened 2019-10-28
multi: add persistent logger

guggero:
This PR replaces #30 now that the `build` package in lnd contains most of what needed to be copy/pasted before.
The diff is so large because the `logger` was renamed to just `log` in all packages as to follow the same convention as in lnd.

The discussion from #30 remains: How should the path for the log files look?
What's currently implemented is the same structure as in lnd (`.loopd/logs/<network>/loopd.log`).
The other possibility would be to set `defaultLogDir` to just `loopDirBase` which will result in the structure `.loopd/<network>/loopd.log`.


Roasbeef:
As far as on disk structure, I'm leaning towards the way it is now for `lnd`. It also gives us the network/chain separation as well, which is future proof.

PR #101, opened 2019-11-01
LSAT aware client interceptor

guggero:
This PR makes the loop daemon aware of the LSAT authentication scheme.
An unary client gRPC connector is added that intercepts server responses and handles payment challenges sent by the server.
An LSAT token is automatically acquired and paid by the interceptor and then stored in the file `.loop/<network>/lsat.token`.


cmd/loopd/config.go
```diff
@@ -27,12 +27,13 @@ type lndConfig struct {
...
-	Insecure    bool   `long:"insecure" description:"disable tls"`
-	Network     string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
-	SwapServer  string `long:"swapserver" description:"swap server address host:port"`
-	RPCListen   string `long:"rpclisten" description:"Address to listen on for gRPC clients"`
-	RESTListen  string `long:"restlisten" description:"Address to listen on for REST clients"`
+	ShowVersion    bool   `short:"V" long:"version" description:"Display version information and exit"`
+	Insecure       bool   `long:"insecure" description:"disable tls"`
+	Network        string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
+	SwapServer     string `long:"swapserver" description:"swap server address host:port"`
+	TLSPathSwapSrv string `long:"tlspathswapserver" description:"Path to swap server tls certificate"`
```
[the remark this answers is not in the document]
Roasbeef: Or this is just for _testing_ purposes?

lsat/interceptor.go
```diff
@@ -0,0 +1,184 @@
...
+	GRPCErrMessage = "payment required"
+
+	// AuthHeader is is the HTTP response header that contains the payment
+	// challenge.
+	AuthHeader = "WWW-Authenticate"
+
+	// MaxRoutingFee is the maximum routing fee in satoshis that we are
+	// going to pay to acquire an LSAT token.
+	// TODO(guggero): make this configurable
+	MaxRoutingFee = 10
```
Roasbeef: % or satoshis?

lsat/interceptor.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return &Interceptor{
+		lnd:   lnd,
+		store: store,
+	}, nil
+}
+
+// UnaryInterceptor is an interceptor method that can be used directly by gRPC
+// for unary calls.
+// If the store contains a token, it is attached as credentials to every call
```
Roasbeef: Style nit: either add a new line above, or wrap with the above line.

lsat/interceptor.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			return err
+		}
+		if err = addLsatCredentials(lsat); err != nil {
+			return err
+		}
+
+		// Execute the same request again, now with the LSAT
+		// token added as an RPC credential.
+		return invoker(ctx, method, req, reply, cc, opts...)
```
Roasbeef: How does this flow handle expiration? So the case that you have a valid token, but it has expired (as they all may for the initial versions, so we aren't locked in). I guess in the end, it depends on how we communicate expiry from the server-side.

lsat/interceptor.go
```diff
@@ -0,0 +1,184 @@
...
+	token, err := NewToken(macBytes, invoice.PaymentHash)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create token: %v", err)
+	}
+
+	// Pay invoice now and wait for the result to arrive.
+	respChan := i.lnd.Client.PayInvoice(
+		context.Background(), invoiceStr, MaxRoutingFee, nil,
+	)
+	result := <-respChan
```
Roasbeef: Should pass in the context and also select on its `Done()` channel so this can be properly cancelled.

lsat/interceptor.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode invoice: %v", err)
+	}
+	token, err := NewToken(macBytes, invoice.PaymentHash)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create token: %v", err)
+	}
+
+	// Pay invoice now and wait for the result to arrive.
+	respChan := i.lnd.Client.PayInvoice(
```
wpaulino: Is there a risk of double paying if two calls that require LSATs are made concurrently and we don't have a token yet?

lsat/store.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return ioutil.WriteFile(f.fileName, bytes, 0600)
+}
+
+// fileExists returns true if the file exists, and false otherwise.
+func fileExists(path string) bool {
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return false
+		}
```
wpaulino: Should we return false `if err != nil && !os.IsNotExist(err)`?

lsat/token.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+	return mac, nil
+}
+
+// serializeToken returns a byte-serialized representation of the token.
+func serializeToken(t *Token) ([]byte, error) {
+	var b bytes.Buffer
+
+	macLen := int32(len(t.BaseMac))
```
wpaulino: Should be `uint32` instead?

wpaulino:
Should we also consider maintaining a history of all LSATs paid for accounting purposes?

cmd/loopd/config.go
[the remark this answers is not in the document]
guggero: This is only to get it working with test-staging. Before, we used `--insecure` there which did HTTP/2 over non-TLS. Kirin can't handle that so I had to drop the `--insecure` and add a self-signed cert.
This won't be a problem for testnet and mainnet as we'll have valid letsencrypt (autocert) certificates.

lsat/interceptor.go
guggero: Yes, that was my assumption also that we would just return "payment required" in that case too and the client would get a new token.
In a future PR we can add a command line option like `--no-auto-lsat`.

lsat/interceptor.go
guggero: The client library currently only supports fixed fee limit in sats. Added the suffix to make it clear.

lsat/interceptor.go
guggero: Yes, definitely. Although it's unlikely there would be parallel requests with the current architecture, I added a mutex anyway.

lsat/store.go
guggero: I thought about that too when I copied the function from lnd.
But when `err != nil` we can't be sure that it doesn't exist, it could also just be a permission issue. So it's probably safer to try to read it and run into an error later then trying to overwrite it.

guggero:
I updated the store to keep old tokens and not just overwrite them.
Also there is now a command that lists all tokens (current and expired) in the command line.

The reason I implemented a file based token store is so people wanting to integrate the interceptor into their code base don't necessarily need a boltDB.
The question is now, should we also add a boltDB compliant store for `loopd` itself or go with the file based store for now?

lsat/store.go
```diff
@@ -22,9 +24,13 @@ type Store interface {
...
-	// Token returns the token that is contained in the store or an error
-	// if there is none.
-	Token() (*Token, error)
+	// CurrentToken returns the token that is currently contained in the
+	// store or an error if there is none.
+	CurrentToken() (*Token, error)
+
+	// AllTokens returns all tokens that the store has knowledge of, even
+	// if they might be expired.
+	AllTokens() ([]*Token, error)
```
Roasbeef: Should update the `StoreToken` `godoc` comment to indicate all tokens should be kept around with in-tact date information.

cmd/loop/lsat.go
```diff
@@ -0,0 +1,73 @@
...
+	PaymentHash     string `json:"payment_hash"`
+	PaymentPreimage string `json:"payment_preimage"`
+	AmountPaid      int64  `json:"amount_paid_msat"`
+	RoutingFeePaid  int64  `json:"routing_fee_paid_msat"`
+	TimeCreated     int64  `json:"time_crated"`
+	Expired         bool   `json:"expired"`
+}
+
+var listTokensCommand = cli.Command{
+	Name:        "listtokens",
```
Roasbeef: `listauth`?

lsat/token.go
```diff
@@ -0,0 +1,188 @@
...
+	token.RoutingFeePaid = routingFeePaid
+	return token, nil
+}
+
+// tokenFromChallenge parses the parts that are present in the challenge part
+// of the LSAT auth protocol which is the macaroon and the payment hash.
+func tokenFromChallenge(baseMac []byte, paymentHash *[32]byte) (*Token, error) {
+	// First, validate that the macaroon is valid and can be unmarshaled.
+	mac := &macaroon.Macaroon{}
+	err := mac.UnmarshalBinary(baseMac)
```
wpaulino: Nit: store the serialized macaroon instead?

lsat/store.go
```diff
@@ -63,33 +69,76 @@ func (f *FileStore) HasToken() bool {
...
 // StoreToken saves a token to the store, overwriting any old token if there is
 // one.
 //
 // NOTE: This is part of the Store interface.
 func (f *FileStore) StoreToken(token *Token) error {
+	// Don't just overwrite existing old tokens but create a backup for
+	// accounting purposes.
+	if f.HasToken() {
+		timestampSuffix := time.Now().Format("2006-01-02T15-04-05")
+		err := os.Rename(f.fileName, f.fileName+timestampSuffix)
```
wpaulino: We'll want to make sure this rename along with the new token write below are atomic.

cmd/loopd/swapclient_server.go
```diff
@@ -342,6 +342,33 @@ func (s *swapClientServer) LoopIn(ctx context.Context,
 	}, nil
 }
 
+// GetLsatTokens returns all tokens that are contained in the LSAT token store.
+func (s *swapClientServer) GetLsatTokens(ctx context.Context,
+	_ *looprpc.TokensRequest) (*looprpc.TokensResponse, error) {
+
+	log.Infof("Get LSAT tokens request received")
```
wpaulino: Nit: demote to debug?

lsat/store.go
Roasbeef: On windows, we also need to _close_ the file (if it's already open) before we rename it. We've hit this a few times before in neutrino and also the way we handle SCB files on disk.

lsat/store.go
guggero: Done.

lsat/token.go
guggero: Yeah, makes sense.

cmd/loopd/swapclient_server.go
guggero: All the other RPCs have this as info so I think we can keep it.

lsat/store.go
guggero: Ok, I tried making this as atomic as possible.

lsat/store.go
```diff
@@ -63,31 +69,83 @@ func (f *FileStore) HasToken() bool {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// We want this operation to be atomic. So if the write of the new file
+	// fails, we want to move the old one back into its position.
+	err = ioutil.WriteFile(f.fileName, bytes, 0600)
+	if err != nil && newFileName != "" {
+		_ = os.Rename(newFileName, f.fileName)
```
wpaulino: Following my other comment, I think we can actually just return the error here and not rename the file back (what you had previously) because we'll have a pending state that we can recover from for the new token we're attempting to store.

lsat/token.go
```diff
@@ -0,0 +1,185 @@
...
+// server.
+func (t *Token) BaseMacaroon() *macaroon.Macaroon {
+	return t.baseMac.Clone()
+}
+
+// PaidMacaroon returns the base macaroon with the proof of payment (preimage)
+// added as a first-party-caveat.
+func (t *Token) PaidMacaroon() (*macaroon.Macaroon, error) {
+	mac := t.BaseMacaroon()
+	caveat := checkers.Condition(
```
wpaulino: Leaving a note here that we'll want to use our library for this to avoid the additional dependency.

lsat/interceptor.go
```diff
@@ -0,0 +1,201 @@
...
+		token, err := NewToken(
+			macBytes, invoice.PaymentHash, result.Preimage,
+			lnwire.NewMSatFromSatoshis(result.PaidAmt),
+			lnwire.NewMSatFromSatoshis(result.PaidFee),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create token: %v",
+				err)
+		}
+		return token, i.store.StoreToken(token)
```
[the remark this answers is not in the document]
guggero: Yes, I agree this could happen.
Unfortunately AFAICT we don't have the `TrackPayment` method exposed in the loops's `lndclient` yet.
And since this is pretty unlikely to happen anyway (the client has to be shut down exactly at the moment when the first request to the server is done), I propose to fix that in a later PR.
I added a TODO comment.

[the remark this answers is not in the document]
guggero: Ah, I must've looked in the wrong place... You are right then, we should address any gaps we can.
I've added the `trackPayment` functionality and also a unit test.

lsat/store.go
guggero: If the write of the new file fails, we do want to move the old file (which contained the pending state) back, otherwise the store won't return true on `HasToken()` the next time. If the rename back fails, we _should_ return the error though, because then we really are in a strange state.

lsat/interceptor.go
```diff
@@ -0,0 +1,292 @@
...
+	// match for us to be able to parse the macaroon and invoice.
+	authHeaderRegex = regexp.MustCompile(
+		"LSAT macaroon='(.*?)' invoice='(.*?)'",
+	)
+)
+
+// Interceptor is a gRPC client interceptor that can handle LSAT authentication
+// challenges with embedded payment requests. It uses a connection to lnd to
+// automatically pay for an authentication token.
+type Interceptor struct {
```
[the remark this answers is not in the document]
Roasbeef: Good idea, this whitelist would also need to be enforced at the proxy though. As is, it's only aware of the service, and but not the individual methods of the service itself.

lsat/interceptor.go
```diff
@@ -0,0 +1,292 @@
...
+				return i.store.StoreToken(token)
+
+			// The payment is still in transit, we'll give it more
+			// time to complete.
+			case routerrpc.PaymentState_IN_FLIGHT:
+
+			// Any other state means either error or timeout.
+			default:
+				return fmt.Errorf("payment tracking failed "+
+					"with state %s", result.State.String())
```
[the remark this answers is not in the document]
Roasbeef: From my reading, right now the client can just re-try again itself. You're suggesting we implement retry automatically in the interceptor? If the client isn't able to send a de minimis amount of sats (say 1 satoshi) or so, then there may be greater underlying issues with their channel set up. 

Zooming out this also serves as an initial reachability test between the client/server as well.

lsat/interceptor.go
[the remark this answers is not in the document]
Roasbeef: One route to implement this would be a new caveat that signals calls which don't need tokens. Or we just extend the existing `freebie` system to examine the request and allow certain calls.

lsat/store.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// StoreToken saves a token to the store. Old tokens should be kept for
+	// accounting purposes but marked as invalid somehow.
+	StoreToken(*Token) error
+}
+
+// FileStore is an implementation of the Store interface that uses a single file
+// to save the serialized token.
+type FileStore struct {
+	fileName string
```
Roasbeef: Possible for multiple swaps to be in flight, so we should embed a `sync.Mutex` within the struct.

lsat/interceptor.go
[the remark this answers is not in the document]
guggero: I've added a PR to the proxy to allow (whitelist) certain calls.

lsat/interceptor.go
```diff
@@ -0,0 +1,292 @@
...
+		// If we receive a state without an error, the payment has been
+		// initiated. Loop until the payment
+		case result := <-payStatusChan:
+			switch result.State {
+			// If the payment was successful, we have all the
+			// information we need and we can return the fully paid
+			// token.
+			case routerrpc.PaymentState_SUCCEEDED:
+				token.Preimage = result.Preimage
+				if result.Route != nil {
```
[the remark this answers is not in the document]
guggero: No, probably not. Removed the if.

lsat/interceptor.go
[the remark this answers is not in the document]
guggero: I'm also in favor of not implementing automatic retry at this moment. I refactored the whole logic with the pending/final token file handling again and added a more explicit feedback to the user where they can find the file and that they can try deleting it to retry. The `listauth` command also shows the file location of the tokens.

lsat/store.go
```diff
@@ -0,0 +1,169 @@
...
+				return err
+			}
+		}
+	}
+
+	// We want this operation to be atomic. So if the write of the new file
+	// fails, we want to move the old one back into its position.
+	err = ioutil.WriteFile(f.fileName, bytes, 0600)
+	if err != nil && newFileName != "" {
+		return os.Rename(newFileName, f.fileName)
```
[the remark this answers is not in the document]
guggero: I think with the new logic it should be clearer now.

lsat/interceptor_test.go
```diff
@@ -0,0 +1,288 @@
...
+			}
+		}
+
+		backendCalled = true
+		return backendErr
+	}
+
+	// Test case 1: The happy path for a call that doesn't require
+	// payment, we expect the backend to be called without any payments
+	// being made.
```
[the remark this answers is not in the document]
guggero: Sure, why not. Looks nicer now.

lsat/store.go
```diff
@@ -50,42 +69,150 @@ func NewFileStore(storeDir string) (*FileStore, error) {
...
+		// Write the new token first, so we still have the pending
+		// around if something goes wrong.
+		err := ioutil.WriteFile(f.fileName, bytes, 0600)
+		if err != nil {
+			return err
+		}
+		return os.Remove(f.fileNamePending)
+
+	// Replace a paid token with a new one by creating a backup first.
+	case !currentToken.isPending() && !newToken.isPending():
```
wpaulino: This case seems unused so far since we always write the pending token first, and such an action is not allowed if we already have a prior valid token. If that's the approach we're going to take for now, I'd rather just remove this.

wpaulino:
Linter failing with:

```
lsat/store_test.go:76:11: S1005: should omit value from range; this loop is equivalent to `for key := range ...` (gosimple)
	for key, _ := range tokens {
	         ^
lsat/store_test.go:111:11: S1005: should omit value from range; this loop is equivalent to `for key := range ...` (gosimple)
	for key, _ := range tokens {
```

lsat/interceptor.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// UnaryInterceptor is an interceptor method that can be used directly by gRPC
+// for unary calls. If the store contains a token, it is attached as credentials
+// to every call before patching it through. The response error is also
+// intercepted for every call. If there is an error returned and it is
+// indicating a payment challenge, a token is acquired and paid for
+// automatically. The original request is then repeated back to the server, now
+// with the new token attached.
+func (i *Interceptor) UnaryInterceptor(ctx context.Context, method string,
```
wpaulino: We should add some context to the LSAT payment errors within this method. Some additional logging here would be nice too.

wpaulino: An issue I ran into is that if we have a pending token but the call we're trying to access doesn't require one, then we'll still block the call from happening. Not sure how we can address this without introducing the whitelist in here as well.

cmd/loop/lsat.go
```diff
@@ -0,0 +1,79 @@
...
+
+type printableToken struct {
+	ID              string `json:"id"`
+	ValidUntil      int64  `json:"valid_until"`
+	BaseMacaroon    string `json:"base_macaroon"`
+	PaymentHash     string `json:"payment_hash"`
+	PaymentPreimage string `json:"payment_preimage"`
+	AmountPaid      int64  `json:"amount_paid_msat"`
+	RoutingFeePaid  int64  `json:"routing_fee_paid_msat"`
+	TimeCreated     int64  `json:"time_crated"`
```
bucko13: typo: I'm guessing should be `time_created`

lsat/interceptor.go
```diff
@@ -161,31 +180,103 @@ func (i *Interceptor) payLsatToken(ctx context.Context, md *metadata.MD) (
...
+			// If the payment was successful, we have all the
+			// information we need and we can return the fully paid
+			// token.
+			case routerrpc.PaymentState_SUCCEEDED:
+				token.Preimage = result.Preimage
+				total := result.Route.TotalAmount
+				fees := result.Route.TotalFees()
+				token.AmountPaid = total - fees
+				token.RoutingFeePaid = fees
+				return i.store.StoreToken(token)
```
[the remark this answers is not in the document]
guggero: Unfortunately the data structure returned from `PayInvoice` is completely different to that of `TrackPayment`. That should be fixed in `lndclient` IMO. I extracted the part in `trackPayment` to a function.

lsat/store.go
```diff
@@ -1,38 +1,57 @@
...
-// FileStore is an implementation of the Store interface that uses a single file
-// to save the serialized token.
+// FileStore is an implementation of the Store interface that files to save the
+// serialized tokens. There is always just one current token that is either
+// pending or fully paid.
 type FileStore struct {
-	fileName string
+	fileName        string
+	fileNamePending string
+	lock            sync.Mutex
```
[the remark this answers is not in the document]
guggero: Yeah, you're right. I added it because @Roasbeef suggested to do so. But as long as we don't use the store outside of the interceptor it doesn't add any value.

cmd/loop/lsat.go
guggero: Yes, thanks!

lsat/interceptor.go
guggero: Good point, if no token token is required for a call then we shouldn't try to resume the token payment at that moment.
We can check it without knowing the whitelist, we just have to wait to run into a 402 first and only then resume the token payment.
I refactored the whole logic to make it (hopefully) easier to follow.

lsat/store.go
```diff
@@ -50,42 +67,131 @@ func NewFileStore(storeDir string) (*FileStore, error) {
...
+				"existing pending token")
+		}
+
+		// Write the new token first, so we still have the pending
+		// around if something goes wrong.
+		err := ioutil.WriteFile(f.fileName, bytes, 0600)
+		if err != nil {
+			return err
+		}
+		return os.Remove(f.fileNamePending)
```
wpaulino: What do you think about not returning the error here? So it's just a best effort operation since we were able to write the token anyway. We can also try removing it on startup within `NewTokenStore` if the file exists and there's a corresponding paid token.

cmd/loop/lsat.go
```diff
@@ -0,0 +1,79 @@
...
+
+type printableToken struct {
+	ID              string `json:"id"`
+	ValidUntil      int64  `json:"valid_until"`
+	BaseMacaroon    string `json:"base_macaroon"`
+	PaymentHash     string `json:"payment_hash"`
+	PaymentPreimage string `json:"payment_preimage"`
+	AmountPaid      int64  `json:"amount_paid_msat"`
+	RoutingFeePaid  int64  `json:"routing_fee_paid_msat"`
+	TimeCreated     int64  `json:"time_created"`
```
wpaulino: Nit: since these are only for the CLI, I'd prefer showing the actual date here.

guggero:
Fixed last nits, this should be good to merge. @alexbosworth want to take a final look?

PR #102, opened 2019-11-02
readme: update slack link

alexbosworth:
Replace non-working slack link with working link


bryanvu:
Not to ruin hours of massive effort, but you can also now link to https://lightning.engineering/slack, which will allow us to update the slack links in a single place when they expire.

alexbosworth:
> Not to ruin hours of massive effort, but you can also now link to https://lightning.engineering/slack.html, which will allow us to update the slack links in a single place when they expire.

Doesn't resolve for me

PR #107, opened 2019-11-09
Make max routing fee rate configurable

NickShelby:


alexbosworth:
Are you still working on this? I think it could be a cool feature

NickShelby:
Ha, thank you @alexbosworth, I'm glad you find this interesting.

I completely forgot I started this, and I was wondering who created this PR for quite a few minutes.

I'm curious why the default `maxRoutingFeeBase` and `maxRoutingFeeRate` was set to such a high limit. My proposal is to lower them to 2 and 4 as suggested in this Draft PR, and have a `max_swap_routing_fee` configurable if people need to bump the fee for urgent loop out.

If you think this sounds reasonable, I'll proceed with the proposal and polish the PR.

Thanks!

alexbosworth:
I don't think lowering the default is that good of an idea because of the lack of user feedback. To a user it could easily just look like they have no possibility to pay, when it should suggest that they can raise their fee limit.

Definitely a user specified routing fee limit is a great idea though

NickShelby:
> I don't think lowering the default is that good of an idea because of the lack of user feedback. To a user it could easily just look like they have no possibility to pay, when it should suggest that they can raise their fee limit.

Makes sense, I was thinking about also updating the failure prompt message if it failed due to the fee was too low. But to keep this PR simple, that'll be for another day then.

My personal experience is that after I used loop out a few times, the fee jumped up very fast, from 3~6 sats, to 2000 sats, which was higher than on-chain fee I paid. It's probably because my node wasn't that well connected, but seeing that 2000 off-chain fee did discourage me from using it freely. I would prefer to have my transaction failed if the off-chain fee was too high.

Anyway, I'll focus on adding the config in this PR then.

PR #110, opened 2019-11-15
lndclient: extend send payment parameters

joostjager:
This PR extends the proxy `SendPayment` method with parameters to allowing payments without an invoice. This is intended to be used for prepay probes.

### DIFF
--- a/lndclient/router_client.go
+++ b/lndclient/router_client.go
@@ -6,19 +6,17 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/lightningnetwork/lnd/lnrpc"
-	"github.com/lightningnetwork/lnd/routing/route"
-
-	"github.com/lightningnetwork/lnd/channeldb"
-	"google.golang.org/grpc/codes"
-
-	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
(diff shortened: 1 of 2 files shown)


joostjager:
There is still some diff left. I also tried to update to `lnd v0.8.1-beta`, but got:

`walletkit_client.go:13:2: module github.com/lightningnetwork/lnd@latest (v0.0.2) found, but does not contain package github.com/lightningnetwork/lnd/lnwallet/chainfee`

lndclient/router_client.go
```diff
@@ -49,6 +47,32 @@ type SendPaymentRequest struct {
 	MaxCltv         *int32
 	OutgoingChannel *uint64
 	Timeout         time.Duration
+
+	// Target is the node in which the payment should be routed towards.
```
halseth: maybe note that all these will be ignored if the invoice is non-empty

halseth: or return an error if both are set.

joostjager: comment added

halseth: I mean, it should be noted on the fields on `SendPaymentRequest`

joostjager: Added

PR #121, opened 2019-12-03
lndclient: fix Value/ValueMsat conversion problem

guggero:
A change in lnd where a field was converted from Sat to MSat wasn't correctly propagated to the `lndclient`. This PR fixes that.


guggero:
We discussed this offline, we'll revert the offending change in lnd to not break backwards compatibility.

guggero:
I was wrong, there is no breaking change in the RPC wire format in lnd. I must've mixed up something. This fix actually is enough tho fix the problem.

guggero:
Yes, I verified that the server is now able to build and run itest with these changes.

PR #123, opened 2019-12-09
lndclient: expose node URIs and message signing RPCs in lightning client

guggero:
For an internal project we need the node URI of the `GetInfo` call exposed in the lightning client.
We also need the ability to sign and verify arbitrary messages.


lndclient/lightning_client.go
```diff
@@ -36,6 +37,11 @@ type LightningClient interface {
 
 	AddInvoice(ctx context.Context, in *invoicesrpc.AddInvoiceData) (
 		lntypes.Hash, string, error)
+
+	SignMessage(ctx context.Context, msg []byte) (string, error)
+
+	VerifyMessage(ctx context.Context, msg []byte, sig string) (bool,
+		*btcec.PublicKey, error)
```
joostjager: It is pre-existing (for the other fields), but we normally document external methods.

test/lightning_client_mock.go
```diff
@@ -119,3 +119,12 @@ func (h *mockLightningClient) AddInvoice(ctx context.Context,
 
 	return hash, payReqString, nil
 }
+
+func (h *mockLightningClient) SignMessage(ctx context.Context, msg []byte) (
+	string, error) {
+
+	// The message "test" signed with
+	// 03f5374b16f0b1f1b49101de1b9d89e0b460bc57ce9c2f9132b73dfc76d3704daa
```
joostjager: Set this key in the mocked `GetInfo`?

test/lightning_client_mock.go
```diff
@@ -119,3 +119,12 @@ func (h *mockLightningClient) AddInvoice(ctx context.Context,
 
 	return hash, payReqString, nil
 }
+
+func (h *mockLightningClient) SignMessage(ctx context.Context, msg []byte) (
+	string, error) {
+
+	// The message "test" signed with
```
joostjager: Assert that `msg` is "test"?

test/lightning_client_mock.go
```diff
@@ -128,3 +129,20 @@ func (h *mockLightningClient) SignMessage(ctx context.Context, msg []byte) (
 	return "rdsuxrzk1dyrq3hzx6496x7tzs5f6k89cz1hecsprgssmku9q5oy6uwoxoras" +
 		"d85igaixb5chtrfxjmb8hri7tqqfqduf7etxgipec7s", nil
 }
+
+func (h *mockLightningClient) VerifyMessage(ctx context.Context, msg []byte,
+	sig string) (bool, *btcec.PublicKey, error) {
+
+	pubKeyHex := "03f5374b16f0b1f1b49101de1b9d89e0b460bc57ce9c2f9132b73df" +
+		"c76d3704daa"
```
joostjager: Same comment, should this be a mock parameter? Or assert that the sig is the same that we returned from `SignMessage` in the mock?

test/lightning_client_mock.go
guggero: Yes, good idea.

guggero:
In discussion with @Roasbeef it turned out that we actually need the custom signing methods added in https://github.com/lightningnetwork/lnd/pull/3812. Should we wait for that PR to be merged and add the wrapper for those too or create a follow-up PR?
Or should we only create a wrapper for the methods we actually need even though this code has already been written?

joostjager:
I would only wrap what you need now. It is not yet that general purpose native golang wrapper.

test/lightning_client_mock.go
```diff
@@ -132,3 +133,23 @@ func (h *mockLightningClient) SignMessage(ctx context.Context, msg []byte) (
 
 	return h.lnd.Signature, nil
 }
+
+func (h *mockLightningClient) VerifyMessage(ctx context.Context, msg []byte,
+	sig string) (bool, *btcec.PublicKey, error) {
+
+	pubkeyBytes, err := hex.DecodeString(h.lnd.NodePubkey)
```
joostjager: If you start decoding here, maybe it is better to make the field `route.Vertex` and using `String()` when needed.

test/lightning_client_mock.go
```diff
@@ -119,3 +127,29 @@ func (h *mockLightningClient) AddInvoice(ctx context.Context,
 
 	return hash, payReqString, nil
 }
+
+func (h *mockLightningClient) SignMessage(ctx context.Context, msg []byte) (
+	string, error) {
+
+	return h.lnd.Signature, nil
```
joostjager: Verify that `msg == signatureMsg`? I don't know how much this mock needs to do, this is just a thought.

guggero:
With lightningnetwork/lnd#3812 now merged, I moved the two methods over to the signer which is what we really need.

test/lightning_client_mock.go
guggero: The pubkey parsing is not necessary now since we don't return it in `VerifyMessage` anymore.

test/lightning_client_mock.go
guggero: I don't think the mock should return any errors that couldn't also happen with the real call. I think it's fine to leave it as is for the moment.

test/lnd_services_mock.go
```diff
@@ -127,6 +130,7 @@ type LndMockServices struct {
 
 	Height     int32
 	NodePubkey string
+	Signature  []byte
```
joostjager: One could still debate whether it should be here or on the mock. But not blocking

lndclient/signer_client.go
```diff
@@ -14,6 +15,17 @@ import (
...
+
+	// SignMessage signs a message with the key specified in the key
+	// locator. The returned signature is fixed-size LN wire format encoded.
+	SignMessage(ctx context.Context, msg []byte,
+		locator keychain.KeyLocator) ([]byte, error)
+
+	// VerifyMessage verifies a signature over a message using the public
+	// key provided. The signature must be fixed-size LN wire format
+	// encoded.
+	VerifyMessage(ctx context.Context, msg, sig, pubkey []byte) (bool,
```
wpaulino: Nit: make `pubkey [33]byte` so that it's clear a compressed public key should be used.

guggero: Yes, good idea. Fixed.

PR #125, opened 2019-12-12
lndclient: expose ListTransactions as part of LightningClient

wpaulino:


test/lightning_client_mock.go
```diff
@@ -119,3 +120,9 @@ func (h *mockLightningClient) AddInvoice(ctx context.Context,
 
 	return hash, payReqString, nil
 }
+
+func (h *mockLightningClient) ListTransactions(
+	ctx context.Context) ([]*wire.MsgTx, error) {
+
+	return nil, nil
```
guggero: Would be nice if we could control what the mock returns by adding a variable to `test/lnd_services_mock.go` and returning it here.

wpaulino: Done, also included in this list the transactions by `SendOutputs` and `PublishTransaction`.

test/lnd_services_mock.go
```diff
@@ -122,6 +122,8 @@ type LndMockServices struct {
 
 	Height int32
 
+	RelevantTransactions []*wire.MsgTx
```
joostjager: Is the word `Relevant` accurate in the mock context? I'd expect something like `AllTransactions`

wpaulino: Fixed.

PR #128, opened 2019-12-30
Add option for max swap off-chain routing fee

### LINKED ISSUE #100: Add max offchain fee parameter to loop out command
When I loop out, it sets a default max offchain fee, but I'd like to be able to configure this value myself.

Free-FreeThinker:
Add a new command line option `max_swap_routing_fee` for specifying max swap routing fee.

Fixes #100

### DIFF
--- a/cmd/loop/loopout.go
+++ b/cmd/loop/loopout.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/btcsuite/btcutil"
 	"github.com/lightninglabs/loop"
 	"github.com/lightninglabs/loop/looprpc"
 	"github.com/lightninglabs/loop/swap"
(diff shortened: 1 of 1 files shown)


cmd/loop/loopout.go
```diff
@@ -115,6 +121,10 @@ func loopOut(ctx *cli.Context) error {
 	}
 
 	limits := getLimits(amt, quote)
+	// If configured, use the specified maximum swap routing fee.
+	if ctx.IsSet("max_swap_routing_fee") {
+		*limits.maxSwapRoutingFee = btcutil.Amount(ctx.Int64("max_swap_routing_fee"))
```
joostjager: Max line len of 80 chars exceeded. We use tab size 8.

Free-FreeThinker: @joostjager, do we have any tool for code formatting?

guggero: There is the `make fmt` command that makes sure that the code is formatted in the way the linter expects.
But long lines aren't automatically wrapped. See [the contribution guidelines](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#48-additional-style-constraints-on-top-of-gofmt) to see how we wrap method calls.

This example would look like this:

```
 		*limits.maxSwapRoutingFee = btcutil.Amount(
 		 		ctx.Int64("max_swap_routing_fee"),
 		) 
```

Free-FreeThinker: Thanks @guggero.

CR feedbacks have been addressed, PTAL, thx!

cc @joostjager

PR #115, opened 2019-11-26
Create loopd package, allow custom listeners

halseth:
This PR lays the foundational work to allow `loopd` to be used as a library. 

It moves the logic currently residing in `cmd/loopd` into a new package `loopd`, and adds configuration options for custom gRPC listeners, allowing `loopd` to both accept incoming connections on custom listeners, and connecting to an `lnd` instance on a custom listener.¨

This is similar to what was done for `lnd` to allow it to be used with in-memory gRPC clients, paving the way for mobile support: https://github.com/lightningnetwork/lnd/pull/3282

PR #135, opened 2020-01-15
loopin: fix handling of incorrect amount in external htlc tx

### LINKED ISSUE #134: Overpaid `loop in --external` htlc. Timeout transaction unable to broadcast.
Accidentally overpaid an external loop in htlc.

I was a bit confused by how much to pay as the instructions displayed at the time of the htlc creation with the loop cli mentioned both the loop in amount and a fee. 

I misunderstood that the total amount sent to the htlc address should JUST be the loop in amount requested and that the loop in fee would be deducted from the lightning payment. I paid BOTH the loop in amount and the loop in fee to the external htlc address.

Just a slight total over-payment of about 1000 sats.

Waited 1000 blocks for htlc timeout. loopd created the timeout tx `LOOP: Publishing timeout tx`, and sent it to lnd for broadcast `[INF] LNWL: Inserting unconfirmed transaction`. 

tx never appeared in mempool. loopd kept submitting timeout transactions building up a larger and larger `unconfirmed_balance` in lnd.

Tried broadcasting the timeout transaction created by loopd manually by obtaining the raw tx from `lncli listchaintxns`. Received the following error:

`sendrawtransaction RPC error: {"code":-26,"message":"non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)"}`

LND on mainnet with neutrino:
`$ lnd --version
lnd version 0.9.0-beta commit=v0.9.0-beta-rc1-30-g269182c9b9f140612ba020b56bd1627c53cf88f0`

Loop:
`$ loopd -V
loopd version 0.3.1-alpha commit=`

joostjager:
Fixes #134

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -265,6 +265,9 @@ func testSuccess(ctx *testContext, amt btcutil.Amount, hash lntypes.Hash,
 	// Publish tick.
 	ctx.expiryChan <- testTime
 
+	// Expect a signing request.
+	<-ctx.Lnd.SignOutputRawChannel
+
 	if !preimageRevealed {
 		ctx.assertStatus(loopdb.StatePreimageRevealed)
 		ctx.assertStorePreimageReveal()
--- a/loopin.go
+++ b/loopin.go
@@ -419,7 +419,7 @@ func (s *loopInSwap) publishOnChainHtlc(ctx context.Context) (bool, error) {
 // the swap invoice is either settled or canceled. If the htlc times out, the
 // timeout tx will be published.
 func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
-	htlc *wire.OutPoint, htlcValue btcutil.Amount) error {
+	htlcOutpoint *wire.OutPoint, htlcValue btcutil.Amount) error {
 
 	// Register the htlc spend notification.
 	rpcCtx, cancel := context.WithCancel(ctx)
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -29,6 +29,9 @@ output*.log
 loop
 !cmd/loop/
 
+loopd
```
guggero: What command do you use to build?
None of the Makefile commands nor `go build ./...` / `go install ./...` actually seem to create these files, that's why we removed the lines in #115.
Not a big deal to add them back, just curious if they are indeed still needed.

halseth: Yeah, this is wrong, since it will ignore everything in the `loopd` folder. 

> None of the Makefile commands nor `go build ./...` or `go install ./...` actually seem to create these files, that's why we removed the lines in #115.

I'm also didn't see the binary appear again, maybe you just have it lying around?

joostjager: I run `go build` inside `cmd/loopd` .

For me, it does not ignore everything in the loopd. If I create a file there, it is picked up by git. Are you sure it does not work for you?

joostjager: Other question is why the same lines for `loop` are still in the file if they cause everything to be ignored.

guggero: Ah, yes, if you go into a directory with a main package then the binary is indeed created there.
I think what we really should do is add the `make build` and `make install` commands and then add that generated binary to the `.gitignore`.

There is a new top-level `loopd` directory (not `cmd/loopd`) that would be ignored/untracked with these changes.

joostjager: Ah, there is a `loopd` directory now. Ok, removed the .gitignore change. Alternatively we could also have added `!loopd` to .gitignore I think. How do you usually run locally then if it isn't with `go build`?

guggero: Yes, that was Johan's suggestion before we decided to remove it altogether.
I always run `go build ./...` in the root directory to check for compilation errors. Or `go install ./...` if I want to update the binaries in my `$GOPATH/bin` directory.

joostjager: Hmm, installing the binaries for local dev isn't something I normally do. Sometimes I also have multiple versions.

PR #139, opened 2020-01-23
version: bump version to 0.4.0-beta-rc1

alexbosworth:


guggero:
I think we should go with the official [Semantic Versioning]() where it says `
<pre-release> ::= <dot-separated pre-release identifiers>`.
So `appPreRelease` should be `rc1.beta` instead of `beta-rc1`.
This will prevent the case where go mod picks the RC tag instead of the final, non-RC version.
We should do the same for all our projects.

alexbosworth:
Can you make an example PR of what you mean @guggero or an alternative to this one with the version you prefer? This one goes off of the LND format

guggero:
Just change the line
```
	appPreRelease = "beta-rc1"
```
to
```
	appPreRelease = "rc1.beta"
```
I'll propose this for the next lnd release as well.

PR #140, opened 2020-01-24
loopd+looprpc: add REST compatible status calls

### LINKED ISSUE #82: Add a REST / sync equivalent for Monitor
I noticed this was missing from the REST RPC, and [saw that there was a TODO in the docs for the gRPC endpoint](https://lightning.engineering/loop/#monitor), but wanted to crack open an issue for this so that I could monitor any progress on this front. Would be great to have this for Joule's upcoming Loop support, to be able to see the status of your loops and possibly notify the user of their loop's status.

guggero:
Fixes #82.

This PR adds REST compatible RPC methods to get a list of all swaps (`GET /v1/loop/swaps`) or the status of a specific swap (`GET /v1/loop/swap/<swaphash>`).
These endpoints can be polled regularly to get synchronous updates.

The PR also adds a REST URL for the monitor streaming call (`GET /v1/loop/monitor`) which is a long-standing HTTP connection that returns one line of JSON for each update.

~Please be aware that the loop REST API is also affected by this missing feature: https://github.com/lightningnetwork/lnd/issues/2087
If a field is missing in the output of the above mentioned calls, then it means its value is falsey (`0, null, "", false, first enum value`).~

Edit: I found out that since we use a newer version of `grpc-gateway` than in `lnd`, we can instruct the marshaler to print the default values.


looprpc/client.proto
```diff
@@ -319,6 +341,23 @@ enum SwapState {
...
+    The list of all currently known swaps and their status.
+    */
+    repeated SwapStatus swaps = 1;
+}
+
+message GetSwapRequest {
+    /**
+    The swap identifier which currently is the hash that locks the HTLCs.
+    */
+    string id = 1;
```
Roasbeef: We should accept raw bytes here (the payment hash) rather than a string here. I get it makes it easier to use for REST (just pass the hex encoded version), but if we use our proto buf struct, then we'll ensure we display the payment hashes as hex all throughout.

loopd/swapclient_server.go
```diff
@@ -234,6 +235,42 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+// status.
+func (s *swapClientServer) GetSwaps(_ context.Context,
+	_ *looprpc.GetSwapsRequest) (*looprpc.GetSwapsResponse, error) {
+
+	var (
+		rpcSwaps = make([]*looprpc.SwapStatus, len(swaps))
+		idx      = 0
+		err      error
+	)
+	for _, swp := range swaps {
```
Roasbeef: Any particular reason you opted to use a new index rather than the index from the range loop here?

loopd/swapclient_server.go
```diff
@@ -234,6 +235,42 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+
+// GetSwap returns all known details about a single swap.
+func (s *swapClientServer) GetSwap(_ context.Context,
+	req *looprpc.GetSwapRequest) (*looprpc.SwapStatus, error) {
+
+	swapHash, err := lntypes.MakeHashFromStr(req.Id)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing swap hash: %v", err)
+	}
+	swp, ok := swaps[swapHash]
```
Roasbeef: 😞

I think this may be a good time to get rid of those global swap state. We should access the state instead directly from the swap client server, or just even read it from disk on-demand.

Roasbeef:
Tested the two new REST endpoints locally, works as advertised 👍

looprpc/client.proto
```diff
@@ -37,9 +37,31 @@ service SwapClient {
...
+        option (google.api.http) = {
+            get: "/v1/loop/monitor"
+        };
+    }
+
+    /**
+    GetSwaps returns a list of all currently known swaps and their current
+    status.
+    */
+    rpc GetSwaps (GetSwapsRequest) returns (GetSwapsResponse) {
```
halseth: `ListSwaps`?

looprpc/client.proto
```diff
@@ -37,9 +37,31 @@ service SwapClient {
...
+        option (google.api.http) = {
+            get: "/v1/loop/swaps"
+        };
+    }
+
+    /**
+    GetSwap returns all known details about a single swap.
+    */
+    rpc GetSwap (GetSwapRequest) returns (SwapStatus) {
```
halseth: `SwapInfo`? Not sure what is the idiomatic way of naming grpc service methods.

loopd/swapclient_server.go
halseth: I think it could be made a server level variable, and then we got get it by querying through the server.

looprpc/client.proto
guggero: I chose `string` because in the existing messages (`SwapResponse` and `SwapStatus`) the id field is already a hex encoded `string` field. I think we should stay consistent in the RPC and either use string for the new messages as well or switch everything to raw bytes (and break existing gRPC users).
I think base64 is quite ugly, as it's also not URL safe. So there's an additional step required for a REST user: `hex -> raw bytes -> base64 -> encode URI component`.

What I think we really should do for `lnd v0.10.0` (and consequentially for loop too) is to patch the `grpc-gateway` library to get rid of base64 completely.

loopd/swapclient_server.go
guggero: Yes, the `swaps` variable is a map.

loopd/swapclient_server.go
guggero: I removed the global swap state and went for serializing the state to the DB more often instead. This results in something like 3 updates per swap until it is completed, which is not too bad in my opinion.

loopd/daemon.go
```diff
@@ -161,13 +152,28 @@ func daemon(config *config, lisCfg *listenerCfg) error {
 		log.Infof("Waiting for updates")
 		for {
 			select {
-			case swap := <-statusChan:
+			case swapInfo := <-statusChan:
 				swapsLock.Lock()
-				swaps[swap.SwapHash] = swap
+
+				updateFn := swapClient.Store.UpdateLoopIn
```
halseth: I would sprinkle some comments here 🤪

loopd/swapclient_server.go
```diff
@@ -147,15 +147,20 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
 
 	log.Infof("Monitor request received")
 
-	send := func(info loop.SwapInfo) error {
-		rpcSwap, err := s.marshallSwap(&info)
+	send := func(info *loop.SwapInfo) error {
```
halseth: might have to do `swap := swap` in the loops where this is used now.

loopd/daemon.go
```diff
@@ -161,13 +152,28 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+				if swapInfo.SwapType == swap.TypeOut {
+					updateFn = swapClient.Store.UpdateLoopOut
+				}
+				log.Debugf("Updating swap %s on disk",
+					swapInfo.SwapHash)
+				err := updateFn(
```
halseth: hm, where/when did we persist before this change? 🤔

loopd/daemon.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/lightninglabs/loop"
 	"github.com/lightninglabs/loop/lndclient"
 	"github.com/lightninglabs/loop/looprpc"
+	"github.com/lightninglabs/loop/swap"
```
halseth: would be nice with an expanded explanation in the commit message  :)

loopd/daemon.go
```diff
@@ -87,7 +87,12 @@ func daemon(config *config, lisCfg *listenerCfg) error {
 	// through REST.
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	mux := proxy.NewServeMux()
+	mux := proxy.NewServeMux(proxy.WithMarshalerOption(
+		proxy.MIMEWildcard, &proxy.JSONPb{
```
halseth: adda comment to this option?

loopd/daemon.go
guggero: I thought that some intermediate states were not persisted as the daemon would go through the swap on startup and find out what the state was. But as it turns out, persisting all updates is a bad idea when I found this comment:
> If an unexpected error happened, report a temporary failure
> but don't persist the error. Otherwise for example a
> connection error could lead to abandoning the swap
> permanently and losing funds.

So I removed the whole persist logic here again. This probably means that the REST interface won't get the same granularity of swap states as the monitor stream. But I think this is less of a problem compared to unwanted side effects.

@joostjager could you please take a look at this anyway?

loopd/swapclient_server.go
guggero: I think we are ok, as we are passing in the loop variable as parameter instead of referencing it from the closure body.

loopd/swapclient_server.go
```diff
@@ -234,6 +235,49 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+		return nil, fmt.Errorf("error reading swaps: %v", err)
+	}
+	swapHash, err := lntypes.MakeHashFromStr(req.Id)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing swap hash: %v", err)
+	}
+	var swp *loop.SwapInfo
+	for _, dbSwap := range swaps {
+		if dbSwap.SwapHash == swapHash {
+			swp = dbSwap
```
joostjager: `return s.marshallSwap(swp)` here and below always return the error?

loopd/swapclient_server.go
```diff
@@ -234,6 +235,49 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+
+// SwapInfo returns all known details about a single swap.
+func (s *swapClientServer) SwapInfo(_ context.Context,
+	req *looprpc.SwapInfoRequest) (*looprpc.SwapStatus, error) {
+
+	swaps, err := s.impl.FetchSwaps()
+	if err != nil {
+		return nil, fmt.Errorf("error reading swaps: %v", err)
+	}
+	swapHash, err := lntypes.MakeHashFromStr(req.Id)
```
joostjager: Could move input validation above fetching the swaps.

looprpc/client.proto
```diff
@@ -37,9 +37,31 @@ service SwapClient {
...
+    rpc ListSwaps (ListSwapsRequest) returns (ListSwapsResponse) {
+        option (google.api.http) = {
+            get: "/v1/loop/swaps"
+        };
+    }
+
+    /**
+    SwapInfo returns all known details about a single swap.
+    */
+    rpc SwapInfo (SwapInfoRequest) returns (SwapStatus) {
```
joostjager: So we specifically want an endpoint that is going to be polled and not extend `monitor` to just return updates for a single swap?

loopd/swapclient_server.go
```diff
@@ -147,15 +147,20 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+	send := func(info *loop.SwapInfo) error {
+		rpcSwap, err := s.marshallSwap(info)
 		if err != nil {
 			return err
 		}
 
 		return server.Send(rpcSwap)
 	}
 
+	swaps, err := s.impl.FetchSwaps()
```
joostjager: Previously the in-memory swap snapshot was updated inside the same lock in which the notification was pushed out. This made sure that we didn't miss an update, or sent a duplicate. How is that situation now?

loopd/daemon.go
joostjager: Yes, that is indeed the idea. We don't store these temp failures in the database. It could just be a lost connection for example. But I do think it is useful to expose these to rpc users, because how do you otherwise find out that the swap (temporarily) failed?

looprpc/client.proto
guggero: Discussed offline that we should deprecate the `string id` and go for a `bytes` version instead.

loopd/swapclient_server.go
```diff
@@ -234,6 +235,49 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
 	}
 }
 
+// ListSwaps returns a list of all currently known swaps and their current
+// status.
+func (s *swapClientServer) ListSwaps(_ context.Context,
+	_ *looprpc.ListSwapsRequest) (*looprpc.ListSwapsResponse, error) {
+
+	swaps, err := s.impl.FetchSwaps()
```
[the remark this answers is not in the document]
guggero: Good point. It's still a couple of 100 thousand swaps probably with the default but setting the default max size as we use in lnd sounds reasonable to me.

looprpc/client.proto
guggero: Yes, because the monitor endpoint is streaming which is difficult to handle in REST. And it also doesn't return all older/completed swaps.

loopd/daemon.go
guggero: I see your point. But having this global state also doesn't feel too nice.
Maybe we should try to make REST users 1st class citizens by allowing them to use streaming RPCs as well with something like https://github.com/tmc/grpc-websocket-proxy .

loopd/swapclient_server.go
guggero: Good point. I moved the fetch into the lock as well.

loopd/daemon.go
joostjager: Does this have anything to do with streaming vs polling? Also for polling, they'd probably want to receive a temp. failure if it happened

loopd/swapclient_server.go
joostjager: Not moved?

loopd/swapclient_server.go
joostjager: Not fixed? Maybe something went wrong with rebase

loopd/swapclient_server.go
```diff
@@ -234,6 +235,45 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+
+// SwapInfo returns all known details about a single swap.
+func (s *swapClientServer) SwapInfo(_ context.Context,
+	req *looprpc.SwapInfoRequest) (*looprpc.SwapStatus, error) {
+
+	swapHash, err := lntypes.MakeHashFromStr(req.Id)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing swap hash: %v", err)
+	}
+	swaps, err := s.impl.FetchSwaps()
```
joostjager: Add newline for clarity

looprpc/client.proto
joostjager: Shouldn't this be `bytes` now then?

loopd/swapclient_server.go
```diff
@@ -165,11 +165,17 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
 	// prevent subscribers from receiving duplicate updates.
 	swapsLock.Lock()
 
+	swaps, err := s.impl.FetchSwaps()
```
joostjager: I don't think this fixes it, because updating the swap in the database isn't happening inside the lock. Just sending out the notifications is in.

loopd/daemon.go
joostjager: I think this regression is something to be sure we want to say yes to.

looprpc/client.proto
guggero: Yes, of course. Fixed.

guggero:
I somehow misunderstood @halseth's comment when I first read it, no idea why.
> I think it could be made a server level variable, and then we got get it by querying through the server.

I think moving the global variables into the `swapClientServer` solves both problems we have. It removes the global package state (which I agree is bad when we start using loop as a library) and we can still server the correct non-persisted state through the REST and gRPC API.
Sorry about the confusion.
@Roasbeef is this acceptable from your point of view also?

loopd/swapclient_server.go
guggero: Yes, I reverted the whole thing, see last comment.

loopd/daemon.go
guggero: Yes, you are right. I was too hung up on the "no global state" that I kind of missed the middle ground that solves both problems (see last comment).

Roasbeef:
> I don't know if moving from global to a singleton struct is such a difference, but I didn't have the impression that the global state wasn't that much of a problem anyway.

Even the existence of global variables can make the code harder to follow as the initialization and lifetime of the variables are unclear.

PR #143, opened 2020-02-04
lndclient: expose retrieval of channel backups through LightningClient

wpaulino:


wpaulino:
> What is the context for this change?

It'll be required for a project in the works.

> Maybe it is now time to move the lnd wrapper into the lnd repo or its own repo?

I'm not opposed to this, I don't think the `loop` repo is the best place to host this code, but ultimately that decision isn't up to me.

lndclient/lightning_client.go
```diff
@@ -384,3 +393,36 @@ func (s *lightningClient) ListTransactions(ctx context.Context) ([]*wire.MsgTx,
...
+
+	req := &lnrpc.ExportChannelBackupRequest{
+		ChanPoint: &lnrpc.ChannelPoint{
+			FundingTxid: &lnrpc.ChannelPoint_FundingTxidBytes{
+				FundingTxidBytes: channelPoint.Hash[:],
+			},
+			OutputIndex: channelPoint.Index,
+		},
+	}
+	resp, err := s.client.ExportChannelBackup(ctx, req)
```
joostjager: In the other calls, we created a per-call timeout context to time out just this request.

wpaulino: Good catch, fixed.

PR #145, opened 2020-02-04
lsat: add stream interceptor

guggero:
We add a new stream intercept method to our interceptor that can attach the LSAT to a stream as well as an unary call.

### DIFF
--- a/lsat/interceptor.go
+++ b/lsat/interceptor.go
@@ -145,6 +145,52 @@ func (i *Interceptor) UnaryInterceptor(ctx context.Context, method string,
 	return invoker(rpcCtx2, method, req, reply, cc, iCtx.opts...)
 }
 
+// StreamInterceptor is an interceptor method that can be used directly by gRPC
+// for streaming calls. If the store contains a token, it is attached as
+// credentials to every stream establishment call before patching it through.
+// The response error is also intercepted for every initial stream initiation.
+// If there is an error returned and it is indicating a payment challenge, a
+// token is acquired and paid for automatically. The original request is then
+// repeated back to the server, now with the new token attached.
+func (i *Interceptor) StreamInterceptor(ctx context.Context,
+	desc *grpc.StreamDesc, cc *grpc.ClientConn, method string,
+	streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream,
+	error) {
+
+	// To avoid paying for a token twice if two parallel requests are
+	// happening, we require an exclusive lock here.
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	// Create the context that we'll use to initiate the real request. This
+	// contains the means to extract response headers and possibly also an
+	// auth token, if we already have paid for one.
+	iCtx, err := i.newInterceptContext(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	// Try establishing the stream now. If anything goes wrong, we only
+	// handle the LSAT error message that comes in the form of a gRPC status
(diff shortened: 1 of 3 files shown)


lsat/interceptor.go
```diff
@@ -105,38 +114,72 @@ func (i *Interceptor) UnaryInterceptor(ctx context.Context, method string,
...
+	if err != nil {
+		return err
+	}
+
+	// Try executing the call now. If anything goes wrong, we only handle
+	// the LSAT error message that comes in the form of a gRPC status error.
+	rpcCtx, cancel := context.WithTimeout(ctx, i.callTimeout)
+	defer cancel()
+	err = invoker(rpcCtx, method, req, reply, cc, iCtx.opts...)
+	if isPaymentRequired(err) {
```
wpaulino: Nit: check `!isPaymentRequired` instead to save on indentation for the fragment below.

lsat/interceptor.go
```diff
@@ -144,6 +144,51 @@ func (i *Interceptor) UnaryInterceptor(ctx context.Context, method string,
...
+	iCtx, err := i.newInterceptContext(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	// Try establishing the stream now. If anything goes wrong, we only
+	// handle the LSAT error message that comes in the form of a gRPC status
+	// error. The context of a stream will be used for the whole lifetime of
+	// it, so we can't really clamp down on the initial call with a timeout.
+	stream, err := streamer(ctx, desc, cc, method, iCtx.opts...)
```
wpaulino: This came out simpler than I thought 👍

lsat/interceptor_test.go
```diff
@@ -240,6 +240,45 @@ func TestInterceptor(t *testing.T) {
...
+		testCases[i].callInterceptor = func() error {
+			return tc.interceptor.UnaryInterceptor(
+				ctx, "", nil, nil, nil, unaryInvoker, nil,
+			)
+		}
+	}
+
+	// Duplicate the test cases now and set the streaming interceptor for
+	// the second set.
+	doubleTestCases := append(testCases, testCases...)
```
wpaulino: I don't think the stream tests are being run in this diff, don't we need to range over them below?

Also, by taking this approach, we'll have duplicate names for the unary and stream interceptors. I think a cleaner approach would be to refactor all common operations within the tests into helpers and creating a separate top level test for the stream interceptors, such that we have `TestUnaryInterceptor` and `TestStreamInterceptor`. I wouldn't consider this a blocker though.

lsat/interceptor.go
guggero: Yes, good idea.

lsat/interceptor_test.go
guggero: Haha, yes. I thought it was too easy that the tests just worked.
I split the two tests to avoid something like that in the future.

lsat/interceptor_test.go
```diff
@@ -300,18 +305,51 @@ func TestInterceptor(t *testing.T) {
...
+}
+
+// TestStreamInterceptor tests that the interceptor can handle LSAT protocol
+// responses in streams and pay the token.
+func TestStreamInterceptor(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+	defer cancel()
+
+	// Run through the test cases.
+	for _, tc := range testCases {
```
wpaulino: Looks like we could refactor this further into another helper `testInterceptor` and invoke that here with the differences between and `TestUnaryInterceptor` and `TestStreamInterceptor`?

PR #152, opened 2020-02-11
multi: allow loop in last hop restriction

joostjager:
This PR adds the ability to loop in to specify a specific last hop for the channel top up payment. It thereby gives more control to users over where they want to increase their liquidity.

Example usage:

`loop in --last_hop 02e563181a29b47700e36f05c99ac9bb1a0b70abb22ea853285c0d05fba4b3616f 50000`

Note that it isn't possible to specify an exact channel. The reason for this is that nodes forward non-strictly and may have their reasons for picking a particular channel.


looprpc/client.proto
```diff
@@ -213,12 +213,10 @@ message LoopInRequest {
...
     /**
-    The channel to loop in. If zero, the channel to loop in is selected based
-    on the lowest routing fee for the swap payment from the server.
-
-    Note: NOT YET IMPLEMENTED
+    The last hop to use for the loop in swap. If empty, the last hop is selected
+    based on the lowest routing fee for the swap payment from the server.
     */
-    uint64 loop_in_channel = 4;
+    bytes last_hop = 4;
```
guggero: While we're at it, I think we should add the `last_hop` (and maybe the `external`) fields to the `SwapStatus` message. Otherwise a user will never be able to see this again through the command line or REST.

loopdb/migration_03_last_hop.go
```diff
@@ -0,0 +1,73 @@
...
+
+		// With the main swap bucket obtained, we'll grab the
+		// raw swap contract bytes.
+		contractBytes := swapBucket.Get(contractKey)
+		if contractBytes == nil {
+			return errors.New("contract not found")
+		}
+
+		const (
+			deprecatedLoopInChannelStart = 142
```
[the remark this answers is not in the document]
joostjager: added

looprpc/client.proto
joostjager: I looked into what this takes, but it isn't as easy as adding the fields in the proto and populating it. The quick fix would be to add loop-type-specific fields like these to `SwapInfo`. This struct however is only supposed to contain the common fields. Having a growing number of unused fields (unused in the other swap type) isn't good for maintainability long term. 

I think it is better to more thoroughly think through what direction we want to go with this and consider exposing the original swap parameters as a separate feature and pr.

guggero: Okay, that sounds good to me. There are other fields that are currently not exposed through the status calls so it makes sense to add them all at once in a follow-up PR.

loopdb/migration_03_last_hop.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		// We now set the new last hop field to all zeroes to indicate
+		// that there is no restriction.
+		var noLastHop [33]byte
+		if _, err := b.Write(noLastHop[:]); err != nil {
+			return err
+		}
+
+		// Append the remaining fields.
```
halseth: Are there any remaining fields here?

loopdb/migration_03_last_hop.go
```diff
@@ -0,0 +1,86 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/coreos/bbolt"
+)
+
+// migrateLastHop migrates the database to v03, replacing the never used loop in
+// channel by a last hop pubkey.
+func migrateLastHop(tx *bbolt.Tx, chainParams *chaincfg.Params) error {
```
halseth: nice no-dep migration :)

looprpc/server.proto
```diff
@@ -78,6 +78,7 @@ message ServerLoopInRequest {
     bytes swap_hash = 2;
     uint64 amt = 3;
     string swap_invoice = 4;
+    bytes last_hop = 5;
```
halseth: nit: server protos could be updated in its own commit

loopdb/migration_03_last_hop.go
joostjager: It is actually just one. Changed comment to mention it by name

loopdb/migration_03_last_hop.go
joostjager: Yes, definitely took care to make it no dep. We are wiser now

looprpc/server.proto
joostjager: As a reviewer, I never like proto updates in their own commit without context :)

PR #155, opened 2020-02-13
lndclient: expose ComputeInputScript in SignerClient

wpaulino:
This is required by an external project to populate the witness/signature script of an input belonging to the backing lnd wallet.


Roasbeef:
`ComputeInputScript` is really just a convenience method, we could leave it off the interace without any loss.

PR #156, opened 2020-02-14
looprpc: remove streaming REST endpoint, fix comment

guggero:
The REST version of the streaming monitoring call is not really user friendly because the current version of the `grpc-gateway` library doesn't support websockets and CORS isn't working. Until we have a solution to do proper websockets we should remove the endpoint in favor of the polling requests.


Roasbeef:
Needs a rebase.

guggero:
Rebased.

PR #158, opened 2020-02-21
Don't fail loop in quote if balance is insufficient for the miner fee estimation

### LINKED ISSUE #157: Loop In Quote Error 'insufficient funds available'
Loop v4.1

If the LND node does not have sufficient funds and UI tries to fetch the quotes from loop/in/quote, the API returns below error:
{
    "error": "insufficient funds available to construct transaction",
    "code": 2,
    "message": "insufficient funds available to construct transaction",
    "details": []
}

Though, we can send the max funds to fetch the quote, UI will be unable to showcase the true maximum capacity of loop-in. So, loop-in quote for maximun term should not have the validation for the onchain balance. Instead, the same validation can be applied while submitting the loop/in request

Sample UI which shows the min & max quote details before user enters the amount:
image

guggero:
Fixes #157.

The fee estimation in loop in works differently from loop out. Because a transaction has to be created that sends the exact amount of BTC to the HTLC, the fee estimation is done by lnd's wallet which tries to construct a real transaction. If not enough funds are available to construct such a sample transaction, the whole quote call would previously fail.

This PR partially fixes this by not failing the quote call but returning a fee of 0 instead. The command line will show a warning to the user in this case. Applications using the gRPC or REST interface should inform the user about the possible reasons if the miner fee shows 0.

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -59,6 +60,11 @@ var (
 	globalCallTimeout = serverRPCTimeout + lsat.PaymentTimeout
 
 	republishDelay = 10 * time.Second
+
+	// MinerFeeEstimationFailed is a magic number that is returned in a
+	// quote call as the miner fee if the fee estimation in lnd's wallet
+	// failed because of insufficient funds.
+	MinerFeeEstimationFailed btcutil.Amount = -1
 )
 
 // Client performs the client side part of swaps. This interface exists to be
@@ -505,10 +511,24 @@ func (s *Client) LoopInQuote(ctx context.Context,
 		}, nil
 	}
 
-	// Get estimate for miner fee.
+	// Get estimate for miner fee. If estimating the miner fee for the
+	// requested amount is not possible because lnd's wallet cannot
+	// construct a sample TX, we just return zero instead of failing the
+	// quote. The user interface should inform the user that fee estimation
+	// was not possible.
+	//
+	// TODO(guggero): Thread through error code from lnd to avoid string
+	// matching.
 	minerFee, err := s.lndServices.Client.EstimateFeeToP2WSH(
 		ctx, request.Amount, request.HtlcConfTarget,
 	)
+	if err != nil && strings.Contains(err.Error(), "insufficient funds") {
+		return &LoopInQuote{
+			SwapFee:   swapFee,
+			MinerFee:  MinerFeeEstimationFailed,
+			CltvDelta: quote.CltvDelta,
+		}, nil
+	}
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/loop/loopin.go
+++ b/cmd/loop/loopin.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcutil"
+	"github.com/lightninglabs/loop"
 	"github.com/lightninglabs/loop/looprpc"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightningnetwork/lnd/routing/route"
@@ -74,6 +75,18 @@ func loopIn(ctx *cli.Context) error {
 	)
 	if err != nil {
 		return err
+	}
+
+	// For loop in, the fee estimation is handed to lnd which tries to
+	// construct a real transaction to sample realistic fees to pay to the
+	// HTLC. If the wallet doesn't have enough funds to create this TX, we
+	// know it won't have enough to pay the real transaction either. It
+	// makes sense to abort the loop in this case.
+	if !external && quote.MinerFee == int64(loop.MinerFeeEstimationFailed) {
+		return fmt.Errorf("miner fee estimation not " +
+			"possible, lnd has insufficient funds to " +
+			"create a sample transaction for selected " +
+			"amount")
 	}
 
 	limits := getInLimits(amt, quote)
--- a/cmd/loop/quote.go
+++ b/cmd/loop/quote.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"context"
+	"fmt"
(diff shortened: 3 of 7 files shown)


cmd/loop/quote.go
```diff
@@ -31,11 +32,22 @@ var quoteCommand = cli.Command{
 				"setting this flag might result in a lower " +
 				"swap fee.",
 		},
+		cli.BoolFlag{
+			Name: "in",
+			Usage: "Indicate you want to get the quote for a " +
+				"loop in instead of loop out.",
+		},
```
joostjager: Should it be like this, or is it better (long term) to convert to something like `loop in --quote`?

client.go
```diff
@@ -505,10 +506,21 @@ func (s *Client) LoopInQuote(ctx context.Context,
...
-	// Get estimate for miner fee.
+	// Get estimate for miner fee. If estimating the miner fee for the
+	// requested amount is not possible because lnd's wallet cannot
+	// construct a sample TX, we just return zero instead of failing the
+	// quote. The user interface should inform the user that fee estimation
+	// was not possible.
 	minerFee, err := s.lndServices.Client.EstimateFeeToP2WSH(
 		ctx, request.Amount, request.HtlcConfTarget,
 	)
+	if err != nil && strings.Contains(err.Error(), "insufficient funds") {
```
joostjager: String matching? Is there no other possibility?

cmd/loop/quote.go
```diff
@@ -97,6 +98,17 @@ func quote(ctx *cli.Context) error {
 		return err
 	}
 
+	// For loop in, the fee estimation is handed to lnd which tries to
+	// construct a real transaction to sample realistic fees to pay to the
+	// HTLC. If the wallet doesn't have enough funds to create this TX, we
+	// don't want to fail the quote. But the user should still be informed
+	// why the fee shows as 0.
+	if in && quoteResp.MinerFee == 0 {
```
joostjager: Can it be an error code on the quote response?

cmd/loop/quote.go
guggero: I added the subcommands `in` and `out` to the `quote` command. Now it's quite similar to how the swaps themselves work.

client.go
guggero: I don't see any other possibility. We get this error over a gRPC boundary so we can't really type match anything. Also the error itself is not exported in `btcsuite/btcwallet/wallet/txauthor/author.go` so we can't even compare the message string directly.

cmd/loop/quote.go
guggero: I added a quote status.

looprpc/client.proto
```diff
@@ -443,6 +449,12 @@ message QuoteResponse {
     On-chain cltv expiry delta
     */
     int32 cltv_delta = 5;
+
+    /**
+    The status of the quote. If this status is not 0 (QUOTE_OK) then an
+    information message should be shown to the user.
+    */
+    Status status = 6;
```
joostjager: I know I suggested this myself, but seeing this now I am not sure anymore. Suppose there would be another field that has a problem, how can that be communicated through `status`? It would need to be a flagged enum. Maybe a magic value for `miner_fee` wasn't that bad. Instead of `0`, we could also have picked `-1` to make it really clear. Although `0` miner fee also should never happen.

client.go
joostjager: It would require an explicit error code to be added to `lnrpc.proto`. Or use one of the http status codes. That doesn't require a proto change.

cmd/loop/loopin.go
```diff
@@ -76,6 +76,18 @@ func loopIn(ctx *cli.Context) error {
...
+	// construct a real transaction to sample realistic fees to pay to the
+	// HTLC. If the wallet doesn't have enough funds to create this TX, we
+	// don't want to fail the quote. But the user should still be informed
+	// why the fee shows as 0.
+	if quote.Status == looprpc.QuoteResponse_MINER_FEE_ESTIMATION_FAILED {
+		fmt.Printf("Warning: Miner fee estimation not " +
+			"possible, lnd has insufficient funds to " +
+			"create a sample transaction for selected " +
+			"amount.\n")
+	}
```
joostjager: Does it make sense to proceed with the swap without funds?

client.go
joostjager: Discussed offline. Change requires lnd change, so if we want to move forward now, string matching is the only option.

joostjager: Could add a TODO there

guggero: Okay, added a TODO.

cmd/loop/quote.go
guggero: Removed the status again and went with fee of -1.

looprpc/client.proto
guggero: Ok, switched back to magic number -1.

cmd/loop/loopin.go
guggero: No, not really. Returning error now instead.

cmd/loop/quote.go
```diff
@@ -59,6 +62,18 @@ func quoteIn(ctx *cli.Context) error {
 		return err
 	}
 
+	// For loop in, the fee estimation is handed to lnd which tries to
+	// construct a real transaction to sample realistic fees to pay to the
+	// HTLC. If the wallet doesn't have enough funds to create this TX, we
+	// don't want to fail the quote. But the user should still be informed
+	// why the fee shows as -1.
+	if quoteResp.MinerFee == int64(loop.MinerFeeEstimationFailed) {
+		_, _ = fmt.Fprintf(os.Stderr, "Warning: Miner fee estimation "+
```
joostjager: Do we generally use `stderr` for such messages in the cli?

guggero: The idea behind this is that because the output is JSON, we would make it invalid if anyone is piping the output into a JSON parser. Then again, people should use the gRPC interface instead of parsing the command line output, so it's perhaps not worth the hassle.

PR #167, opened 2020-03-09
lsat: allow adding macaroons to insecure requests if proxied through tor

guggero:
Depends on #165, only last commit is new.

Fixes the issue of getting the error `cannot send secure credentials on an insecure connection` when trying to connect to the loop server using Tor for a call where an LSAT is required.

By default, authentication tokens in gRPC shouldn't be sent over insecure (non-TLS) communication channels. But when using Tor through a SOCKS proxy, the transport is encrypted even if it looks insecure to gRPC. We want to allow attaching the LSAT macaroon to the request if we're using a Tor proxy.

### DIFF
--- a/lsat/credential.go
+++ b/lsat/credential.go
@@ -1,0 +1,52 @@
+package lsat
+
+import (
+	"context"
+	"encoding/hex"
+
+	"gopkg.in/macaroon.v2"
+)
+
+// MacaroonCredential wraps a macaroon to implement the
+// credentials.PerRPCCredentials interface.
+type MacaroonCredential struct {
+	*macaroon.Macaroon
+
+	// AllowInsecure specifies if the macaroon is allowed to be sent over
+	// insecure transport channels. This should only ever be set to true if
+	// the insecure connection is proxied through tor and the destination
+	// address is an onion service.
+	AllowInsecure bool
+}
+
+// RequireTransportSecurity implements the PerRPCCredentials interface.
+func (m MacaroonCredential) RequireTransportSecurity() bool {
+	return !m.AllowInsecure
+}
+
+// GetRequestMetadata implements the PerRPCCredentials interface. This method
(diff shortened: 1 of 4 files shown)


lsat/credential.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// MacaroonCredential wraps a macaroon to implement the
+// credentials.PerRPCCredentials interface.
+type MacaroonCredential struct {
+	*macaroon.Macaroon
+
+	// AllowInsecure specifies if the macaroon is allowed to be sent over
+	// insecure transport channels. This should only ever be set to true if
+	// the insecure connection is proxied through tor.
```
wpaulino: Not just proxied through Tor, but the destination must be an onion service.

swap_server_client.go
```diff
@@ -58,10 +58,17 @@ func newSwapServerClient(address, proxyAddress string, insecure bool,
 	tlsPath string, lsatStore lsat.Store, lnd *lndclient.LndServices,
 	maxLSATCost, maxLSATFee btcutil.Amount) (*grpcSwapServerClient, error) {
 
+	// If a proxy address is set, we know tor is being used to secure the
+	// transport of our credentials (no other proxies are allowed
+	// currently). We need to set the flag accordingly, otherwise gRPC won't
+	// add the LSAT macaroon to the request.
+	allowInsecure := proxyAddress != ""
```
wpaulino: This change isn't safe yet until we include the swap server onion addresses. When proxying over Tor to a clearnet HTTP server, the transport is not encrypted.

guggero: Yes, you're right, forgot about that.

joostjager:
Can be rebased

guggero:
Rebased.

carlaKC:
Is this something we still want to get in?

guggero:
I don't think there has been any demand for this. Can always re-open later.

PR #173, opened 2020-04-01
loopd: add missing mutex unlock and fix subtle reace condition

bhandras:
This PR adds a missing mutex unlock upon return and fixes a subtle race condition when the subscriber may still write to the concurrent queue but the concurrent queue is closed, which can block the write.


loopd/swapclient_server.go
```diff
@@ -459,6 +459,7 @@ func (s *swapClientServer) processStatusUpdates(mainCtx context.Context) {
 				select {
 				case subscriber <- swp:
 				case <-mainCtx.Done():
+					s.swapsLock.Unlock()
```
Roasbeef, line 462: 👍

PR #174, opened 2020-04-06
test: update chainnotifier mock to fwd notifications for matching txs

bhandras:
This PR implements support for multiple tx notification subscribers in the
chain notifier mock. This is needed to make it possible to test scenarios
where we're watching the chain for different versions of the same transaction
(eg. NP2WSH and P2WSH)

### DIFF
--- a/test/chainnotifier_mock.go
+++ b/test/chainnotifier_mock.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"bytes"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Roasbeef:
Ignore the assignments, was testing out the native auto assign in Github now.

bhandras:
Sorry, for some reason I failed to amend the actual fix which is in this PR: https://github.com/lightninglabs/loop/pull/175/commits PTAL

PR #175, opened 2020-04-07
test: followup update chainnotifier mock to fwd notifications for matching txs

bhandras:
This is the remainder of the previous PR (https://github.com/lightninglabs/loop/pull/174) which was left out by my mistake. This is required to make it fully possible to test scenarios where we're watching the chain for different versions of the same transaction (eg. NP2WSH and P2WSH)

### DIFF
--- a/test/chainnotifier_mock.go
+++ b/test/chainnotifier_mock.go
@@ -12,8 +12,10 @@ import (
 )
 
 type mockChainNotifier struct {
-	lnd *LndMockServices
-	wg  sync.WaitGroup
+	sync.Mutex
+	lnd               *LndMockServices
+	confRegistrations []*ConfRegistration
+	wg                sync.WaitGroup
 }
 
 // SpendRegistration contains registration details.
@@ -29,6 +31,7 @@ type ConfRegistration struct {
 	PkScript   []byte
 	HeightHint int32
 	NumConfs   int32
+	ConfChan   chan *chainntnfs.TxConfirmation
 }
 
 func (c *mockChainNotifier) RegisterSpendNtfn(ctx context.Context,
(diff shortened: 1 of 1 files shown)


test/chainnotifier_mock.go
```diff
@@ -112,26 +123,25 @@ func (c *mockChainNotifier) RegisterConfirmationsNtfn(ctx context.Context,
...
 		select {
 		case m := <-c.lnd.ConfChannel:
-			if bytes.Equal(m.Tx.TxOut[0].PkScript, pkScript) {
-				select {
-				case confChan <- m:
-				case <-ctx.Done():
+			for _, r := range c.confRegistrations {
+				// Whichever conf notifier catches the confirmation
+				// will forward it to all matching subscibers.
+				if bytes.Equal(m.Tx.TxOut[0].PkScript, r.PkScript) {
```
joostjager, line 129: Remove registration. Or not?

test/chainnotifier_mock.go
```diff
@@ -103,7 +105,16 @@ func (c *mockChainNotifier) RegisterConfirmationsNtfn(ctx context.Context,
...
-	confChan := make(chan *chainntnfs.TxConfirmation, 1)
+	reg := &ConfRegistration{
+		PkScript:   pkScript,
+		TxID:       txid,
+		HeightHint: heightHint,
+		NumConfs:   numConfs,
+		ConfChan:   make(chan *chainntnfs.TxConfirmation, 1),
+	}
+
+	c.confRegistrations = append(c.confRegistrations, reg)
```
joostjager, line 116: Thread safety?

bhandras: Good catch, I guess I didn't notice this as there weren't concurrent registrations in our current tests, not even after adding native segwit.

test/chainnotifier_mock.go
bhandras: Also good suggestion. Pretty simple to remove so why not.

test/chainnotifier_mock.go
```diff
@@ -112,26 +126,35 @@ func (c *mockChainNotifier) RegisterConfirmationsNtfn(ctx context.Context,
...
+			c.Lock()
+			for i := 0; i < len(c.confRegistrations); i++ {
+				r := c.confRegistrations[i]
+
+				// Whichever conf notifier catches the confirmation
+				// will forward it to all matching subscibers.
+				if bytes.Equal(m.Tx.TxOut[0].PkScript, r.PkScript) {
+					// Unregister the "notifier".
+					c.confRegistrations = append(
+						c.confRegistrations[:i], c.confRegistrations[i+1:]...,
```
joostjager, line 138: Isn't `i+1` a crash if it is the last reg? Probably not as I see similar examples on internet

test/chainnotifier_mock.go
```diff
@@ -112,26 +126,35 @@ func (c *mockChainNotifier) RegisterConfirmationsNtfn(ctx context.Context,
...
+					)
+					i--
```
joostjager, line 140: I am thinking about a better way to go about this, but can't think of it. It works.

test/chainnotifier_mock.go
bhandras: it's valid (https://play.golang.org/p/TXP0B4ZtCNw)

test/chainnotifier_mock.go
bhandras: linked list could fit better

PR #180, opened 2020-04-16
cmd/loop+README: explain fees and optimize fee display

### LINKED ISSUE #176: units for on-chain fees off by a factor of 100
If I run `loop quote out` it says the `miner_fee` is `2440`. If I run `loop out` and choose `x` to expand fee detail, the `Max on-chain fee` is `244000`. What are the units for each of these fees, and why are they different? The other fees all match up between the two commands.

guggero:
Fixes #176.

Optimizes how the fees are displayed and called in the command line output and then documents them in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ USAGE:
 
 DESCRIPTION:
 
-  Attempts loop out the target amount into either the backing lnd's
+  Attempts to loop out the target amount into either the backing lnd's
   wallet, or a targeted address.
 
   The amount is to be specified in satoshis.
@@ -158,9 +158,12 @@ DESCRIPTION:
   specified. If not specified, a new wallet address will be generated.
 
 OPTIONS:
-   --channel value  the 8-byte compact channel ID of the channel to loop out (default: 0)
-   --addr value     the optional address that the looped out funds should be sent to, if left blank the funds will go to lnd's wallet
-   --amt value      the amount in satoshis to loop out (default: 0)
+   --channel value               the 8-byte compact channel ID of the channel to loop out (default: 0)
+   --addr value                  the optional address that the looped out funds should be sent to, if let blank the funds will go to lnd's wallet
+   --amt value                   the amount in satoshis to loop out (default: 0)
+   --conf_target value           the number of blocks from the swap initiation height that the on-chain HTLC should be swept within (default: 6)
+   --max_swap_routing_fee value  the max off-chain swap routing fee in satoshis, if not specified, a default max fee will be used (default: 0)
+   --fast                        Indicate you want to swap immediately, paying potentially a higher fee. If not set the swap server might choose to wait up to 30 minutes before publishing the swap HTLC on-chain, to save on its chain fees. Not setting this flag therefore might result in a lower swap fee.
 ```
 
 It's possible to receive more inbound capacity on a particular channel
@@ -177,6 +180,79 @@ This will take some time, as it requires an on-chain confirmation. When the
 swap is initiated successfully, `loopd` will see the process through.
 
 To query in-flight swap statuses, run `loop monitor`.
+
+### Fees explained
+
+The following is an example output of a 0.01 BTC fast (non-batched) Loop Out
+swap from `testnet`:
+
+```bash
+$ loop out --amt 1000000 --fast
+Max swap fees for 1000000 sat Loop Out: 36046 sat
+Fast swap requested.
+CONTINUE SWAP? (y/n), expand fee detail (x): x
+
+Estimated on-chain sweep fee:        149 sat
+Max on-chain sweep fee:              14900 sat
+Max off-chain swap routing fee:      20010 sat
+Max off-chain prepay routing fee:    36 sat
+Max no show penalty (prepay):        1337 sat
+Max swap fee:                        1100 sat
+CONTINUE SWAP? (y/n):
+```
+
+Explanation:
+
+- **Max swap fees for <x> sat Loop Out** (36046 sat): The absolute maximum in
+  fees that need to be paid. This includes on-chain and off-chain fees. This
+  represents the ceiling or worst-case scenario. The actual fees will likely be
+  lower. This is the sum of `14900 + 20010 + 36 + 1100` (see below).
+- **Estimated on-chain sweep fee** (149 sat): The estimated cost to sweep the
+  HTLC in case of success, calculated based on the _current_ on-chain fees.
+  This value is called `miner_fee` in the gRPC/REST responses.
+- **Max on-chain sweep fee** (14900 sat): The maximum on-chain fee the daemon
+  is going to allow for sweeping the HTLC in case of success. A fee estimation
+  based on the `--conf_target` flag is always performed before sweeping. The
+  factor of `100` times the estimated fee is applied in case the fees spike
+  between the time the swap is initiated and the time the HTLC can be swept. But
+  that is the absolute worst-case fee that will be paid. If there is no fee
+  spike, a normal, much lower fee will be used.
+- **Max off-chain swap routing fee** (20010 sat): The maximum off-chain
+  routing fee that the daemon should pay when finding a route to pay the
+  Lightning invoice. This is a hard limit. If no route with a lower or equal fee
+  is found, the payment (and the swap) is aborted. This value is calculated
+  statically based on the swap amount (see `maxRoutingFeeBase` and
+  `maxRoutingFeeRate` in `cmd/loop/main.go`).
+- **Max off-chain prepay routing fee** (36 sat): The maximum off-chain routing
+  fee that the daemon should pay when finding a route to pay the prepay fee.
+  This is a hard limit. If no route with a lower or equal fee is found, the
+  payment (and the swap) is aborted. This value is calculated statically based
+  on the prepay amount (see `maxRoutingFeeBase` and `maxRoutingFeeRate` in
+  `cmd/loop/main.go`).
+- **Max no show penalty (prepay)** (1337 sat): This is the amount that has to be
+  pre-paid (off-chain) before the server publishes the HTLC on-chain. This is
+  necessary to ensure the server's on-chain fees are paid if the client aborts
+  and never completes the swap _after_ the HTLC has been published on-chain.
+  If the swap completes normally, this amount is counted towards the full swap
+  amount and therefore is actually a pre-payment and not a fee. This value is
+  called `prepay_amt` in the gRPC/REST responses.
+- **Max swap fee** (1100 sat): The maximum amount of service fees we allow the
+  server to charge for executing the swap. The client aborts the swap if the
+  fee proposed by the server exceeds this maximum. It is therefore recommended
+  to obtain the maximum by asking the server for a quote first. The actual fees
+  might be lower than this maximum if user specific discounts are applied. This
+  value is called `swap_fee` in the gRPC/REST responses.
+
+#### Fast vs. batched swaps
+
+By default, Loop Outs are executed as normal speed swaps. This means the server
+will wait up to 30 minutes until it publishes the HTLC on-chain to improve the
+chances that it can be batched together with other user's swaps to reduce the
+on-chain footprint and fees. The server offers a reduced swap fee for slow swaps
+to incentivize users to batch more.
+
+If a swap should be executed immediately, the `--fast` flag can be used. Fast
+swaps won't benefit from a reduced swap fee.
 
 ### Loop In Swaps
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -158,9 +158,12 @@ DESCRIPTION:
...
-   --addr value     the optional address that the looped out funds should be sent to, if left blank the funds will go to lnd's wallet
-   --amt value      the amount in satoshis to loop out (default: 0)
+   --channel value               the 8-byte compact channel ID of the channel to loop out (default: 0)
+   --addr value                  the optional address that the looped out funds should be sent to, if left blank the funds will go to lnd's wallet
+   --amt value                   the amount in satoshis to loop out (default: 0)
+   --conf_target value           the number of blocks from the swap initiation height that the on-chain HTLC should be swept within (default: 6)
+   --max_swap_routing_fee value  the max off-chain swap routing fee in satoshis, if let blank a default max fee will be used (default: 0)
+   --fast                        Indicate you want to swap immediately, paying potentially a higher fee. If not set the swap server might choose to wait up to 30 minutes before publishing the swap HTLC on-chain, to save on chain fees. Not setting this flag might result in a lower swap fee.
```
alexbosworth, line 166: The 30 min may be dynamic, I might not mention it in a static comment

README.md
```diff
@@ -178,6 +181,72 @@ swap is initiated successfully, `loopd` will see the process through.
...
+  amount and therefore is actually a pre-payment and not a fee. This value is
+  called `prepay_amt` in the gRPC/REST responses.
+- **Max swap fee** (1'100 sat): The maximum amount of service fees the server is
+  going to charge for executing the swap. The actual fees might be lower if user
+  specific discounts are applied. This value is called `swap_fee` in the
+  gRPC/REST responses.
+
+#### Fast vs. batched swaps
+
+By default, Loop Outs are executed as slow swaps. This means the server will
```
alexbosworth, line 241: I would say they are normal speed swaps

alexbosworth:
How about moving to an estimate terminology instead of the current "maximum" terminology? I'd rather know what I am most likely to pay rather than the absolute maximum I might pay in the worst situation

README.md
guggero: This is already hard-coded in the description of the flag. This just represents the current output of the `loop out --help` command. I think we should change this text once the command itself doesn't have a static value anymore.

guggero:
The described "maximum" values are really that: Hard limits. The swap will fail if any of those are exceeded. It would be nice to go with an estimation model, but that would mean we would actually have to implement more estimations, which is not part of this PR.

alexbosworth:
> The described "maximum" values are really that: Hard limits. The swap will fail if any of those are exceeded. It would be nice to go with an estimation model, but that would mean we would actually have to implement more estimations, which is not part of this PR.

I just don't know if it's even worth it to show the hard limit of something like on-chain fee, it just seems like confusing information but sure it could be out of scope

README.md
```diff
@@ -178,6 +181,72 @@ swap is initiated successfully, `loopd` will see the process through.
...
+  fees that need to be paid. This includes on-chain and off-chain fees. This
+  represents the ceiling or worst-case scenario. The actual fees will likely be
+  lower. This is the sum of `14900 + 20010 + 36 + 1100` (see below).
+- **Estimated on-chain sweep fee** (149 sat): The estimated cost to sweep the
+  HTLC in case of success, calculated based on the _current_ on-chain fees.
+  This value is called `miner_fee` in the gRPC/REST responses.
+- **Max on-chain sweep fee** (14'900 sat): The maximum on-chain fee the daemon
+  is going to allow for sweeping the HTLC in case of success. The factor of
+  `100` to the estimated fee is applied in case the fees spike between the time
+  the swap is initiated and the time the HTLC can be swept.
```
joostjager, line 216: Maybe some explanation to comfort the user that it is unlikely that this is going to be paid.

README.md
```diff
@@ -178,6 +181,72 @@ swap is initiated successfully, `loopd` will see the process through.
...
+- **Max off-chain swap routing fee** (20'010 sat): The maximum off-chain
+  routing fee that the daemon should pay when finding a route to pay the
+  Lightning invoice. This is a hard limit. If no route with a lower or equal fee
+  is found, the payment is aborted. This value is calculated statically based
```
joostjager, line 220: And also the swap is aborted.

README.md
```diff
@@ -178,6 +181,72 @@ swap is initiated successfully, `loopd` will see the process through.
...
+  amount (base fee 10 sat, fee rate of 2%).
+- **Max no show penalty (prepay)** (1'337 sat): This is the amount that has to be
+  pre-paid (off-chain) before the server publishes the HTLC on-chain. This is
+  necessary to ensure the server's on-chain fees are paid if the client aborts
+  and never completes the swap _after_ the HTLC has been published on-chain.
+  If the swap completes normally, this amount is counted towards the full swap
+  amount and therefore is actually a pre-payment and not a fee. This value is
+  called `prepay_amt` in the gRPC/REST responses.
+- **Max swap fee** (1'100 sat): The maximum amount of service fees the server is
+  going to charge for executing the swap. The actual fees might be lower if user
```
joostjager, line 235: It is really the maximum swap fee proposed by the server that we are going to accept

README.md
```diff
@@ -178,6 +181,72 @@ swap is initiated successfully, `loopd` will see the process through.
...
+- **Max swap fees for <x> sat Loop Out** (36'046 sat): The absolute maximum in
+  fees that need to be paid. This includes on-chain and off-chain fees. This
+  represents the ceiling or worst-case scenario. The actual fees will likely be
+  lower. This is the sum of `14900 + 20010 + 36 + 1100` (see below).
+- **Estimated on-chain sweep fee** (149 sat): The estimated cost to sweep the
+  HTLC in case of success, calculated based on the _current_ on-chain fees.
+  This value is called `miner_fee` in the gRPC/REST responses.
+- **Max on-chain sweep fee** (14'900 sat): The maximum on-chain fee the daemon
+  is going to allow for sweeping the HTLC in case of success. The factor of
+  `100` to the estimated fee is applied in case the fees spike between the time
```
joostjager, line 215: to -> times?

README.md
```diff
@@ -178,6 +181,72 @@ swap is initiated successfully, `loopd` will see the process through.
...
+- **Max off-chain swap routing fee** (20'010 sat): The maximum off-chain
+  routing fee that the daemon should pay when finding a route to pay the
+  Lightning invoice. This is a hard limit. If no route with a lower or equal fee
+  is found, the payment is aborted. This value is calculated statically based
+  on the swap amount (base fee 10 sat, fee rate of 2%).
+- **Max off-chain prepay routing fee** (36 sat): The maximum off-chain routing
+  fee that the daemon should pay when finding a route to pay the prepay fee.
+  This is a hard limit. If no route with a lower or equal fee is found, the
+  payment is aborted. This value is calculated statically based on the prepay
+  amount (base fee 10 sat, fee rate of 2%).
```
joostjager, line 226: It is maintenance, to hard code these in the readme

cmd/loop/main.go
```diff
@@ -176,26 +176,30 @@ func displayLimits(swapType swap.Type, amt btcutil.Amount, l *limits,
...
 	case "x":
 		fmt.Println()
-		if swapType != swap.TypeIn || !externalHtlc {
-			fmt.Printf("Max on-chain fee:                 %d\n",
-				l.maxMinerFee)
+		f := "%-36s %d sat\n"
+		if swapType == swap.TypeOut {
+			fmt.Printf(f, "Estimated on-chain sweep fee:",
+				minerFees)
+			fmt.Printf(f, "Max on-chain sweep fee:", l.maxMinerFee)
```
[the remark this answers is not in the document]
joostjager: I guess no, because they specify a conf target

joostjager:
Maybe some comments don't really apply, because text is copied from comment flag documentation. But maybe some flags could be better described.

cmd/loop/main.go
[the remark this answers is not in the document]
guggero: The miner fee is still displayed in the next `if`. It is now just called differently (`on-chain HTLC fee` instead of just `on-chain fee`).

README.md
```diff
@@ -158,9 +158,12 @@ DESCRIPTION:
   specified. If not specified, a new wallet address will be generated.
 
 OPTIONS:
-   --channel value  the 8-byte compact channel ID of the channel to loop out (default: 0)
-   --addr value     the optional address that the looped out funds should be sent to, if left blank the funds will go to lnd's wallet
-   --amt value      the amount in satoshis to loop out (default: 0)
+   --channel value               the 8-byte compact channel ID of the channel to loop out (default: 0)
+   --addr value                  the optional address that the looped out funds should be sent to, if left blank the funds will go to lnd's wallet
+   --amt value                   the amount in satoshis to loop out (default: 0)
+   --conf_target value           the number of blocks from the swap initiation height that the on-chain HTLC should be swept within (default: 6)
```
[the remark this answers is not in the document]
guggero: No, this is the section for loop out. This only applies to the successful sweep.

cmd/loop/main.go
```diff
@@ -176,26 +176,30 @@ func displayLimits(swapType swap.Type, amt btcutil.Amount, l *limits,
 		return nil
 	case "x":
 		fmt.Println()
-		if swapType != swap.TypeIn || !externalHtlc {
-			fmt.Printf("Max on-chain fee:                 %d\n",
-				l.maxMinerFee)
+		f := "%-36s %d sat\n"
```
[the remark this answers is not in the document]
guggero: I like the idea but for these few lines it's probably overkill.

cmd/loop/main.go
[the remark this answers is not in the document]
joostjager: I meant the _maximum_ miner fee. But because there is no time pressure on the client side for loop in, there is no max.

cmd/loop/main.go
```diff
@@ -176,26 +176,30 @@ func displayLimits(swapType swap.Type, amt btcutil.Amount, l *limits,
...
+		if swapType == swap.TypeOut {
+			fmt.Printf(f, "Estimated on-chain sweep fee:",
+				minerFees)
+			fmt.Printf(f, "Max on-chain sweep fee:", l.maxMinerFee)
+		}
+		if swapType == swap.TypeIn && !externalHtlc {
```
joostjager, line 185: Could be a `switch` statement to prevent evaluation of this if when type is out.

README.md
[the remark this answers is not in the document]
joostjager: Ah I see it now

README.md
```diff
@@ -178,6 +181,77 @@ swap is initiated successfully, `loopd` will see the process through.
...
+Max off-chain swap routing fee:      20010 sat
+Max off-chain prepay routing fee:    36 sat
+Max no show penalty (prepay):        1337 sat
+Max swap fee:                        1100 sat
+CONTINUE SWAP? (y/n):
+```
+
+Explanation:
+
+- **Max swap fees for <x> sat Loop Out** (36'046 sat): The absolute maximum in
```
joostjager, line 206: What is the `'` there? Another notation :)

README.md
```diff
@@ -178,6 +181,77 @@ swap is initiated successfully, `loopd` will see the process through.
...
+- **Max off-chain swap routing fee** (20'010 sat): The maximum off-chain
+  routing fee that the daemon should pay when finding a route to pay the
+  Lightning invoice. This is a hard limit. If no route with a lower or equal fee
+  is found, the payment (and the swap) is aborted. This value is calculated
+  statically based on the swap amount (see `maxRoutingFeeBase` and
+  `maxRoutingFeeRate` in `cmd/loop/main.go`).
+- **Max off-chain prepay routing fee** (36 sat): The maximum off-chain routing
+  fee that the daemon should pay when finding a route to pay the prepay fee.
+  This is a hard limit. If no route with a lower or equal fee is found, the
+  payment is aborted. This value is calculated statically based on the prepay
```
joostjager, line 229: payment (and the swap)

README.md
```diff
@@ -178,6 +181,77 @@ swap is initiated successfully, `loopd` will see the process through.
...
+- **Max no show penalty (prepay)** (1'337 sat): This is the amount that has to be
+  pre-paid (off-chain) before the server publishes the HTLC on-chain. This is
+  necessary to ensure the server's on-chain fees are paid if the client aborts
+  and never completes the swap _after_ the HTLC has been published on-chain.
+  If the swap completes normally, this amount is counted towards the full swap
+  amount and therefore is actually a pre-payment and not a fee. This value is
+  called `prepay_amt` in the gRPC/REST responses.
+- **Max swap fee** (1'100 sat): The maximum amount of service fees we allow the
+  server to charge for executing the swap. This value must be obtained by asking
+  the server for a quote first. If this doesn't match the quoted fee, the server
```
joostjager, line 240: "The value is obtained via a quote" and then "if this (the value obtained via a quote) doesn't match the quoted fee" -> can never happen?

What happens is that the swap is aborted if the fee proposed by the server after initiating the swap exceeds the max

README.md
```diff
@@ -208,6 +283,32 @@ A Loop In swap can be executed a follows:
...
+```
+
+Explanation:
+
+- **Estimated on-chain HTLC fee** (154 sat): The estimated on-chain fee that the
+  daemon has to pay to publish the HTLC. This is an estimation from `lnd`'s
+  wallet based on the available UTXOs and current network fees. This value is
+  called `miner_fee` in the gRPC/REST responses.
+- **Max swap fee** (110 sat): The maximum amount of service fees the server is
+  going to charge for executing the swap. The actual fees might be lower if user
```
joostjager, line 308: Also here this is the maximum that we accept as a client. If the server proposes a higher amount, we don't publish the htlc and cancel the swap.

README.md
guggero: Ah, I forgot that other localizations use commas or other characters to separate thousands steps. Removed it.

PR #178, opened 2020-04-14
loopout: allow multi-loop

joostjager:
This PR moves over the loop out code to use `routerrpc` for sending payments. Only with `routerrpc` it is possible to make payments consisting of multiple shards.

The maximum number of shards is fixed to `5` in this PR. In a follow-up, this can be refined further.

Fixes #117

PR #181, opened 2020-04-17
Make loop aware of lnd's version

guggero:
`lnd` version `v0.10.0` introduced some breaking API changes to the `routerrpc`. To detect this and fail gracefully, we rely on the new `GetVersion` RPC method of the `verrpc` subserver of `lnd` and pin every `loop` version to a compatible `lnd` version from now on.

This PR also assumes that the next version of the`loop` client release will be `v0.6.0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,25 +57,36 @@ problems. Community support is also available in the
 LND and the loop client are using Go modules. Make sure that the `GO111MODULE`
 env variable is set to `on`.
 
-In order to execute a swap, **You need to run lnd 0.7.1+, or master built with
-sub-servers enabled.**
+In order to execute a swap, **you need to run a compatible lnd version built
+with the correct sub-servers enabled.**
 
 ### LND
 
-If you are building from source, and not using a 0.7.1 or higher release of
-lnd, make sure that you are using the `master` branch of lnd. You can get this
-by git cloning the repository
+To run loop, you need a compatible version of `lnd` running. It is generally
+recommended to always keep both `lnd` and `loop` updated to the most recent
+released version. If you need to run an older version of `lnd`, please consult
+the following table for supported versions.
+
+Loop Version | Compatible LND Version(s)
+------------------|------------------
+`>= v0.6.0-beta`  | `v0.10.x-beta`
+`<= 0.5.1-beta`   | `v0.7.1-beta` - `v0.10.x-beta`
+
+If you are building from source make sure you are using the latest tagged
+version of lnd. You can get this by git cloning the repository and checking out
+a specific tag:
 
 ```
 git clone https://github.com/lightningnetwork/lnd.git
+cd lnd
+git checkout v0.10.0-beta
 ```
 
 Once the lnd repository is cloned, it will need to be built with special build
 tags that enable the swap. This enables the required lnd rpc services.
 
 ```
-cd lnd
-make install tags="signrpc walletrpc chainrpc invoicesrpc routerrpc"
+make install tags="signrpc walletrpc chainrpc invoicesrpc"
 ```
 
 Check to see if you have already installed lnd. If you have, you will need to
--- a/client.go
+++ b/client.go
@@ -2,9 +2,7 @@ package loop
 
 import (
 	"context"
-	"encoding/hex"
 	"errors"
-	"fmt"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -211,13 +209,9 @@ func (s *Client) Run(ctx context.Context,
 	}
 
 	// Log connected node.
-	info, err := s.lndServices.Client.GetInfo(ctx)
-	if err != nil {
-		return fmt.Errorf("GetInfo error: %v", err)
-	}
-	log.Infof("Connected to lnd node %v with pubkey %v",
-		info.Alias, hex.EncodeToString(info.IdentityPubkey[:]),
-	)
+	log.Infof("Connected to lnd node '%v' with pubkey %x (version %s)",
+		s.lndServices.NodeAlias, s.lndServices.NodePubkey,
+		lndclient.VersionString(s.lndServices.Version))
 
 	// Setup main context used for cancelation.
 	mainCtx, mainCancel := context.WithCancel(ctx)
--- a/client_test.go
+++ b/client_test.go
@@ -35,14 +35,13 @@ var (
 	prepayInvoiceDesc = "prepay"
 )
 
-// TestSuccess tests the uncharge happy flow.
+// TestSuccess tests the loop out happy flow.
 func TestSuccess(t *testing.T) {
 	defer test.Guard(t)()
 
 	ctx := createClientTestContext(t, nil)
 
-	// Initiate uncharge.
-
+	// Initiate loop out.
 	hash, _, err := ctx.swapClient.LoopOut(context.Background(), testRequest)
 	if err != nil {
 		t.Fatal(err)
@@ -54,7 +53,7 @@ func TestSuccess(t *testing.T) {
 	signalSwapPaymentResult := ctx.AssertPaid(swapInvoiceDesc)
 	signalPrepaymentResult := ctx.AssertPaid(prepayInvoiceDesc)
 
-	// Expect client to register for conf
+	// Expect client to register for conf.
 	confIntent := ctx.AssertRegisterConf()
 
 	testSuccess(ctx, testRequest.Amount, *hash,
@@ -228,7 +227,7 @@ func testResume(t *testing.T, expired, preimageRevealed, expectSuccess bool) {
 	signalSwapPaymentResult := ctx.AssertPaid(swapInvoiceDesc)
 	signalPrepaymentResult := ctx.AssertPaid(prepayInvoiceDesc)
 
-	// Expect client to register for conf
+	// Expect client to register for conf.
 	confIntent := ctx.AssertRegisterConf()
 
 	signalSwapPaymentResult(nil)
--- a/lndclient/lnd_services.go
+++ b/lndclient/lnd_services.go
@@ -2,7 +2,6 @@ package lndclient
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net"
 	"path/filepath"
@@ -92,13 +91,6 @@ func NewLndServicesWithDialer(dialer dialerFunc, lndAddress, network,
 		}
 	}
 
-	// Now that we've ensured our macaroon directory is set properly, we
-	// can retrieve our full macaroon pouch from the directory.
-	macaroons, err := newMacaroonPouch(macaroonDir)
-	if err != nil {
-		return nil, fmt.Errorf("unable to obtain macaroons: %v", err)
-	}
-
 	// Setup connection with lnd
 	log.Infof("Creating lnd connection to %v", lndAddress)
 	conn, err := getClientConn(dialer, lndAddress, tlsPath)
@@ -113,24 +105,35 @@ func NewLndServicesWithDialer(dialer dialerFunc, lndAddress, network,
 		return nil, err
 	}
 
+	// We are going to check that the connected lnd is on the same network
+	// and is a compatible version with all the required subservers enabled.
+	// For this, we make two calls, both of which only need the readonly
+	// macaroon. We don't use the pouch yet because if not all subservers
+	// are enabled, then not all macaroons might be there and the user would
+	// get a more cryptic error message.
+	readonlyMac, err := newSerializedMacaroon(
+		filepath.Join(macaroonDir, defaultReadonlyFilename),
+	)
+	if err != nil {
+		return nil, err
+	}
+	err = checkLndCompatibility(conn, chainParams, readonlyMac, network)
+	if err != nil {
+		return nil, err
+	}
+
+	// Now that we've ensured our macaroon directory is set properly, we
+	// can retrieve our full macaroon pouch from the directory.
+	macaroons, err := newMacaroonPouch(macaroonDir)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain macaroons: %v", err)
+	}
+
+	// With the macaroons loaded and the version checked, we can now create
+	// the real lightning client which uses the admin macaroon.
 	lightningClient := newLightningClient(
 		conn, chainParams, macaroons.adminMac,
 	)
-
-	// With our macaroons obtained, we'll ensure that the network for lnd
-	// matches our expected network.
-	info, err := lightningClient.GetInfo(context.Background())
-	if err != nil {
-		conn.Close()
-		return nil, fmt.Errorf("unable to get info for lnd "+
-			"node: %v", err)
-	}
-	if network != info.Network {
-		conn.Close()
-		return nil, errors.New(
-			"network mismatch with connected lnd instance",
-		)
-	}
 
 	// With the network check passed, we'll now initialize the rest of the
 	// sub-server connections, giving each of them their specific macaroon.
@@ -142,7 +145,10 @@ func NewLndServicesWithDialer(dialer dialerFunc, lndAddress, network,
 
 	cleanup := func() {
 		log.Debugf("Closing lnd connection")
-		conn.Close()
+		err := conn.Close()
+		if err != nil {
+			log.Errorf("Error closing client connection: %v", err)
+		}
 
 		log.Debugf("Wait for client to finish")
 		lightningClient.WaitForFinished()
@@ -183,6 +189,32 @@ func (s *GrpcLndServices) Close() {
 	log.Debugf("Lnd services finished")
 }
 
+// checkLndCompatibility makes sure the connected lnd instance is running on the
+// correct network.
+func checkLndCompatibility(conn *grpc.ClientConn, chainParams *chaincfg.Params,
+	readonlyMac serializedMacaroon, network string) error {
+
+	// We use our own client with a readonly macaroon here, because we know
+	// that's all we need for the checks.
+	lightningClient := newLightningClient(conn, chainParams, readonlyMac)
+
+	// With our readonly macaroon obtained, we'll ensure that the network
+	// for lnd matches our expected network.
+	info, err := lightningClient.GetInfo(context.Background())
+	if err != nil {
+		conn.Close()
+		return fmt.Errorf("unable to get info for lnd "+
+			"node: %v", err)
+	}
+	if network != info.Network {
+		conn.Close()
+		return fmt.Errorf("network mismatch with connected lnd "+
+			"node, got '%s', wanted '%s'", info.Network, network)
+
+	}
+	return nil
+}
+
 var (
 	defaultRPCPort         = "10009"
 	defaultLndDir          = btcutil.AppDataDir("lnd", false)
@@ -199,6 +231,7 @@ var (
 	defaultWalletKitMacaroonFilename = "walletkit.macaroon"
 	defaultRouterMacaroonFilename    = "router.macaroon"
 	defaultSignerFilename            = "signer.macaroon"
+	defaultReadonlyFilename          = "readonly.macaroon"
 
 	// maxMsgRecvSize is the largest gRPC message our client will receive.
 	// We set this to 200MiB.
--- a/lndclient/lnd_services.go
+++ b/lndclient/lnd_services.go
@@ -2,6 +2,7 @@ package lndclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"path/filepath"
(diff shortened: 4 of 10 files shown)


README.md
```diff
@@ -57,14 +57,25 @@ problems. Community support is also available in the
...
+
+LND Version | Compatible Loop Version
+------------|------------------------
+`v0.10.x-beta`                 | `>= v0.6.0-beta`
+`v0.7.1-beta` - `v0.9.2-beta` | `v0.5.1-beta`
+`<= v0.7.0-beta`              | not supported for loop!
+
+
+If you are building from source make sure that you are using the `master` branch
+of lnd. You can get this by git cloning the repository:
```
carlaKC, line 78: I know this was already in the readme, but do we want to be telling people to run lnd on master? Wouldn't we want them to build the most recent tag?

guggero: You are right. I don't think we should advise people to run master. I changed the text.

lndclient/versioner_client.go
```diff
@@ -0,0 +1,41 @@
...
+// GetVersion returns the version and build information of the lnd
+// daemon.
+//
+// NOTE: This method is part of the VersionerClient interface.
+func (v *versionerClient) GetVersion(ctx context.Context) (*verrpc.Version,
+	error) {
+
+	rpcCtx, cancel := context.WithTimeout(ctx, rpcTimeout)
+	defer cancel()
+	return v.client.GetVersion(rpcCtx, &verrpc.VersionRequest{})
```
joostjager, line 40: I think that the macaroon needs to be attached here? I am getting

`[ERR] LOOPD: GetVersion error: rpc error: code = Unknown desc = expected 1 macaroon, got 0`

client.go
```diff
@@ -219,7 +243,27 @@ func (s *Client) Run(ctx context.Context,
...
+		// The version service has only been added in lnd v0.10.0. If
+		// we get an unimplemented error, it means the lnd version is
+		// definitely older than that.
+		s, ok := status.FromError(err)
+		if ok && s.Code() == codes.Unimplemented {
+			return ErrLndVersionIncompatible
+		}
+		return fmt.Errorf("GetVersion error: %v", err)
+	}
+	err = assertLndVersionCompatible(
```
joostjager, line 258: For loop it is also important to check the required sub servers

client_test.go
```diff
@@ -228,7 +357,7 @@ func testResume(t *testing.T, expired, preimageRevealed, expectSuccess bool) {
 	signalSwapPaymentResult := ctx.AssertPaid(swapInvoiceDesc)
 	signalPrepaymentResult := ctx.AssertPaid(prepayInvoiceDesc)
 
-	// Expect client to register for conf
+	// Expect client to register for conf.
```
joostjager, line 360: Nit: comment fixes can be separate commit. But only if you want to be a purist :)

joostjager:
#178 has been merged, this pr can be rebased

lndclient/versioner_client.go
guggero: Hmm... I seem to only have tested the negative case and missed this. Thanks for testing!

client.go
guggero: Yes, I forgot about that. Added the check.

client_test.go
guggero: Sure, purist sounds good :D

lndclient/versioner_client.go
joostjager: I tried out this, but not sure if it is better (non-blocking of course)
```
rpcCtx, cancel := context.WithTimeout(
	v.readonlyMac.WithMacaroonAuth(ctx), rpcTimeout,
)
```

client.go
```diff
@@ -219,7 +219,7 @@ func (s *Client) Run(ctx context.Context,
 		info.Alias, hex.EncodeToString(info.IdentityPubkey[:]),
 	)
 
-	// Setup main context used for cancelation.
+	// Setup main context used for cancellation.
```
joostjager, line 292: This is a hot debate. @cfromknecht, @carlaKC, ... ?

client.go
```diff
@@ -546,3 +593,35 @@ func (s *Client) LoopInTerms(ctx context.Context) (
 
 	return s.Server.GetLoopInTerms(ctx)
 }
+
+// assertLndVersionCompatible makes sure the detected lnd version is compatible
+// with our current version requirements.
+func assertLndVersionCompatible(actualVersion *verrpc.Version, expectedMajor,
```
joostjager, line 622: I can see this code being used elsewhere. Isn't this something to move into `LndServices`? There is already a `GetInfo` call being executed always.

client.go
```diff
@@ -88,6 +95,14 @@ var (
 	// quote call as the miner fee if the fee estimation in lnd's wallet
 	// failed because of insufficient funds.
 	MinerFeeEstimationFailed btcutil.Amount = -1
+
+	// LndRequiredActiveSubservers is a list of subservers that need to be
+	// activated in lnd for loop to work properly. These are still
+	// experimental and are only activated through a build flag. Therefore
+	// they might not be present if the user manually compiled.
+	LndRequiredActiveSubservers = []string{
+		"signrpc", "walletrpc", "chainrpc", "invoicesrpc",
```
joostjager, line 104: Maybe string constants? Probably more necessary if the version checking code is moved into a shared module like `lndclient`.

client.go
```diff
@@ -259,11 +274,19 @@ func (s *Client) Run(ctx context.Context,
 		return err
 	}
 
+	// Make sure all required subservers are active.
+	enabled := strings.Join(version.BuildTags, ",")
+	for _, required := range LndRequiredActiveSubservers {
+		if !strings.Contains(enabled, required) {
+			return ErrLndSubserversMissing
+		}
+	}
```
joostjager, line 283: Extract to method to be consistent with version checking?

README.md
```diff
@@ -57,25 +57,38 @@ problems. Community support is also available in the
 LND and the loop client are using Go modules. Make sure that the `GO111MODULE`
 env variable is set to `on`.
 
-In order to execute a swap, **You need to run lnd 0.7.1+, or master built with
-sub-servers enabled.**
+In order to execute a swap, **you need to run a compatible lnd version built
+with sub-servers enabled.**
```
joostjager, line 61: with _the correct_ sub-servers enabled?

README.md
```diff
@@ -57,25 +57,38 @@ problems. Community support is also available in the
...
+To run loop, you need a compatible version of `lnd` running. It is generally
+recommended to always keep both `lnd` and `loop` updated to the most recent
+released version. If you need to run an older version of `lnd`, please consult
+the following table for supported versions.
+
+LND Version | Compatible Loop Version
+------------|------------------------
+`v0.10.x-beta`                 | `>= v0.6.0-beta`
+`v0.7.1-beta` - `v0.9.2-beta` | `v0.5.1-beta`
+`<= v0.7.0-beta`              | not supported for loop!
```
joostjager, line 74: Nice

README.md
```diff
@@ -57,25 +57,38 @@ problems. Community support is also available in the
...
-by git cloning the repository
+To run loop, you need a compatible version of `lnd` running. It is generally
+recommended to always keep both `lnd` and `loop` updated to the most recent
+released version. If you need to run an older version of `lnd`, please consult
+the following table for supported versions.
+
+LND Version | Compatible Loop Version
+------------|------------------------
+`v0.10.x-beta`                 | `>= v0.6.0-beta`
+`v0.7.1-beta` - `v0.9.2-beta` | `v0.5.1-beta`
```
joostjager, line 73: I don't know whether it is useful to also specify a loop version range?

client.go
```diff
@@ -259,11 +274,19 @@ func (s *Client) Run(ctx context.Context,
 		return err
 	}
 
+	// Make sure all required subservers are active.
+	enabled := strings.Join(version.BuildTags, ",")
+	for _, required := range LndRequiredActiveSubservers {
+		if !strings.Contains(enabled, required) {
+			return ErrLndSubserversMissing
```
joostjager, line 281: If the user never ran with the right sub-servers, this code may never be hit and a macaroon not found error will terminate the app first. Maybe the version check should be executed before loading all the other macaroons. Perhaps it is a step too far for now.

client.go
carlaKC: The great debate rages on 🔥

README.md
guggero: I don't think so. Users should always use the most recent compatible version of loop.

client.go
guggero: The `versionrpc` uses the default `readonly.macaroon` which should always be there if the node starts correctly.

joostjager: But when we load `readonly.macaroon`, we also try to load all the other macaroons. This happens before we check the tags.

guggero: Ah, that's what you mean. Yeah, possibly. But I think that does exceed the scope of this PR.

joostjager: But it does make the checking of tags less useful. Especially new users that don't know about build tags will get a macaroon error. I think the whole version/tags check could be integrated nicely in `LndServices` where the `GetInfo` calls is made too (version is info too). At that point, you may be able to read the readonly macaroon outside of the pouch just for this.

guggero:
The suggestion to do the version check when creating a new client and to not use the macaroon pouch for that yet required a lot of changes. So this is almost a new PR at this point, sorry about that. But I am also much happier with how things look/work now.

lndclient/versioner_client.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// The actual version and expected version are identical.
+	return nil
+}
+
+// assertBuildTagsEnabled makes sure all required build tags are set.
+func assertBuildTagsEnabled(actualVersion *verrpc.Version,
+	requiredTags []string) error {
+
+	enabled := strings.Join(actualVersion.BuildTags, ",")
```
joostjager, line 170: Probably cleaner not to use a string for this, but store the actual tags in a map.

lndclient/lnd_services.go
```diff
@@ -12,20 +11,33 @@ import (
...
 	"google.golang.org/grpc/credentials"
 )
 
-var rpcTimeout = 30 * time.Second
+var (
+	rpcTimeout = 30 * time.Second
+
+	// minimalCompatibleVersion is the version of lnd we require at least
+	// when checking compatibility. This will be used as a fall back version
+	// if none is specified in the configuration.
```
joostjager, line 24: Maybe add that this is the minimum required version if you want to be able to use all functionality. When only a subset is required, a lower compatible version can be specified.

lndclient/lnd_services.go
```diff
@@ -39,6 +39,7 @@ type LndServices struct {
 	Signer        SignerClient
 	Invoices      InvoicesClient
 	Router        RouterClient
+	Versioner     VersionerClient
```
joostjager, line 54: Does it need to be exported? We can store the version info when we retrieve it on startup and expose that. It won't ever change, so there is no need to re-retrieve?

lndclient/versioner_client.go
```diff
@@ -0,0 +1,179 @@
...
+	return v.client.GetVersion(rpcCtx, &verrpc.VersionRequest{})
+}
+
+// CheckVersionCompatibility makes sure the connected lnd node has the correct
+// version and required build tags enabled.
+//
+// NOTE: This check will **never** return a non-nil error for a version of
+// lnd < 0.10.0 because any version previous to 0.10.0 doesn't have the version
+// endpoint implemented!
+func CheckVersionCompatibility(client VersionerClient, major, minor,
```
joostjager, line 73: Does this need to be exported? Is there a use case to check the version again after checking it on startup? If not exported, it can probably be moved near the startup code.

lndclient/versioner_client.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// All check positive, version is fully compatible.
+	return nil
+}
+
+// VersionString returns a nice, human readable string of a version returned by
+// the VersionerClient, including all build tags.
+func VersionString(version *verrpc.Version) string {
+	if version == nil {
+		return "<nil>"
```
joostjager, line 107: Or is this pre-0.10?

lndclient/lnd_services.go
```diff
@@ -12,20 +11,33 @@ import (
...
+	rpcTimeout = 30 * time.Second
+
+	// minimalCompatibleVersion is the version of lnd we require at least
+	// when checking compatibility. This will be used as a fall back version
+	// if none is specified in the configuration.
+	minimalCompatibleVersion = &verrpc.Version{
+		AppMajor: 0,
+		AppMinor: 10,
+		AppPatch: 0,
+	}
```
joostjager, line 29: Minimum required tags to use all functionality?

lndclient/lnd_services.go
```diff
@@ -173,6 +198,9 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
 			Router:        routerClient,
 			Versioner:     versionerClient,
 			ChainParams:   chainParams,
+			NodeAlias:     nodeAlias,
```
joostjager, line 201: This looks like an independent change that could be isolated in a commit?

lndclient/lnd_services.go
```diff
@@ -192,29 +220,74 @@ func (s *GrpcLndServices) Close() {
...
-	readonlyMac serializedMacaroon, network string) error {
+	readonlyMac serializedMacaroon, network string,
+	minVersion *verrpc.Version) (string, [33]byte, *verrpc.Version, error) {
 
-	lightningClient := newLightningClient(
-		conn, chainParams, readonlyMac,
-	)
+	// onErr is a closure that simplifies returning multiple values in the
+	// error case.
+	onErr := func(err error) (string, [33]byte, *verrpc.Version, error) {
```
joostjager, line 231: Instead of this, maybe `checkLndCompatibiltiy` (which now does more than just checking) can be an `init` method on `LndServices`, so that the retrieved data can be written into fields directly

lndclient/macaroon_pouch.go
```diff
@@ -104,5 +107,12 @@ func newMacaroonPouch(macaroonDir string) (*macaroonPouch, error) {
 		return nil, err
 	}
 
+	m.readonlyMac, err = newSerializedMacaroon(
+		filepath.Join(macaroonDir, defaultReadonlyFilename),
+	)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
guggero: Yes, for the `Versioner` client. I wanted the `Versioner` to behave like every other subserver client we have. Treating it differently than the others just because for now the returned information is static probably leads to more questions in the long run. And would need more explanatory comments everywhere.
For example, I plan to add an `Unlocked` status to the version RPC that will tell the user if the wallet is unlocked or not. This information will be very useful to have dynamically.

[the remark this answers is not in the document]
joostjager: Interesting, yes that status would be nice to have. I hope the versioner subserver is already up before wallet unlock?

lndclient/lnd_services.go
guggero: See comment above.

lndclient/versioner_client.go
guggero: I added this only because a test didn't specify a version in the mock. I fixed the test and removed the extra check.

lndclient/lnd_services.go
guggero: The `LndServices` is not instantiated here yet. I feel like I've refactored enough with this PR already.

lndclient/lnd_services.go
```diff
@@ -18,6 +18,19 @@ import (
 
 var rpcTimeout = 30 * time.Second
 
+// LndServicesConfig holds all configuration settings that are needed to connect
+// to an lnd node.
+type LndServicesConfig struct {
```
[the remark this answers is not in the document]
guggero: I would agree if more of these options had default values. But at the moment it's only the `Dialer` and `CheckVersion`.

lndclient/macaroon_pouch.go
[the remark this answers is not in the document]
guggero: I haven't started with the PR yet. But yes, the idea is to load selected subservers together with the walletunlocker.

lndclient/lnd_services.go
guggero: No, we load and store the info we get from `GetInfo` and `GetVersion` in this commit. Or what do you mean?

lndclient/lnd_services.go
```diff
@@ -113,25 +106,33 @@ func NewLndServicesWithDialer(dialer dialerFunc, lndAddress, network,
...
-		return nil, fmt.Errorf("unable to get info for lnd "+
-			"node: %v", err)
+		return nil, err
 	}
-	if network != info.Network {
-		conn.Close()
-		return nil, errors.New(
-			"network mismatch with connected lnd instance",
-		)
+	err = checkLndCompatibility(conn, chainParams, readonlyMac, network)
```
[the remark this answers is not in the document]
guggero: Oops, I forgot to add it in this commit because it's changed again in a later one. Fixed it.

lndclient/lnd_services.go
joostjager: Fair enough

test/versioner_mock.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var (
+	defaultMockBuildTags = []string{
+		"signrpc", "walletrpc", "chainrpc", "invoicesrpc",
+	}
+)
+
+type mockVersioner struct {
+	lnd     *LndMockServices
```
joostjager, line 28: unused?

lndclient/lnd_services_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Next, make sure an older version than what we want is rejected.
+	oldVersion := &mockVersioner{
+		version: &verrpc.Version{
+			AppMajor: 0,
+			AppMinor: 10,
+			AppPatch: 0,
+		},
+	}
+	err = checkVersionCompatibility(oldVersion, &verrpc.Version{
```
joostjager, line 48: This function is not found if running this commit.

lndclient/lnd_services.go
joostjager: I mean that the `GetInfo` call was already there, so the node alias and key were also already available. Before introducing the version call.

lndclient/lnd_services.go
```diff
@@ -198,28 +247,146 @@ func (s *GrpcLndServices) Close() {
...
-			"network mismatch with connected lnd instance",
-		)
+		return onErr(fmt.Errorf(
+			"network mismatch with connected lnd instance. "+
+				"wanted '%s', got '%s'", network, info.Network,
+		))
+	}
+
+	// Now let's also check the version of the connected lnd node.
+	err = checkVersionCompatibility(versionerClient, minVersion)
```
joostjager, line 291: `GetVersion` is called twice now. Could also pass the retrieved version into `checkVersionCompatibility`

lndclient/lnd_services.go
```diff
@@ -183,6 +246,150 @@ func (s *GrpcLndServices) Close() {
...
+	if err != nil {
+		return onErr(fmt.Errorf(
+			"unable to get version for lnd node: %v", err,
+		))
+	}
+
+	// Return the static part of the info we just queried from the node so
+	// it can be cached for later use.
+	return info.Alias, info.IdentityPubkey, version, nil
+}
```
joostjager, line 305: While testing, I get the following message on lnd 0.9:
`error checking connected lnd version. at least version "v0.10.0, build tags 'signrpc,walletrpc,chainrpc,invoicesrpc'" is required: version check not implemented, need minimum lnd version of v0.10.0-beta`

not the most concise message for novice users

lndclient/lnd_services.go
```diff
@@ -183,6 +190,32 @@ func (s *GrpcLndServices) Close() {
...
+	info, err := lightningClient.GetInfo(context.Background())
+	if err != nil {
+		conn.Close()
+		return fmt.Errorf("unable to get info for lnd "+
+			"node: %v", err)
+	}
+	if network != info.Network {
+		conn.Close()
+		return errors.New(
+			"network mismatch with connected lnd instance",
```
carlaKC: nit: might as well add network values?

lndclient/lnd_services.go
```diff
@@ -18,6 +18,19 @@ import (
 
 var rpcTimeout = 30 * time.Second
 
+// LndServicesConfig holds all configuration settings that are needed to connect
+// to an lnd node.
+type LndServicesConfig struct {
+	LndAddress  string
+	Network     string
+	MacaroonDir string
```
carlaKC: Godocs for these fields?

lndclient/lnd_services.go
guggero: Ok, I changed the error message a bit. It will now say: `lnd compatibility check failed: version check not implemented, need minimum lnd version of v0.10.0-beta`.

For the other errors (version mismatch or build tag missing), it will return the same, more detailed message: `error checking connected lnd version. at least version "v0.10.0, build tags 'signrpc,walletrpc,chainrpc,invoicesrpc'" is required`

lndclient/lnd_services.go
guggero: Okay, fair enough. I added an extra commit.

guggero:
I changed the compatibility matrix again, because we now don't break backwards compatibility with `lnd v0.10.0`.

lndclient/lnd_services.go
joostjager: Tested, looks good

lndclient/lnd_services.go
joostjager: Maybe it felt a bit over the top, but it is definitely nice now that I am re-reviewing

lndclient/lnd_services.go
```diff
@@ -214,25 +263,36 @@ func (s *GrpcLndServices) Close() {
...
 		if closeErr != nil {
 			log.Errorf("Error closing lnd connection: %v", closeErr)
 		}
 
-		return "", [33]byte{}, err
+		// Make static error messages a bit less cryptic by adding the
+		// version or build tag that we expect.
+		newErr := fmt.Errorf("lnd compatibility check failed: %v", err)
+		if err == ErrVersionIncompatible || err == ErrBuildTagsMissing {
+			newErr = fmt.Errorf("error checking connected lnd "+
```
joostjager, line 284: Now you could just as well return this error from `checkVersionCompatibility` directly? Maybe this just exists for the sake of the unit tests now (nb).

guggero: Yeah, but I didn't want to rewrite the unit tests and fall back to string matching there, so I left it. This would be a nice use case for the error wrapping, but I'm not sure if we already want to use that broadly?

joostjager: I haven't seen it yet. Let's merge this thing

PR #182, opened 2020-04-17
Slight build clarification

instagibbs:
output of `go install` is extremely sparse, was accidentally trying `go install ./..` and it seemed perfectly happy aside from not outputting any artifacts in the expected place

PRing in case it's not just me :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ install the command line interface and swap client service.
 ```
 git clone https://github.com/lightninglabs/loop.git
 cd loop/cmd
-go install ./...
+go install ./... # Note this is a `...` wildcard
 ```
 
 ## Execute a Swap


Roasbeef:
If you tack on a `-v`, then you'll see which packages n stuff are being compiled. If you want even _more_ log output, then you can use `-x` which shows the actually _commands_ that are being executed.

README.md
```diff
@@ -105,7 +105,7 @@ install the command line interface and swap client service.
 ```
 git clone https://github.com/lightninglabs/loop.git
 cd loop/cmd
-go install ./...
+go install ./... # Note this is a `...` wildcard
```
joostjager, line 108: >was accidentally trying go install ./.. and it seemed perfectly happy aside from not outputting any artifacts in the expected place

Does this comment make that part more clear?

instagibbs: would have saved me a few minutes yeah, up to you if it's worth it. Feel free to close, just a personal anecdote.

joostjager: What i try to understand is how that comment that you add in this pr would've saved you the minutes? I would expect the comment to either describe the install location or describe the verbose flags.

PR #183, opened 2020-04-22
lndclient: move to V2 payment methods

joostjager:
Otherwise mpp isn't supported.


joostjager:
I tested this on regtest and works with mpp

PR #185, opened 2020-04-27
loopd: make maximum number of payment parts configurable

joostjager:
As we are still in the experimentation phase with MPP, it is probably safer to expose this setting to the user. They can set it to `1` in case they don't want multi-loops.

A further step is to make `max_parts` a swap parameter that is passed in through the client rpc interface. This however also requires a database change. It therefore seems to be wiser to await how the current implementation of MPP will be received first.

### DIFF
--- a/client.go
+++ b/client.go
@@ -81,12 +81,38 @@ type Client struct {
 	clientConfig
 }
 
-// NewClient returns a new instance to initiate swaps with.
-func NewClient(dbDir string, serverAddress, proxyAddress string, insecure bool,
-	tlsPathServer string, lnd *lndclient.LndServices, maxLSATCost,
-	maxLSATFee btcutil.Amount) (*Client, func(), error) {
+// ClientConfig is the exported configuration structure that is required to
+// instantiate the loop client.
+type ClientConfig struct {
+	// ServerAddress is the loop server to connect to.
+	ServerAddress string
 
-	store, err := loopdb.NewBoltSwapStore(dbDir, lnd.ChainParams)
+	// ProxyAddress is the SOCKS proxy that should be used to establish the
+	// connection.
+	ProxyAddress string
+
+	// Insecure skips TLS when set.
+	Insecure bool
+
+	// TLSPathServer is the path to the TLS certificate that is required to
+	// connect to the server.
+	TLSPathServer string
+
+	// Lnd is an instance of the lnd proxy.
+	Lnd *lndclient.LndServices
+
+	// MaxLsatCost is the maximum price we are willing to pay to the server
+	// for the token.
+	MaxLsatCost btcutil.Amount
+
+	// MaxLsatFee is the maximum that we are willing to pay in routing fees
+	// to obtain the token.
+	MaxLsatFee btcutil.Amount
+}
+
+// NewClient returns a new instance to initiate swaps with.
+func NewClient(dbDir string, cfg *ClientConfig) (*Client, func(), error) {
+	store, err := loopdb.NewBoltSwapStore(dbDir, cfg.Lnd.ChainParams)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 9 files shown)


client.go
```diff
@@ -81,30 +81,60 @@ type Client struct {
...
+	ServerAddress string
+
+	// ProxyAddress is the SOCKS proxy that should be used to establish the
+	// connection.
+	ProxyAddress string
+
+	// Insecure skips TLS when set.
+	Insecure bool
+
+	// TlsPathServer is the path to the TLS certificate that is required to
```
bhandras, line 100: nit: TlsPathServer => TLSPathServer

client.go
```diff
@@ -81,30 +81,60 @@ type Client struct {
...
 
-// NewClient returns a new instance to initiate swaps with.
-func NewClient(dbDir string, serverAddress, proxyAddress string, insecure bool,
-	tlsPathServer string, lnd *lndclient.LndServices, maxLSATCost,
-	maxLSATFee btcutil.Amount) (*Client, func(), error) {
+// ClientConfig is the exported configuration structure that is required to
+// instantiate the loop client.
+type ClientConfig struct {
+	// DbDir is the directory where the loopd database resides.
+	DbDir string
```
bhandras, line 88: nit: DBDir?

loopd/utils.go
```diff
@@ -10,19 +10,26 @@ import (
...
 	if err != nil {
 		return nil, nil, err
 	}
 
-	swapClient, cleanUp, err := loop.NewClient(
-		storeDir, config.SwapServer, config.Proxy, config.Insecure,
-		config.TLSPathSwapSrv, lnd, btcutil.Amount(config.MaxLSATCost),
-		btcutil.Amount(config.MaxLSATFee),
-	)
+	cfg := &loop.ClientConfig{
```
bhandras, line 21: nit: maybe better to call this `clientConfig` and leave the other `config` or `cfg` as that fits better with the type name.

client.go
```diff
@@ -111,6 +111,10 @@ type ClientConfig struct {
 	// MaxLSATFee is the maximum that we are willing to pay in routing fees
 	// to obtain the token.
 	MaxLSATFee btcutil.Amount
+
+	// LoopOutMaxParts defines that maximum number of part that may be used
```
bhandras, line 115: nit: "that maximum number of part" => "the maximum number of parts"

bhandras: Maybe it'd be also good to indicate that "parts" refer to multipart payment.

client.go
joostjager: I don't know. I don't like those abbreviations to stand out so much. Fixed the TLS one because of the linter, but this one I am getting no linter complaints about. I guess I don't like all those capitals. `RPCServer` vs `RpcServer`. I prefer the latter because it doesn't attract attention needlessly when glancing over the code.

joostjager: It don't think that abbreviated things deserve more attention just because they're abbreviated.

client.go
joostjager: fixed

loopd/utils.go
joostjager: fixed

client.go
joostjager: clarified

lndclient/router_client.go
```diff
@@ -96,7 +96,7 @@ type SendPaymentRequest struct {
 
 	// The maximum number of partial payments that may be used to complete
 	// the full amount.
-	MaxShards uint32
+	MaxParts uint32
```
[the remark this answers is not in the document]
joostjager: fixed

client.go
joostjager: Decapitalized.

client.go
```diff
@@ -81,30 +81,60 @@ type Client struct {
...
-	lsatStore, err := lsat.NewFileStore(dbDir)
+	lsatStore, err := lsat.NewFileStore(cfg.DbDir)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	swapServerClient, err := newSwapServerClient(
-		serverAddress, proxyAddress, insecure, tlsPathServer, lsatStore,
-		lnd, maxLSATCost, maxLSATFee,
+		cfg.ServerAddress, cfg.ProxyAddress, cfg.Insecure,
```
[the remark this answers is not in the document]
joostjager: Nice suggestion. I only excluded `dbDir`, because it is none of `Client`'s business.

loopd/config.go
```diff
@@ -57,16 +59,17 @@ const (
...
+	RPCListen:       "localhost:11010",
+	RESTListen:      "localhost:8081",
+	Insecure:        false,
+	LogDir:          defaultLogDir,
+	MaxLogFiles:     defaultMaxLogFiles,
+	MaxLogFileSize:  defaultMaxLogFileSize,
+	DebugLevel:      defaultLogLevel,
+	MaxLSATCost:     lsat.DefaultMaxCostSats,
+	MaxLSATFee:      lsat.DefaultMaxRoutingFeeSats,
+	LoopOutMaxParts: 5,
```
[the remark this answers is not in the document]
joostjager: fixed

PR #186, opened 2020-04-27
multi: report protocol version

joostjager:
To keep supporting older clients, it is required for the server to know the protocol version that a client adheres to. This PR adds that version as a field to the loop in and out initiation calls.


looprpc/server.proto
```diff
@@ -19,6 +19,11 @@ service SwapServer {
         returns (ServerLoopInQuoteResponse);
 }
 
+enum ProtocolVersion {
```
bhandras, line 22: I'd add some comments to help readers understand what these enums mean.

looprpc/server.proto
```diff
@@ -80,6 +88,9 @@ message ServerLoopInRequest {
     uint64 amt = 3;
     string swap_invoice = 4;
     bytes last_hop = 5;
+
+    /// The protocol version that the client adheres to.
+    ProtocolVersion protocol_version = 6;
```
bhandras, line 93: Question is, should we version all RPC calls? Or just loops?
Should we use server reflection instead? (https://github.com/grpc/grpc-go/tree/master/examples/features/reflection)

looprpc/server.proto
joostjager: Added

looprpc/server.proto
joostjager: Can you get a version number with server reflection?

What is versioned here is just the protocol version (scripts etc), not the RPC interface. For that, we'd need to have a separate endpoint like `GetVersion` in `lnd`.

looprpc/server.proto
```diff
@@ -19,6 +19,20 @@ service SwapServer {
...
+this is not a flagged enum. If a particular protocol version adds a feature,
+then in general all the preceding features are also supported. Exception to this
+is when features get deprecated.
+*/
+enum ProtocolVersion {
+    /// No protocol version reported at all.
+    LEGACY = 0;
+
+    /// Client reports its supported protocol version.
+    REPORT_VERSION = 1;
```
carlaKC, line 33: So this enum will be extended to have additional protocol versions? If that's the case the name `REPORT_VERSION` and its comment don't make sense to me - makes me think we're going to provide a specific version value somewhere.

Also, how are these protocol versions tracked? By adding an enum to this proto?
Thinking about the case of a non-loop client (if we're accounting for that), how do I know what version I'm supporting?

looprpc/server.proto
joostjager: Going to look into the grpc context, maybe that can carry the version number.

joostjager: Discussed with @guggero and grpc metadata doesn't seem a great idea. It is all string based (no nice enum) and also proxies may change the keys :fearful:

joostjager: Added version to all RPC calls.

bhandras: Server reflection may be used for protocol discovery, which would be useful for clients to check if their protocol description is compatible with the server protocol description. Afaik context may be used to store a version number (it's similar to http headers) but that also is somewhat hidden (as in not reflected in the proto definitions).

looprpc/server.proto
bhandras: I can agree with @carlaKC here, the naming is a bit confusing. Perhaps instead of calling it `ProtocolVersion` we could call itt `SwapProtocol` (if I understand the purpose correctly), and have better naming for the individual protocol identifiers.

joostjager: I think that the confusion is that `REPORT_VERSION` itself is a feature that a client can support? 

The idea is indeed that with every protocol change we do on the server, we add a value to the enum. Then the next client release will start sending that enum value.

joostjager: Loop is all about swaps, so `SwapProtocol` doesn't add more information than just `Protocol` in my opinion.

Suggestions for what to rename `REPORT_VERSION` to?

looprpc/server.proto
joostjager: The point is that not every aspect of the protocol (which for example also includes the on-chain htlc bitcoin script), can be observed from the proto signatures.

Using the grpc metadata in the context seems not very attractive (see comment above)

looprpc/server.proto
bhandras: I think a better question is how will we use `REPORT_VERSION`? Isn't `LEGACY` the only version we use now?

joostjager: Can also call it `MULTI_LOOP_OUT`?

bhandras: That is much better :) I'd be happy with that.

joostjager: Updated

PR #190, opened 2020-04-30
lndclient: add preimage to invoice subscription and lookup invoice

carlaKC:
Add invoice preimage to our current subscription (field nil for non-settle updates) so that subscribing clients can immediately react to the preimage without needing an additional lookup. We don't strictly need this with `LookupInvoice` added, but I think it's nice to have the preimage on hand. 

Add `LookupInvoice` to lndclient to allow race-safe invoice state checks.
This is required in addition to adding the preimage to subscription to cover the following case:
- We settle an invoice in lnd
- lnd dispatches an update for the invoice 
- we take action on our currently known state after settling but before receiving the update

This is prevented by getting a consistent view fromlnd's db with the lookup.

### DIFF
--- a/lndclient/invoices_client.go
+++ b/lndclient/invoices_client.go
@@ -30,6 +30,10 @@ type InvoicesClient interface {
 type InvoiceUpdate struct {
 	State   channeldb.ContractState
 	AmtPaid btcutil.Amount
+
+	// Preimage is the preimage for the invoice that has just been
+	// paid. This field will be nil for updates that are not settles.
+	Preimage *lntypes.Preimage
 }
 
 type invoicesClient struct {
@@ -111,11 +115,26 @@ func (s *invoicesClient) SubscribeSingleInvoice(ctx context.Context,
 				return
 			}
 
-			select {
-			case updateChan <- InvoiceUpdate{
+			update := InvoiceUpdate{
 				State:   state,
 				AmtPaid: btcutil.Amount(invoice.AmtPaidSat),
-			}:
+			}
+
+			// If the invoice is settled, we expect the preimage
+			// to be present so we include it.
+			if state == channeldb.ContractSettled {
+				preimage, err := lntypes.MakePreimage(
+					invoice.RPreimage,
+				)
+				if err != nil {
+					errChan <- err
+					return
+				}
+				update.Preimage = &preimage
+			}
+
+			select {
+			case updateChan <- update:
 			case <-ctx.Done():
 				return
 			}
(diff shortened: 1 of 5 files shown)


lndclient/invoices_client.go
```diff
@@ -111,10 +115,17 @@ func (s *invoicesClient) SubscribeSingleInvoice(ctx context.Context,
 				return
 			}
 
+			preimage, err := lntypes.MakePreimage(invoice.RPreimage)
```
joostjager, line 118: Doesn't this return an error when the preimage is still unknown?

loopin_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/lightningnetwork/lnd/lntypes"
+
```
joostjager, line 8: \n

loopin_test.go
```diff
@@ -231,7 +239,8 @@ func testLoopInTimeout(t *testing.T, externalValue int64) {
 
 	// Signal the the invoice was canceled.
 	subscription.Update <- lndclient.InvoiceUpdate{
-		State: channeldb.ContractCanceled,
+		State:    channeldb.ContractCanceled,
+		Preimage: nil,
```
joostjager, line 243: Can be omitted

loopin_test.go
```diff
@@ -82,11 +84,17 @@ func TestLoopInSuccess(t *testing.T) {
...
+	}
+
 	// Server has already paid invoice before spending the htlc. Signal
 	// settled.
 	subscription.Update <- lndclient.InvoiceUpdate{
-		State:   channeldb.ContractSettled,
-		AmtPaid: 49000,
+		State:    channeldb.ContractSettled,
+		AmtPaid:  49000,
+		Preimage: &typedPreimage,
```
joostjager, line 97: Do we update this test already now as a preparation for when we implement fast preimage reveal to the client in the future?

lndclient/invoices_client.go
bhandras: I guess we need to check for `ContractSettled` here?

loopin_test.go
carlaKC: I'm just always in favor of updating tests with what we expect in real life. Otherwise you get test failures down the line because a field isn't set somewhere. 

Fields aren't needed in this PR, but personally preferred.

carlaKC:
This change isn't required for my change anymore, closing for now.

PR #191, opened 2020-05-04
golangci: update linter reference point and add comment

carlaKC:
Update linter reference point to the 0.6 tag's commit to fix linter failure due to travis not being able to find the reference point commit in its shallow clone. 

Alternate solution is to set travis to deep clone as done in lnd.


carlaKC:
Closing in favor of #192

PR #193, opened 2020-05-05
loopout: push preimage to server to speed up off chain claim

carlaKC:
This PR adds a best-effort preimage push to the loop out client to speed up off-chain claim on htlcs, thus acquiring outgoing liquidity more quickly. 

The preimage is disclosed to the server after the on chain htlc has confirmed and after the client has published a sweep to claim the htlc on chain with the preimage. The preimage is **already** disclosed through the sweep publish, so this push does not change the security model of the client.


looprpc/server.proto
```diff
@@ -137,3 +140,9 @@ message ServerLoopInTerms {
     uint64 min_swap_amount = 1;
     uint64 max_swap_amount = 2;
 }
+
+message ServerLoopOutPushPreimageRequest {
+    bytes preimage = 1;
+}
+
+message ServerLoopOutPushPreimageResponse {}
```
joostjager, line 148: As a reviewer, I like the proto squashed with the commit that uses it

loopout.go
```diff
@@ -666,6 +666,28 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+	// Track whether we have successfully pushed the preimage to the server
+	// so that we do not needlessly continue pushing preimages.
+	var preimagePushed bool
+
+	// pushPreimage pushes our preimage to the server directly if we
+	// have already disclosed our preimage in a sweep and we have not yet
+	// successfully pushed a preimage to the server. This trustlessly speeds
+	// up our swap.
+	pushPreimage := func() {
+		if !preimagePushed && s.state == loopdb.StatePreimageRevealed {
```
joostjager, line 678: Rewrite if to return early: `if preimagePush || state != Revealed { return }`

loopout.go
```diff
@@ -693,6 +715,10 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
 				return nil, err
 			}
 
+			// If we successfully published a spend, we push the
+			// preimage to the server.
```
joostjager, line 719: So we do it every block. Sounds okay, but is there another reason that this is a good interval besides that it is easy to implement?

loopout.go
joostjager, line 678: Probably good to discuss with @alexbosworth the preferred timing for disclosure. Before or after the first tx publication.

alexbosworth: I don't have a strong preference, but I would say after if pressed

Ideally it will be making no difference since we won't depend on the server ACK

carlaKC: Will stick with after. 

> Ideally it will be making no difference since we won't depend on the server ACK

Going to have the client lookup the invoice with lnd, so there's no dependence on the server 👍

loopout.go
carlaKC: If we are only going to reveal our preimage after we have revealed it through a sweep, after a sweep attempt is the only time we need to re-evaluate whether to push again? I tried it with its own ticker, but then we just have to wait for the `timerChan` to tick first, followed by the preimge push ticker (which just makes htlc claim slower) 

There is the off chance that the sever fails, but I don't think we should form our logic based on the server maybe being down? And if it is, every block is a reasonable enough random retry delay.

joostjager: Yes it is also worth something to have only a single timer

joostjager: Seems that the current version of the pr does have the separate timer?

loopout.go
```diff
@@ -706,7 +706,77 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+		// If our payment is complete, there is no need to proceed in
+		// attempting to push the preimage.
+		if paymentComplete {
+			return nil
+		}
+
+		// If we have not yet revealed our preimage through a sweep, we
+		// do not push the preimage because we may choose to never sweep
+		// if fees are too high.
+		if s.state != loopdb.StatePreimageRevealed {
```
joostjager, line 720: By the time we get here, haven't we always revealed already?

loopout.go
```diff
@@ -706,7 +706,77 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+		}
+
+		// Create a child context for a one time lookup of our payment.
+		// We defer cancelling it so that lnd will terminate the stream.
+		trackCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		// Lookup our payment in lnd before we push to the server, this
+		// allows us to exit early if the payment is already settled.
+		statusChan, errChan, err := s.lnd.Router.TrackPayment(
```
joostjager, line 731: Alternatively this call can be done once before we enter the republish loop. We just keep listening to the same status channel.

loopout.go
carlaKC: Yeah I changed my mind after testing and seeing the double wait, should have marked as do not review yet, sorry!

loopout.go
```diff
@@ -706,7 +706,77 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+		case status := <-statusChan:
+			if status.State == lnrpc.Payment_SUCCEEDED {
+				s.log.Infof("Payment succeeded, no need to " +
+					"push preimage")
+
+				paymentComplete = true
+				return nil
+			}
+
+			// If our payment has not succeeded, we fallthrough
```
bhandras, line 753: Could it be that the payment is still in flight? Should we wait it out in that case?

loopout.go
carlaKC: We might have called our sweep function but not actually decided to sweep because fees are too high.

joostjager: Ah, that is also what the comment says indeed.

And what if we persist the revealed state and crash before we actually publish (reveal)? On restart we are going to disclose the unpublished preimage to the server?

loopout.go
carlaKC: Since the hodl htlc is locked in, I would expect the payment to be in flight at this stage?

loopout.go
carlaKC: I don't think it matters because our current sweep logic will take note of the reveal and publish at all costs? 
If we restart with preimage revealed state, the sweeper will always publish our sweep with the max fee. 

There is the case where the sweep still doesn't get in (maybe high fees, maybe bitcoind borked) but I think that should be addressed by more aggressive client sweeping? We also won't sweep after preimage push (not that that should be relied on).

joostjager: Suppose we have no more backend after the crash?

carlaKC: Then the off chain htlc will be pulled and the on chain will not be claimed. 
But I think this falls into the category of "exceptional errors" that we can't really account for. In that case the swapper would need to manually sweep their output. 

We have no way to guarantee publish/mempool propagation other than to just keep trying to sweep, which we already do.

joostjager: Yes strong point. Publish isn't a guarantee either indeed.

carlaKC: Then the off chain htlc will be pulled and the on chain will not be claimed. 
But I think this falls into the category of "exceptional errors" that we can't really account for. In that case the swapper would need to manually sweep their output. 

We have no way to guarantee publish/mempool propagation other than to just keep trying to sweep, which we already do.

looprpc/server.proto
```diff
@@ -143,3 +146,9 @@ message ServerLoopInTerms {
     uint64 min_swap_amount = 1;
     uint64 max_swap_amount = 2;
 }
+
+message ServerLoopOutPushPreimageRequest {
+    bytes preimage = 1;
```
joostjager, line 151: Add some comments, even though we don't yet have comments elsewhere?

swap_server_client.go
```diff
@@ -215,6 +218,19 @@ func (s *grpcSwapServerClient) NewLoopOutSwap(ctx context.Context,
 	}, nil
 }
 
+// PushLoopOutPreimage pushes a preimage to the server.
+func (s *grpcSwapServerClient) PushLoopOutPreimage(ctx context.Context,
+	preimage lntypes.Preimage) error {
+
+	_, err := s.server.LoopOutPushPreimage(ctx,
```
joostjager, line 225: Create timeout ctx?

test/context.go
```diff
@@ -94,6 +94,24 @@ func (ctx *Context) AssertRegisterSpendNtfn(script []byte) {
...
+// returns the track payment message so that it can be used to send updates
+// to the test.
+func (ctx *Context) AssertTrackPayment() TrackPaymentMessage {
+	ctx.T.Helper()
+
+	var msg TrackPaymentMessage
+	select {
+	case msg = <-ctx.Lnd.TrackPaymentChannel:
+
+	case <-time.After(Timeout):
```
joostjager, line 107: I think we generally use `defer timeout()()`, so this may not be necessary. Same for timeouts below.

loopout.go
joostjager: Thinking about this again, isn't this inconsistent? So the happy flow is:
1. Mark revealed in db
2. Reveal in sweep
3. Reveal to server endpoint

However if we crash right after 1 and restart, the order becomes:
1. Mark revealed in db
2. Reveal to server endpoint
3. Reveal in sweep

To minimize variation, isn't it better to always stick to the second sequence?

One other thing we haven't discussed is failure to publish the tx because of a bug in loop and/or wallet code. In that case, the first sequence is safer. To always do the first sequence, I think we'd need two states: `PreimageRevealPublish` and  `PreimageRevealed`. Can also be useful in certain debug scenarios perhaps.

lndclient/router_client.go
```diff
@@ -226,17 +231,27 @@ func (r *routerClient) trackPayment(ctx context.Context,
...
+
+				// If the stream has exited on the server side,
+				// which may be the case when the payment has
+				// reached a final state, we return a stream
+				// closed error so that callers can handle this
+				// case specifically.
+				case code == codes.Unknown &&
+					err.Error() == "EOF":
+
+					err = ErrStreamComplete
```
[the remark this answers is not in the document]
carlaKC: If we close both channels, we can't keep listening on the error channel in a select once the stream is done, select on a closed channel will always trigger. I think it's preferable to leave the behavior as is, and add a special case so that we can identify this case when we want to.

loopout.go
```diff
@@ -706,6 +706,50 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
 		return nil, fmt.Errorf("register spend ntfn: %v", err)
 	}
 
+	// Create a child context for a one time lookup of our payment.
+	// We defer cancelling it so that lnd will terminate the stream.
```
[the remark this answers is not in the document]
carlaKC: Yes? We want to lnd that we no longer need this subscription once this function exits? Otherwise we'll have a bunch of unused track payment subscriptions in lnd.

lndclient/router_client.go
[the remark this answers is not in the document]
joostjager: Ok good point about bigger selects. And how about not sending anything at all? The success or failure result is already the EOF marker really.

loopout.go
carlaKC: How would the second sequence you list ever happen? If we mark preimage as revealed and then crash, we'll call `sweep` with our state set to `PreimageRevealed` on restart. We will broadcast a sweep tx again before we push preimage to server?

loopout.go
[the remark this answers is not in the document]
joostjager: I mean, is `lnd` the entity that terminates the stream? I think our client is still terminating the stream at the end of the function?

loopout.go
joostjager: I was assuming the push preimage closure was ran even before the main loop, but ofc that is just the definition of the closure. Resolved

loopout_test.go
```diff
@@ -376,3 +396,187 @@ func TestCustomSweepConfTarget(t *testing.T) {
...
+	expiryChan <- time.Now()
+
+	// Expect a signing request for the HTLC success transaction.
+	<-ctx.Lnd.SignOutputRawChannel
+
+	// This is the first time we have swept, so we expect our preimage
+	// revealed state to be set.
+	cfg.store.(*storeMock).assertLoopOutState(loopdb.StatePreimageRevealed)
+	status := <-statusChan
+	if status.State != loopdb.StatePreimageRevealed {
```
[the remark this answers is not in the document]
carlaKC: Our state enums are not actually of type `SwapState` (can't use require.equal) because we don't use iota/set type here:
https://github.com/lightninglabs/loop/blob/master/loopdb/swapstate.go#L22

This is the case in a few places, so I think that should be addressed in its own PR? Might have some db implications since the type is a uint and the consts are ints.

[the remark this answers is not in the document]
joostjager: Also not `require.Equal` with with a cast? I agree that changing the data type itself is not something for this pr.

loopout.go
```diff
@@ -719,6 +763,24 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
 		case err := <-spendErr:
 			return nil, err
 
+		// Receive status updates for our payment so that we can detect
+		// whether we've successfully pushed our preimage.
+		case status := <-statusChan:
+			if status.State == lnrpc.Payment_SUCCEEDED {
+				s.log.Infof("Payment succeeded, preimage " +
+					"push successful")
```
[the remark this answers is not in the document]
carlaKC: Indeed, comment updated!

loopout.go
carlaKC: Moving this out of a closure into its own function so it reads a bit better as well.

loopout_test.go
```diff
@@ -376,3 +396,187 @@ func TestCustomSweepConfTarget(t *testing.T) {
...
+
+	// Assert that we made a query to track our payment, as required for
+	// preimage push tracking.
+	trackPayment := ctx.AssertTrackPayment()
+
+	// Tick the expiry channel, we are still using our client confirmation
+	// target at this stage which has fees higher than our max acceptable
+	// fee. We do not expect a sweep attempt at this point. Since our
+	// preimage is not revealed, we also do not expect a preimage push.
+	expiryChan <- time.Now()
```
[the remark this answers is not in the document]
carlaKC: Nope, just a lazy copy, updated to a const.

lndclient/router_client.go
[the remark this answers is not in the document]
carlaKC: That would be a behaviour change for other calls using this endpoint (although I don't know how much it matters, since the update would always arrive first). Could also move the `EOF` check into our loop which has this specific use case and leave the client untouched?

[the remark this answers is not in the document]
joostjager: Behavior change is also a valid point. It just feels a bit strange to always end the happy flow with an error. Close channel  to me is the most natural way to map the grpc stream close still. Another alternative for bigger selects is to set the status channel to nil once you've read a close, so that the event loop can keep going.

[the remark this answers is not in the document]
carlaKC: > It just feels a bit strange to always end the happy flow with an error

This is already the case with the EOF if you keep listening after settle? This just makes it an error that we specifically match. Thinking more about this I think we should just catch it in the caller, since this is a very specific case. 

> set the status channel to nil once you've read a close

This feels like it would be racey? We replace the channel with nil at the same time we try to read from it? It won't be an issue in practice, but I suspect it will trigger unit test race detection.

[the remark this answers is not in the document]
joostjager: >This is already the case with the EOF if you keep listening after settle?

If the channel is closed, you will at least get a `!ok` in `update, ok := <- updateChan`. It maps better to go channels than a custom err?

> We replace the channel with nil at the same time we try to read from it?

I mean replace the channel with `nil` in the caller select. So:
```
case update, ok := <- updateChan:
   if !ok {
      updateChan = nil
      continue
   }
````

File: test/context.go (hunk 9 above)
Comment (carlaKC) [reply test/context.go:107]: Following the assert pattern in this file, I don't think it hurts to have timeouts here in case we don't defer timeout somewhere.

File: lndclient/router_client.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (carlaKC) [reply lndclient/router_client.go:254]: Ok yeah nice, I thought you meant in the client itself.

File: lndclient/router_client.go [hunk 15]
```diff
@@ -226,6 +229,17 @@ func (r *routerClient) trackPayment(ctx context.Context,
...
+				// If we get an EOF error, the payment has
+				// reached a final state and the server is
+				// finished sending us updates. We close both
+				// channels to signal that we are done sending
+				// values on them and return.
+				if err == io.EOF {
+					close(statusChan)
+					close(errorChan)
+					return
+				}
```
joostjager, line 241: Passing back your own question: are changes needed anywhere we use `trackPayment` because of this?

loopout_test.go
```diff
@@ -376,3 +396,190 @@ func TestCustomSweepConfTarget(t *testing.T) {
...
+	require.Equal(
+		t, status.State, loopdb.SwapState(loopdb.StatePreimageRevealed),
+	)
+
+	// We expect the sweep tx to have been published.
+	ctx.ReceiveTx()
+
+	// Once we have published an on chain sweep, we expect a preimage to
+	// have been pushed to the server after the sweep.
+	<-server.preimagePush
```
joostjager, line 542: Assert that it is the correct preimage?

loopout_test.go
```diff
@@ -376,3 +396,190 @@ func TestCustomSweepConfTarget(t *testing.T) {
...
+// Finally, we make a last sweep attempt, push the preimage (because we have
+// not detected our settle) and settle the off chain htlc, indicating that the
+// server successfully settled using the preimage push. In this test, we need
+// to start with a fee rate that will be too high, then progress to an
+// acceptable one. We do this by starting with a high confirmation target with
+// a high fee, and setting the default confirmation fee (which our swap will
+// drop down to if it is not confirming in time) to a lower fee. This is not
+// intuitive (lower confs having lower fees), but it allows up to mock fee
+// changes.
+func TestPreimagePush(t *testing.T) {
```
joostjager, line 414: Isn't this just a `Success` test, as preimage push is part of the happy flow?

test/context.go
joostjager: No, it doesn't hurt. Maybe something to revisit in the test refactor pr. Or not.

loopout.go
```diff
@@ -719,6 +733,45 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+		// Receive status updates for our payment so that we can detect
+		// whether we've successfully pushed our preimage.
+		case status, ok := <-trackChan:
+			// If our channel has been closed, indicating that the
+			// server is finished providing updates because the
+			// payment has reached a terminal state, we replace
+			// the closed channel with nil so that we will no longer
+			// listen on it.
+			if !ok {
+				trackChan = nil
```
joostjager, line 745: Do we need to check here that we actually reached a terminal state?

loopout.go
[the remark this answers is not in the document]
joostjager: Ok, reusing the existing cancel context.

lndclient/router_client.go
carlaKC: Did a quick audit of our usage of  `TrackPayment` and `trackPayment` (called by `SendPayment`):
`lsat/interceptor.trackPayment`: exits on success/timeout, ok
`waitForHtlcSpendConfirmed`: this change, accounts for channel close
`payInvoiceAsync`: exits on success/timeout, ok

Also went through the other projects that use `lndclient` and they all follow the pattern of exiting once the payment is complete.

loopout_test.go
carlaKC: In `TestSuccess` we push preimage once and then our payment is settled. This tests the case where we push and the payment is not settled, so we push again. It also fails to sweep on first try becuase fees are too high, whereas our success tests always sweep first thing. I can cut it, but I think it adds a few nice coverage improvements.

loopout.go
carlaKC: I don't think so, grpc only sends an EOF when the stream exits cleanly. Checking state here is checking for lnd trackpayments bugs in loop, I think it's a safe assumption to make.

looprpc/server.proto
```diff
@@ -9,6 +9,9 @@ service SwapServer {
 
     rpc NewLoopOutSwap (ServerLoopOutRequest) returns (ServerLoopOutResponse);
 
+    rpc LoopOutPushPreimage (ServerLoopOutPushPreimageRequest)
```
joostjager, line 12: Transfer a fresh copy of the proto from the server pr once that is merged

carlaKC:
Pushed updated protos and nit, holding off on merge for now while waiting for server.

PR #199, opened 2020-05-12
version: bump version to 0.6.1-beta

joostjager:


joostjager:
Yes it is unfortunate. Apparently we can't modify the tag anymore once pushed, so thinking what to do now...

PR #200, opened 2020-05-12
version: bump to v0.6.2-beta

joostjager:
We can't remove the previous tag, so need to bump again unfortunately.

Before next release, we need to improve the release scripts to add checks for all these things like we recently did for the lnd release script.


joostjager:
Yes

PR #201, opened 2020-05-13
release: add tag and version verification to release script

joostjager:
To prevent some common human errors when building a release. Changes taken from lnd release script, not invented here.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -4,9 +4,11 @@
 # we support with the golang cross-compiler.
 #
 # Copyright (c) 2016 Company 0, LLC.
-# Copyright (c) 2019 Lightning Labs
 # Use of this source code is governed by the ISC
 # license.
+
+# Exit on errors.
+set -e
 
 # If no tag specified, use date + version otherwise use tag.
 if [[ $1x = x ]]; then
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -4,17 +4,62 @@
...
+    # Use a regex to isolate the version string.
+    LOOP_VERSION_REGEX="version (.+) "
+    if [[ $LOOP_VERSION_OUTPUT =~ $LOOP_VERSION_REGEX ]]; then
+        # Prepend 'v' to match git tag naming scheme.
+        LOOP_VERSION="v${BASH_REMATCH[1]}"
+        echo "version: $LOOP_VERSION"
+
+        # If tag contains a release candidate suffix, append this suffix to the
+        # loop reported version before we compare.
+        RC_REGEX="-rc[0-9]+$"
```
guggero, line 49: I'm not sure we should do this. Maybe it would be better if we declare that the release manager or product owner is allowed to create and merge a PR to add the RC suffix without reviews. Then that merge commit would also get the tag.
And anyway, the RC tag should use a dot and not a hyphen (`vX.Y.Z-beta.rcU`).

joostjager: Removed. We don't do RCs (yet) anyway

guggero: Ah, I just noticed this was copied from `lnd`. Perhaps we should remove it there too or at least fix the dash vs dot issue.

release.sh
```diff
@@ -4,17 +4,55 @@
...
 if [[ $1x = x ]]; then
     DATE=`date +%Y%m%d`
     VERSION="01"
     TAG=$DATE-$VERSION
 else
     TAG=$1
+
+    # If a tag is specified, ensure that that tag is present and checked out.
+    if [[ $TAG != $(git describe) ]]; then
+        echo "tag $TAG not checked out"
```
bhandras, line 24: nit: s/not/is not/
same below

release.sh
```diff
@@ -4,17 +4,55 @@
...
+    fi
+
+    # Verify that it is signed.
+    if ! git verify-tag $TAG; then 
+        echo "tag $TAG not signed"
+        exit 1
+    fi
+
+    # Build loop to extract version.
+    make build
```
bhandras, line 35: i wonder if it's a good idea to also make the `build` target the default in our make file. Kind of weird that we need to always do a `make build`

release.sh
```diff
@@ -4,17 +4,55 @@
 # we support with the golang cross-compiler.
 #
 # Copyright (c) 2016 Company 0, LLC.
-# Copyright (c) 2019 Lightning Labs
+# Copyright (c) 2020 Lightning Labs
```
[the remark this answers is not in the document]
joostjager: Removed line. It isn't present either in the lnd release script where this is copied from.

release.sh
joostjager: This is copied from `lnd`. I think this shorter statement is not uncommon for error messages

release.sh
joostjager: added default target

Issue #194, opened 2020-05-06
rpc: add tls and macaroon auth for loopd clients


Roasbeef:
As is today, one can only hit the loopd API over local host on the same machine. As we want to be able to support more advanced configuration with remote lnd and loopd nodes, we should expand the client RPC server to support TLS and also macaroons. Much of this can be borrowed from lnd. To start a simple read and write macaroon should suffice. In the future we can also leverage the bakery to bake more granular macaroons.

joostjager:
Partial duplicate (TLS part) of #13. Can you move the TLS part into #13, so we have two distinct issues? They are related of course, but TLS can be implement first

joostjager:
Moved TLS description to #13

sako0938:
I am interested in this issue. I have scripts for interacting with the LND rest API, and that connects to a remote LND instance. It is a lot less useful to have to SSH into my node server in order to run my REST loop script. I end up just using the CLI most of the time, instead of the API because of this.

joostjager:
Are you interested in macaroons (this issue) or TLS (#13)? I guess it must be TLS?

sako0938:
I would think macaroons, whatever would allow the loop REST api to run on 0.0.0.0:8081, instead of localhost:8081

joostjager:
Restored issue description to original description after realizing that a server certificate on its own doesn't prevent anyone from accessing the server. So as originally stated, we need both macaroons and tls.

Issue #13, opened 2019-03-07
rpc: add TLS for loopd clients


Roasbeef:
As is today, one can only hit the loopd API over local host on the same machine. As we want to be able to support more advanced configuration with remote lnd and loopd nodes, we should expand the client RPC server to support TLS. Much of this can be borrowed from lnd

joostjager:
Copied desc from #194

joostjager:
Closing because just adding a server certificate doesn't offer any form of authentication of the client.

PR #202, opened 2020-05-15
loopd+config: prepare to be used as a library

guggero:
To be able to run loopd as a library, we need to export some config related elements and create the possibility to instantiate and run the RPC server without it creating its own grpc.Server server or REST proxy.ServeMux instance.

### DIFF
--- a/cmd/loop/main.go
+++ b/cmd/loop/main.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strconv"
 	"time"
@@ -14,11 +15,14 @@ import (
 	"github.com/lightninglabs/protobuf-hex-display/json"
 	"github.com/lightninglabs/protobuf-hex-display/jsonpb"
 	"github.com/lightninglabs/protobuf-hex-display/proto"
+	"github.com/lightningnetwork/lnd/macaroons"
 
 	"github.com/btcsuite/btcutil"
 
 	"github.com/urfave/cli"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"gopkg.in/macaroon.v2"
 )
 
 var (
@@ -34,6 +38,21 @@ var (
 	// maxMsgRecvSize is the largest message our client will receive. We
 	// set this to 200MiB atm.
 	maxMsgRecvSize = grpc.MaxCallRecvMsgSize(1 * 1024 * 1024 * 200)
+
+	// defaultMacaroonTimeout is the default macaroon timeout in seconds
+	// that we set when sending it over the line.
+	defaultMacaroonTimeout int64 = 60
+
+	tlsCertFlag = cli.StringFlag{
+		Name: "tlscertpath",
+		Usage: "path to loop's TLS certificate, only needed if loop " +
+			"runs in the same process as lnd",
+	}
+	macaroonPathFlag = cli.StringFlag{
+		Name: "macaroonpath",
+		Usage: "path to macaroon file, only needed if loop runs " +
+			"in the same process as lnd",
+	}
 )
 
 func printJSON(resp interface{}) {
@@ -84,6 +103,8 @@ func main() {
 			Value: "localhost:11010",
 			Usage: "loopd daemon address host:port",
 		},
+		tlsCertFlag,
+		macaroonPathFlag,
 	}
 	app.Commands = []cli.Command{
 		loopOutCommand, loopInCommand, termsCommand,
@@ -99,7 +120,9 @@ func main() {
 
 func getClient(ctx *cli.Context) (looprpc.SwapClientClient, func(), error) {
 	rpcServer := ctx.GlobalString("rpcserver")
-	conn, err := getClientConn(rpcServer)
+	tlsCertPath := ctx.GlobalString(tlsCertFlag.Name)
+	macaroonPath := ctx.GlobalString(macaroonPathFlag.Name)
+	conn, err := getClientConn(rpcServer, tlsCertPath, macaroonPath)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -256,10 +279,34 @@ func logSwap(swap *looprpc.SwapStatus) {
 	fmt.Println()
 }
 
-func getClientConn(address string) (*grpc.ClientConn, error) {
+func getClientConn(address, tlsCertPath, macaroonPath string) (*grpc.ClientConn,
+	error) {
+
 	opts := []grpc.DialOption{
-		grpc.WithInsecure(),
 		grpc.WithDefaultCallOptions(maxMsgRecvSize),
+	}
+
+	switch {
+	// If a TLS certificate file is specified, we need to load it and build
+	// transport credentials with it.
+	case tlsCertPath != "":
+		creds, err := credentials.NewClientTLSFromFile(tlsCertPath, "")
+		if err != nil {
+			fatal(err)
+		}
+
+		// Macaroons are only allowed to be transmitted over a TLS
+		// enabled connection.
+		if macaroonPath != "" {
+			opts = append(opts, readMacaroon(macaroonPath))
+		}
+
+		opts = append(opts, grpc.WithTransportCredentials(creds))
+
+	// By default, if no certificate is supplied, we assume the RPC server
+	// runs without TLS.
+	default:
+		opts = append(opts, grpc.WithInsecure())
 	}
 
 	conn, err := grpc.Dial(address, opts...)
@@ -268,4 +315,43 @@ func getClientConn(address string) (*grpc.ClientConn, error) {
 	}
 
 	return conn, nil
+}
+
+// readMacaroon tries to read the macaroon file at the specified path and create
+// gRPC dial options from it.
+func readMacaroon(macPath string) grpc.DialOption {
+	// Load the specified macaroon file.
+	macBytes, err := ioutil.ReadFile(macPath)
+	if err != nil {
+		fatal(fmt.Errorf("unable to read macaroon path : %v", err))
+	}
+
+	mac := &macaroon.Macaroon{}
+	if err = mac.UnmarshalBinary(macBytes); err != nil {
+		fatal(fmt.Errorf("unable to decode macaroon: %v", err))
+	}
+
+	macConstraints := []macaroons.Constraint{
+		// We add a time-based constraint to prevent replay of the
+		// macaroon. It's good for 60 seconds by default to make up for
+		// any discrepancy between client and server clocks, but leaking
+		// the macaroon before it becomes invalid makes it possible for
+		// an attacker to reuse the macaroon. In addition, the validity
+		// time of the macaroon is extended by the time the server clock
+		// is behind the client clock, or shortened by the time the
+		// server clock is ahead of the client clock (or invalid
+		// altogether if, in the latter case, this time is more than 60
+		// seconds).
+		macaroons.TimeoutConstraint(defaultMacaroonTimeout),
+	}
+
+	// Apply constraints to the macaroon.
+	constrainedMac, err := macaroons.AddConstraints(mac, macConstraints...)
+	if err != nil {
+		fatal(err)
+	}
+
+	// Now we append the macaroon credentials to the dial options.
+	cred := macaroons.NewMacaroonCredential(constrainedMac)
+	return grpc.WithPerRPCCredentials(cred)
 }
--- a/loopd/config.go
+++ b/loopd/config.go
@@ -29,7 +29,7 @@ type lndConfig struct {
 type viewParameters struct{}
 
 type Config struct {
-	ShowVersion    bool   `short:"V" long:"version" description:"Display version information and exit"`
+	ShowVersion    bool   `long:"version" description:"Display version information and exit"`
 	Insecure       bool   `long:"insecure" description:"disable tls"`
 	Network        string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
 	SwapServer     string `long:"swapserver" description:"swap server address host:port"`
@@ -42,7 +42,7 @@ type Config struct {
 	MaxLogFiles    int    `long:"maxlogfiles" description:"Maximum logfiles to keep (0 for no rotation)"`
 	MaxLogFileSize int    `long:"maxlogfilesize" description:"Maximum logfile size in MB"`
 
-	DebugLevel  string `short:"d" long:"debuglevel" description:"Logging level for all subsystems {trace, debug, info, warn, error, critical} -- You may also specify <subsystem>=<level>,<subsystem2>=<level>,... to set the log level for individual subsystems -- Use show to list available subsystems"`
+	DebugLevel  string `long:"debuglevel" description:"Logging level for all subsystems {trace, debug, info, warn, error, critical} -- You may also specify <subsystem>=<level>,<subsystem2>=<level>,... to set the log level for individual subsystems -- Use show to list available subsystems"`
 	MaxLSATCost uint32 `long:"maxlsatcost" description:"Maximum cost in satoshis that loopd is going to pay for an LSAT token automatically. Does not include routing fees."`
 	MaxLSATFee  uint32 `long:"maxlsatfee" description:"Maximum routing fee in satoshis that we are willing to pay while paying for an LSAT token."`
 
--- a/loopd/config.go
+++ b/loopd/config.go
@@ -28,7 +28,7 @@ type lndConfig struct {
 
 type viewParameters struct{}
 
-type config struct {
+type Config struct {
 	ShowVersion    bool   `short:"V" long:"version" description:"Display version information and exit"`
 	Insecure       bool   `long:"insecure" description:"disable tls"`
 	Network        string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
@@ -60,8 +60,8 @@ const (
 )
 
 // DefaultConfig returns all default values for the Config struct.
-func DefaultConfig() config {
-	return config{
+func DefaultConfig() Config {
+	return Config{
 		Network:         "mainnet",
 		RPCListen:       "localhost:11010",
 		RESTListen:      "localhost:8081",
--- a/loopd/config.go
+++ b/loopd/config.go
@@ -59,19 +59,22 @@ const (
 	testnetServer = "test.swap.lightning.today:11010"
 )
 
-var defaultConfig = config{
-	Network:         "mainnet",
-	RPCListen:       "localhost:11010",
-	RESTListen:      "localhost:8081",
-	Insecure:        false,
-	LogDir:          defaultLogDir,
-	MaxLogFiles:     defaultMaxLogFiles,
-	MaxLogFileSize:  defaultMaxLogFileSize,
-	DebugLevel:      defaultLogLevel,
-	MaxLSATCost:     lsat.DefaultMaxCostSats,
-	MaxLSATFee:      lsat.DefaultMaxRoutingFeeSats,
-	LoopOutMaxParts: defaultLoopOutMaxParts,
-	Lnd: &lndConfig{
-		Host: "localhost:10009",
-	},
+// DefaultConfig returns all default values for the Config struct.
+func DefaultConfig() config {
+	return config{
+		Network:         "mainnet",
+		RPCListen:       "localhost:11010",
+		RESTListen:      "localhost:8081",
+		Insecure:        false,
+		LogDir:          defaultLogDir,
+		MaxLogFiles:     defaultMaxLogFiles,
+		MaxLogFileSize:  defaultMaxLogFileSize,
+		DebugLevel:      defaultLogLevel,
+		MaxLSATCost:     lsat.DefaultMaxCostSats,
+		MaxLSATFee:      lsat.DefaultMaxRoutingFeeSats,
+		LoopOutMaxParts: defaultLoopOutMaxParts,
+		Lnd: &lndConfig{
+			Host: "localhost:10009",
+		},
+	}
 }
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -38,34 +38,57 @@ type listenerCfg struct {
 	getLnd func(string, *lndConfig) (*lndclient.GrpcLndServices, error)
 }
 
-// daemon runs loopd in daemon mode. It will listen for grpc connections,
+// Daemon is the struct that holds one instance of the loop client daemon.
+type Daemon struct {
+	// swapClientServer is the embedded RPC server that satisfies the client
+	// RPC interface. We embed this struct so the Daemon itself can be
+	// registered to an existing grpc.Server to run as a subserver in the
+	// same process.
+	swapClientServer
+
+	cfg         *Config
+	listenerCfg *listenerCfg
+
+	lnd *lndclient.GrpcLndServices
+}
+
+// New creates a new instance of the loop client daemon.
+func New(config *Config, lisCfg *listenerCfg) *Daemon {
+	return &Daemon{
+		cfg:         config,
+		listenerCfg: lisCfg,
+	}
+}
+
+// Run runs loopd in daemon mode. It will listen for grpc connections,
 // execute commands and pass back swap status information.
-func daemon(config *Config, lisCfg *listenerCfg) error {
-	lnd, err := lisCfg.getLnd(config.Network, config.Lnd)
+func (d *Daemon) Run() error {
+	var err error
+	d.lnd, err = d.listenerCfg.getLnd(d.cfg.Network, d.cfg.Lnd)
 	if err != nil {
 		return err
 	}
-	defer lnd.Close()
+	defer d.lnd.Close()
 
 	// If no swap server is specified, use the default addresses for mainnet
 	// and testnet.
-	if config.SwapServer == "" {
+	if d.cfg.SwapServer == "" {
 		// TODO(wilmer): Use onion service addresses when proxy is
 		// active.
-		switch config.Network {
+		switch d.cfg.Network {
 		case "mainnet":
-			config.SwapServer = mainnetServer
+			d.cfg.SwapServer = mainnetServer
 		case "testnet":
-			config.SwapServer = testnetServer
+			d.cfg.SwapServer = testnetServer
 		default:
 			return errors.New("no swap server address specified")
 		}
 	}
 
-	log.Infof("Swap server address: %v", config.SwapServer)
+	log.Infof("Swap server address: %v", d.cfg.SwapServer)
 
 	// Create an instance of the loop client library.
-	swapClient, cleanup, err := getClient(config, &lnd.LndServices)
+	swapClient, cleanup, err := getClient(d.cfg, &d.lnd.LndServices)
 	if err != nil {
 		return err
 	}
@@ -83,9 +106,9 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
 	}
 
 	// Instantiate the loopd gRPC server.
-	server := swapClientServer{
+	d.swapClientServer = swapClientServer{
 		impl:        swapClient,
-		lnd:         &lnd.LndServices,
+		lnd:         &d.lnd.LndServices,
 		swaps:       swaps,
 		subscribers: make(map[int]chan<- interface{}),
 		statusChan:  make(chan loop.SwapInfo),
@@ -93,14 +116,14 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
 
 	serverOpts := []grpc.ServerOption{}
 	grpcServer := grpc.NewServer(serverOpts...)
-	looprpc.RegisterSwapClientServer(grpcServer, &server)
+	looprpc.RegisterSwapClientServer(grpcServer, d)
 
 	// Next, start the gRPC server listening for HTTP/2 connections.
 	log.Infof("Starting gRPC listener")
-	grpcListener, err := lisCfg.grpcListener()
+	grpcListener, err := d.listenerCfg.grpcListener()
 	if err != nil {
 		return fmt.Errorf("RPC server unable to listen on %s",
-			config.RPCListen)
+			d.cfg.RPCListen)
 
 	}
 	defer grpcListener.Close()
@@ -122,24 +145,24 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
 	defer cancel()
 	mux := proxy.NewServeMux(customMarshalerOption)
 	var restHandler http.Handler = mux
-	if config.CORSOrigin != "" {
-		restHandler = allowCORS(restHandler, config.CORSOrigin)
+	if d.cfg.CORSOrigin != "" {
+		restHandler = allowCORS(restHandler, d.cfg.CORSOrigin)
 	}
 	proxyOpts := []grpc.DialOption{
 		grpc.WithInsecure(),
 		grpc.WithDefaultCallOptions(maxMsgRecvSize),
 	}
 	err = looprpc.RegisterSwapClientHandlerFromEndpoint(
-		ctx, mux, config.RPCListen, proxyOpts,
+		ctx, mux, d.cfg.RPCListen, proxyOpts,
 	)
 	if err != nil {
 		return err
 	}
 
-	restListener, err := lisCfg.restListener()
+	restListener, err := d.listenerCfg.restListener()
 	if err != nil {
 		return fmt.Errorf("REST proxy unable to listen on %s",
-			config.RESTListen)
+			d.cfg.RESTListen)
 	}
 
 	// A nil listener indicates REST is disabled.
@@ -170,7 +193,7 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
 		defer wg.Done()
 
 		log.Infof("Starting swap client")
-		err := swapClient.Run(mainCtx, server.statusChan)
+		err := swapClient.Run(mainCtx, d.statusChan)
 		if err != nil {
 			log.Error(err)
 		}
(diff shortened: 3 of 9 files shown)


joostjager:
Is there a branch that allows me to test the operation of loopd inside a parent process?

loopd/config.go
```diff
@@ -28,7 +28,7 @@ type lndConfig struct {
 
 type viewParameters struct{}
 
-type config struct {
+type Config struct {
 	ShowVersion    bool   `short:"V" long:"version" description:"Display version information and exit"`
```
joostjager: Looking at this now I am wondering whether this isn't a layering violation. We have a library that also defines command line options. It probably is a practical choice to prevent duplication the struct in the cli.

loopd/daemon.go
```diff
@@ -85,7 +112,7 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
 	// Instantiate the loopd gRPC server.
 	server := swapClientServer{
 		impl:        swapClient,
-		lnd:         &lnd.LndServices,
+		lnd:         &d.lnd.LndServices,
```
joostjager: Didn't `New` already create this `swapClientServer`?

joostjager: Ok, I see changes in the next commit. Commit structure can probably be optimized

loopd/daemon.go
```diff
@@ -259,6 +229,58 @@ func (d *Daemon) Run() error {
...
+	var err error
+	d.impl, d.clientCleanup, err = getClient(d.cfg, &d.lnd.LndServices)
+	if err != nil {
+		return err
+	}
+
+	// Retrieve all currently existing swaps from the database.
+	swapsList, err := d.impl.FetchSwaps()
+	if err != nil {
+		return err
```
joostjager, line 253: is `clientCleanup` called in this case?

loopd/daemon.go
```diff
@@ -259,6 +229,58 @@ func (d *Daemon) Run() error {
...
+	for _, s := range swapsList {
+		d.swaps[s.SwapHash] = *s
+	}
+
+	return nil
+}
+
+// stop tries to gracefully shut down the daemon.
+func (d *Daemon) stop() error {
+	if d.lnd != nil {
```
joostjager, line 333: Shouldn't this first take down the grpc server?

loopd/daemon.go
```diff
@@ -271,6 +287,12 @@ func (d *Daemon) initializeClient() error {
 
 // stop tries to gracefully shut down the daemon.
 func (d *Daemon) stop() error {
+	// Trying to stop the daemon multiple times isn't disallowed, we just
+	// need to make sure the logic itself is only run once.
```
joostjager, line 303: Is this expected to happen? Maybe we can at least log something, to possibly catch a bug around this. Same for `Run`

cmd/loop/main.go
```diff
@@ -256,16 +280,79 @@ func logSwap(swap *looprpc.SwapStatus) {
...
+
 	opts := []grpc.DialOption{
-		grpc.WithInsecure(),
 		grpc.WithDefaultCallOptions(maxMsgRecvSize),
 	}
 
+	switch {
+	// If a TLS certificate file is specified, we need to load it and build
+	// transport credentials with it.
+	case tlsCertPath != "":
```
joostjager, line 293: Maybe it is a good idea to first do a preparatory PR that adds TLS and macaroons to `loop`? That would also fix #194 and #13

cmd/loop/main.go
```diff
@@ -34,6 +38,22 @@ var (
...
+		Name: "tlscertpath",
+		Usage: "path to loop's TLS certificate, only " +
+			"needed if loop runs in the same process " +
+			"as lnd (GrUB)",
+	}
+	macaroonPathFlag = cli.StringFlag{
+		Name: "macaroonpath",
+		Usage: "path to macaroon file, only needed if loop runs " +
+			"in the same process as lnd (GrUB)",
+	}
```
joostjager, line 56: If you are going to move the permissions from the grub pr to here, it also makes sense to require them for stand alone use

loopd/daemon.go
```diff
@@ -161,75 +179,184 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+	return nil
+}
 
-		cancel()
-	}()
+// StartAsSubserver is an alternative to Start where the RPC server does not
+// create its own gRPC server but registers to an existing one. The same goes
+// for REST (if enabled), instead of creating an own mux and HTTP server, we
+// register to an existing one.
+func (d *Daemon) StartAsSubserver(lndGrpc *lndclient.GrpcLndServices) error {
```
joostjager, line 204: Can this be called by the stand-alone `Start`?

loopd/daemon.go
```diff
@@ -38,72 +42,85 @@ type listenerCfg struct {
...
+	restCtxCancel func()
+}
+
+// New creates a new instance of the loop client daemon.
+func New(config *Config, lisCfg *listenerCfg) *Daemon {
+	return &Daemon{
+		swapClientServer: swapClientServer{
+			swaps:       make(map[lntypes.Hash]loop.SwapInfo),
+			subscribers: make(map[int]chan<- interface{}),
+			statusChan:  make(chan loop.SwapInfo),
```
joostjager, line 81: Part of `swapClientServer` is initialized here and more happens in `Start`. Maybe cleaner to do it all in `Start`

loopd/daemon.go
joostjager: Same for other errors that may happen in `Start`

loopd/daemon.go
```diff
@@ -119,38 +136,39 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+	if d.restListener != nil {
 		log.Infof("Starting REST proxy listener")
 
-		defer restListener.Close()
-		proxy := &http.Server{Handler: restHandler}
+		d.restServer = &http.Server{Handler: restHandler}
 
 		go func() {
-			err := proxy.Serve(restListener)
+			log.Infof("REST proxy listening on %s",
```
joostjager, line 169: waitgroup add? pre-existing

loopd/daemon.go
```diff
@@ -161,75 +179,184 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
 	go func() {
-		defer wg.Done()
+		defer d.wg.Done()
 
-		log.Infof("Starting swap client")
-		err := swapClient.Run(mainCtx, server.statusChan)
+		log.Infof("RPC server listening on %s", d.grpcListener.Addr())
+		err = d.grpcServer.Serve(d.grpcListener)
 		if err != nil {
 			log.Error(err)
```
joostjager, line 193: What happens to the rest server in this case?

loopd/daemon.go
```diff
@@ -119,38 +136,39 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
 
-		defer restListener.Close()
-		proxy := &http.Server{Handler: restHandler}
+		d.restServer = &http.Server{Handler: restHandler}
 
 		go func() {
-			err := proxy.Serve(restListener)
+			log.Infof("REST proxy listening on %s",
+				d.restListener.Addr())
+			err := d.restServer.Serve(d.restListener)
```
joostjager, line 171: Error is just logged, no further stopping?

loopd/daemon.go
```diff
@@ -161,75 +179,184 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+	// Next, shut down the connections to lnd and the swap server.
+	if d.lnd != nil {
+		d.lnd.Close()
+	}
+	if d.clientCleanup != nil {
+		d.clientCleanup()
+	}
+
+	// The servers should already have stopped gracefully. If not, we
+	// force them by closing its underlying listeners.
```
joostjager, line 341: How can this happen?

loopd/daemon.go
```diff
@@ -171,34 +176,8 @@ func (d *Daemon) Run() error {
...
-	// Start a goroutine that broadcasts swap updates to clients.
-	d.wg.Add(1)
-	go func() {
-		defer d.wg.Done()
-
-		log.Infof("Waiting for updates")
-		d.processStatusUpdates(mainCtx)
-	}()
+	// Start the main event handlers.
+	d.startEventHandlers()
```
joostjager, line 183: The name 'event handlers' maybe doesn't do right to the core swap logic that is started in this function

loopd/daemon.go
```diff
@@ -242,6 +208,10 @@ func (d *Daemon) Run() error {
 	select {
 	case <-interruptChannel:
 		log.Infof("Received SIGINT (Ctrl+C).")
+		err := d.stop()
```
joostjager: With the prior `defer` calls, the cleanup was always performed (also for errors). Now it is only when an interrupt happens?

loopd/daemon.go
```diff
@@ -259,6 +229,58 @@ func (d *Daemon) Run() error {
...
+// stop tries to gracefully shut down the daemon.
+func (d *Daemon) stop() error {
+	if d.lnd != nil {
+		d.lnd.Close()
+	}
+
+	if d.clientCleanup != nil {
+		d.clientCleanup()
+	}
+	return nil
```
carlaKC: nit: add newline in this commit rather than later

loopd/daemon.go
```diff
@@ -102,8 +107,8 @@ func (d *Daemon) Run() error {
 	// With our client created, let's now finish setting up and start our
 	// RPC server.
 	serverOpts := []grpc.ServerOption{}
-	grpcServer := grpc.NewServer(serverOpts...)
-	looprpc.RegisterSwapClientServer(grpcServer, d)
+	d.grpcServer = grpc.NewServer(serverOpts...)
```
carlaKC, line 113: could this be moved to `New`? That way we don't need to check whether it's nil in `stop`

loopd/daemon.go
```diff
@@ -327,7 +314,26 @@ func (d *Daemon) stop() error {
 		d.clientCleanup()
 	}
 
-	return nil
+	// The servers should already have stopped gracefully. If not, we
+	// force them by closing its underlying listeners.
+	if d.grpcListener != nil {
+		_ = d.grpcListener.Close()
```
carlaKC, line 343: log these errors?

loopd/start.go
```diff
@@ -168,8 +169,17 @@ func Start(rpcCfg RPCConfig) error {
...
-		return daemon.Run()
+		if err := daemon.Start(); err != nil {
+			return err
+		}
+		<-signal.ShutdownChannel()
+		log.Infof("Received SIGINT (Ctrl+C).")
+		if err := daemon.Stop(); err != nil {
+			return err
+		}
+		return nil
```
carlaKC, line 183: return `daemon.Stop`? 
and nit: add some newlines here

loopd/daemon.go
```diff
@@ -259,6 +229,58 @@ func (d *Daemon) Run() error {
...
+
+	for _, s := range swapsList {
+		d.swaps[s.SwapHash] = *s
+	}
+
+	return nil
+}
+
+// stop tries to gracefully shut down the daemon.
+func (d *Daemon) stop() error {
```
carlaKC: If we only call `stop` on sigterm, is it possible that any of these fields are nil? If we're going to `defer stop()` then ignore this comment, just wondering if the checks are necessary.

loopd/daemon.go
guggero: Ah yes, I missed that duplicated initialization of `swapClientServer`. Fixed the commit.

loopd/daemon.go
guggero: Yes, you're right. I added a commit that calls `d.Stop()` before returning an error. Because everything in `Stop()` is guarded against being nil, we can do that.

loopd/daemon.go
guggero: At this point `stop` only contains what was done through the `defer` statements before. The final order will be correct which will be established in later commits.

loopd/daemon.go
guggero: Yes, explained the reason more clearly and also moved this to the commit where it makes more sense.

cmd/loop/main.go
guggero: Yes, I think when introducing macaroons to loop, it would make sense to also export the required permissions. But that's also not in the scope of this PR.

cmd/loop/main.go
guggero: We can do this later, this isn't in the scope of this PR.

loopd/daemon.go
guggero: `Serve` blocks until the server is shut down. So the error will only be returned if we're already in shutdown anyway. The REST server will also be shut down then.

loopd/daemon.go
guggero: Yes, same here. `Serve` only returns on shutdown of the server.

loopd/daemon.go
guggero: No, because we want to have the guard against starting it twice in there. I refactored some things so this got even smaller. I don't think we gain any readability by refactoring this into smaller methods.

loopd/daemon.go
guggero: It shouldn't. There was a `defer` before. But looking at `d.grpcServer.Stop()` and `d.restServer.Close()` they both close the underlying listeners in any case.

loopd/daemon.go
guggero: I added a `Stop` to each error case. So a proper shutdown should now always be performed, not only on interrupt.

loopd/daemon.go
guggero: There won't be a `grpcServer` if we start the daemon as a subserver. So we need the nil check anyway.

loopd/daemon.go
guggero: These would only return `use of closed network connection` because the servers themselves already closed the listeners. That's why I removed these lines completely.

loopd/daemon.go
guggero: I added `Stop()` to each error case so yes, they can be nil. Just forgot to do that before. Also if we start as a subserver, there won't be a `grpcServer` nor `restServer`. So all the checks definitely make sense.

loopd/config.go
guggero: The command line library is very nice that it allows config structs to easily be hierarchical. This is a big part of what makes the Grub possible. But short command line options must be unique because they only consist of one letter and cannot be hierarchical. I don't think this is a big issue. I don't like the short versions anyway because I can never remember which tool uses which (is `-v` version or verbose? is it lowercase or uppercase? and so on).

loopd/daemon.go
carlaKC: Got it, thanks for explaining!

loopd/config.go
joostjager: I just mean that ideally the names of command lines flags don't belong in a library. Someone else may use a different command line parsing library and then those don't work anyway.

loopd/daemon.go
joostjager: I was thinking about other errors than the shutdown error

loopd/daemon.go
joostjager: Isn't this now already cleaning up dependencies before the main `cancel()` is called?

loopd/daemon.go
joostjager: But in this commit, it is incorrect? I think the commits should be logical units of change that don't leave open ends

loopd/daemon.go
```diff
@@ -72,9 +75,9 @@ func New(config *Config, lisCfg *listenerCfg) *Daemon {
 	}
 }
 
-// Run runs loopd in daemon mode. It will listen for grpc connections,
+// Start starts loopd in daemon mode. It will listen for grpc connections,
 // execute commands and pass back swap status information.
-func (d *Daemon) Run() error {
+func (d *Daemon) Start() error {
```
joostjager, line 85: I am not too sure about the commit structure. It feels that I am seeing the lines change multiple times too often. Perhaps some squashing is appropriate. As a reviewer, I moved to reviewing the full diff because of this. And that defeats the purpose of having a commit structure.

loopd/daemon.go
joostjager: And how about other errors?

cmd/loop/main.go
joostjager: Isn't it double work to first introduce the loop macaroons/permissions in grub and then later move them here? Also this 'only needed if loop runs in the same process' looks a bit strange to me, hinting towards an optimal order of implementation that is different

loopd/daemon.go
```diff
@@ -161,75 +184,191 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+
+	// Now finally fully initialize the swap client RPC server instance.
+	d.swapClientServer = swapClientServer{
+		impl:        swapclient,
+		lnd:         &d.lnd.LndServices,
+		swaps:       make(map[lntypes.Hash]loop.SwapInfo),
+		subscribers: make(map[int]chan<- interface{}),
+		statusChan:  make(chan loop.SwapInfo),
+	}
+	d.clientCleanup = clientCleanup
```
joostjager, line 263: Move this line up?

loopd/daemon.go
```diff
@@ -38,72 +42,85 @@ type listenerCfg struct {
...
 
 	// Next, start the gRPC server listening for HTTP/2 connections.
 	log.Infof("Starting gRPC listener")
-	grpcListener, err := lisCfg.grpcListener()
+	d.grpcListener, err = d.listenerCfg.grpcListener()
 	if err != nil {
+		_ = d.Stop()
 		return fmt.Errorf("RPC server unable to listen on %s",
-			config.RPCListen)
+			d.cfg.RPCListen)
```
joostjager, line 121: Include actual `err`? (pre-existing)

loopd/daemon.go
```diff
@@ -119,38 +136,44 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
 		return err
 	}
 
-	restListener, err := lisCfg.restListener()
+	d.restListener, err = d.listenerCfg.restListener()
 	if err != nil {
+		_ = d.Stop()
 		return fmt.Errorf("REST proxy unable to listen on %s",
-			config.RESTListen)
+			d.cfg.RESTListen)
```
joostjager, line 161: Also pass on the `err` (pre-existing)?

loopd/daemon.go
```diff
@@ -38,72 +42,85 @@ type listenerCfg struct {
...
+	if atomic.AddInt32(&d.started, 1) != 1 {
+		return errOnlyStartOnce
+	}
 
-	// Create an instance of the loop client library.
-	swapClient, cleanup, err := getClient(config, &lnd.LndServices)
+	var err error
+	d.lnd, err = d.listenerCfg.getLnd(d.cfg.Network, d.cfg.Lnd)
 	if err != nil {
+		_ = d.Stop()
```
joostjager, line 96: There is a lot of stop calling on error. Probably everywhere. Create a wrapping function that calls this function, inspects the error and call stop if it is not nil?

loopd/daemon.go
```diff
@@ -161,75 +184,191 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
-	// Run until the users terminates loopd or an error occurred.
-	select {
-	case <-interruptChannel:
-		log.Infof("Received SIGINT (Ctrl+C).")
+	// First of all, cancel the main context that all event handlers are
+	// using. This should stop all swap activity and all event handlers
+	// should exit.
+	if d.mainCtxCancel != nil {
+		d.mainCtxCancel()
+	}
```
joostjager, line 334: Can you just do this, stop everything at the same time? I would think that services need to be stopped outside to inside and wait for each one to have stopped properly.

In general I feel that the `Start`/`Stop` pattern makes the whole shutdown sequence less readable and forces us to resort to these nil checks, where as otherwise we could just keep adding defers.

How about keeping a synchronous Run() method that does all its clean up and adding a wrapping `Start`/`Stop` construction that just initiates a goroutine?

cmd/loop/main.go
joostjager: Discussed offline, leaving this for later because it isn't trivial in grub to "merge" those macaroons. Only remove the reference to `GrUB`

cmd/loop/main.go
joostjager: Discussed offline

loopd/daemon.go
```diff
@@ -161,75 +184,191 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+	if err != nil {
+		return err
+	}
 
-	// Start the grpc server.
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	return nil
+}
```
joostjager, line 227: Discussed offline: what happens if something goes wrong between `Start` and `Stop`

loopd/run.go
```diff
@@ -168,7 +170,17 @@ func Start(rpcCfg RPCConfig) error {
...
 	// Execute command.
 	if parser.Active == nil {
-		return daemon(&config, lisCfg)
+		signal.Intercept()
+
+		daemon := New(&config, lisCfg)
+		if err := daemon.Start(); err != nil {
+			return err
+		}
+
```
joostjager, line 179: Discussed offline: If a daemon error happens here, the cli will hang indefinitely.

loopd/daemon.go
guggero: I've added the stop to the commit where it's first introduced. I also created the `stopAndReturnErr` helper.

loopd/daemon.go
guggero: Discussed offline. The stop now first stops accepting gRPC connections, then cancels the main context, shuts down the gRPC server and then its dependencies.

loopd/daemon.go
guggero: They will now be sent over the `errChan`.

loopd/daemon.go
guggero: Discussed this offline. Decided to stay with the `Start`/`Stop` model but add an error channel that can be used to detect runtime errors.

I did some manual tests and made sure that the daemon stops correctly if the swap handler crashes and also if a `monitor` subscription is connected.
For the second case I also fixed it to return a nice error message instead of just force closing the gRPC server and returning a "connection is terminating" error.

loopd/daemon.go
guggero: Addressed with `errChan`.

loopd/run.go
guggero: Addressed by listening on `errChan`.

loopd/run.go
```diff
@@ -168,7 +170,27 @@ func Start(rpcCfg RPCConfig) error {
...
+		select {
+		case <-signal.ShutdownChannel():
+			log.Infof("Received SIGINT (Ctrl+C).")
+			return daemon.Stop()
+
+		case err := <-errChan:
+			log.Errorf("Daemon sent unexpected error: %v", err)
+			stopErr := daemon.Stop()
```
joostjager, line 188: I would think that this shouldn't be necessary. The daemon should have shut down cleanly if it sends the error, to keep responsibilities clean.

loopd/daemon.go
```diff
@@ -119,117 +135,291 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
...
+			// the errChan later, after we've stopped the daemon.
 			log.Error(err)
 		}
 		log.Infof("Swap client stopped")
 
-		log.Infof("Stopping gRPC server")
-		grpcServer.Stop()
+		// We need to mark this goroutine to be done explicitly here.
+		// Otherwise the stop below will wait for the defer which will
+		// end in a deadlock.
```
joostjager, line 303: To me this comment signals that we are not on the optimal design yet. There are currently four goroutines running: rest, grpc, swap client and status updates. Maybe a main goroutine is needed to listen on all error channels and coordinate the stopping process.

loopd/daemon.go
```diff
@@ -119,117 +135,291 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+	}()
+
+	return nil
+}
+
+// StartAsSubserver is an alternative to Start where the RPC server does not
+// create its own gRPC server but registers to an existing one. The same goes
+// for REST (if enabled), instead of creating an own mux and HTTP server, we
+// register to an existing one.
+func (d *Daemon) StartAsSubserver(lndGrpc *lndclient.GrpcLndServices) error {
```
joostjager, line 205: So this function for now ignores the err channel?

loopd/daemon.go
```diff
@@ -119,117 +135,291 @@ func daemon(config *config, lisCfg *listenerCfg) error {
...
+	}
 
-		if restListener != nil {
-			log.Infof("REST proxy listening on %s", restListener.Addr())
-		}
+	// First of all, tell the gRPC server to stop accepting connections.
+	// This method blocks until all pending requests are completed. Because
+	// of our streaming subscriptions, those won't be canceled until the
+	// main context is canceled, so we'd be stuck here. That's why we need
+	// to run this in a goroutine.
```
joostjager, line 354: If this is the case anyway, it looks indeed to be better to first cancel the main context and not start this goroutine. Maybe this only makes sense if there would be a delay before cancelling the main context. But that may be seen as annoying if there are not pending grpc connections. Maybe it should give `GracefulStop` up to a specific time to stop gracefully and thereafter we hard-cancel the context.

loopd/daemon.go
guggero: Yes, that makes sense, I see it now. It does look cleaner that way.

loopd/daemon.go
guggero: No. You still have to instantiate the daemon with the channel and react on an error arriving there. But the `d.stopAndReturnErr` was missing here.

loopd/daemon.go
guggero: But the worry was that while the main context is being canceled and the swap handlers are shutting down, we would still accept incoming requests. This is handled by this now.
Or do you think we should add an explicit delay after the goroutine with the graceful shutdown was started?

joostjager: Is it really handled? To me it looks like we call `GracefulStop` and at the exact same moment we cancel the main context. Which basically means that the order isn't really changed from before? There may also still a request "in flight", but don't know how that works internally exactly.

To me the cleanest approach is what I described above. Give GracefulStop *up to* a specific time to stop gracefully and thereafter we hard-cancel the context.

But we can also decide not to deal with this at all and keep it simple without goroutines. First  cancel context, then stop grpc.

loopd/daemon.go
joostjager: Ah right, same `NewDaemon` call.

loopd/daemon.go
```diff
@@ -119,117 +140,291 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
...
+			defer d.wg.Done()
+
+			log.Infof("REST proxy listening on %s",
+				d.restListener.Addr())
+			err := d.restServer.Serve(d.restListener)
 			// ErrServerClosed is always returned when the proxy is
 			// shut down, so don't log it.
 			if err != nil && err != http.ErrServerClosed {
 				log.Error(err)
+				d.internalErrChan <- err
```
joostjager, line 183: The log statements in the goroutines (also below) can be moved to the main error handling goroutine. Maybe add a comment everywhere that the channel has enough buffer to not block.

loopd/daemon.go
```diff
@@ -119,117 +140,291 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
...
+		}()
 
-		if restListener != nil {
-			log.Infof("REST proxy listening on %s", restListener.Addr())
-		}
+		select {
+		case err := <-d.internalErrChan:
+			log.Errorf("Runtime error in daemon, shutting down: "+
+				"%v", err)
+			d.mainErrChan <- err
```
joostjager, line 336: This blocks if it isn't read or buffered. I think the usual pattern in `lnd` is to instantiatie the channel in the unit that writes to it, so that it can be created with buffer size `1`.

loopd/daemon.go
```diff
@@ -38,72 +42,89 @@ type listenerCfg struct {
...
+		return errOnlyStartOnce
 	}
 
-	swaps := make(map[lntypes.Hash]loop.SwapInfo)
-	for _, s := range swapsList {
-		swaps[s.SwapHash] = *s
+	var err error
+	d.lnd, err = d.listenerCfg.getLnd(d.cfg.Network, d.cfg.Lnd)
+	if err != nil {
+		return d.stopAndReturnErr(err)
```
joostjager, line 103: My suggestion was, instead of creating `stopAndReturnErr`, to create a wrapping method for `Start`. For example rename `Start` to `start` and add a new `Start` method that looks like:
```
err := start()
if err != nil {
   log(err)
   stop()
}
```

The advantage is that future devs can never forgot to call `stopAndReturnErr`

loopd/daemon.go
```diff
@@ -119,117 +140,291 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
...
-	case <-interruptChannel:
-		log.Infof("Received SIGINT (Ctrl+C).")
+// Stop tries to gracefully shut down the daemon.
+func (d *Daemon) Stop() error {
+	// If the swap client crashes, stop is called to make sure everything
+	// is cleaned up. But this also means we get here twice in case of the
+	// normal shutdown because after the swap client exists normally, stop
+	// is called again. We don't want to log an error in that case.
+	if atomic.AddInt32(&d.stopped, 1) != 1 {
+		return nil
```
joostjager, line 350: Isn't this a problem if a first stop is still in progress? It looks to the caller as if we already stopped.

loopd/daemon.go
```diff
@@ -119,117 +140,291 @@ func daemon(config *Config, lisCfg *listenerCfg) error {
...
+		defer d.wg.Done()
 
 		log.Infof("Waiting for updates")
-		server.processStatusUpdates(mainCtx)
+		d.processStatusUpdates(d.mainCtx)
 	}()
 
-	// Start the grpc server.
-	wg.Add(1)
+	// Last, start our internal error handler. This will return at most one
```
joostjager, line 317: This isn't strictly a swap handler. The goroutines of the grpc server also write to this? I mean that the name of the function doesn't cover what's inside.

loopd/daemon.go
guggero: Okay, removed the goroutine.

loopd/daemon.go
guggero: I don't see a scenario where this would be a problem. If a caller calls `Stop` just after the daemon crashed with an error then that error would be delivered over the error channel and handled there.

loopd/daemon.go
guggero: Okay, that makes sense.

loopd/daemon.go
joostjager: Does `startHandlers` cover it then? There are also other handlers started outside of this.

loopd/daemon.go
joostjager: Maybe they haven't read the error from the channel yet, because a ctrl-c was issued right before. The internal error handler is then still waiting for `Stop` to complete, but the ctrl-c stop call returns early here and may exit the process immediately, preventing clean shut down?

loopd/daemon.go
```diff
@@ -159,100 +181,258 @@ func (d *Daemon) Run() error {
...
+	// error on the main error channel to inform the caller that something
+	// went wrong.
+	d.wg.Add(1)
 	go func() {
-		defer wg.Done()
+		defer d.wg.Done()
 
-		log.Infof("RPC server listening on %s", grpcListener.Addr())
+		defer func() {
+			err := d.Stop()
```
joostjager, line 363: Does this work? Waiting (in Stop) on the wait group inside a goroutine that is added to the wait group?

loopd/daemon.go
guggero: That might be possible, yes. I added a wait on the wait group here. So the second call to `Stop` will return a nil error but only _after_ the first invocation is done.

loopd/daemon.go
guggero: Yes, because of the `defer` above. The order of the defers is kept. I added a comment to clarify.

guggero: Hmm... I thought it would because of the order of the `defer`s. But you are right, it didn't work.
Solved it with a helper function `ensureShutdown`. All my manual test cases now work correctly.

loopd/daemon.go
```diff
@@ -38,72 +38,113 @@ type listenerCfg struct {
...
+
+// Start starts loopd in daemon mode. It will listen for grpc connections,
 // execute commands and pass back swap status information.
-func daemon(config *config, lisCfg *listenerCfg) error {
-	lnd, err := lisCfg.getLnd(config.Network, config.Lnd)
+func (d *Daemon) Start() error {
+	err := d.start()
 	if err != nil {
+		log.Errorf("Error while starting daemon: %v", err)
+		stopErr := d.Stop()
```
joostjager, line 100: If the daemon was already started, we end up here and stop the running daemon

loopd/daemon.go
```diff
@@ -20,6 +21,10 @@ var (
 	// maxMsgRecvSize is the largest message our REST proxy will receive. We
 	// set this to 200MiB atm.
 	maxMsgRecvSize = grpc.MaxCallRecvMsgSize(1 * 1024 * 1024 * 200)
+
+	// errOnlyStartOnce is the error that is returned if the daemon is
+	// started more than once.
+	errOnlyStartOnce = fmt.Errorf("daemon can only be started once")
```
joostjager, line 27: Move this to previous commit?

loopd/daemon.go
```diff
@@ -159,100 +172,221 @@ func (d *Daemon) Run() error {
...
-	signal.Notify(interruptChannel, os.Interrupt)
+	return nil
+}
 
-	// Run until the users terminates loopd or an error occurred.
-	select {
-	case <-interruptChannel:
-		log.Infof("Received SIGINT (Ctrl+C).")
+// Stop tries to gracefully shut down the daemon. A caller needs to wait for a
+// message on the main error channel indicating that the shutdown is completed.
```
carlaKC, line 383: Why don't we wait for the error channel in this function? So we stop+block waiting for our error, then the called knows we're done once this function unblocks.

guggero:
Thanks a lot for all the review work!
It was a lot of back and forth but I think we can use this design in other services too, now that we have worked out all the possible edge cases.

PR #203, opened 2020-05-18
lndclient: extend LightningClient with ListChannels and RouterClient with keysend

bhandras:
This PR adds `ListChannels` to `LightningClient` and extends `RouterClient.SendPayment` with keysend.

### DIFF
--- a/lndclient/lightning_client.go
+++ b/lndclient/lightning_client.go
@@ -17,6 +17,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnrpc/invoicesrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
 	"github.com/lightningnetwork/lnd/zpay32"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -46,6 +47,9 @@ type LightningClient interface {
 	// node.
 	ListTransactions(ctx context.Context) ([]*wire.MsgTx, error)
 
+	// ListChannels retrieves all channels of the backing lnd node.
+	ListChannels(ctx context.Context) ([]ChannelInfo, error)
+
 	// ChannelBackup retrieves the backup for a particular channel. The
 	// backup is returned as an encrypted chanbackup.Single payload.
 	ChannelBackup(context.Context, wire.OutPoint) ([]byte, error)
@@ -63,6 +67,29 @@ type Info struct {
 	Alias          string
 	Network        string
 	Uris           []string
+}
+
+// ChannelInfo stores unpacked per-channel info.
+type ChannelInfo struct {
+	// Active indecates whether the channel is active.
+	Active bool
+
+	// ChannelID holds the unique channel ID for the channel. The first 3 bytes
+	// are the block height, the next 3 the index within the block, and the last
+	// 2 bytes are the /output index for the channel.
+	ChannelID uint64
+
+	// PubKeyBytes is the raw bytes of the public key of the remote node.
+	PubKeyBytes route.Vertex
+
+	// Capacity is the total amount of funds held in this channel.
+	Capacity btcutil.Amount
+
+	// LocalBalance is the current balance of this node in this channel.
+	LocalBalance btcutil.Amount
+
+	// RemoteBalance is the counterparty's current balance in this channel.
+	RemoteBalance btcutil.Amount
 }
 
 var (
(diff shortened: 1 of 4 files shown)


lndclient/lightning_client.go
```diff
@@ -394,6 +397,24 @@ func (s *lightningClient) ListTransactions(ctx context.Context) ([]*wire.MsgTx,
 	return txs, nil
 }
 
+// ListChannels retrieves all channels of the backing lnd node.
+func (s *lightningClient) ListChannels(ctx context.Context) (
+	[]*lnrpc.Channel, error) {
```
joostjager, line 402: The idea of the wrapper is to not expose things like string based pubkeys. Ideally we'd create a strongly-typed struct to return a slice of. Can for now just contains the fields that we need.

lndclient/router_client.go
```diff
@@ -102,6 +104,9 @@ type SendPaymentRequest struct {
 	// MaxParts is the maximum number of partial payments that may be used
 	// to complete the full amount.
 	MaxParts uint32
+
+	// KeySend is set to true if the payment request will encode the preimage.
```
joostjager, line 108: `payment request` generally refers to the hex-encoded invoice. You probably mean the tlv payload?

lndclient/router_client.go
```diff
@@ -140,6 +145,17 @@ func (r *routerClient) SendPayment(ctx context.Context,
...
+	if request.KeySend {
+		var preimage lntypes.Preimage
+		if _, err := rand.Read(preimage[:]); err != nil {
+			return nil, nil, err
+		}
+
+		// Override the payment hash.
+		rpcReq.DestCustomRecords = make(map[uint64][]byte)
+		rpcReq.DestCustomRecords[record.KeySendType] = preimage[:]
+		request.PaymentHash = preimage.Hash()
```
joostjager, line 157: Assert that `PaymentHash` wasn't set, to catch unintended use?

bhandras: done

lndclient/router_client.go
bhandras: ptal

lndclient/lightning_client.go
bhandras: done

lndclient/lightning_client.go
```diff
@@ -394,6 +421,41 @@ func (s *lightningClient) ListTransactions(ctx context.Context) ([]*wire.MsgTx,
...
+		s.adminMac.WithMacaroonAuth(rpcCtx),
+		&lnrpc.ListChannelsRequest{},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]ChannelInfo, len(response.Channels))
+	for i, channel := range response.Channels {
+		remoteVertex, err := route.NewVertexFromStr(channel.RemotePubkey)
```
joostjager, line 441: Future devs will thank you for the luxury of not having to parse everywhere :)

lndclient/lightning_client.go
```diff
@@ -61,6 +65,29 @@ type Info struct {
...
+	// ChannelID holds the unique channel ID for the channel. The first 3 bytes
+	// are the block height, the next 3 the index within the block, and the last
+	// 2 bytes are the /output index for the channel.
+	ChannelID uint64
+
+	// PubKeyBytes is the raw bytes of the public key of the remote node.
+	PubKeyBytes route.Vertex
+
+	// Capacity is the total amount of funds held in this channel.
+	Capacity int64
```
joostjager, line 82: Can be type `btcutil.Amount`, just as the balances below

lndclient/router_client.go
```diff
@@ -140,6 +149,22 @@ func (r *routerClient) SendPayment(ctx context.Context,
 	if request.LastHopPubkey != nil {
 		rpcReq.LastHopPubkey = request.LastHopPubkey[:]
 	}
+	if request.KeySend {
+		if !bytes.Equal(request.PaymentHash[:], emptyHash[:]) {
```
joostjager, line 153: If payment hash is optional, it would be better to make it type `*lntypes.Hash` to avoid the magic value.

lndclient/lightning_client.go
bhandras: indeed, that is much better

lndclient/router_client.go
bhandras: agree, much nicer to have the `*lntypes.Hash`, done

lndclient/router_client.go
```diff
@@ -140,6 +145,23 @@ func (r *routerClient) SendPayment(ctx context.Context,
...
+		}
+
+		var preimage lntypes.Preimage
+		if _, err := rand.Read(preimage[:]); err != nil {
+			return nil, nil, err
+		}
+
+		// Override the payment hash.
+		rpcReq.DestCustomRecords = make(map[uint64][]byte)
+		rpcReq.DestCustomRecords[record.KeySendType] = preimage[:]
```
joostjager, line 161: Alternative. Fewer characters but more lines :) Non-blocking ofc
```
rpcReq.DestCustomRecords = map[uint64][]byte {
   record.KeySendType: preimage[:],
}
```

bhandras: done, probably compiles to the same code, but more readable :)

lndclient/lightning_client.go
```diff
@@ -42,6 +43,9 @@ type LightningClient interface {
 	// node.
 	ListTransactions(ctx context.Context) ([]*wire.MsgTx, error)
 
+	// ListChannels retrieves all channels of the backing lnd node.
+	ListChannels(ctx context.Context) ([]ChannelInfo, error)
```
carlaKC, line 47: `[]*ChannelInfo`?

PR #205, opened 2020-05-19
loop out: allow outbound channel set for multi-loops

joostjager:
Currently loop out supports multi-loops (looping out multiple channels in one operation using MPP). But there is no control over which channels are used for the payment.

This issue adds the ability for the user to indicate a set of channels which are all allowed to be used in the swap.

Depends https://github.com/lightningnetwork/lnd/pull/4257


joostjager:
This is now implemented as an on-the-fly migration, but I am not sure whether it is the best choice. Especially the question of how to test this. Dragging in all the old serialization code mostly brings us back to the old-style migrations. And neither do I like the handling of legacy fields to be around forever.

loopd/view.go
```diff
@@ -64,13 +64,18 @@ func viewOut(swapClient *loop.Client, chainParams *chaincfg.Params) error {
...
-			unchargeChannel = strconv.FormatUint(
-				*s.Contract.UnchargeChannel, 10,
-			)
+		var unchargeChannels string
+		if len(s.Contract.OutgoingChanSet) > 0 {
+			for i, chanID := range s.Contract.OutgoingChanSet {
+				if i > 0 {
+					unchargeChannels += ", "
+				}
+				unchargeChannels += strconv.FormatUint(chanID, 10)
```
carlaKC, line 73: Same here, maybe make a short helper func which we can just call inline in the log since we do this more than once? Th

cmd/loop/loopout.go
```diff
@@ -25,10 +27,10 @@ var loopOutCommand = cli.Command{
 	Optionally a BASE58/bech32 encoded bitcoin destination address may be
 	specified. If not specified, a new wallet address will be generated.`,
 	Flags: []cli.Flag{
-		cli.Uint64Flag{
+		cli.StringFlag{
```
carlaKC, line 30: `StringSliceFlag` so we don't need to parse ourselves?

cmd/loop/loopout.go
```diff
@@ -25,10 +27,10 @@ var loopOutCommand = cli.Command{
...
 			Name: "channel",
-			Usage: "the 8-byte compact channel ID of the channel " +
-				"to loop out",
+			Usage: "the comma-separated list of 8-byte compact " +
```
carlaKC, line 32: Is "compact channel ID" a term that end users are familiar with?

joostjager:
>Tangential, but do we have an official deprecation plan for loop?

Just created this morning: https://github.com/lightninglabs/loop/pull/206

loopdb/store.go
```diff
@@ -216,6 +198,35 @@ func (s *boltSwapStore) fetchSwaps(bucketKey []byte,
...
 }
 
+// deserializeUpdates deserializes the list of swap updates that are stored as a
+// key of the given bucket.
+func deserializeUpdates(swapBucket *bbolt.Bucket) ([]*LoopEvent, error) {
+	// Once we have the raw swap, we'll also need to decode
+	// each of the past updates to the swap itself.
+	stateBucket := swapBucket.Bucket(updatesBucketKey)
+	if stateBucket == nil {
+		return nil, errors.New("updates bucket not found")
```
bhandras, line 258: It may be the right time to also create predefined errors instead of creating them inline (I know this is a common pattern in this file).

loopdb/store.go
```diff
@@ -227,59 +244,68 @@ func deserializeUpdates(swapBucket *bbolt.Bucket) ([]*LoopEvent, error) {
...
-	)
-	if err != nil {
-		return nil, err
-	}
+		// We'll now traverse the root bucket for all active swaps. The
+		// primary key is the swap hash itself.
+		return rootBucket.ForEach(func(swapHash, v []byte) error {
+			// Only go into things that we know are sub-bucket
+			// keys.
+			if v != nil {
```
bhandras, line 298: Not part of this review, just leaving a side note for the future: would be nice to have a sub bucket containing the swap buckets instead.

lndclient/router_client.go
```diff
@@ -64,10 +64,14 @@ type SendPaymentRequest struct {
 	// are only processed when the Invoice field is empty.
 	Invoice string
 
-	MaxFee          btcutil.Amount
-	MaxCltv         *int32
-	OutgoingChannel *uint64
-	Timeout         time.Duration
+	MaxFee  btcutil.Amount
```
bhandras, line 67: nb: Now that you're adding to this code it is perhaps a good opportunity to extend the documentation for other public fields as well :)

loopdb/store_test.go
```diff
@@ -37,31 +37,10 @@ var (
 	testTime = time.Date(2018, time.January, 9, 14, 00, 00, 0, time.UTC)
 )
 
-// TestLoopOutStore tests all the basic functionality of the current bbolt
+// testLoopOutStore tests all the basic functionality of the current bbolt
```
bhandras, line 40: nit: testLoopOutStore -> TestLoopOutStore

client.go
```diff
@@ -359,9 +360,16 @@ func (s *Client) resumeSwaps(ctx context.Context,
 func (s *Client) LoopOut(globalCtx context.Context,
 	request *OutRequest) (*lntypes.Hash, btcutil.Address, error) {
 
-	log.Infof("LoopOut %v to %v (channel: %v)",
-		request.Amount, request.DestAddr,
-		request.LoopOutChannel,
+	var channels string
+	for i, chanID := range request.OutgoingChanSet {
+		if i > 0 {
+			channels += ","
```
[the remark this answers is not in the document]
bhandras: +1 for the `strings.Join`, would also not add a `,` at the end.

loopdb/loopout.go
```diff
@@ -34,9 +34,9 @@ type LoopOutContract struct {
 	// client sweep tx.
 	SweepConfTarget int32
 
-	// TargetChannel is the channel to loop out. If zero, any channel may
-	// be used.
-	UnchargeChannel *uint64
+	// OutgoingChanSet is the set of channels that may be used. If empty,
+	// any channel may be used.
+	OutgoingChanSet []uint64
```
bhandras, line 39: Consider changing the doc such that it says the items are short channel ids (like above).

loopout.go
```diff
@@ -136,6 +135,7 @@ func newLoopOutSwap(globalCtx context.Context, cfg *swapConfig,
 			MaxMinerFee:      request.MaxMinerFee,
 			MaxSwapFee:       request.MaxSwapFee,
 		},
+		OutgoingChanSet: request.OutgoingChanSet,
```
bhandras, line 138: Do we need to make sure that the there's no duplication in the set?

cmd/loop/loopout.go
```diff
@@ -145,9 +147,14 @@ func loopOut(ctx *cli.Context) error {
 		return err
 	}
 
-	var unchargeChannel uint64
-	if ctx.IsSet("channel") {
-		unchargeChannel = ctx.Uint64("channel")
+	chanStrings := strings.Split(ctx.String("channel"), ",")
+	var outgoingChanSet []uint64
+	for _, chanString := range chanStrings {
+		chanID, err := strconv.ParseUint(chanString, 10, 64)
```
bhandras, line 153: What if the the passed list/set contains duplicate items?

loopdb/store.go
joostjager: I am not convinced that we should replace every error with a constant. Some errors also have parameters and for those we'd need to create separate error structs to store the parameters. It is useful to assert on in unit tests, but that is not the case for this error.

loopdb/store.go
```diff
@@ -216,6 +198,35 @@ func (s *boltSwapStore) fetchSwaps(bucketKey []byte,
...
+// key of the given bucket.
+func deserializeUpdates(swapBucket *bbolt.Bucket) ([]*LoopEvent, error) {
+	// Once we have the raw swap, we'll also need to decode
+	// each of the past updates to the swap itself.
+	stateBucket := swapBucket.Bucket(updatesBucketKey)
+	if stateBucket == nil {
+		return nil, errors.New("updates bucket not found")
+	}
+
+	// De serialize and collect each swap update into our
```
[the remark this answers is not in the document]
joostjager: fixed

loopdb/store.go
```diff
@@ -216,6 +198,35 @@ func (s *boltSwapStore) fetchSwaps(bucketKey []byte,
...
+	// slice of swap events.
+	var updates []*LoopEvent
+	err := stateBucket.ForEach(func(k, v []byte) error {
```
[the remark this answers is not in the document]
joostjager: fixed

loopdb/store.go
joostjager: How would that be advantageous? Maybe if we need to add loop-type-specific data that isn't related to a particular swap? Workaround is to just create a new top level bucket. I wouldn't really want to store anything other than swap buckets in the `rootBucket` in this function.

loopdb/store.go
```diff
@@ -227,59 +244,68 @@ func deserializeUpdates(swapBucket *bbolt.Bucket) ([]*LoopEvent, error) {
...
+
+			loop := LoopIn{
+				Loop: Loop{
+					Events: updates,
+				},
 				Contract: contract,
-				Loop:     loop,
-			})
+			}
+			copy(loop.Hash[:], swapHash)
```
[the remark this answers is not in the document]
joostjager: I think it is a copy because we don't need the validation. Otherwise it would be:
```
var err error
loop.Hash, err = lntypes.NewHash(swapHash)
if err != nil {
   return err
}
```

lndclient/router_client.go
joostjager: commit added

loopdb/store_test.go
joostjager: fixed

client.go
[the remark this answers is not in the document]
joostjager: This doesn't add `, ` at the end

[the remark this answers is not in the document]
joostjager: Created `ChannelSet` struct with `String` method

loopd/view.go
joostjager: helper added, see above

loopdb/loopout.go
joostjager: added short ids

loopout.go
joostjager: Good point, check added

cmd/loop/loopout.go
joostjager: replaced by 'short' which I think is mostly used

cmd/loop/loopout.go
joostjager: Never heard of it. Added! To bad there seems to be no Uint64SliceFlag

joostjager: Added, but doesn't work as expected. For every channel, the complete flag needs to be added. Left as is with more compact csv

joostjager: Feature request https://github.com/urfave/cli/issues/1134

cmd/loop/loopout.go
joostjager: This is now handled in `loopd`

loopdb/store.go
bhandras: Yes, true. I missed the part this is not the root bucket but a "root bucket". In this case it really is not of any advantage.

joostjager: Hehe, yes it is the local root :)

bhandras:
Looks good. It would be nice if you could cover the added functionality with at least one test in the PR too.

joostjager:
Added unit test and test for on-the-fly migration

joostjager:
Before release, I want to follow up with an itest covering multi-loop: https://github.com/lightninglabs/nautilus/issues/250

loopout_test.go
```diff
@@ -15,6 +15,105 @@ import (
...
+		store:  store,
+		server: server,
+	}
+
+	sweeper := &sweep.Sweeper{Lnd: &lnd.LndServices}
+
+	blockEpochChan := make(chan interface{})
+	statusChan := make(chan SwapInfo)
+
+	const maxParts = 5
```
bhandras, line 48: how about extending this test just a little bit, such that max parts is a parameter and we test that with zero it fails, and then 1, 2, say 5?

joostjager: We can't just test that at this level. There is no real lnd or pathfinding involved. The test just asserts passing through the parameter. This more suitable for an itest, although it partially redoes the lnd integration test. In loop we can assume the lnd part of mpp to be working.

bhandras: Thanks for explaining, I missed that. Maybe if LND mock could be parametrized then we could still test our side a bit better. I agree it is much more convenient with itest though.

loopout_test.go
```diff
@@ -15,6 +15,105 @@ import (
 	"github.com/lightninglabs/loop/test"
 )
 
+// TestLoopOutPayment tests the first part of the loop out process up to the
+// point where the off-chain payments are made.
+func TestLoopOutPayment(t *testing.T) {
```
carlaKC, line 21: `TestLoopOutPaymentSplit`? The current name is a bit vague.

loopdb/store.go
```diff
@@ -318,44 +330,56 @@ func (s *boltSwapStore) createLoop(bucketKey []byte, hash lntypes.Hash,
...
-//
-// NOTE: Part of the loopdb.SwapStore interface.
-func (s *boltSwapStore) CreateLoopIn(hash lntypes.Hash,
-	swap *LoopInContract) error {
+		// From the root bucket, we'll make a new sub swap bucket using
+		// the swap hash.
+		swapBucket, err := rootBucket.CreateBucket(hash[:])
+		if err != nil {
+			return err
+		}
```
carlaKC, line 448: Could this all be pulled out into a `createLoopBucket(rootKey)` call to reduce duplication? Or is that the opposite of what we're going for splitting these out.

loopdb/store.go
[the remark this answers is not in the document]
carlaKC: I don't think it hurts to have it, so our read function's correctness is independent of the write func. But it is pre-existing, and chances of writing to disk incorrectly are slim so fine to leave as is.

cmd/loop/loopout.go
carlaKC: You can specify `channel={chan1,chan2}` in braces then it'll parse correctly, but a single comma separated flag may be more user-friendly, happy with either.

cmd/loop/loopout.go
```diff
@@ -87,6 +89,17 @@ func loopOut(ctx *cli.Context) error {
 		return err
 	}
 
+	// Parse outgoing channel set.
+	chanStrings := strings.Split(ctx.String("channel"), ",")
+	var outgoingChanSet []uint64
+	for _, chanString := range chanStrings {
```
carlaKC, line 95: Is there any need to check that the set of channels provided are unique? I doubt it, but wanted to check that it won't behave strangely with duplicates on the lnd side.

loopout_test.go
joostjager: It doesn't really test the split itself, but really just the payment parameters. Renamed to `TestLoopOutPaymentParameters`

loopdb/store.go
joostjager: Done. Didn't think about this variant.

loopdb/store.go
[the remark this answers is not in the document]
joostjager: Changed to constructor

cmd/loop/loopout.go
joostjager: Ok didn't know that. But the braces pattern isn't very common indeed, so leaving as is

cmd/loop/loopout.go
joostjager: It is checked in `loopd`

carlaKC: I don't know how seriously we account for the alternate client case?

joostjager: The alternate client would still connect to `loopd`?

PR #206, opened 2020-05-19
readme: document api support level

joostjager:


README.md
```diff
@@ -336,3 +336,16 @@ Its location is `~/.loopd/<network>/loop.db`.
...
+### Server
+
+The Loop server API and on-chain scripts are kept backwards compatible as long
+as reasonably possible.
+
+### Client
+
+When breaking changes to the Loop client daemon API are made, old fields will be
+marked as deprecated. Deprecated fields will remain supported until the next
+minor release.
```
guggero, line 351: This is only the case as long as we are in the semver special case of major version `0`. Afterwards breaking changes should only happen in major releases. Not sure if we should clarify this or if it's obvious to those who care.

joostjager: Yes, I think we can change it if that happens. I am wondering why we don't just release major versions? There will always be improvements. @alexbosworth do you have an opinion on this?

guggero: I don't think we should do a major version as long as we have the `-beta` suffix. And as long as lnd (and LN in general) is considered beta (#reckless), we probably shouldn't give users the wrong sense of security by removing the `-beta` suffix in loop.
So probably no major releases for a few more years, IMO.

PR #209, opened 2020-05-23
cmd/loop: fix channel parsing

joostjager:


githorray:
Tested, happily looping again. 🤗

cmd/loop/loopout.go
```diff
@@ -89,15 +89,20 @@ func loopOut(ctx *cli.Context) error {
...
-	for _, chanString := range chanStrings {
-		chanID, err := strconv.ParseUint(chanString, 10, 64)
-		if err != nil {
-			return err
+	if ctx.IsSet("channel") {
+		chanStrings := strings.Split(ctx.String("channel"), ",")
+		for _, chanString := range chanStrings {
+			chanID, err := strconv.ParseUint(chanString, 10, 64)
+			if err != nil {
+				return fmt.Errorf("error parsing \"%v\"",
```
bhandras, line 101: nit: some more contextual info may work better, eg. "error parsing channel id"

joostjager: fixed

PR #210, opened 2020-05-29
loopin: mpp pre-swap probe

joostjager:
Let the loop client create a probe invoice so that the server can do a multi-path probe on it.


loopin.go
```diff
@@ -135,17 +135,55 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+			Value:  lnwire.NewMSatFromSatoshis(swapInvoiceAmt),
+			Memo:   "probe",
+			Expiry: 3600 * 24 * 365,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a cancellable context that is used for monitoring the probe.
```
bhandras, line 157: nit: cancelable

loopin.go
```diff
@@ -209,6 +247,63 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+		ctx, probeHash,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait in the background for the probe to arrive.
+	probeResult := make(chan error, 1)
+
+	go func() {
```
bhandras, line 266: nit:  you could reduce the nesting a bit by extracting this into a closure and go running that separately.

loopin.go
```diff
@@ -209,6 +247,63 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+					log.Infof("Server probe successful")
+					probeResult <- nil
+
+					// Cancel probe invoice so that the
+					// server will know that its probe was
+					// successful.
+					err := lnd.Invoices.CancelInvoice(
+						ctx, probeHash,
+					)
+					if err != nil {
```
bhandras, line 281: Since this error is only logged but doesn't change the execution flow, we could just `Warnf` instead of `Errorf`, wdyt?

loopin.go
```diff
@@ -209,6 +247,63 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+						log.Errorf("Cancel probe invoice: %v", err)
+					}
+
+					return
+
+				case channeldb.ContractSettled:
+					log.Errorf("Impossible that probe " +
```
bhandras, line 288: Even though impossible, will this hang the client, since we never write to `probeResult`? Also, should we panic here instead?

loopin.go
```diff
@@ -209,6 +247,63 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+						"invoice was settled")
+				}
+
+			case err := <-errChan:
+				log.Errorf("Probe invoice: %v", err)
```
bhandras, line 293: nit: We're logging here and where we read from `probeResult` above too.

bhandras (commented):
Some nits, overall looks good.

release_notes.md
```diff
@@ -13,6 +13,11 @@ This file tracks release notes for the loop client.
 
 #### New Features
 
+* Multi-path payment has been enabled for Loop In. This means that it is now possible
+  to replenish multiple channels via a single Loop In request and a single on-chain htlc.
+  This has to potential to greatly reduce chain fee costs. Note that it is not yet possible
+  to select the channels that will be looped in.
```
carlaKC, line 19: nit: note that you can only select the peer, not the channels?

carlaKC (approved):
Tested with server side, LGTM! 

Also nice that there are no in-flight upgrade concerns with this, since the probe is part of creating a new swap.

loopin.go
joostjager: Using that spelling on the rpc is bad enough already.

loopin.go
joostjager: Doesn't a closure result in exactly the same amount of nesting?

loopin.go
joostjager: It is something that should never happen, so I think `Errorf` is most appropriate.

loopin.go
joostjager: I don't think this hangs the client, because after the probe init call returns from the server, we cancel the context and exit this goroutine.

joostjager: `panic` would bring down the whole server. Wouldn't want to risk that if this would happen because of a bug somewhere, especially because the probe result isn't critical.

loopin.go
```diff
@@ -135,17 +135,55 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+	// can be sure that we don't know the preimage.
+	probeHash := lntypes.Hash(sha256.Sum256(swapHash[:]))
+	probeHash[0] ^= 1
+
+	log.Infof("Creating probe invoice %v", probeHash)
+	probeInvoice, err := cfg.lnd.Invoices.AddHoldInvoice(
+		globalCtx, &invoicesrpc.AddInvoiceData{
+			Hash:   &probeHash,
+			Value:  lnwire.NewMSatFromSatoshis(swapInvoiceAmt),
+			Memo:   "probe",
```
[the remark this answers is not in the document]
joostjager: changed

loopin.go
```diff
@@ -135,17 +135,55 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+			Expiry: 3600 * 24 * 365,
```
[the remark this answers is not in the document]
joostjager: Yes it is a lot, especially given that the server currently only tries for 60 seconds max. Changed to one hour.

loopin.go
joostjager: For consistency I changed it to also write to `probeResult` when the invoice is settled

loopin.go
joostjager: log removed

release_notes.md
joostjager: fixed

loopin.go
bhandras: > `panic` would bring down the whole server. Wouldn't want to risk that if this would happen because of a bug somewhere, especially because the probe result isn't critical.

sgtm

loopin.go
bhandras: If I'm not mistaken you can still get rid of the `for` since we return from every case of the select.

bhandras: Ok, discussed this offline, the for loop is required indeed.

bhandras (approved):
LGTM :+1:

PR #214, opened 2020-06-02
loopd: close error channel on shutdown

guggero:
To avoid listeners of the error channel to block after the daemon
has already shut down, we close the channel once we're finished.


guggero:
Discussed offline, can be solved differently.

PR #220, opened 2020-06-06
lndclient: expose basic coin selection RPCs for WalletKitClient

wpaulino:


lndclient/walletkit_client.go
```diff
@@ -47,6 +77,107 @@ func newWalletKitClient(conn *grpc.ClientConn,
...
+func (m *walletKitClient) ListUnspent(ctx context.Context, minConfs,
+	maxConfs uint32) ([]*lnwallet.Utxo, error) {
+
+	rpcCtx, cancel := context.WithTimeout(ctx, rpcTimeout)
+	defer cancel()
+
+	rpcCtx = m.walletKitMac.WithMacaroonAuth(rpcCtx)
+	resp, err := m.client.ListUnspent(rpcCtx, &walletrpc.ListUnspentRequest{
+		MinConfs: int32(minConfs),
+		MaxConfs: int32(maxConfs),
```
joostjager, line 91: This cast is problematic when the caller wants to pass in `math.MaxUint32` (after looking at the function signature).

wpaulino: Yes, I noticed this issue as well on my end. Fixed.

PR #222, opened 2020-06-10
loopout: scale htlc confirmations above threshold

carlaKC:
This change adds scaling of the confirmations the client requires for the server's loop out on chain htlc based on swap amount.


joostjager:
Quick question: can weird things happen if a restart happens with pending swaps and different config parameters? So that a confirmed swap isn't confirmed anymore or vv?

carlaKC:
Closing this for now, not sure it's the right approach and I don't have capacity for it this sprint.

PR #224, opened 2020-06-11
cmd/loop: set exit code 1 on error

### LINKED ISSUE #223: LND compat check failure returning exit code 0
```
2020-06-11 18:46:27.135 [INF] LNDC: Connected to lnd
lnd compatibility check failed: unable to get info for lnd node: rpc error: code = Unimplemented desc = unknown service lnrpc.Lightning
```

This seems wrong and means restart policy using whatever system must be "always" and not "on failure". This error can be returned in many common cases such as lnd not being yet unlocked, etc.

guggero:
Fixes #223.

### DIFF
--- a/cmd/loopd/main.go
+++ b/cmd/loopd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/lightninglabs/loop/loopd"
 )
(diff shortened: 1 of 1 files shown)


cmd/loopd/main.go
```diff
@@ -11,5 +12,6 @@ func main() {
 	err := loopd.Run(cfg)
 	if err != nil {
 		fmt.Printf("loopd exited with an error: %v\n", err)
+		os.Exit(1)
```
joostjager, line 15: One question is why the print statement above doesn't show up in the issue log. Maybe diff version?

guggero: Yes, the error was added in a PR that didn't make it into a release yet.

PR #226, opened 2020-06-12
multi: consume and log sever state updates

carlaKC:
This PR adds a subscription to server side events for each swap and logs the updates as they arrive. This is just a first step while we decide what states we want to push through from the server and how to present them. 

Some things to note: 
- On server restart, we log a "err transport is closing" the error is unfortunately `internal` in the grpc package so it's string matched and silenced so we don't log it on restarts
- The current implementation does not recover from server restarts, we just won't log if there happens to be a server restart. Since we just use it for logging, best effort is ok.


joostjager:
On bullet 2: indeed, we could handle a subscription request for a completed swap by sending back the final state from the db.

swap_server_client.go
```diff
@@ -273,8 +299,109 @@ func (s *grpcSwapServerClient) NewLoopInSwap(ctx context.Context,
...
+	Update looprpc.ServerSwapState
+
+	// Timestamp is the time of the server update.
+	Timestamp time.Time
+}
+
+// LoopInState subscribes to loop in server state and pipes updates into the
+// channel provided.
+func (s *grpcSwapServerClient) LoopInState(ctx context.Context,
+	hash lntypes.Hash, updateChan chan<- *ServerUpdate) (<-chan error,
```
joostjager, line 317: in `lndclient` we generally use a pattern that returns both channels.

executor.go
```diff
@@ -124,6 +124,28 @@ func (s *executor) run(mainCtx context.Context,
...
 			}()
 
+			// Start a goroutine which will consume state updates
+			// from the server and pipe them into our updates
+			// channel.
+			s.wg.Add(1)
+			go func() {
+				defer s.wg.Done()
+
+				errChan, err := newSwap.subscribeServerUpdates(
```
joostjager, line 134: I would move all of this into the swap execution itself. Keep it together. In the future it is likely that we need the server updates for more than just informing the user. For example fast preimage reveal on loop in.

looprpc/client.swagger.json
```diff
@@ -25,6 +21,12 @@
             "schema": {
               "$ref": "#/definitions/looprpcSwapResponse"
             }
+          },
```
carlaKC, line 24: Hm I think I need to downgrade protoc, @guggero loop is still on the older version, right?

guggero: We updated now, I think.

swap_server_client.go
carlaKC: We need to pass this in so that the daemon can consume from the updates channel (see next commit). If we don't do that, we need to return a channel here then fan into another channel; added complexity with not much benefit imo, and we already have this pattern with `statusChan`

executor.go
carlaKC: I'm unsure that the swap needs to know about this subscription. Since we never want to rely on these messages for swap execution (that would make our setup trusted), they're only really ever going to be for monitoring? Eg payment shards arriving, there's nothing the swap execution can do with that extra info, it's just nice for the user to know. 

I'd say that preimage reveal should be a separate endpoint. Avoid the temptation to overload this one subscription.

loopout.go
```diff
@@ -257,6 +257,20 @@ func (s *loopOutSwap) execute(mainCtx context.Context,
 	s.executeConfig = *cfg
 	s.height = height
 
+	// Create context for our state subscription which we will cancel once
+	// swap execution has completed, ensuring that we kill the subscribe
+	// goroutine.
+	subCtx, cancel := context.WithCancel(mainCtx)
+	defer cancel()
```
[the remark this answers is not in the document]
guggero: Ah, no, I guess the main context passed in here already cancels on shutdown.

carlaKC:
> I guess the local database state will be updated as well with a future PR?

Yes, going to do some requirements gathering to see exactly what the client needs (combine server/client state, persist state etc) and then upgrade from just logging.

loopout.go
```diff
@@ -254,9 +257,25 @@ func (s *loopOutSwap) sendUpdate(ctx context.Context) error {
...
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		subscribeAndLogUpdates(
```
joostjager, line 274: Inside `subscribeAndLogUpdates`, it seems that another goroutine is started. Isn't one goroutine enough?

joostjager: Discussed offline that two goroutines is indeed ok. It is in line with how we build other grpc wrappers that convert `Recv` into channel sends

swap_server_client.go
```diff
@@ -275,8 +301,151 @@ func (s *grpcSwapServerClient) NewLoopInSwap(ctx context.Context,
...
+	receive := func() (*ServerUpdate, error) {
+		response, err := resp.Recv()
+		if err != nil {
+			return nil, err
+		}
+
+		return &ServerUpdate{
+			State:     response.State,
+			Timestamp: time.Unix(0, response.TimestampNs),
+		}, err
```
guggero, line 368: nit: explicit nil instead of implicit?

swap_server_client.go
```diff
@@ -275,8 +301,151 @@ func (s *grpcSwapServerClient) NewLoopInSwap(ctx context.Context,
...
+	go func() {
+		defer s.wg.Done()
+
+		for {
+			// Try to receive from our stream. If there are no items
+			// to consume, this call will block. If our stream is
+			// cancelled by the server we will receive an error.
+			response, err := receive()
+			switch err {
+			// If we get a nil error, we fallthrough to deliver
```
guggero, line 400: nit: `fallthrough` is perhaps the wrong word to use here as there's also a keyword that does the exact opposite of what's happening here. Perhaps "If we get a nil error, we proceed with delivering the update we have just received"?

joostjager:
I think it would be good to create an issue for making the update stream resume after a server restart and link it in this PR.

swap_server_client.go
```diff
@@ -275,8 +301,151 @@ func (s *grpcSwapServerClient) NewLoopInSwap(ctx context.Context,
...
+			// the update we have just received.
+			case nil:
+
+			// If we get an EOF error, the server is finished
+			// sending us updates, so we return with a non-nil
+			// error. This is possible for swaps where the server
+			// marks the swap as a success before we do.
+			case io.EOF:
+				log.Trace("server subscription finished")
```
joostjager, line 409: Does `Trace` surface at all? I don't think we have log levels implemented yet

swap_server_client.go
```diff
@@ -275,8 +301,151 @@ func (s *grpcSwapServerClient) NewLoopInSwap(ctx context.Context,
...
+				errChan <- nil
+				return
+
+			// If we receive a non-nil error, we exit.
+			default:
+				// If we get a transport is closing error, we
+				// return a nil error because the server is
+				// restarting.
+				if isErrConClosing(err) {
+					log.Trace("server subscription exit")
```
joostjager, line 420: Same comment about trace. This is probably important to show to the user, so that they know they won't be kept up to date anymore.

swap_server_client.go
```diff
@@ -82,6 +107,7 @@ func newSwapServerClient(cfg *ClientConfig, lsatStore lsat.Store) (
 	return &grpcSwapServerClient{
 		conn:   serverConn,
 		server: server,
+		quit:   make(chan struct{}),
```
joostjager, line 110: Isn't quit already handled through the contexts that are passed in on every method?

swap_server_client.go
joostjager: Maybe it isn't necessary (and also tricky) to do string match and the error can be passed back as is?

carlaKC: I really don't like exposing rpc errors to a user, they're very unfriendly. Would rather return a nicely formatted shutting down error.

joostjager: Ok, yes remapping the error is also a possibility. Although we don't know if it is really the server shutting down or the user's internet having a hick up.

swap_server_client.go
carlaKC: indeed, can just rely on ctx

updates.go
```diff
@@ -0,0 +1,55 @@
...
+		case update := <-subscribeChan:
+			log.Infof("Server update: %v received, "+
+				"timestamp: %v", update.State, update.Timestamp)
+
+		// If we get an error from the server, we check whether it is
+		// due to server exit, or restart, and log this information
+		// for the client. Otherwise, we just log non-nil errors.
+		case err := <-errChan:
+			switch err {
+			case errServerSubscriptionComplete:
```
joostjager, line 36: `case errServerSubscriptionComplete, errServerRestarting:` ?

swap_server_client.go
joostjager: What do you think about this? The current error message seems to narrow for the range of events that can trigger this

carlaKC: Can make it more generic? "subscription failed, no further updates will be sent"

joostjager: Sounds good to me. At least a lot less specific than 'server restart'

updates.go
```diff
@@ -0,0 +1,53 @@
...
+			log.Infof("Server update: %v received, "+
+				"timestamp: %v", update.State, update.Timestamp)
+
+		// If we get an error from the server, we check whether it is
+		// due to server exit, or restart, and log this information
+		// for the client. Otherwise, we just log non-nil errors.
+		case err := <-errChan:
+			switch err {
+			case errServerSubscriptionComplete,
+				errSubscriptionFailed:
```
joostjager, line 37: Isn't `errSubscriptionFailed` just an error that should be logged on the error log level?

PR #230, opened 2020-06-15
loopd: group and clarify swap server parameters

### LINKED ISSUE #225: --insecure could use a bit of elaboration
```
--insecure                                 disable tls
```

I had set it thinking I was otherwise needing tls cert for `loop`-`loopd` communication, which of course interfered with connection with default loop server. Making it clear it's for loop server and not gRPC would be 👌

joostjager:
Config parameters can be confusing because `loopd` is both a client and a server. This PR adds a config section to isolate the loop server parameters.

Fixes #225

### DIFF
--- a/loopd/config.go
+++ b/loopd/config.go
@@ -26,17 +26,22 @@ type lndConfig struct {
 	TLSPath     string `long:"tlspath" description:"Path to lnd tls certificate"`
 }
 
+type loopServerConfig struct {
+	Host  string `long:"host" description:"Loop server address host:port"`
+	Proxy string `long:"proxy" description:"The host:port of a SOCKS proxy through which all connections to the loop server will be established over"`
+
+	NoTLS   bool   `long:"notls" description:"Disable tls for communication to the loop server [testing only]"`
+	TLSPath string `long:"tlspath" description:"Path to loop server tls certificate [testing only]"`
+}
+
 type viewParameters struct{}
 
 type Config struct {
-	ShowVersion    bool   `long:"version" description:"Display version information and exit"`
-	Insecure       bool   `long:"insecure" description:"disable tls"`
-	Network        string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
-	SwapServer     string `long:"swapserver" description:"swap server address host:port"`
-	TLSPathSwapSrv string `long:"tlspathswapserver" description:"Path to swap server tls certificate. Only needed if the swap server uses a self-signed certificate."`
-	RPCListen      string `long:"rpclisten" description:"Address to listen on for gRPC clients"`
-	RESTListen     string `long:"restlisten" description:"Address to listen on for REST clients"`
-	CORSOrigin     string `long:"corsorigin" description:"The value to send in the Access-Control-Allow-Origin header. Header will be omitted if empty."`
+	ShowVersion bool   `long:"version" description:"Display version information and exit"`
+	Network     string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
+	RPCListen   string `long:"rpclisten" description:"Address to listen on for gRPC clients"`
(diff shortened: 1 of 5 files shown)


loopd/config.go
```diff
@@ -26,17 +26,22 @@ type lndConfig struct {
 	TLSPath     string `long:"tlspath" description:"Path to lnd tls certificate"`
 }
 
+type loopServerConfig struct {
+	Host  string `long:"host" description:"Loop server address host:port"`
```
bhandras, line 30: missing punctuation?

loopd/config.go
```diff
@@ -26,17 +26,22 @@ type lndConfig struct {
...
 type Config struct {
-	ShowVersion    bool   `long:"version" description:"Display version information and exit"`
-	Insecure       bool   `long:"insecure" description:"disable tls"`
-	Network        string `long:"network" description:"network to run on" choice:"regtest" choice:"testnet" choice:"mainnet" choice:"simnet"`
-	SwapServer     string `long:"swapserver" description:"swap server address host:port"`
-	TLSPathSwapSrv string `long:"tlspathswapserver" description:"Path to swap server tls certificate. Only needed if the swap server uses a self-signed certificate."`
-	RPCListen      string `long:"rpclisten" description:"Address to listen on for gRPC clients"`
-	RESTListen     string `long:"restlisten" description:"Address to listen on for REST clients"`
-	CORSOrigin     string `long:"corsorigin" description:"The value to send in the Access-Control-Allow-Origin header. Header will be omitted if empty."`
+	ShowVersion bool   `long:"version" description:"Display version information and exit"`
```
bhandras, line 40: missing punctuations

loopd/config.go
joostjager: I was in doubt. Do we or don't we do punctuation? It seems we don't if the sentence is 'short enough' :grimacing:

bhandras: It doesn't really matter imo, but we could be consistent. Checked LND for inspiration and it mostly doesn't have punctuations unless it does :)

guggero: I vote for consistently not using punctuation unless there's more than one sentence (which probably should be avoided anyway).

loopd/utils.go
```diff
@@ -19,10 +19,10 @@ func getClient(config *Config, lnd *lndclient.LndServices) (*loop.Client,
 	}
 
 	clientConfig := &loop.ClientConfig{
-		ServerAddress:   config.SwapServer,
-		ProxyAddress:    config.Proxy,
-		Insecure:        config.Insecure,
-		TLSPathServer:   config.TLSPathSwapSrv,
+		ServerAddress:   config.Server.Host,
```
guggero, line 22: I think we should instead pass in the `*loopServerConfig` struct directly.

loopd/config.go
joostjager: Sounds like a rule. Applied.

loopd/utils.go
joostjager: It creates a circular dependency and pushes down the annotated flags even deeper into the stack. Took a brief look at unifying `/` and `/loopd`, but that outgrows the scope of this pr.

joostjager:
I think this will surely affect end users, but if we want the change (I think it is good) it needs to happen at some point anyway. @alexbosworth what is your thought about this?

alexbosworth:
> I think this will surely affect end users, but if we want the change (I think it is good) it needs to happen at some point anyway. @alexbosworth what is your thought about this?

I prefer not to break end users, but if it must happen some time I guess better sooner than later

joostjager:
We could go through the process of deprecating the flags first, but even for `lnd` we don't do that generally.

What do you think about the change itself?

alexbosworth:
> We could go through the process of deprecating the flags first, but even for `lnd` we don't do that generally.
> 
> What do you think about the change itself?

I don't really understand why it's necessary but maybe I'm missing something

joostjager:
Discussed offline. Virtually everyone will use the default settings anyway. Just users that use `--proxy` will need to update their command line or config file. Seems worth it to clarify these settings for all other users.

Issue #164, opened 2020-03-06
Document update request for loop/out


ShahanaFarooqui:
https://lightning.engineering/loopapi/rest/index.html#post-v1-loop-out

max_swap_routing_fee is in 'sats' rather than 'msats'.

Roasbeef:
I think it's intended that this value be in sats... @joostjager?

PR #231, opened 2020-06-15
looprpc: fix doc on routing fee units

joostjager:
Fixes https://github.com/lightninglabs/loop/issues/164

### DIFF
--- a/looprpc/client.pb.gw.go
+++ b/looprpc/client.pb.gw.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net/http"
 
+	"github.com/golang/protobuf/descriptor"
 	"github.com/golang/protobuf/proto"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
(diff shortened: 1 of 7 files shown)


looprpc/client.pb.gw.go
```diff
@@ -45,6 +48,23 @@ func request_SwapClient_LoopOut_0(ctx context.Context, marshaler runtime.Marshal
 
 }
 
+func local_request_SwapClient_LoopOut_0(ctx context.Context, marshaler runtime.Marshaler, server SwapClientServer, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
```
guggero, line 51: We should also update the versions of the `grpc-gateway`, `genproto` and `grpc` libraries to reflect these changes. Otherwise we could have a divergence between the generated code and the runtime library. Optimally this would be caught by compiler errors but we never know if any silent behavior changes could break our code.

joostjager: Fixed. Only the `grpc-gateway` ref needed an update.

Issue #228, opened 2020-06-13
loopd 0.6.4 fails to start if lnd is in master (0.10.99-beta)


beeforbacon:
After updating to newest LND and LOOPD versions, loopd fails to start: 
```
$ loopd
2020-06-13 14:46:50.595 [INF] LOOPD: Version: 0.6.4-beta commit=
2020-06-13 14:46:50.596 [INF] LNDC: Creating lnd connection to localhost:10009
2020-06-13 14:46:50.598 [INF] LNDC: Connected to lnd
loopd exited with an error: error checking connected lnd version. at least version "v0.10.1, build tags 'signrpc,walletrpc,chainrpc,invoicesrpc'" is required
```

If  make install tags="signrpc walletrpc chainrpc invoicesrpc" is run, then LND fails to start due to version regression unable. 
Attempted to edit line 47 in the version.go in the build directory to reflect "1" to try to brute force loopd to accept the version, but no change.

joostjager:
I can't reproduce this. This is my output for lnd master and loop 0.6.4:

`
2020-06-14 08:12:23.025 [INF] LOOP: Connected to lnd node 'tester' with pubkey 02fd0d67e115a8a02cd4983fa96f1fc784bb2c2a79006a170af24d7ed9602b5745 (version v0.10.99-beta, build tags 'signrpc,walletrpc,chainrpc,invoicesrpc')
`

Are you sure you are connecting indeed to an instance of lnd master?

beeforbacon:
Hi, Joost. $ lnd --version  is master. I have  updated LND and loop with the usual readme method that has not caused issue in the past.  When the tags are downloaded LND fails to spool up because it is in master and being asked to run 0.10.1. If I update LND back using git pull(thus clearing out the tags), it works correctly. So it may not be a loop issue as much as some sort of issue with the tags that I am dealing with. Not sure where to go from there. If the tags are 0.10.1 specific, and I already have master running, should the tags be accepted?

beeforbacon:
So if it is not a loop issue it may be a tag issue. The following is what happens. If this is not a loop issue then feel free to close the issue and I will work with the slack for troubleshooting. Thanks!
`bacon@baconmachine:~$ cd lnd
bacon@baconmachine:~/lnd$ make install tags="signrpc walletrpc chainrpc invoicesrpc"
 Installing lnd and lncli.
[...]

PR #232, opened 2020-06-15
readme: update compatibility table

joostjager:


joostjager:
Tacked on commit to log version to help diagnose issues like https://github.com/lightninglabs/loop/issues/228

PR #184, opened 2020-04-24
support for native segwit loop-in

bhandras:
This PR adds support for native segwit loop-in. When the loop-in htlc is published internally then by default the P2WSH (native segwit) htlc will be selected, otherwise users can pay to any address returned in the `SwapResponse`. The PR also deprecates `SwapResponse.htlc_address` and `SwapStatus.htlc_address`.

PR #236, opened 2020-06-17
loopd: return missing p2wsh loop out address

joostjager:
Probably introduced https://github.com/lightninglabs/loop/pull/184

### DIFF
--- a/loopd/swapclient_server.go
+++ b/loopd/swapclient_server.go
@@ -110,9 +110,10 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 	}
 
 	return &looprpc.SwapResponse{
-		Id:          hash.String(),
-		IdBytes:     hash[:],
-		HtlcAddress: htlc.String(),
+		Id:               hash.String(),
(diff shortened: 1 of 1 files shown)


loopd/swapclient_server.go
```diff
@@ -110,9 +110,10 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 	}
 
 	return &looprpc.SwapResponse{
-		Id:          hash.String(),
-		IdBytes:     hash[:],
-		HtlcAddress: htlc.String(),
+		Id:               hash.String(),
+		IdBytes:          hash[:],
+		HtlcAddress:      htlc.String(),
+		HtlcAddressP2Wsh: htlc.String(),
```
guggero, line 116: Doesn't this now return a `P2SH` address in the `HtlcAddressP2Wsh` field for old swaps?

joostjager: That is okay and intentional. The `HtlcAddress` is deprecated.

PR #237, opened 2020-06-17
lndclient: move to github.com/lightninglabs/lndclient

joostjager:
Extracted code that is shared by multiple projects to its own repository.


go.mod
```diff
@@ -8,16 +8,15 @@ require (
 	github.com/coreos/bbolt v1.3.3
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/golang/protobuf v1.3.2
-	github.com/google/go-cmp v0.3.1 // indirect
 	github.com/grpc-ecosystem/grpc-gateway v1.14.3
 	github.com/jessevdk/go-flags v1.4.0
+	github.com/lightninglabs/lndclient v0.0.0-20200618064231-15628713b198
```
guggero, line 13: We shouldn't use a version of `lndclient` that has a dependency back to the loop repo. That should be removed first, otherwise I guarantee that it will come back and bite us.
Maybe some more code needs to be copied so we can remove the dependency.

joostjager: Oh really, I missed that. Yes that is bad

joostjager: https://github.com/lightninglabs/lndclient/pull/2

PR #238, opened 2020-06-23
loopout: record htlc tx hash

joostjager:


loopdb/migration_04_updates_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/coreos/bbolt"
+	"github.com/stretchr/testify/require"
+)
+
+// TestMigrationUpdates asserts that the swap updates migration is carried out
+// correctly.
+func TestMigrationUpdates(t *testing.T) {
```
bhandras, line 16: q: Should we also test migration failure? Would it be better if migration success/failure would be tested with actual dbs?

loopdb/migration_04_updates_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Fetch the legacy loop out swap and assert that the updates are still
+	// there.
+	outSwaps, err := store.FetchLoopOutSwaps()
+	if err != nil {
+		t.Fatal(err)
+	}
+	outSwap := outSwaps[0]
+	require.Len(t, outSwap.Events, 2)
+	require.Equal(t, StateSuccess, outSwap.Events[1].State)
```
bhandras, line 81: If I understand correctly the idea is that fetch is successful, so we know that the fetched swap contains the data we expect it and we assert only on the state?

loopdb/store.go
```diff
@@ -482,24 +495,30 @@ func (s *boltSwapStore) updateLoop(bucketKey []byte, hash lntypes.Hash,
...
 
 		// Each update for this swap will get a new monotonically
 		// increasing ID number that we'll obtain now.
-		id, err := updateBucket.NextSequence()
+		id, err := updatesBucket.NextSequence()
 		if err != nil {
 			return err
 		}
 
+		updateBucket, err := updatesBucket.CreateBucket(itob(id))
```
bhandras, line 524: nit: just to make it easier to read, I'd call this `nextUpdateBucket`.

loopdb/migration_04_updates_test.go
joostjager: It is tested with actual dbs? We've tried so many styles of migrations over the past year. Which specific variant do you have in mind exactly?

loopdb/migration_04_updates_test.go
joostjager: Yes, it is like a sample that the migration went alright. Of course this test can be extended to test more, but given that we just move the value, I don't think it adds that much

loopdb/migration_04_updates_test.go
bhandras: What I meant is that we could have actual db files for all versions, and do the update tests on those in a sequence. Then test that the end result is key/value equivalent between steps/in the end.

joostjager: Ah, so start with a v1 database instead of a v3 in this case, to cover all migrations in one swoop? Only thing is then that we can't test additions after v1?

loopdb/swapstate.go
```diff
@@ -19,44 +19,44 @@ const (
...
+	StateFailTemporary SwapState = 7
 
 	// StateHtlcPublished means that the client published the on-chain htlc.
-	StateHtlcPublished = 8
+	StateHtlcPublished SwapState = 8
 
 	// StateInvoiceSettled means that the swap invoice has been paid by the
 	// server.
-	StateInvoiceSettled = 9
+	StateInvoiceSettled SwapState = 9
```
carlaKC, line 62: nicee

loopdb/migration_04_updates.go
```diff
@@ -0,0 +1,112 @@
...
+	return nil
+}
+
+// migrateSwapTypeUpdates migrates updates for swaps in the specified bucket.
+func migrateSwapTypeUpdates(rootBucket *bbolt.Bucket) error {
+	var swaps [][]byte
+	err := rootBucket.ForEach(func(swapHash, v []byte) error {
+		// Only go into things that we know are sub-bucket
+		// keys.
+		if v == nil {
```
carlaKC, line 35: `swapBucket := rootBucket.Bucket(swapHash)` like we have below rather than a nil check on value?

loopdb/migration_04_updates.go
```diff
@@ -0,0 +1,112 @@
...
+			swaps = append(swaps, swapHash)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Do not modify inside the for each.
```
carlaKC, line 45: Nit: this comment belongs above the for each imo.

loopdb/migration_04_updates_test.go
```diff
@@ -0,0 +1,92 @@
...
+					Hex("0000000000000001"): Hex("161b252a770e649b01000000000000053900000000000000000000000000000001"),
+					Hex("0000000000000002"): Hex("161b252ab671bdd90200000000000005f10000000000001a9c0000000000000003"),
+				},
+			},
+		},
+	}
+
+	// Restore a legacy database.
+	tempDirName, err := ioutil.TempDir("", "clientstore")
+	if err != nil {
```
carlaKC, line 49: require?

loopdb/migration_04_updates_test.go
```diff
@@ -0,0 +1,92 @@
...
+	tempPath := filepath.Join(tempDirName, dbFileName)
+	db, err := bbolt.Open(tempPath, 0600, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = db.Update(func(tx *bbolt.Tx) error {
+		return RestoreDB(tx, legacyDb)
+	})
+	if err != nil {
+		t.Fatal(err)
```
carlaKC, line 63: if our update fail, `db.Close()` won't run? Unlikely to be an issue though.

loopdb/store.go
```diff
@@ -265,8 +268,18 @@ func deserializeUpdates(swapBucket *bbolt.Bucket) ([]*LoopEvent, error) {
...
 	// events.
 	var updates []*LoopEvent
-	err := stateBucket.ForEach(func(_, v []byte) error {
-		event, err := deserializeLoopEvent(v)
+	err := stateBucket.ForEach(func(k, v []byte) error {
+		if v != nil {
+			return fmt.Errorf("expected state sub-bucket for %x", k)
+		}
+
+		updateBucket := stateBucket.Bucket(k)
```
carlaKC, line 280: Could also remove the `v!=nil` and check `updateBucket!=nil` instead, slightly more intuitive?

loopdb/migration_04_updates.go
joostjager: I think it is the same result, but calling `Bucket` again is slower than just checking `v` for nilness?

loopdb/migration_04_updates.go
joostjager: moved

loopdb/migration_04_updates_test.go
joostjager: fixed

loopdb/migration_04_updates_test.go
joostjager: Moved close to above the assertion

loopdb/store.go
joostjager: good suggestion, fixed

loopdb/store.go
joostjager: fixed

PR #240, opened 2020-06-25
loopin: record htlx tx hash

joostjager:
Fix similar to https://github.com/lightninglabs/loop/pull/238 but for loop in.

Make sure we always keep tracking our own htlc, even if other htlcs are published on-chain.

Also adds a failure state for htlcs with an incorrect amount.

### DIFF
--- a/loopdb/swapstate.go
+++ b/loopdb/swapstate.go
@@ -60,6 +60,10 @@ const (
 	// StateInvoiceSettled means that the swap invoice has been paid by the
 	// server.
 	StateInvoiceSettled SwapState = 9
+
+	// StateFailIncorrectHtlcAmt indicates that the amount of an externally
+	// published loop in htlc didn't match the swap amount.
+	StateFailIncorrectHtlcAmt SwapState = 10
 )
 
 // SwapStateType defines the types of swap states that exist. Every swap state
@@ -126,6 +130,9 @@ func (s SwapState) String() string {
 
 	case StateInvoiceSettled:
 		return "InvoiceSettled"
+
+	case StateFailIncorrectHtlcAmt:
+		return "IncorrectHtlcAmt"
 
 	default:
 		return "Unknown"
--- a/loopin.go
+++ b/loopin.go
@@ -406,20 +406,19 @@ func (s *loopInSwap) waitForHtlcConf(globalCtx context.Context) (
 		return nil, err
 	}
 
-	for {
+	var conf *chainntnfs.TxConfirmation
+	for conf == nil {
 		select {
 
 		// P2WSH htlc confirmed.
-		case conf := <-confChanP2WSH:
+		case conf = <-confChanP2WSH:
 			s.htlc = s.htlcP2WSH
 			s.log.Infof("P2WSH htlc confirmed")
-			return conf, nil
 
 		// NP2WSH htlc confirmed.
-		case conf := <-confChanNP2WSH:
+		case conf = <-confChanNP2WSH:
 			s.htlc = s.htlcNP2WSH
 			s.log.Infof("NP2WSH htlc confirmed")
-			return conf, nil
 
 		// Conf ntfn error.
 		case err := <-confErrP2WSH:
@@ -438,6 +437,8 @@ func (s *loopInSwap) waitForHtlcConf(globalCtx context.Context) (
 			return nil, globalCtx.Err()
 		}
 	}
+
+	return conf, nil
 }
 
 // publishOnChainHtlc checks whether there are still enough blocks left and if
--- a/loopin.go
+++ b/loopin.go
@@ -694,19 +694,23 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
 // update notification.
 func (s *loopInSwap) persistAndAnnounceState(ctx context.Context) error {
 	// Update state in store.
-	err := s.store.UpdateLoopIn(
+	if err := s.persistState(); err != nil {
+		return err
+	}
+
+	// Send out swap update
+	return s.sendUpdate(ctx)
+}
+
+// persistState updates the swap state on disk.
+func (s *loopInSwap) persistState() error {
+	return s.store.UpdateLoopIn(
 		s.hash, s.lastUpdateTime,
 		loopdb.SwapStateData{
 			State: s.state,
 			Cost:  s.cost,
 		},
 	)
-	if err != nil {
-		return err
-	}
-
-	// Send out swap update
-	return s.sendUpdate(ctx)
 }
 
 // setState updates the swap state and last update timestamp.
--- a/loopin.go
+++ b/loopin.go
@@ -9,17 +9,16 @@ import (
 
 	"github.com/btcsuite/btcutil"
 
-	"github.com/lightningnetwork/lnd/chainntnfs"
-	"github.com/lightningnetwork/lnd/channeldb"
-	"github.com/lightningnetwork/lnd/lnwire"
-
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/lightningnetwork/lnd/lnrpc/invoicesrpc"
-
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/loop/loopdb"
 	"github.com/lightninglabs/loop/swap"
+	"github.com/lightningnetwork/lnd/chainntnfs"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnrpc/invoicesrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
 )
 
 var (
@@ -57,6 +56,9 @@ type loopInSwap struct {
 	htlcP2WSH *swap.Htlc
 
 	htlcNP2WSH *swap.Htlc
+
+	// htlcTxHash is the confirmed htlc tx id.
+	htlcTxHash *chainhash.Hash
 
 	timeoutAddr btcutil.Address
 }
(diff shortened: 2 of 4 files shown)


loopin.go
```diff
@@ -406,20 +406,19 @@ func (s *loopInSwap) waitForHtlcConf(globalCtx context.Context) (
 		return nil, err
 	}
 
-	for {
+	var conf *chainntnfs.TxConfirmation
```
bhandras, line 434: nit: Imho the original version was just as clean as this one if not a bit cleaner. Of course it's subjective.

loopin.go
```diff
@@ -690,8 +690,9 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
 	return nil
 }
 
-// persistState updates the swap state and sends out an update notification.
-func (s *loopInSwap) persistState(ctx context.Context) error {
+// persistAndAnnounceState updates the swap state on disk and sends out an
+// update notification.
+func (s *loopInSwap) persistAndAnnounceState(ctx context.Context) error {
```
bhandras, line 744: Wouldn't it be better if we would just have `setState(announce bool)` which would set `lastUpdate`, persist and announce (when requested).

loopin.go
```diff
@@ -694,19 +694,23 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
...
+	if err := s.persistState(); err != nil {
+		return err
+	}
+
+	// Send out swap update
+	return s.sendUpdate(ctx)
+}
+
+// persistState updates the swap state on disk.
+func (s *loopInSwap) persistState() error {
```
bhandras, line 755: nit: I'd probably merge this commit and the previous one.

loopdb/swapstate.go
```diff
@@ -127,6 +131,9 @@ func (s SwapState) String() string {
 	case StateInvoiceSettled:
 		return "InvoiceSettled"
 
+	case StateFailIncorrectHtlcAmt:
```
bhandras, line 134: nit: wdyt about adding External to the name, so it's clear immediately that this refers to external htlc?

loopin_test.go
```diff
@@ -61,6 +62,10 @@ func TestLoopInSuccess(t *testing.T) {
 	// Expect htlc to be published.
 	htlcTx := <-ctx.lnd.SendOutputsChannel
 
+	// Expect the same state to be written again with the htlc tx hash.
+	state := ctx.store.assertLoopInState(loopdb.StateHtlcPublished)
+	require.NotNil(t, state.HtlcTxHash)
```
bhandras, line 67: should we check the hash too?
```
require.NotNil(t, state.HtlcTxHash)
require.Equal(t, htlcTx.Hash(), state.HtlcTxHash)
```

store_mock_test.go
```diff
@@ -215,13 +215,19 @@ func (s *storeMock) assertLoopInStored() {
...
-func (s *storeMock) assertLoopInState(expectedState loopdb.SwapState) {
+// assertLoopInState asserts that a specified state transition is persisted to
+// disk.
+func (s *storeMock) assertLoopInState(
+	expectedState loopdb.SwapState) loopdb.SwapStateData {
+
 	s.t.Helper()
 
 	state := <-s.loopInUpdateChan
 	if state.State != expectedState {
```
bhandras, line 226: nit: could refactor this to use testify.

loopin.go
```diff
@@ -491,7 +520,20 @@ func (s *loopInSwap) publishOnChainHtlc(ctx context.Context) (bool, error) {
...
+	txHash := tx.TxHash()
+	s.log.Infof("Published on chain HTLC tx %v", txHash)
+
+	// Persist the htlc hash so that after a restart we are still waiting
+	// for our own htlc. We don't need to announce to clients, because the
+	// state remains unchanged.
+	//
+	// TODO(joostjager): Store tx hash before calling SendOutputs. This is
+	// not yet possible with the current lnd api.
+	s.htlcTxHash = &txHash
```
bhandras, line 532: nb/nit: consider extracting a method for this, so it's one less place where we update `lastUpdateTime` and persist manually.

loopin.go
joostjager: I need this for the change in the next commit

loopin.go
joostjager: We don't always persist immediately if we set state. Every call to the db will result in a new update row being added.

loopdb/swapstate.go
joostjager: Technically it can also happen because of a bug... But yes, maybe better to rename.

loopin.go
joostjager: I thought this would make it easier to review, especially because there is no rewriting of prior changes

loopdb/swapstate.go
joostjager: Tried it, but it also becomes quite a long name causing early line wrapping in some places. I'd say the failure reason is accurate also without External added?

loopin.go
joostjager: method extracted

loopin_test.go
joostjager: added

store_mock_test.go
joostjager: fixed

loopin.go
carlaKC: Does this need to be split out at all if we always call `persistAndAnnounce`?

loopin.go
bhandras: Indeed, we append a new state with a sequence id. Should state maybe have a "dirty" flag so that we only serialize when the value changed? Would be nice to have one update function which works in every occasion.

loopin.go
bhandras: Another vote for making state updates cleaner :)

loopdb/swapstate.go
bhandras: Should we decide names based on line length or functionality? :speak_no_evil:

joostjager: Hehe, there is always more detail to add to a name. `StateFailIncorrectExternalLoopInHtlcAmt`

joostjager: I'd say both are factors

loopin.go
joostjager: @carlaKC in this PR, `persistState` is also called without announce.

loopin.go
joostjager: Then we should still get more updates than needed? Every change will immediately result in a db update, even if it is not necessary to persist the change yet.

I don't really want to change that mechanism in this PR. It will affect all swap state db actions and possibly also what happens on resume.

loopdb/swapstate.go
carlaKC: Generally I'm on the side of more descriptive names (even if it costs us line length), but in this case I'd leave external off because it _could_ happen with an internal htlc.

loopin.go
```diff
@@ -394,20 +397,35 @@ func (s *loopInSwap) executeSwap(globalCtx context.Context) error {
 func (s *loopInSwap) waitForHtlcConf(globalCtx context.Context) (
 	*chainntnfs.TxConfirmation, error) {
 
+	// Register for confirmation of the htlc. It is essential to specify not
+	// just the pk script, because an attacker may publish the same htlc
+	// with a lower value and we don't want to follow through with that tx.
+	// In the unlikely event that our call to SendOutputs crashes and we
+	// restart, htlcTxHash will be nil at this point. Then only register
```
carlaKC, line 404: Not for this PR, and depending on how quickly this gets addresses in lnd, but would it be possible for us to try to recover our txid from lnd in the case of this crash? Since we know our pksrcipt, we can `listChainTxns` and check out the outputs on the raw tx?

loopin.go
```diff
@@ -445,6 +463,17 @@ func (s *loopInSwap) waitForHtlcConf(globalCtx context.Context) (
...
+	// Store htlc tx hash for accounting purposes. Usually this call is a
+	// no-op because the htlc tx hash was already known. Exceptions are:
+	//
+	// - Old pending swaps that were initiated before we persisted the htlc
+	//   tx hash directly after publish.
+	//
+	// - Swaps that experienced a crash during their call to SendOutputs. In
+	//   that case, we weren't able to record the tx hash.
+	txHash := conf.Tx.TxHash()
+	s.htlcTxHash = &txHash
```
carlaKC, line 475: Is it worth saving state here if our hash previously was not stored?
Thinking about one of these edge cases where our htlc confirms, then we restart and the txid is lost. We then run through this code again, and might pick up another htlc (which we could avoid if we'd saved here). 

This is probably the edge of edge cases though.

loopin.go
joostjager: I believe the real fix here is to separate construction of the tx and publication. With the new utxo locking API in 0.11 that will be possible. Then we'll always first persist the tx hash and then publish and this gap will disappear.

loopin.go
joostjager: I thought about that, but we are indeed already in an unsafe situation with us not having persisted the htlc tx hash in the first place. This would also only be useful if the conf notifications are returned non-deterministically, which doesn't seem to be the case. I think it doesn't outweigh having a fingerprint of this in the database.

loopin.go
bhandras: SGTM

loopdb/swapstate.go
bhandras: How could this happen with internal htlc?

joostjager: I think only if there is a bug.

joostjager: Leaving short(er) name in place. I do agree with the idea of making it specific, but being complete takes a lot of characters. I think an underlying problem is that we share states between loop in and loop out. There will probably be a split somewhere in the future.

PR #242, opened 2020-06-30
multi: expose server message to clients

joostjager:
Surface the initiation message that is received from the server:

```
joost@joost-desktop $ loop out 50000
Max swap fees for 50000 sat Loop Out: 739767 sat
Regular swap speed requested, it might take up to 30m0s for the swap to be executed.
CONTINUE SWAP? (y/n), expand fee detail (x): y
Swap initiated
ID:             3a663f27c122fa19af8a8d8eb76874fc2e59ab21c772d86f2df047a7369d6c7a
HTLC address:   bcrt1qvsmhc5anwv23ar6mqelje8pfj6ks9ugtawc63saq5pmmhptsp7lsh8wksk
Server message: your loop out is being served
```


cmd/loop/loopin.go
```diff
@@ -157,6 +157,7 @@ func loopIn(ctx *cli.Context) error {
 		fmt.Printf("HTLC address (NP2WSH): %v\n", resp.HtlcAddressNp2Wsh)
 	}
 	fmt.Printf("HTLC address (P2WSH): %v\n", resp.HtlcAddressP2Wsh)
+	fmt.Printf("Server message: %v\n", resp.ServerMessage)
```
bhandras, line 160: This will always print. How about printing only if `strings.Trim(resp.ServerMessage, " ") != ""`.

joostjager: Added

looprpc/server.proto
```diff
@@ -72,6 +72,9 @@ message ServerLoopOutResponse {
     bytes sender_key = 3;
 
     int32 expiry = 4;
+
+    // A human-readable message for the user of the loop server.
```
carlaKC, line 76: Looking at this from the client repo, I think "message from the loop server" makes more sense. But really semantics here.

interface.go
```diff
@@ -250,6 +250,17 @@ type LoopInSwapInfo struct { // nolint
 	HtlcAddressNP2WSH btcutil.Address
 }
 
+// LoopOutSwapInfo contains essential information of a loop-out swap after the
+// swap is initiated.
+type LoopOutSwapInfo struct { // nolint
```
carlaKC, line 259: nit: `// nolint:golint` rather than give this struct a full linter free pass.

loopout.go
```diff
@@ -75,7 +75,8 @@ type executeConfig struct {
 
 // loopOutInitResult contains information about a just-initiated loop out swap.
 type loopOutInitResult struct {
-	swap *loopOutSwap
+	swap          *loopOutSwap
+	serverMessage string
```
carlaKC, line 79: idk about adding a struct for our struct rather than just returning `loopOutSwap, string, error`

loopin.go
```diff
@@ -180,7 +187,12 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
 		return nil, fmt.Errorf("cannot store swap: %v", err)
 	}
 
-	return swap, nil
+	swap.log.Infof("Server message: %v", swapResp.serverMessage)
```
carlaKC, line 190: Perhaps if `!=""`, otherwise we log `Server message: `

looprpc/server.proto
joostjager: fixed

loopin.go
joostjager: added

loopout.go
joostjager: I had that first and changed it to this. It felt that the server message got more attention than it is worth, but totally subjective

interface.go
joostjager: fixed

cmd/loop/loopin.go
joostjager: Removed trim part again. It is a bit ugly. I think I will do a follow up pr on the server to clear after all, just in case we want to get rid of it quickly without any trace.

PR #243, opened 2020-07-02
loopdb: add native segwit version bump to prevent client downgrade

carlaKC:
If a client with native segwit swaps downgrades mid-swap, they will
no longer recognize their own swap, and will not be able to complete
the timeout sweep (if required). We create this migration as its own
migration (rather than adding one generic bump version migration) so
that migrations of this type follow our existing pattern of ascending
numbers, and have names that indicate what they are for.


carlaKC:
Already bumped with updates, closing.

Issue #229, opened 2020-06-14
Max_sat_per_byte adjustable and no hurry swaps for 300sats


HTransistor:
### Background
Tried to do a loop out, way too expensive to be useful.
This is what i got in console:
loop out -max_swap_routing_fee=300 250000
Max swap fees for 250000 sat Loop Out: 15861 sat
Regular swap speed requested, it might take up to 30m0s for the swap to be executed.
CONTINUE SWAP? (y/n), expand fee detail (x): x

Estimated on-chain sweep fee:        149 sat
Max on-chain sweep fee:              14900 sat
Max off-chain swap routing fee:      300 sat
Max no show penalty (prepay):        1337 sat
Max off-chain prepay routing fee:    36 sat
Max swap fee:                        625 sat

Describe your issue here.
Is it possible to set the max on-chain sweep fee?
Even though it is a max fee, but I would not like it to reach it in any case. Can I set my own max, that it cancels, when it would exceed it? I mean basically I want to my max pay fee to be in the region of 300sat (I would like to be able to set the maximum of all the listed payfees (especially the most expensive ones).
I mean setting up channels is nothing what I am doing in a hurry. My on-chain tx gets sometimes 1sat/byte so 200sat/tx in 30min (sometimes 2days). I would like to have similiar tx fee for channel funding/loop in / loop out, because then I would have  an increase in speed (for free) and therefore have a reason to use the lightning network. I would have an advantage even in the case when i suddenly want to withdraw my funds.

### Your environment

* version of `lnd` lncli version 0.10.1-beta commit=v0.10.1-beta
* which operating system (`uname -a` on *Nix) Linux dsche-pc 5.6.16-1-MANJARO #1 SMP PREEMPT Wed Jun 3 14:26:28 UTC 2020 x86_64 GNU/Linux

* version of `btcd`, `bitcoind`, or other backend Bitcoin Core version v0.20.0
* any other relevant environment details

### Steps to reproduce

Tell us how to reproduce this issue. Please provide stacktraces and links to code in question.

### Expected behaviour
I would like to be able to set the maximum of all the listed payfees (epecially the most expensive ones).

Tell us what should happen

### Actual behaviour
Can't use loop as it is too expensive (in worst-case).
Tell us what happens instead

joostjager:
[...]

Issue #207, opened 2020-05-19
Add sat_per_byte parameter


bb2k16:
I understand that this was previously requested, and confirmation time parameter was added for loopout. To some extend it worked, but with the limit only around 80 blocks, this is only sufficient to certain extend, but doesnt crater for clients that would rather pay at the exact satoshi per byte who doesn't mind the wait.

is this possible to add or is this a bad idea?

joostjager:
I've added info about whether it is a bad idea to https://github.com/lightninglabs/loop/issues/229. Closing this issue, so we can centralize the discussion there.

PR #248, opened 2020-07-14
loopout: user-specified expiry

joostjager:
This PR allows the user to select the cltv delta of the on-chain htlc, within boundaries set by the client. The selection takes place implicitly through the sweep confirmation target swap parameter.

The range of accepted cltv delta is observable through the `terms` command.

Related: #229 #207


looprpc/server.proto
```diff
@@ -177,3 +206,76 @@ message ServerLoopOutPushPreimageRequest {
 
 message ServerLoopOutPushPreimageResponse {
 }
+
+message SubscribeUpdatesRequest {
```
carlaKC, line 210: Shouldn't these rather be added in the subscribe PR since they're not used?

looprpc/server.proto
```diff
@@ -177,3 +206,76 @@ message ServerLoopOutPushPreimageRequest {
...
+    FAILED_SWAP_DEADLINE = 8;
+
+    // The server could not publish the loop out on chain htlc.
+    FAILED_HTLC_PUBLICATION = 9;
+
+    // The server has published the loop out on chain timeout tx.
+    TIMEOUT_PUBLISHED = 10;
+
+    // The swap has failed for unknown reasons, it will not be completed.
+    UNEXPECTED_FAILURE = 11;
```
carlaKC, line 264: If we are keeping these, need the htlc confirmed state that was just added.

client.go
```diff
@@ -386,6 +390,26 @@ func (s *Client) LoopOut(globalCtx context.Context,
...
+// getExpiry returns an absolute expiry height based on the sweep confirmation
+// target, constrained by the server terms.
+func (s *Client) getExpiry(height int32, terms *LoopOutTerms,
+	confTarget int32) (int32, error) {
+
+	cltvDelta := confTarget
+
+	switch {
+	case confTarget < terms.MinCltvDelta:
+		cltvDelta = terms.MinCltvDelta
```
carlaKC, line 402: could just return inline here then we don't need cltvDelta

bhandras: won't the linter complain?

carlaKC: looks ok?
```
	switch {
	case confTarget < terms.MinCltvDelta:
		return height + terms.MinCltvDelta, nil

	case confTarget > terms.MaxCltvDelta:
		return 0, fmt.Errorf("...")
	}

	return height + confTarget, nil
```

bhandras: oh yes, true!

client.go
```diff
@@ -386,6 +390,26 @@ func (s *Client) LoopOut(globalCtx context.Context,
...
+
+	case confTarget > terms.MaxCltvDelta:
+		return 0, fmt.Errorf("confirmation target %v exceeds maximum "+
```
bhandras, line 405: Alternatively we could just clamp to max just like in the min case?

looprpc/server.proto
bhandras: +1

client.go
joostjager: That is what I first had, but realized that we also need to protect users against sweeping with very high conf targets while the expiry is much sooner

looprpc/server.proto
joostjager: added

client.go
joostjager: fixed

client.go
```diff
@@ -386,6 +390,24 @@ func (s *Client) LoopOut(globalCtx context.Context,
...
 }
 
+// getExpiry returns an absolute expiry height based on the sweep confirmation
+// target, constrained by the server terms.
+func (s *Client) getExpiry(height int32, terms *LoopOutTerms,
+	confTarget int32) (int32, error) {
+
+	switch {
+	case confTarget < terms.MinCltvDelta:
+		return terms.MinCltvDelta, nil
```
carlaKC, line 400: height + MinCltvDelta?

carlaKC:
This is pretty much there for me, just need to fix the unit tests. Still uncertain about adding protos that aren't used in the PR though.

client.go
joostjager: Ah, that is why I didn't do the early retuns I remember now.

joostjager: Fixed

PR #249, opened 2020-07-14
looprpc: split client messages

joostjager:
Preparation for #248 where loop-out specific fields are going to be added.


cmd/loop/main.go
```diff
@@ -136,53 +136,63 @@ func getMaxRoutingFee(amt btcutil.Amount) btcutil.Amount {
...
 
 		maxSwapFee:   btcutil.Amount(quote.SwapFee),
-		maxPrepayAmt: &maxPrepayAmt,
+		maxPrepayAmt: maxPrepayAmt,
 	}
 }
 
-func displayLimits(swapType swap.Type, amt, minerFees btcutil.Amount, l *limits,
-	externalHtlc bool, warning string) error {
+func displayInLimits(amt, minerFees btcutil.Amount,
```
carlaKC, line 183: Do we need the warning string here, looks like we don't use it for loop in?

cmd/loop/main.go
```diff
@@ -136,53 +136,63 @@ func getMaxRoutingFee(amt btcutil.Amount) btcutil.Amount {
...
+type inLimits struct {
+	maxMinerFee btcutil.Amount
+	maxSwapFee  btcutil.Amount
+}
+
+func getInLimits(quote *looprpc.InQuoteResponse) *inLimits {
+	return &inLimits{
+		// Apply a multiplier to the estimated miner fee, to not get
+		// the swap canceled because fees increased in the mean time.
+		maxMinerFee: btcutil.Amount(quote.MinerFee) * 3,
```
carlaKC, line 148: Why has this changed from *100 to *3 for loopin? Perhaps pull the multiplier out and add an explanatory comment.

looprpc/client.proto
```diff
@@ -492,12 +516,7 @@ message QuoteResponse {
...
     An estimate of the on-chain fee that needs to be paid to sweep the HTLC for
-    a loop out or to pay to the HTLC for loop in. If a miner fee of 0 is
-    returned, it means the external_htlc flag was set for a loop in and the fee
-    estimation was skipped. If a miner fee of -1 is returned, it means lnd's
-    wallet tried to estimate the fee but was unable to create a sample
-    estimation transaction because not enough funds are available. An
-    information message should be shown to the user in this case.
+    a loop out or to pay to the HTLC for loop in.
     */
     int64 miner_fee = 3;
```
carlaKC, line 521: while we're splitting these out, we could change the field names to be loop type specific? Since grpc doesn't care if we change names, and people will already have compile-time errors when they upgrade.

Something like `sweep_fee` for out, `htlc_publish_fee` for in.

Perhaps also mention that these fees are sats in the comments. I know it's assumed for on chain, but no harm being over-specific.

looprpc/client.proto
```diff
@@ -419,7 +419,22 @@ message SwapInfoRequest {
 message TermsRequest {
 }
 
-message TermsResponse {
+message InTermsResponse {
+
+    reserved 1, 2, 3, 4, 7;
```
bhandras, line 424: This is to make it compatible with the old `TermsResponse`? Alternatively you could use a `oneof` and loop type specific sub messages. (I'm slightly in favor of the split though)

cmd/loop/main.go
bhandras: I think it didn't, as it's just moved. +1 for not having magic numbers

cmd/loop/main.go
joostjager: Removed. This unified concept has been going on for too long already.

looprpc/client.proto
joostjager: Good suggestion, fixed

cmd/loop/main.go
joostjager: It hasn't changed. Previously `getLimits` was only called for loop out.

looprpc/client.proto
joostjager: Yes, to keep it backwards compatible at runtime.

cmd/loop/main.go
joostjager: I can add a constant, but if it's name will be `multiplier`, it doesn't add much. Suggestions?

bhandras: Yeah, if this is a simple empirical multiplier then the comment we currently have is probably enough.

PR #255, opened 2020-07-20
preliminary work for integrating the v2 HTLC script

bhandras:
This PR lays the ground work for integrating the v2 HTLC script by adding support to for using it in the `swap` module.

PR #257, opened 2020-07-23
multi: use the new htlc v2 scripts for loop in/out with fallback to htlc v1 for old clients

bhandras:
This PR builds (and is rebased) on top of https://github.com/lightninglabs/loop/pull/255 (first two commits)

The PR integrates the new htlc v2 script to loop in/out while keeping compatibility for old clients. Compatibility is ensured by keeping track of the protocol version for swaps, with old swaps defaulting the the latest released protocol.


bhandras:
Looking for concept ACK at this point as the test coverage is not fully extended yet.

loopdb/codec.go
```diff
@@ -1,8 +1,45 @@
...
 // itob returns an 8-byte big endian representation of v.
 func itob(v uint64) []byte {
 	b := make([]byte, 8)
 	byteOrder.PutUint64(b, v)
 	return b
 }
+
+// MarshalProtocolVersion marshals a looprpc.ProtocolVersion value to a byte
+// slice.
+func MarshalProtocolVersion(version looprpc.ProtocolVersion) []byte {
```
carlaKC, line 18: Server side we went for a separate enum for DB when storing protocol version. Reasoning was that a value not written in the DB isn't actually the same as a legacy rpc protcol version (they may be on preimage push version, for example, and we just weren't storing at that time). I was also concerned about a screw up on rpc level leaking through to the DB, although in practice it's unlikely to happen.

loopdb/store.go
```diff
@@ -71,6 +81,8 @@ var (
 	byteOrder = binary.BigEndian
 
 	keyLength = 33
+
+	defaultProtocolVersion = looprpc.ProtocolVersion_USER_EXPIRY_LOOP_OUT
```
carlaKC, line 85: So if protocol version is not written for a swap we use the latest version, couldn't that report a legacy client which isn't on this version yet as supporting it?

loopdb/store.go
```diff
@@ -357,9 +382,22 @@ func (s *boltSwapStore) FetchLoopInSwaps() ([]*LoopIn, error) {
 				return err
 			}
 
+			// Try to unmarshal the protocol version for the swap.
+			// If the protocol version is not stored (which is
+			// the case for old clients), we'll return the
+			// passed defaultProtocolVersion instead.
+			protocolVersion, err := UnmarshalProtocolVersion(
```
halseth, line 389: Do we need to pass in the default version? Could be more readable to just set the default first, then only call the method if anything is found in the bucket.

loopdb/codec.go
bhandras: Please note that using the proto generated enum is preexisting in the loop code.

In this change we don't fall back to the legacy version, but instead to `ProtocolVersion_USER_EXPIRY_LOOP_OUT` (which is the previous one) as the exact previous version does not matter as we only serialize for new swaps, and in that case the saved version will be `ProtocolVersion_HTLC_V2` (see the commit changing the current version).

The reasoning for not creating an extra enum wrapper over the proto generated enums is that it adds extra complexity where imo is not needed.

So to summarize:
- for old swaps: protocol version will be only used to decide between htlc v1/v2 scripts
- for newly created swaps: protocol version will be `ProtocolVersion_HTLC_V2` (see commit changing the current version for completeness)

loopdb/store.go
bhandras: For all we know this is undecidable, but also because we don't actually serialize or send the version for swaps that don't have it serialized, it's just a matter of preference. We can safely change it to legacy as well, as it really only is used to decide between the two script versions.

loopdb/store.go
bhandras: The reason this is passed is to avoid repeating the same checking and defaulting code where we unmarshal the protocol version. Can slice up if you wish but imo it's cleaner to keep together. WDYT?

loopdb/store.go
```diff
@@ -550,6 +596,14 @@ func (s *boltSwapStore) CreateLoopIn(hash lntypes.Hash,
 			return err
 		}
 
+		// Store the current protocol version.
```
carlaKC, line 599: I know I juust did this with labels, 😅 , but now that we have 2 fields that are common the the swap types but not in the swap contract, it may be nice to have a common function which takes the bucket and adds these fields. If we choose not to migrate swap contract, and continue with adding new keys it will be nice to have them in the same place. n/b

loopin.go
```diff
@@ -176,6 +177,7 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
 	}
 
 	swapKit := newSwapKit(
+		looprpc.CurrentProtocolVersion,
```
carlaKC, line 180: nit, linebreak is before 80

loopdb/store.go
halseth: I mean
```go
protocolVersion := default
versionBytes := swapBucket.Get(protocolVersionKey)
if versionBytes != nil {
   protocolVersion, err = UnmarshalProtocolVersion(versionBytes)
   if err != nil {
      return err
   }
}
```
the checks on the passed version bytes could still be performed.

loopdb/codec.go
bhandras: I strongly think it'd better to use proto generated enums but I also decided to instead go with the more common opinion and translate between RPC and DB enums to also keep the code more compatible with the server side. PTAL

loopdb/store.go
bhandras: Decided to instead go with the enum translation and add the the unrecorded similarly as on the server side. PTAL

loopdb/store.go
bhandras: done

loopin.go
bhandras: done

loopin_test.go
```diff
@@ -326,11 +347,18 @@ func testLoopInResume(t *testing.T, state loopdb.SwapState, expired bool) {
 					},
 				},
 			},
-			Hash: testPreimage.Hash(),
+			Hash:            testPreimage.Hash(),
+			ProtocolVersion: storedVersion,
 		},
 	}
 
+	scriptVersion := swap.HtlcV1
```
carlaKC, line 355: If we are setting the script version in the test, why don't we just pass in the htlc version? Since we're not actually testing our version -> htlc type mapping, just duplicating it in a test.

swap.go
```diff
@@ -39,18 +42,25 @@ func newSwapKit(hash lntypes.Hash, swapType swap.Type, cfg *swapConfig,
...
+		contract:        contract,
+		swapType:        swapType,
+		protocolVersion: protocolVersion,
 	}
 }
 
 // getHtlc composes and returns the on-chain swap script.
 func (s *swapKit) getHtlc(outputType swap.HtlcOutputType) (*swap.Htlc, error) {
+	scriptVersion := swap.HtlcV1
+	if s.protocolVersion >= loopdb.ProtocolVersionHtlcV2 {
```
carlaKC, line 58: Won't this set HtlcV2 for s.protocolVersion = `ProtocolVersionUnrecorded`? Which would mean any swap that is resumed after upgrade would have the wrong htlc type.

client.go
```diff
@@ -175,6 +175,7 @@ func (s *Client) FetchSwaps() ([]*SwapInfo, error) {
 
 	for _, swp := range loopOutSwaps {
 		htlc, err := swap.NewHtlc(
+			swap.HtlcV1,
```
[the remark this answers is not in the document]
bhandras: done

loopdb/store.go
```diff
@@ -178,9 +189,18 @@ func NewBoltSwapStore(dbPath string, chainParams *chaincfg.Params) (
 		return nil, err
 	}
 
+	// Convert the RPC porotocol verson to DB protocol version.
```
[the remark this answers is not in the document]
bhandras: done

swap.go
bhandras: Good catch! This was left when changed to db protocol version from proto protocol version.

loopin_test.go
bhandras: done, ptal

swap.go
```diff
@@ -51,6 +51,7 @@ func newSwapKit(hash lntypes.Hash, swapType swap.Type, cfg *swapConfig,
 // getHtlc composes and returns the on-chain swap script.
 func (s *swapKit) getHtlc(outputType swap.HtlcOutputType) (*swap.Htlc, error) {
 	return swap.NewHtlc(
+		swap.HtlcV1,
```
carlaKC: still on their own newlines?

swap/htlc.go
```diff
@@ -32,6 +32,9 @@ type ScriptVersion uint8
 const (
 	// HtlcV1 refers to the original version of the HTLC script.
 	HtlcV1 ScriptVersion = iota
+
+	// HtlcV2 refers to the improved or V2 version of the HTLC script.
```
carlaKC, line 36: sorry to keep nitting thing. "improved or V2" sounds like this represents two different things to me. 
My previous comment was more to briefly describe how the script is different form V1, but non-blocking.

loopdb/store.go
bhandras: I was thinking about this, and I'm unsure if that'd make the code more readable or harder to follow. Maybe I tend to be more on the "keeping them separate" side to keep them explicit rather then hide behind a function.

swap.go
```diff
@@ -39,19 +42,41 @@ func newSwapKit(hash lntypes.Hash, swapType swap.Type, cfg *swapConfig,
...
+	}
+
+	return scriptVersion
+}
+
 // getHtlc composes and returns the on-chain swap script.
 func (s *swapKit) getHtlc(outputType swap.HtlcOutputType) (*swap.Htlc, error) {
 	return swap.NewHtlc(
-		swap.HtlcV1, s.contract.CltvExpiry, s.contract.SenderKey,
+		GetHtlcScriptVersion(s.contract.ProtocolVersion),
```
halseth, line 78: nit: might as well let this method return the script directly now? Maybe we don't need to define `swap.ScriptVersion`

loopdb/store.go
carlaKC: indeed, was on the fence about it myself so let's leave as is.

carlaKC:
New changes LGTM, having protocol version in `SwapContract` cleans things up nicely.

looprpc/version.go
```diff
@@ -0,0 +1,5 @@
+package looprpc
+
+// CurrentProtocolVersion defines the version of the protocol that is currently
+// supported by the loop client.
+const CurrentProtocolVersion = ProtocolVersion_USER_EXPIRY_LOOP_OUT
```
joostjager: Isn't it better to move it to the `loopdb` package, because that are the types that we use internally?

swap.go
joostjager: That would mix protocol version knowledge into the htlc type. I think it is better like it is.

loopout.go
```diff
@@ -223,7 +232,8 @@ func resumeLoopOutSwap(reqContext context.Context, cfg *swapConfig,
 	log.Infof("Resuming loop out swap %v", hash)
 
 	swapKit := newSwapKit(
-		hash, swap.TypeOut, cfg, &pend.Contract.SwapContract,
+		pend.Contract.ProtocolVersion, hash, swap.TypeOut, cfg,
```
joostjager, line 235: This PR really has two changes: store protocol version and add htlc v2 (dependent on stored protocol version). You could order the commits in the same way.

loopin_test.go
```diff
@@ -282,20 +282,48 @@ func testLoopInTimeout(t *testing.T,
 
 // TestLoopInResume tests resuming swaps in various states.
 func TestLoopInResume(t *testing.T) {
-	t.Run("initiated", func(t *testing.T) {
-		testLoopInResume(t, loopdb.StateInitiated, false)
-	})
+	storedVersion := []loopdb.ProtocolVersion{
```
joostjager, line 285: No loop out test?

swap/htlc.go
```diff
@@ -25,8 +26,17 @@ const (
 	HtlcNP2WSH
 )
 
+// ScriptVerison defines the HTLC script version.
+type ScriptVersion uint8
+
+const (
+	// HtlcV1 refers to the original version of the HTLC script.
+	HtlcV1 ScriptVersion = iota
```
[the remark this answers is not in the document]
bhandras: I'd consider constants to be always better than magic numbers (even if in this case these are not magic in any sense).

loopdb/protocol_version.go
```diff
@@ -0,0 +1,104 @@
...
+	// out preimages to the sever to speed up claim.
+	ProtocolVersionPreimagePush ProtocolVersion = 3
+
+	// ProtocolVersionUserExpiryLoopOut indicates that the client will
+	// proposed a cltv expiry height for loop out.
+	ProtocolVersionUserExpiryLoopOut ProtocolVersion = 4
+
+	// ProtocolVersionUnrecorded is set for swaps were created before we
+	// started saving protocol version with swaps.
+	ProtocolVersionUnrecorded ProtocolVersion = math.MaxUint32
```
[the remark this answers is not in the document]
bhandras: Yeah, we don't need, but it's nicer to keep an unrecorded "version". The first incarnation of this PR simply used the protocol version enum generated from the proto definition and I fixed anything that's not recorded to the previous protocol version. That isn't entirely correct either as we have no information about which version of the client started that swap. One could say it's not a problem either as we didn't use the version before for swaps, but it's not nice either.
Now that the separation of proto generated and DB protocol version is done, I'd say we can keep as is as at least this is more extendable whereas the proto generated is fixed.

looprpc/version.go
bhandras: The reason it's in the `looprpc` package is because the constant's source is from proto and also there's no circular dependency that is created this way. I'd say it's better to keep it here.

loopdb/protocol_version.go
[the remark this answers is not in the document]
joostjager: Ah yes indeed, also the client can have older swaps in the db.

loopout.go
bhandras: makes sense, ptal

looprpc/version.go
bhandras: Also decided to remove the `ConvertProtocolVersion` function as we only used it to convert the compile time constant. Added a test to keep constants in sync and sane though.

loopdb/protocol_version.go
```diff
@@ -28,20 +28,24 @@ const (
 	ProtocolVersionPreimagePush ProtocolVersion = 3
 
 	// ProtocolVersionUserExpiryLoopOut indicates that the client will
-	// proposed a cltv expiry height for loop out.
+	// propose a cltv expiry height for loop out.
```
joostjager, line 31: Fix in previous pr?

swap.go
```diff
@@ -39,19 +42,41 @@ func newSwapKit(hash lntypes.Hash, swapType swap.Type, cfg *swapConfig,
...
+	scriptVersion := swap.HtlcV1
+	if protocolVersion != loopdb.ProtocolVersionUnrecorded &&
+		protocolVersion >= loopdb.ProtocolVersionHtlcV2 {
+
+		// Use HTLC v2 script only if we know the swap was initiated
+		// with a client that supports HTLC v2. Unrecorded protocol
+		// version implies that there was no protocol version stored
+		// along side a serialized swap that we're resuming in which
+		// case the swap was initiated with HTLC v1 script.
+		scriptVersion = swap.HtlcV2
```
joostjager, line 69: Can be an early return

loopdb/store.go
```diff
@@ -401,6 +422,18 @@ func (s *boltSwapStore) FetchLoopInSwaps() ([]*LoopIn, error) {
...
+			// If the protocol version is not stored (which is
+			// the case for old clients), we'll assume the
+			// ProtocolVersionUnrecorded instead.
+			contract.ProtocolVersion, err =
+				UnmarshalProtocolVersion(
+					swapBucket.Get(protocolVersionKey),
+				)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
bhandras: Do you mean something like this:
```
func readProtocolVersion(swapBucket) (ProtocolVersion, error) {
        return UnmarshalProtocolVersion(swapBucket.Get(protocolVersionKey)
}
``` 
Would this make the code any more readable or simpler?

swap_server_client.go
```diff
@@ -331,7 +328,7 @@ func (s *grpcSwapServerClient) SubscribeLoopInUpdates(ctx context.Context,
 
 	resp, err := s.server.SubscribeLoopInUpdates(
 		ctx, &looprpc.SubscribeUpdatesRequest{
-			ProtocolVersion: protocolVersion,
+			ProtocolVersion: loopdb.CurrentRPCProtocolVersion,
```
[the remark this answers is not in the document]
bhandras: We don't use the protocol version for these RPCs, they're just forwarded, as we only care about the hash preimage. We can still update as it's more consistent. I'm ok with updating actually. @carlaKC ?

swap_server_client.go
```diff
@@ -268,7 +265,7 @@ func (s *grpcSwapServerClient) PushLoopOutPreimage(ctx context.Context,
 
 	_, err := s.server.LoopOutPushPreimage(rpcCtx,
 		&looprpc.ServerLoopOutPushPreimageRequest{
-			ProtocolVersion: protocolVersion,
+			ProtocolVersion: loopdb.CurrentRPCProtocolVersion,
```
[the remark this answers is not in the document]
bhandras: See my answer below.

loopdb/protocol_version.go
bhandras: done

swap.go
bhandras: done

loopdb/protocol_version.go
```diff
@@ -0,0 +1,76 @@
...
+	// started saving protocol version with swaps.
+	ProtocolVersionUnrecorded ProtocolVersion = math.MaxUint32
+
+	// CurrentRpcProtocolVersion defines the version of the RPC protocol
+	// that is currently supported by the loop client.
+	CurrentRPCProtocolVersion = looprpc.ProtocolVersion_USER_EXPIRY_LOOP_OUT
+
+	// CurrentInteranlProtocolVersionInternal defines the RPC current
+	// protocol in the internal representation.
+	CurrentInternalProtocolVersion = ProtocolVersionUserExpiryLoopOut
```
joostjager: Somehow I keep thinking of a hard cast here instead of always updating two fields...

loopdb/store.go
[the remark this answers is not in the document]
joostjager: I was thinking
```
func readCommonSwapProperties(&contract, swapBucket) error {
   contract.ProtocolVersion, err = UnmarshalProtocolVersion(
	swapBucket.Get(protocolVersionKey),
   )
   return err
}
```

But with only one property it is not very clearly better. It does make it easier to add the second shared property.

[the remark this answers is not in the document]
joostjager: It is non-blocking

swap.go
```diff
@@ -39,19 +42,40 @@ func newSwapKit(hash lntypes.Hash, swapType swap.Type, cfg *swapConfig,
...
+	}
+
+	return swap.HtlcV1
+}
+
 // getHtlc composes and returns the on-chain swap script.
 func (s *swapKit) getHtlc(outputType swap.HtlcOutputType) (*swap.Htlc, error) {
 	return swap.NewHtlc(
-		swap.HtlcV1, s.contract.CltvExpiry, s.contract.SenderKey,
+		GetHtlcScriptVersion(s.contract.ProtocolVersion),
```
joostjager, line 77: It looks like the quote call still always uses htlc v1. I don't think it matters much, but then we need to change it when we eventually phase out v1.

bhandras: Indeed, fixed.

loopdb/protocol_version.go
bhandras: agree, fixed

loopdb/store.go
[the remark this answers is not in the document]
bhandras: I honestly think it's better to keep as is, as it's easier to follow. Unless you insist you like the extracted method better, I'd keep it.

joostjager (approved):
LGTM but blocked on server release

swap/htlc.go
```diff
@@ -207,6 +223,50 @@ func swapHTLCScriptV1(cltvExpiry int32, senderHtlcKey,
...
+
+	builder := txscript.NewScriptBuilder()
+	builder.AddData(receiverHtlcKey[:])
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	builder.AddOp(txscript.OP_NOTIF)
+
+	builder.AddOp(txscript.OP_DUP)
+	builder.AddOp(txscript.OP_HASH160)
+	builder.AddData(input.Ripemd160H(senderHtlcKey[:]))
```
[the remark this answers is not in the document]
bhandras: fixed

loopdb/codec_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestProtocolVersionMarshalUnMarshal tests that marshalling and unmarshalling
+// looprpc.ProtocolVersion works correctly.
+func TestProtocolVersionMarshalUnMarshal(t *testing.T) {
+	t.Parallel()
+
+	testVersions := [...]ProtocolVersion{
```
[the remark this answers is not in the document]
bhandras: :rofl:

swap/htlc.go
```diff
@@ -144,19 +161,19 @@ func NewHtlc(cltvExpiry int32, senderKey, receiverKey [33]byte,
 	}, nil
 }
 
-// SwapHTLCScript returns the on-chain HTLC witness script.
+// SwapHTLCScriptV1 returns the on-chain HTLC witness script.
```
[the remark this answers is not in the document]
bhandras: I'll update this before merge. (sparing a few rebases)

cfromknecht:
@bhandras latest pass looks very much there 🎉

there was one lingering comment i had from the preceding PR which maybe we want to tack on here? 
https://github.com/lightninglabs/loop/pull/255#discussion_r478624667

PR #258, opened 2020-07-24
loopout: allow per-swap confirmation targets for server HTLC

carlaKC:
This PR allows setting of the number of confirmations we require for the loop out htlc that is published by the server on a per-swap basis. This allows setting of higher confirmation targets for larger value swaps (reducing counterparty risk of being double spent). 

On the cli, we fail for 0 values because one would need to specifically set a 0 value (the flag defaults to 1), but rpc side we just bump 0 values up to the default of 1 conf so that older clients don't start failing when the server introduces this new parameter. Conf targets are persisted so that we remember them across restarts.


cmd/loop/loopout.go
```diff
@@ -41,6 +42,12 @@ var loopOutCommand = cli.Command{
 			Name:  "amt",
 			Usage: "the amount in satoshis to loop out",
 		},
+		cli.Uint64Flag{
```
halseth, line 45: nit: split cli changes into its own commit?

cmd/loop/loopout.go
```diff
@@ -41,6 +42,12 @@ var loopOutCommand = cli.Command{
...
+			Name: "htlc_confs",
+			Usage: "the number of of confirmations, in blocks " +
+				"that we require for the htlc extended by the server",
```
halseth, line 48: add "before we reveal the preimage"?

cmd/loop/loopout.go
bhandras: +1

carlaKC: done 👌

cmd/loop/loopout.go
```diff
@@ -41,6 +42,13 @@ var loopOutCommand = cli.Command{
 			Name:  "amt",
 			Usage: "the amount in satoshis to loop out",
 		},
+		cli.Uint64Flag{
+			Name: "htlc_confs",
```
bhandras, line 46: nit: I wonder if naming this `htlc_conf_target` would be better? Just so that it aligns with the `conf_target` flag. Maybe it's even worse as then users might mix the two.

carlaKC: These two have slightly different meaning. The `conf_target` flag is the number of confirmations we target in fee estimations for our sweep vs the `htlc_confs` which is the number of blocks that we definitely will wait for the htlc to confirm on chain.

Perhaps `conf_target` should actually be `sweep_conf_target` but I'm not sure it's worth potentially breaking some scripts that people may have out there by changing  the name.

interface.go
```diff
@@ -64,6 +64,10 @@ type OutRequest struct {
 	// client sweep tx.
 	SweepConfTarget int32
 
+	// HtlcConfTarget specifies the number of confirmations we require for
+	// on chain loop out htlcs.
+	HtlcConfTarget int32
```
halseth, line 69: related to the discussion on the CLI flag name: why do we call it conf target here? Make it sound more like a fee estimation param, maybe we should be consistent in using something like `HtlcRequiredConfs`?

cmd/loop/loopout.go
bhandras: Yeah, maybe `sweep_conf_target` might be a better name for the `conf_target` but also wouldn't change it.

interface.go
carlaKC: oh yeah def, I think my mind just saw sweep target and went with it.

carlaKC:
@halseth renamed that last var, PTAL :)

interface.go
```diff
@@ -175,7 +175,7 @@ type LoopInRequest struct {
 	// call.
 	MaxMinerFee btcutil.Amount
 
-	// HtlcConfTarget specifies the targeted confirmation target for the
+	// HtlcConfirmations specifies the targeted confirmation target for the
```
halseth, line 182: field name not actually changed

interface.go
```diff
@@ -216,7 +216,7 @@ type LoopInQuoteRequest struct {
 	// include the swap and miner fee.
 	Amount btcutil.Amount
 
-	// HtlcConfTarget specifies the targeted confirmation target for the
+	// HtlcConfirmations specifies the targeted confirmation target for the
```
halseth, line 223: same

interface.go
halseth: also, the comment is a bit weird, maybe use something similar to the one for loop out below?

interface.go
halseth: should this one be renamed as well?

looprpc/client.proto
```diff
@@ -182,6 +182,12 @@ message LoopOutRequest {
     */
     int32 sweep_conf_target = 9;
 
+    /*
+    The number of confirmations that we require for the on chain htlc that will
+    be published by the server.
```
halseth, line 187: add "before we reveal the preimage"?

interface.go
carlaKC: ah, this is actually find and replace gone wild, shouldn't have been renamed in the first place, removed.

PR #259, opened 2020-07-24
looprpc: surface all failure reasons for failed swaps

### LINKED ISSUE #252: Surface all client swap states on the client rpc
Currently the following client swap states are all flattened into a single `Failed` state:

* **FailOffchainPayments**: indicates that it wasn't possible to find
routes for one or both of the off-chain payments to the server that
satisfied the payment restrictions (fee and timelock limits).

* **FailTimeout**: indicates that the on-chain htlc wasn't confirmed
before its expiry or confirmed too late (MinPreimageRevealDelta
violated).

* **FailSweepTimeout**: indicates that the on-chain htlc wasn't swept
before the server revoked the htlc. The server didn't pull the
off-chain htlc (even though it could have) and we timed out the
off-chain htlc ourselves. No funds lost.

* **FailInsufficientValue**: indicates that the published on-chain htlc
had a value lower than the requested amount.

* **FailTemporary**: indicates that the swap cannot progress because
of an internal error. This is not a final state. Manual intervention
(like a restart) is required to solve this problem.

* **FailIncorrectHtlcAmt**: indicates that the amount of an externally
published loop in htlc didn't match the swap amount.

This issue is about surfacing these more specific client state.

carlaKC:
This PR surfaces more granular swap failure reasons so that end users are provided with more information regarding why their swap failed. 

An open question with this PR is how we deprecate the old `Failed` field, since there is a risk that existing code that checks for the failed value will not know to update to check for new failure enum values. Options here:
1. Remove the field so that we break at compile time, chosen in this PR because it will fail loudly and is a once off fix
2. Deprecate the field, and add new fields, this risks the change going undetected
3. Leave the failed state as is and add a `FAILURE_REASON` which enriches the failure with our different failure types, but might be inconvenient to support going forward because our db state is now split across two rpc states

@alexbosworth which do you think is preferable? 

Fixes #252


alexbosworth:
> This PR surfaces more granular swap failure reasons so that end users are provided with more information regarding why their swap failed.
> 
> An open question with this PR is how we deprecate the old `Failed` field, since there is a risk that existing code that checks for the failed value will not know to update to check for new failure enum values. Options here:
> 
> 1. Remove the field so that we break at compile time, chosen in this PR because it will fail loudly and is a once off fix
> 2. Deprecate the field, and add new fields, this risks the change going undetected
> 3. Leave the failed state as is and add a `FAILURE_REASON` which enriches the failure with our different failure types, but might be inconvenient to support going forward because our db state is now split across two rpc states
> 
> @alexbosworth which do you think is preferable?
> 
> Fixes #252

Sure there are tradeoffs for these choices, I'd go with deprecating the field and adding new fields. The pace of integration and updating that we see is slow enough that if things break hard the issue is that instead of fixing the issue the support is just dropped

carlaKC:
> if things break hard the issue is that instead of fixing the issue the support is just dropped

Wouldn't it be worse to break at runtime though? Sinking time into debugging vs a compile time error.

alexbosworth:
> > if things break hard the issue is that instead of fixing the issue the support is just dropped
> 
> Wouldn't it be worse to break at runtime though? Sinking time into debugging vs a compile time error.

Is it really breaking? Maybe I don't get what's happening? What breaks exactly at runtime?

carlaKC:
> Is it really breaking? 

If anybody is using the rpc and checking `state==looprpc.FAILED`, this check will no longer detect failed swaps because we're setting our new values. If we just deprecate, this check will still seem ok vs removing it and breaking at compile time so people know to change to the new set of states.

alexbosworth:
> If anybody is using the rpc and checking `state==looprpc.FAILED`, this check will no longer detect failed swaps because we're setting our new values. If we just deprecate, this check will still seem ok vs removing it and breaking at compile time so people know to change to the new set of states.

Ah ok I get it

Yeah I guess it doesn't work with the normal gRPC model of deprecation. What I was thinking is that if you supported the old state modeling in the deprecated field but added in that new field that encapsulates the more detailed states, but I can see how that's trickier

I think hard breaking isn't too bad, but another delineation may be a split between server failure reason and client failure reason. The existing field could be kept for internal failure codes and the new field could be server informed failures? That's what's really being added right, server informed failures?

carlaKC:
> I think hard breaking isn't too bad, but another delineation may be a split between server failure reason and client failure reason. The existing field could be kept for internal failure codes and the new field could be server informed failures? That's what's really being added right, server informed failures?

This isn't the server failure reason, that would be another change and I agree would probably add `server_state` or something like that. The client failure reason is squashed, we this PR just un-squashes it into all the different failures that the client sets.

alexbosworth:
OK

> Existing code that checks for the failed value will not know to update to check for new failure enum values. Options here:

What about just extending the enum to new states and never using the old states? Is that not preferable because it fails silently?

carlaKC:
> Is that not preferable because it fails silently?

Yeah, I think it's worse implementer side.

alexbosworth:
> > Is that not preferable because it fails silently?

Isn't it inherent in an enum that it is going to have states that you don't know about though? If I were implementing it I would show the swap in "unknown state" which is kind of like a soft break

I don't feel super strongly about any of these options, just that my preference is to support implementers by minimizing their update workload

carlaKC:
Thinking more about cases like [this](https://github.com/tdickman/auto-loop/blob/master/autoloop/run.py#L63) where the failure state is checked to determine next action. Although deleting the field wouldn't work in this specific case, since the code references the numerical value rather than the rpc enum. Perhaps just advertising widely is enough, given that we're likely in touch with most implementors.

guggero:
I think option 1 doesn't really work because for most clients there probably is no "compile time". If you use JavaScript for example, you might have a copy of the proto in your project and compile it every once in a while. If you don't notice there's a need to update the copy and compile it, your program will still look for `state==looprpc.FAILED` and that would just never be true anymore.

IMO splitting the failure into a `FAILURE_REASON` isn't too bad. And that would keep backward compatibility.

carlaKC:
> there probably is no "compile time"

Ah I didn't realize this, failure reason def the way to go then. Some sample output:

Monitor:
```
2020-07-23T17:00:20+02:00 LOOP_OUT FAILED (FAILURE_REASON_OFFCHAIN) 0.0001 BTC - bcrt1qsmlm8y4weq4lfpyvuqchsz0x335q48ml6l02gqxnaa3q6m45jmfsqcm76s (cost: server 1337, onchain 0, offchain 0)
2020-07-28T14:01:37+02:00 LOOP_OUT SUCCESS 0.0001 BTC - bcrt1qxufnpgj5ncdff4jy25xdt3x8w3fwhdvsz6gjfpscy0zph28npu2s5kf067 (cost: server 17, onchain 6812, offchain 2)
```

List swaps:
```
{
            "amt": "10000",
            "id": "f7c0ac8f26dddc36e911cf6df4bff2f3360e3042d40363eb6e8a80e4368e1f6b",
            "id_bytes": "f7c0ac8f26dddc36e911cf6df4bff2f3360e3042d40363eb6e8a80e4368e1f6b",
            "type": "LOOP_OUT",
            "state": "FAILED",
            "failure_reason": "FAILURE_REASON_OFFCHAIN",
            "initiation_time": "1595511767354882000",
            "last_update_time": "1595511767404167000",
            "htlc_address": "bcrt1q8765qg25pqe20lqqwrq3k74shjjumxhhjkn4mh8pk95z25w4wqxs32kcx4",
            "htlc_address_p2wsh": "bcrt1q8765qg25pqe20lqqwrq3k74shjjumxhhjkn4mh8pk95z25w4wqxs32kcx4",
            "htlc_address_np2wsh": "",
            "cost_server": "0",
            "cost_onchain": "0",
            "cost_offchain": "0"
        },
               {
            "amt": "10000",
            "id": "730a7e0e9cf1e39dcc9a66ce3edbfdaead9118f6ac24bd8b12bf3a8c8b55f98c",
            "id_bytes": "730a7e0e9cf1e39dcc9a66ce3edbfdaead9118f6ac24bd8b12bf3a8c8b55f98c",
            "type": "LOOP_OUT",
            "state": "SUCCESS",
            "failure_reason": "FAILURE_REASON_NONE",
            "initiation_time": "1595928499836479000",
            "last_update_time": "1595937697281581000",
            "htlc_address": "bcrt1qgwlmkskc0waej0nmy33qzvy7n74ugj8rqh79mx4juc4s4wp28rjqywwslh",
            "htlc_address_p2wsh": "bcrt1qgwlmkskc0waej0nmy33qzvy7n74ugj8rqh79mx4juc4s4wp28rjqywwslh",
            "htlc_address_np2wsh": "",
            "cost_server": "17",
            "cost_onchain": "6812",
            "cost_offchain": "2"
        },
```

PR #263, opened 2020-07-28
multi: add labelling to swaps

carlaKC:
When we start producing automated loops, we will want a way to distinguish these from manually initiated swaps. This PR adds a label field to our swap contract which can be used to identify automatically created loops. Since this is a useful feature for users, we expose this value for clients to set for manual swaps as well. To allow ourselves the ability to reserve future labels for internal use, we use a reserved prefix which swap labels may not start with so that we can extend our list of internal labels without worrying about clients having used the label in the meantime.

There are a few other options for how we add a new field, requiring various degrees of migrations or hackeyness:
1. Migrate `SwapContract` to use TLV: 
\+ solves this issue forever
\- large migration

2. Once off migration to add field: 
\+ solves issue for now
\- does not address root issue

3. Add field under new key in swaphash bucket: 
\+ easy, no migration
\- we have a field in `SwapContract` that is not serialized/deserialized with the contract

4. Add a new common field to swaps, and a new bucket that contains TLV records (or one key one value) with this information: 
\+ no migration
\- splits swap information into different places


labels/labels.go
```diff
@@ -0,0 +1,47 @@
...
+	// If the label is shorter than our reserved prefix, it cannot contain
+	// it.
+	if len(label) < len(Reserved) {
+		return nil
+	}
+
+	// Check if our label begins with our reserved prefix. We don't mind if
+	// it has our reserved prefix in another case, we just need to be able
+	// to reserve a subset of labels with this prefix.
+	if strings.Contains(label[0:len(Reserved)], Reserved) {
```
bhandras, line 42: could be simply `label[0:len(Reserved)] == Reserved`

loopdb/loopin.go
```diff
@@ -109,9 +110,35 @@ func serializeLoopInContract(swap *LoopInContract) (
...
 	return b.Bytes(), nil
 }
 
+// serializeLabel writes a label to disk with its length and value. If the label
+// length is 0, we just write a 0 length.
+func serializeLabel(w io.Writer, label string) error {
+	// Write our label length followed by the value of our label. This will
+	// allow us to optionally include the label field.
+	labelLength := len(label)
+	if err := binary.Write(w, byteOrder, uint8(labelLength)); err != nil {
```
bhandras, line 126: this will limit the max length to 0xFF vs 500 defined above

loopdb/loopin.go
```diff
@@ -181,5 +208,48 @@ func deserializeLoopInContract(value []byte) (*LoopInContract, error) {
...
+
+// deserializeLabel attempts to read an optional label field, serialized with
+// its length followed by value. If no length field is present, the swap was
+// created before we started writing labels, so we return an empty label. If we
+// do have a length field present, we read the label out if it's length is non
+// zero.
+func deserializeLabel(r io.Reader) (string, error) {
+	// Try to read our label length. If we hit an EOF, there are no more
+	// fields present so we can just return an empty contract.
+	var labelLength uint8
```
bhandras, line 227: similarly, you will want to read `uint16` at least

labels/labels.go
```diff
@@ -0,0 +1,47 @@
+package labels
+
+import (
+	"errors"
+	"strings"
+)
+
+const (
+	// MaxLength is the maximum length we allow for labels.
+	MaxLength = 500
```
bhandras, line 10: q: why 500?

loopdb/loopin.go
```diff
@@ -181,5 +208,48 @@ func deserializeLoopInContract(value []byte) (*LoopInContract, error) {
...
+	err := binary.Read(r, byteOrder, &labelLength)
+	if err == io.EOF {
```
bhandras, line 229: This is clever, but this also means that we cannot extend the serialized contract anymore as there's no way to express an empty label right?

looprpc/client.proto
```diff
@@ -190,6 +190,12 @@ message LoopOutRequest {
     lower total swap fee.
     */
     uint64 swap_publication_deadline = 10;
+
+    /*
+    An optional label for this swap. This field is limited to 500 characters
+    and may not be one of the reserved values in loop/labels Reserved list.
```
bhandras, line 196: We could also mention that it cannot start with the `[Reserved]` prefix.

loopdb/loopin.go
carlaKC: An empty label would be written with a 0 length? 

Legacy: EOF
No label: [0]
With Label: [len][label]

If we wanted to extend, we would be able to just add another field with the same format?

bhandras: Yeah, the problem is more theoretical than practical, but if we were to add another lenght/value field then if the client didn't set the first one (the label), then we need to insert it on the fly (or migrate lazily) in order to be able to deserialize correctly.

labels/labels.go
carlaKC: Just seemed like a sane limit to put on them?

looprpc/client.proto
carlaKC: Ah yes indeed, this is leftover from a previous version.

carlaKC:
>  I'm slightly in favor of serializing under it's own key though as adding it to the end of the contract is equivalent to that. 

I was between those two options, so happy to move over to this approach. We've also run into some issues lnd side with EOF checks before, so probably best to be safe. Only remaining thought here is whether we should bit the bullet and migrate the `SwapContract`, because adding this ad-hoc key now makes for a more involved migration in the future if we need one.

loopdb/store.go
```diff
@@ -60,6 +60,15 @@ var (
 	// value: time || rawSwapState
 	contractKey = []byte("contract")
 
+	// labelKey is the key that stores an optional label for the swap. If
+	// a swap was created before we started adding labels, or was created
+	// without a label, this key will not be present.
+	//
+	// path: loopInBucket/loopOutBucket -> swapBucket[hash] -> labelKey
```
bhandras, line 67: much cleaner! :cool:

cmd/loop/loopin.go
```diff
@@ -24,6 +25,14 @@ var (
 			"confirm within",
 	}
 
+	labelFlag = cli.StringFlag{
+		Name: "label",
+		Usage: fmt.Sprintf("an optional label for this swap,"+
+			"limitmed to %v characters. The label may not start "+
```
bhandras, line 31: s/limitmed/limited

labels/labels.go
```diff
@@ -0,0 +1,46 @@
...
+// Validate checks that a label is of appropriate length and is not in our list
+// of reserved labels.
+func Validate(label string) error {
+	if len(label) > MaxLength {
+		return ErrLabelTooLong
+	}
+
+	// If the label is shorter than our reserved prefix, it cannot contain
+	// it.
+	if len(label) < len(Reserved) {
```
halseth, line 34: can the valid label be empty?

carlaKC: yes, in the case where a user doesn't provide a swap label, we can allow empty ones.

PR #265, opened 2020-08-11
liquidity: add liquidity manager with updatable runtime parameters

carlaKC:
This PR is part [1 of 2] for adding swap suggestions, part of the greater autoloop-saga:
1. Add liquidity manager boilerplate**
2. Swap recommendations endpoint

---

This defines a desired liquidity situation using the following:
1. Target: an entity that we apply a liquidity rule to, this PR specifies the following targets
 - `Node`: asses total liquidity across all channels
 - `Peer`: asses liquidity on a per-peer level
 - `Channel`: asses liquidity of each channel individually
 Future iterations of this subsystem will allow setting of custom targets using channelID/pubkey, which will overwrite the rules set by the generalized targets (or can be set without generalized targets). 

2. Rules: a set of desired conditions for liquidity, this PR implements threshold percentages for incoming/outgoing (future possible rules: threshold incoming/outgoing amount, exclude a target)


loopd/config.go
```diff
@@ -34,6 +38,18 @@ type loopServerConfig struct {
 	TLSPath string `long:"tlspath" description:"Path to loop server tls certificate [testing only]"`
 }
 
+type autoLoopConfig struct {
```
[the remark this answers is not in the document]
carlaKC: Yeah I was also wondering about this. Perhaps a rename to `liquiditymanager` would make sense?

Follow up PR would have the opt-in bools for loop in /out, but I think the rename would make it more clear regardless. We still want people who don't want autoloop to set thresholds and use the static endpoint (with false opt-in bools).

carlaKC:
> should this and follow-up PRs with the "full" automation be released in one go?

I think it makes sense to design them so that they can be separately released, so that we can get early feedback. Will definitely make the config names more clear as you suggest!  

> maybe add to LiT

This would be the bestt, but I'm not sure the timing for V2 of LiT will work out. Will chat to @jamaljsr though 🚀

halseth:
High level this looks very good! I would maybe suggest trying to break the PR up a bit, since in total it gets quite large.

The crucial logic to review to begin with is around liquidity calculations and swap suggestions (this is also what is what would be most debated how to do in a design doc). If we add the minimal amount of coded to test and review this (unit tests, logging of the results perhaps), we could add the simpler, more massive parts as a follow-up: RPC endpoints to get and set config, RPC to get suggested swaps, CLI commands.

carlaKC:
Thanks for having a look @halseth!

> we could add the simpler, more massive parts as a follow-up: RPC endpoints to get and set config, RPC to get suggested swaps, CLI commands.

SGTM, def a better way to split it up.

liquidity/threshold_rule.go
```diff
@@ -0,0 +1,229 @@
...
+// This function calculates the portion of our total capacity we should shift
+// from the surplus side to deficit side without unbalancing the surplus side.
+// This is important, because we do not want to recommend a swap in one
+// direction that just results in our needing to produce a swap in the other
+// direction.
+//
+// If we do have enough surplus, we aim to fall in the midpoint of our two
+// minimums so that we do not tip our channel balance too far in one direction
+// or the other.
+func calculateSwapAmount(capacity, deficitCurrent, deficitMinimum,
```
joostjager, line 189: `deficitMinimum` as a name confused me a bit. I read it as the 'minimally required deficit', while it really is 'minimum balance on the deficit side'?.

liquidity/threshold_rule.go
```diff
@@ -0,0 +1,229 @@
...
+	// Get the maximum shift allowed on our surplus side that is possible
+	// before it dips beneath its minimum.
+	available := surplusCurrent - surplusMinimum
+
+	// If the amount of surplus we have is less than the minimum amount we
+	// need to address our deficit, we split the excess between our
+	// directions so that we move towards our desired amount, even if we
+	// can't reach our threshold.
+	if available <= required {
+		return available / 2
```
joostjager, line 205: I find this function not so easy to understand, but maybe I think about the problem in a too simplistic way. One question that I have is whether this function is always continuous, no steps?

Isn't the swap amount not just the `midpointDelta`, capped by `available`?

looprpc/client.proto
```diff
@@ -666,3 +697,124 @@ message LsatToken {
...
+    /*
+    Additional reasoning to accompany our action.
+    TODO(carla): make this an enum
+    */
+    string reason = 1;
+
+    /*
+    The set of recommended loop outs.
+    */
+    repeated LoopOutParameters loop_out= 2;
```
joostjager, line 798: Would it be strange to directly return a `LoopOutRequest` message here? To make the gap to auto-loop as close as it can be?

looprpc/client.proto
```diff
@@ -666,3 +697,124 @@ message LsatToken {
...
+    current config,so fields that you do not want to change should be set to
+    their current values.
+    */
+    LiquidityConfig config = 1;
+}
+
+message SuggestSwapsRequest{
+}
+
+message SuggestSwapsResponse{
```
joostjager, line 771: Has a discussion been started on how to possibly integrate this into Lightning Terminal?

looprpc/client.proto
```diff
@@ -104,6 +104,37 @@ service SwapClient {
...
+
+    /*
+    SuggestSwaps returns a list of recommended swaps based on the current
+    state of your node's channels and the autolooper configuration set.
+    */
+    rpc SuggestSwaps(SuggestSwapsRequest) returns (SuggestSwapsResponse){
+        option (google.api.http) = {
+            get: "/v1/auto/swaps"
+        };
+    }
```
joostjager, line 137: Add comments that these calls are experimental and can change any time?

looprpc/client.proto
```diff
@@ -666,3 +697,124 @@ message LsatToken {
...
+    /*
+    The target to apply liquidity rules to.
+    */
+    LiquidityTarget target = 2;
+
+    /*
+    A rule to be applied to the target. This value may not be set if setting
+    target to none, and must be supplied otherwise.
+    */
+    LiquidityRule rule = 3;
```
joostjager, line 719: Is it necessary to have a separate rule message already? I am worried about trying to design for something that may work out different anyway

looprpc/client.proto
```diff
@@ -666,3 +697,124 @@ message LsatToken {
...
+    NONE = 0;
+    NODE = 1;
+    PEER = 2;
+    CHANNEL =3;
+}
+
+enum LiquidityRuleType{
+    UNKNOWN = 0;
+    THRESHOLD = 1;
+}
```
joostjager, line 732: Necessary?

looprpc/client.proto
```diff
@@ -666,3 +697,124 @@ message LsatToken {
...
+
+    /*
+    The target that the rule was applied to.
+    */
+    LiquidityTarget target = 2;
+
+    /*
+    A map of target ID to suggested swaps.
+    */
+    map<string, SwapSuggestion> swaps = 3;
```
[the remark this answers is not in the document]
carlaKC: The reasoning here was that we may want to expand to allow people to set specific rules for a channel/peer. Which would look something like `loop setrule 029f9d70dfdae96333bae711eb23dce5e13188bcbc6463e73eac19a513361afcd6 {rule values}`. I thought it would be nice to display per-target, especially for custom rules, which is why it's exposed, wdyt?

looprpc/client.proto
carlaKC: I think we should design the proto in a way that lets us add more rules easily. I want to add the ability to exclude a peer/channel (requirement for most users I've spoken to) as another rule type. Perhaps we want to add absolute amounts rather than ratios etc.

liquidity/threshold_rule.go
```diff
@@ -0,0 +1,229 @@
...
+	default:
+		return nil, fmt.Errorf("unknown swap type: %v", *swapType)
+	}
+
+	// At this stage, we know that we need to perform a swap, and we know
+	// the amount of our total capacity that we need to move. Before we
+	// proceed, we do a quick check that the amount we need to move is more
+	// than the minimum swap amount, returning if it is not.
+	if amt < restrictions.MinimumAmount {
+		return newSwapSet(ReasonMinimumAmount, nil), nil
```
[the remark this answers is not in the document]
carlaKC: I specifically wanted these for the case where we do not recommend any action. If you run a suggestions endpoint and it returns an empty response, it's not unreasonable to wonder whether it worked at all. We can send errors for some of these, but the case where liquidity is ok (for example) is difficult to express without some reason field.

looprpc/client.proto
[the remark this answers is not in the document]
joostjager: Yes it may develop in that direction. But may comment was just about the data type of a multi-formatted string. Isn't there a stronger typed alternative, like multiple maps with different key types?

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,233 @@
...
+		newRule.MinimumInbound = uint32(ctx.Int("mininbound"))
+		newRule.Type = looprpc.LiquidityRuleType_THRESHOLD
+	}
+
+	if ctx.IsSet("minoutbound") {
+		newRule.MinimumOutbound = uint32(ctx.Int("minoutbound"))
+		newRule.Type = looprpc.LiquidityRuleType_THRESHOLD
+	}
+
+	ruleSet := newRule.Type != looprpc.LiquidityRuleType_UNKNOWN
```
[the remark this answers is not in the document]
carlaKC: To reserve the default 0 value as a state that we do not recognize. Without an unknown state, we assume a type when it could just be that the caller did not set it.

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,233 @@
...
+		return err
+	}
+
+	// If our config was not set before this, we set it to a non-nil value
+	// now.
+	if cfg == nil {
+		cfg = &looprpc.LiquidityConfig{}
+	}
+
+	if ctx.IsSet("inclprivate") {
```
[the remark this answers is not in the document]
carlaKC: `--inclprivate=false` works for bool flag

liquidity/swaps.go
```diff
@@ -0,0 +1,161 @@
...
+	maxSwapAmount lnwire.MilliSatoshi) []SwapRecommendation {
+
+	var swaps []SwapRecommendation
+
+	for _, channel := range channels {
+		availableAmt := channel.amount
+
+		// If the available amount is smaller than the minimum amount we
+		// can swap, we cannot use this chanel.
+		if availableAmt < minSwapAmount {
```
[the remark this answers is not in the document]
carlaKC: Going to leave for a follow up 👍

client.go
```diff
@@ -143,13 +146,45 @@ func NewClient(dbDir string, cfg *ClientConfig) (*Client, func(), error) {
...
+
+			inTerms, err := swapServerClient.GetLoopInTerms(ctx)
+			if err != nil {
+				return nil, err
+			}
+
+			return liquidity.NewRestrictions(
+				inTerms.MinSwapAmount, inTerms.MaxSwapAmount,
+			), nil
+		},
```
joostjager, line 175: Isn't it possible to just pass in `swapServerClient`?

client.go
```diff
@@ -246,6 +281,16 @@ func (s *Client) Run(ctx context.Context,
 	mainCtx, mainCancel := context.WithCancel(ctx)
 	defer mainCancel()
 
+	// Start our liquidity manager. This loop will exit when our main ctx
+	// is cancelled.
+	s.wg.Add(1)
+	go func() {
+		if err := s.manager.Run(mainCtx); err != nil {
```
joostjager, line 288: Isn't it the case that with just suggestions, that the liquidity manager only needs to react to rpc calls? I am wondering why it needs its own thread. Or is that only to synchronize the config? If we move to persistent storage of that, the sync would be handled by the db.

client.go
```diff
@@ -644,6 +644,21 @@ func (s *Client) LoopInQuote(ctx context.Context,
...
+// GetLiquidityConfig gets the currently set parameters for our liquidity
+// manager.
+func (s *Client) GetLiquidityConfig(ctx context.Context) (*liquidity.Parameters,
+	error) {
+
+	return s.manager.GetParameters(ctx)
+}
+
+// SetLiquidityParameters attempts to set our liquidity manager's parameters.
+func (s *Client) SetLiquidityParameters(ctx context.Context,
```
joostjager, line 656: Naming inconsistency, "config" vs "parameters"?

client.go
```diff
@@ -644,6 +644,21 @@ func (s *Client) LoopInQuote(ctx context.Context,
 	}, nil
 }
 
+// GetLiquidityConfig gets the currently set parameters for our liquidity
+// manager.
+func (s *Client) GetLiquidityConfig(ctx context.Context) (*liquidity.Parameters,
+	error) {
+
+	return s.manager.GetParameters(ctx)
```
joostjager, line 652: Is this pass through really needed or can `swapClientServer` also invoke manager directly?

looprpc/client.proto
joostjager: But with the proto structure as is, it doesn't seem to be possible to add another rule?

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,196 @@
...
+		Config: cfg,
+	}
+
+	_, err = client.SetLiquidityConfig(context.Background(), req)
+	return err
+}
+
+var setLiquidityRuleCommand = cli.Command{
+	Name:  "setrule",
+	Usage: "set liquidity manger rule for a target",
```
joostjager, line 93: This does suggest there can be multiple targets active?

cmd/loop/liquidity.go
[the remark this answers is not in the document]
joostjager: In `lnd`, most of the enums don't have this field. And for some that have it, it means that it is actually unknown. In the case of these config settings, there is no real reason to have an unknown value.

Choosing between raising questions to rpc users about the meaning of 'unknown' vs reminding users about setting a field, I'd prefer the former.

joostjager (commented):
Main comment is still about trying to design for the future. I think that if we do it, we should have pretty solid specs on what that future looks like exactly.

cmd/loop/liquidity.go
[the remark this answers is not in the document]
joostjager: ok

client.go
carlaKC: This needs a refactor at the moment because `swapServerClient` is in the loop package, so we'd have circular dependencies. Could also just move all of this into the loop package (?), or directly import `looprpc` (I don't think this is a good idea, since we have a mock for the interface already).

looprpc/client.proto
carlaKC: The aim here is to allow a single `LiquidityRule` message, which we expect different fields to be set for depending on the type field that's set. The alternative is a `oneOf`, which I think we want to avoid.

> Choosing between raising questions to rpc users about the meaning of 'unknown' vs reminding users about setting a field, I'd prefer the former.

Can't reply to the other [comment](https://github.com/lightninglabs/loop/pull/265#discussion_r478294590) on this because GH is being strange, but I really feel that we should have the unknown zero value to distinguish between default zero value and intentionally set types. We'd never need to for returning values, only to make sure the user knows what they are setting.

client.go
carlaKC: Just to synchronize cfg, reduced this to a lock for now.

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,174 @@
...
+	ArgsUsage:   "shortchanid",
+	Flags: []cli.Flag{
+		cli.IntFlag{
+			Name: "mininbound",
+			Usage: "the minimum percentage of inbound liquidity " +
+				"to total capacity beneath which to " +
+				"recommend loop out to acquire inbound.",
+		},
+		cli.IntFlag{
+			Name: "minoutbound",
```
joostjager, line 52: This, or is `min_balance` / `max_balance` more clear? Maybe it is also good to connect with how LiT communicates about channel state.

liquidity/liquidity.go
```diff
@@ -0,0 +1,120 @@
...
+	// updated at runtime.
+	params *Parameters
+
+	// paramsLock is a lock for our current set of parameters.
+	paramsLock sync.Mutex
+
+	// done is closed when our main event loop is shutting down. This allows
+	// us to cancel requests sent to our main event loop that cannot be
+	// served.
+	done chan struct{}
```
joostjager, line 81: Unused in this commit

client.go
```diff
@@ -65,6 +66,8 @@ type Client struct {
 	sweeper     *sweep.Sweeper
 	executor    *executor
 
+	*liquidity.Manager
```
joostjager, line 69: I wouldn't expect this to be embedded in `Client`, but rather an object that exists side-by-side with `Client`? This is also what I meant in https://github.com/lightninglabs/loop/pull/265#discussion_r478287585

cmd/loop/liquidity.go
guggero: I don't think "balance" is the best terminology to use here since that implies absolute values to me instead of percentages. Since it's a threshold rule, it could be called `inbound_threshold`/`outbound_threshold` for example?

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,174 @@
...
+		return fmt.Errorf("could not parse channel ID: %v", err)
+	}
+
+	client, cleanup, err := getClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	// We need to set all values in the config that we params the server to
```
guggero, line 83: nit: that first sentence got scrambled somehow, not sure what you want to say here :)

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func setParam(ctx *cli.Context) error {
+	// We require that a channel ID is set for this rule update.
+	if ctx.NArg() != 1 {
+		return fmt.Errorf("please set a channel id for the rule " +
+			"update")
+	}
+
+	chanID, err := strconv.ParseInt(ctx.Args().First(), 10, 64)
```
guggero, line 72: Can be `ParseUint` as channel IDs can never be smaller than 0.

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,174 @@
...
+		ruleSet     bool
+		otherRules  []*looprpc.LiquidityRule
+	)
+
+	// Run through our current set of rules and check whether we have a rule
+	// currently set for this channel. We also track a slice containing all
+	// of the rules we currently have set for other channels, because we
+	// want to leave these rules untouched.
+	for _, rule := range params.Rules {
+		rule := rule
```
guggero, line 105: I think this is only necessary if you call a goroutine in the loop.

guggero (commented):
I think it was a wise decision to split the PR. Did a second pass, looks pretty good. I like the abstractions and naming, makes more sense to me now than auto loop before.

carlaKC:
Last set of changes [e2640ed ...f341a5b](https://github.com/lightninglabs/loop/compare/2e7cc2fd1545ca2ee439552f1b25610aa5132d7a..f341a5b36ea0095c2e98e2806519bddc943f575d).

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,176 @@
...
+		return err
+	}
+
+	// If we are setting a rule for this channel (not clearing it), check
+	// that at least one value is set.
+	switch {
+	case inboundSet:
+	case outboundSet:
+	default:
+		return fmt.Errorf("provide at least one flag to set rules or " +
```
joostjager, line 142: `if !inboundSet && !outboundSet`?

cmd/loop/liquidity.go
```diff
@@ -0,0 +1,176 @@
...
+	// Create a new rule which will be used to overwrite our current rule.
+	newRule := &looprpc.LiquidityRule{
+		ChannelId: chanID,
+	}
+
+	if inboundSet {
+		newRule.IncomingThreshold = uint32(
+			ctx.Int("incoming_threshold"),
+		)
+		newRule.Type = looprpc.LiquidityRuleType_THRESHOLD
```
joostjager, line 155: Can be moved to the initializer of `newRule`?

client.go
joostjager: Looks better outside of `Client` now!

liquidity/liquidity.go
```diff
@@ -0,0 +1,125 @@
...
+	"strings"
+	"sync"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+var (
+	// ErrNoParameters is returned when a request is made to lookup manager
+	// parameters, but none are set.
+	ErrNoParameters = errors.New("no parameters set for manager")
```
joostjager, line 19: Maybe we can just start with an empty list of rules and avoid this special case?

joostjager (approved):

guggero (approved):
LGTM :taco:

loopd/swapclient_server.go
```diff
@@ -549,6 +548,103 @@ func (s *swapClientServer) GetLsatTokens(ctx context.Context,
...
+
+	params := liquidity.Parameters{
+		ChannelRules: make(
+			map[lnwire.ShortChannelID]*liquidity.ThresholdRule,
+			len(in.Parameters.Rules),
+		),
+	}
+
+	for _, rule := range in.Parameters.Rules {
+		if rule.ChannelId == 0 {
```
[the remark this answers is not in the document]
carlaKC: moved 👍

carlaKC:
Last few comments addressed [here](https://github.com/lightninglabs/loop/compare/f341a5b36ea0095c2e98e2806519bddc943f575d..81e4005bd2a89765b5b09e693b1666b6fc2d137d). Added non-nil parameters on startup as suggested [here](https://github.com/lightninglabs/loop/pull/265#discussion_r481775258), which means we can remove the pointer entirely for `Parameters`. We still need to deep-copy because of the channels map, but I think it simplifies things a bit.

PR #266, opened 2020-08-12
loopd: allow custom database path

### LINKED ISSUE #241: Feature Request: Allow setting db location

carlaKC:
This PR adds config options which can be used to set a custom location for loopdb:
* `--datadir` allows setting of a custom database path without affecting where we put our logs
* `--loopdir` sets a single directory for logs, database and config, overwriting the other config values if they are present

We could probably achieve this with just `loopdir` but I included `datadir` just incase somebody has a use need to split logs and db.

Fixes #241

### DIFF
--- a/loopd/config.go
+++ b/loopd/config.go
@@ -1,10 +1,13 @@
 package loopd
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/btcsuite/btcutil"
 	"github.com/lightninglabs/loop/lsat"
+	"github.com/lightningnetwork/lnd/lncfg"
 )
 
 var (
@@ -14,6 +17,7 @@ var (
 	defaultLogDirname  = "logs"
 	defaultLogFilename = "loopd.log"
 	defaultLogDir      = filepath.Join(loopDirBase, defaultLogDirname)
+	defaultConfigFile  = filepath.Join(loopDirBase, defaultConfigFilename)
 
 	defaultMaxLogFiles     = 3
 	defaultMaxLogFileSize  = 10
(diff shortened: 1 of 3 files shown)


loopd/config.go
```diff
@@ -86,3 +94,29 @@ func DefaultConfig() Config {
 		},
 	}
 }
+
+// CleanAndExpandPath expands environment variables and leading ~ in the
+// passed path, cleans the result, and returns it.
+// This function is taken from https://github.com/btcsuite/btcd
+func CleanAndExpandPath(path string) string {
```
guggero, line 101: Since we pull in lnd anyway, can't we just use the function `lncfg.CleanAndExpandPath`?

carlaKC: indeed!

carlaKC:
> Perhaps consider not allowing loop and data dir together, but it's good as is too.

Nice idea, done!

carlaKC:
Added config file to this for completeness, if you want to have a look at the changes @bhandras @guggero, had to move some things around.

loopd/run.go
```diff
@@ -109,12 +110,21 @@ func Run(rpcCfg RPCConfig) error {
...
-		return err
+	loopDir := lncfg.CleanAndExpandPath(config.LoopDir)
+	configFile := lncfg.CleanAndExpandPath(config.ConfigFile)
+
+	// If our loop directory is set and the config file parameter is not
+	// set, we assume that they want to point to a config file in their
+	// loop dir. However, if the config file has a non-default value, then
+	// we leave the config parameter as its custom value.
+	if loopDir != loopDirBase {
+		if configFile == defaultConfigFile {
```
bhandras, line 121: nit: you can join the to conditions and spare an `if`, nb

PR #267, opened 2020-08-14
loopd: wait for lnd to be synced before starting

### LINKED ISSUE #215: loopd: wait for lnd to be ready on startup
If `loopd` is started directly after `lnd` it is possible that we'll exit with an error in `ChainNotifier.RegisterBlockEpochNtfn` because the notifier is not ready yet.

A workaround in another project currently solves the problem like this:

```
	err = wait.NoError(func() error {
		ctxt, cancel := context.WithTimeout(
			context.Background(), defaultStartupTimeout,
		)
		defer cancel()

		notifier := g.lndClient.ChainNotifier
		resChan, errChan, err := notifier.RegisterBlockEpochNtfn(
			ctxt,
		)
		if err != nil {
			return err
		}

		// Block until we get a positive/negative answer or the timeout
		// is reached.
		select {
		case <-resChan:
			return nil

		case err := <-errChan:
			return err

		case <-ctxt.Done():
			return fmt.Errorf("wait for chain notifier to be " +
				"ready timed out")
		}
	}, defaultStartupTimeout)
	if err != nil {
		return err
	}
```

But this is not optimal as it creates an error entry in lnd's log every 20 milliseconds.

Additionally, we should also check whether loop can work while lnd's chain is not yet fully synced. Because loop involves on-chain operations, it's probably advisable to also wait until the chain is fully synced.

carlaKC:
This PR updates our lndclient config to block until lnd is fully synced to the graph. Some sample output:
Kill loopd before sync completes. 
```
2020-08-14 09:07:29.517 [INF] LNDC: Waiting for lnd to be fully synced to its chain backend, this might take a while
^C
2020-08-14 09:07:31.973 [DBG] LNDC: Closing lnd connection
2020-08-14 09:07:31.973 [DBG] LNDC: Wait for client to finish
2020-08-14 09:07:31.973 [DBG] LNDC: Wait for chain notifier to finish
2020-08-14 09:07:31.973 [DBG] LNDC: Wait for invoices to finish
2020-08-14 09:07:31.973 [DBG] LNDC: Lnd services finished
loopd exited with an error: error waiting for chain to be synced: context canceled

```

Sync completes:
```
2020-08-14 09:07:47.303 [INF] LNDC: Waiting for lnd to be fully synced to its chain backend, this might take a while
2020-08-14 09:07:52.333 [INF] LNDC: lnd is now fully synced to its chain backend
```

It requires that we bump our lndclient version to use the changes added in lightninglabs/lndclient/pull/7. This version of lndclient requires lnd `v0.11` because it includes the coin selection and sweeps endpoints. We do not bump our required version in this PR, because loop does not make use of these new endpoints. If we do start to use these endpoints, the minimum version **must** be bumped. 

Since we update `sweeper.go`, which the server uses, we will need to update lndclient on server side as well.

Fixes #215


test/lnd_services_mock.go
```diff
@@ -10,7 +10,6 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightningnetwork/lnd/chainntnfs"
-	"github.com/lightningnetwork/lnd/input"
```
bhandras, line 13: nit: could be a separate commit

guggero: I think we generally want each commit to compile on its own. And since these changes were necessary because of the `lndclient` update, they should be in the same commit.

loopd/run.go
```diff
@@ -168,10 +194,14 @@ func Run(rpcCfg RPCConfig) error {
 
 	lisCfg := newListenerCfg(&config, rpcCfg)
 
+	// Start listening for signal interrupts regardless of which command
+	// we are running. When our command tries to get a lnd connection, it
+	// blocks until lnd is synced. We listen for interrupts so that we can
+	// shutdown the daemon while waiting for sync to complete.
+	signal.Intercept()
```
guggero, line 201: This currently has the effect that the view command cannot be canceled anymore because it does not react to the shutdown. But I don't think that matters hugely as it's a command that should normally terminate quickly.

carlaKC: We can still terminate view if lnd isn't synced, which is what we need the interrupt for? 
Could also add intercept checks to the view function between the various calls, but I don' think that's required as you say.

test/lnd_services_mock.go
carlaKC: Yeah, wanted it to compile.

loopd/run.go
guggero: Yeah, you're right, without this you couldn't cleanly shut down the daemon on a view command if lnd wasn't synced. And view probably doesn't take very long anyway. It's not like the server side view command that has to interact with `bitcoind` and could potentially take a while.

carlaKC: Yeah def, if it becomes more long lived we'll need to check for interrupts 👍

PR #268, opened 2020-08-14
loopout: more user friendly temporary server error

carlaKC:
This PR adds special handing for temporary server errors. Since we use the `OutOfRange` error code for bad user params when creating a loop out, and for temporary server errors, we need to string match this error to identify it and provide a more friendly and informative message. As seen in lightninglabs/lightning-terminal/issues/92. 

Old error:
(logs): `LOOPD: LoopOut: cannot initiate swap: rpc error: code = OutOfRange desc = amount above current maximum`
(loop): `[loop] rpc error: code = Unknown desc = cannot initiate swap: rpc error: code = OutOfRange desc = amount above current maximum`

New error:
(logs): `[ERR] LOOPD: LoopOut: cannot initiate swap: temporary server error, please retry later`
(loop): `[loop] rpc error: code = Unknown desc = cannot initiate swap: temporary server error, please retry later`

### DIFF
--- a/loopout.go
+++ b/loopout.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bhandras:
Could a perhaps better approach be making these server side errors more descriptive?

guggero:
This does get rid of the `rpc error: code = OutOfRange desc =` part of the error. Maybe we should in general just unwrap all RPC errors (because the code almost never matters to the user) and adjust the specific error message on the server as @bhandras suggested?

carlaKC:
Yeah indeed, server side makes more sense! I had client side in my head from the meeting for some reason.

carlaKC:
Fixing this server side.

PR #272, opened 2020-08-18
loopd: add network to default config file path

carlaKC:
This PR fixes a regression introduced in #266 which removed the network path from the default location of our config file.

### DIFF
--- a/loopd/run.go
+++ b/loopd/run.go
@@ -137,17 +137,7 @@ func Run(rpcCfg RPCConfig) error {
 
 	// Parse ini file.
 	loopDir := lncfg.CleanAndExpandPath(config.LoopDir)
-	configFile := lncfg.CleanAndExpandPath(config.ConfigFile)
-
-	// If our loop directory is set and the config file parameter is not
-	// set, we assume that they want to point to a config file in their
-	// loop dir. However, if the config file has a non-default value, then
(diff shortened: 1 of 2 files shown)


loopd/run.go
```diff
@@ -237,3 +227,25 @@ func Run(rpcCfg RPCConfig) error {
...
+	// If the config file path provided by the user is set, then we just
+	// use this value.
+	if cfg.ConfigFile != defaultConfigFile {
+		return lncfg.CleanAndExpandPath(cfg.ConfigFile)
+	}
+
+	// If the user has set a loop directory that is different to the default
+	// we will use this loop directory as the location of our config file.
+	// We do not namespace by network, because this is a custom loop dir.
+	if loopDir != loopDirBase {
```
bhandras, line 243: Thinking a bit about this, I think we could still namespace by network to make this even more simple.

bhandras (approved):
Sorry we missed this in the original review. LGTM

joostjager (approved):
Tested and works

loopd/run.go
carlaKC: Also thought about this, but if you're setting your own base dir, I think it makes sense that we leave everything up to the user? I can imagine using base dir to tell loop that everything it needs is in one place, so I wouldn't want to then have to include network in my own setup.

PR #278, opened 2020-08-26
liquidity: add swap suggestions endpoint

carlaKC:
This PR is part [2 of 2] for adding swap suggestions, part of the greater autoloop-saga:
1. Add liquidity manager boilerplate, see #265
2. Swap recommendations endpoint

**Do not review first 3 commits**

---

The suggestions added in this PR are not yet *self aware*. This means that if you get suggestions, swap and then get suggestions again before we lock in htlcs, the same suggestion will be given. We assume that there is a sane human operating these suggestions that is aware that they have just dispatched a swap. 

Some sample output:
```
$ loop suggestswaps
{
	"loop_out": [
		{
			"amt": 1000000,
			"outgoing_chan_set": [
				124244814004224
			]
		}
	]
}

```


carlaKC:
Since the PR was split, copying over some comments that apply to this section:

> [[1](https://github.com/lightninglabs/loop/pull/265#discussion_r475409570)] Is it necessary to return the config every time suggestions are requested?

I think it's nice to have the rule and target there so that you know why the suggestion was made, but we can also cut these fields and add them later if we want?

> [[2](https://github.com/lightninglabs/loop/pull/265#discussion_r475459370)] Would it be strange to directly return a LoopOutRequest message here? 

I think this is a nice idea, have updated the suggestions endpoint to return requests to close the gap. 

> [[3](https://github.com/lightninglabs/loop/pull/265#discussion_r475388886)] To reduce the size of the PR, is it worth only implementing one of these? 

These being `SwapRecommendation` interface for loop out/in. Do you still suggest this after splitting the PR? I don't think it saves us too many lines.

liquidity/balances.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// balances summarizes the state of the balances of a channel. Channel reserve,
+// fees and pending htlc balances are not included in these balances.
+type balances struct {
+	// capacity is the total capacity in all of our channels.
+	capacity btcutil.Amount
+
+	// incoming is the total remote balance across all channels.
```
joostjager, line 16: 'across all channels' doesn't apply anymore

liquidity/threshold_rule.go
```diff
@@ -59,3 +62,115 @@ func (r *ThresholdRule) validate() error {
...
+	}
+
+	switch {
+	// If we have too little inbound and too little outbound, we cannot
+	// do anything to help our liquidity situation. This will happen in the
+	// case where we have a lot of pending htlcs on our channels.
+	case inbound.hasDeficit() && outbound.hasDeficit():
+		return 0, nil
+
+	// If we have too little inbound, but not too little outbound, it is
```
joostjager, line 140: not too little -> sufficient?

liquidity/threshold_rule.go
```diff
@@ -59,3 +62,115 @@ func (r *ThresholdRule) validate() error {
...
+	// TODO(carla): do a best-effort split attempt here?
+	if available < midpointDelta {
+		return 0
+	}
+
+	// Otherwise, we have sufficient surplus to shift our deficit side all
+	// the way to the midpoint. We return the amount we need to reach this
+	// midpoint so that we do not overshoot and produce a deficit in the
+	// opposite direction.
+	return midpointDelta
```
joostjager, line 220: It still seems that the code in this file can be simplified further. I'd expect something like:
```
midpoint = (min_balance + max_balance) / 2
if balance < min_balance { loop_in(midpoint-balance)}
if balance > max_balance {loop_out(balance - midpoint)}

File: liquidity/threshold_rule.go [hunk 4]
```diff
@@ -63,6 +63,51 @@ func (r *ThresholdRule) validate() error {
...
+	// If a nil swap type is returned, we do not need to recommend any swaps
+	// so we return an empty swap set.
+	if swapType == nil {
+		return nil, nil
+	}
+
+	// We currently only support loop out, so return nil if a loop in is
+	// recommended.
+	switch *swapType {
+	case swap.TypeOut:
```
joostjager, line 85: Technically the code could be simplified further to return just a bool from shouldswap instead of a swap type.

liquidity/threshold_rule.go
```diff
@@ -63,6 +63,51 @@ func (r *ThresholdRule) validate() error {
...
+	// the amount of our total capacity that we need to move. Before we
+	// proceed, we do a quick check that the amount we need to move is more
+	// than the minimum swap amount, returning if it is not.
+	if amount < outRestrictions.Minimum {
+		return nil, nil
+	}
+
+	// If we exceed our maximum amount, then reduce to the maximum amount we
+	// can dispatch.
+	if amount > outRestrictions.Maximum {
```
joostjager, line 104: More efficient with a switch statement

liquidity/liquidity.go
```diff
@@ -25,6 +30,9 @@ type Config struct {
 	// LoopOutRestrictions returns the restrictions that the server applies
 	// to loop out swaps.
 	LoopOutRestrictions func(ctx context.Context) (*Restrictions, error)
+
+	// LND provides us with access to lnd's main rpc.
+	LND lndclient.LightningClient
```
joostjager, line 35: This is pretty aggressive. Is that that annoying linter rule again?

liquidity/threshold_rule.go
```diff
@@ -59,3 +62,115 @@ func (r *ThresholdRule) validate() error {
...
+	surplusSide balanceRequirement) btcutil.Amount {
+
+	// Get the maximum shift allowed on our surplus side that is possible
+	// before it dips beneath its minimum.
+	available := surplusSide.currentAmount - surplusSide.minimumAmount
+
+	// Get the midpoint between our two minimum amounts, expressed from
+	// the perspective of the side that we currently have a deficit of
+	// liquidity.
+	midpoint := (deficitSide.minimumAmount +
```
joostjager, line 201: We only do loop out now, but the reason that we still have a minimum inbound is to properly calculate the midpoint? Alternatively the mid point could be set directly, `ideal_balance`.

carlaKC: I think that specifying thresholds allows us to be a bit more expressive as to when to suggest a swap? 

Take the simple case of a channel with 100 capacity, with thresholds one could set the following, both with a midpoint of 50:
A: `MinIncoming=40` `MinOutgoing=40` 
B: `MinIncoming=10` `MinOutgoing=10`

However, they have quite different meanings, A wants to swap whenever incoming drops below 40%, whereas B is happy to wait till they get all the way down to 10%. 

We also manage to capture the idea of a "sends reserve" with a `minOutbound` which (I think) is lost when we set a single value?

liquidity/liquidity.go
carlaKC: Meaning `LND` vs `Lnd`, no linter problems, just conventional go?

liquidity/threshold_rule.go
carlaKC: Simplified this a bit, but we still do need to do the available amount checks. It's not the case that we can always just swap `midpoint-balance` because if we have some pending htlcs, this may tip us below our outgoing threshold anyway.

liquidity/threshold_rule.go
```diff
@@ -59,3 +61,58 @@ func (r *ThresholdRule) validate() error {
 
 	return nil
 }
+
+// loopOutSwapAmount determines whether we can perform a loop out swap, and
+// returns the amount we need to swap to reach the desired liquidity balance
+// specified by the incoming and outgoing thresholds.
+func loopOutSwapAmount(balances *balances, incomingThreshold,
```
joostjager, line 96: Maybe `minimumIncomingPerc` is a more consistent name

liquidity/threshold_rule.go
```diff
@@ -59,3 +61,58 @@ func (r *ThresholdRule) validate() error {
...
+	// cannot take any further action.
+	case balances.outgoing <= minimumOutgoing:
+		return 0
+
+	}
+
+	// Express our minimum outgoing amount as a maximum incoming amount.
+	// We will use this value to limit the amount that we swap, so that we
+	// do not dip below our outgoing threshold.
+	maxIncoming := balances.capacity - minimumOutgoing
```
joostjager, line 122: inconsistency `minimum` vs `max` (short)

liquidity/threshold_rule.go
```diff
@@ -59,3 +61,58 @@ func (r *ThresholdRule) validate() error {
...
+	available := balances.outgoing - minimumOutgoing
+
+	// If we do not have enough balance available to reach our midpoint, we
+	// take no action. This is the case when we have a large portion of
+	// pending htlcs.
+	if available < required {
+		return 0
+	}
+
+	return required
```
joostjager, line 145: This function reads very well

liquidity/liquidity.go
joostjager: We already have for example `LndServices`

liquidity/liquidity.go
```diff
@@ -123,3 +131,53 @@ func (m *Manager) SetParameters(params Parameters) error {
...
+
+	m.paramsLock.Lock()
+	defer m.paramsLock.Unlock()
+
+	if m.params == nil {
+		return nil, ErrNoParameters
+	}
+
+	if len(m.params.ChannelRules) == 0 {
+		return nil, ErrNoRules
```
joostjager, line 149: I'd think no rules means no suggestion but not an error?

joostjager (approved):
LGTM

liquidity/threshold_rule.go
guggero: Why `Perc`? It's an absolute amount, not a percentage. How about `requiredInboundAmount`? Or just leave it the way it is, I think `loopOutSwapAmount` also fits.

loopd/swapclient_server.go
```diff
@@ -645,6 +645,34 @@ func rpcToRule(rule *looprpc.LiquidityRule) (*liquidity.ThresholdRule, error) {
...
+	swaps, err := s.liquidityMgr.SuggestSwaps(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var loopOut []*looprpc.LoopOutRequest
+
+	for _, swap := range swaps {
+		loopOut = append(loopOut,
+			&looprpc.LoopOutRequest{
```
guggero, line 662: nit: could move `&looprpc.LoopOutRequest` to previous line, would get rid of one level of indentation.

guggero (approved):
Very clean PR, looks good to me!
Also tested it on regtest, got them swap suggestions coming in :rocket:

liquidity/threshold_rule.go
joostjager: It does look like a percentage to me. Below it is converted to an absolute amount?

carlaKC:
Last diff [here](https://github.com/lightninglabs/loop/compare/d949c14556e95d59e4c1308c7522c7cdf659328a..ef183fff1588b1c8a9e332be24fb71bb59461b40), just nits for this PR (the changes are on the parent PR).

PR #279, opened 2020-08-26
liquidity: expand swap suggestions to allow custom rules (example)

carlaKC:
This PR provides examples of how the existing swap suggestions PRs can be expanded to allow for more customization. 
**Don't actually look at the code on this, more of an idea/POC**

The previous PRs add the concept of a target/rule, which allows us to examine liquidity per node/peer/channel.
This PR expands this rule set to allow setting of specific rules per peer/channel which override this general target/rule pair.

Take the following example config, I am a node operator that just wants some inbound for my node overall, but I don't want to perform any swaps with a specific peer. I can set an overall node rule, and just add a custom rule to exclude that peer. Alternately, if I have one peer I receive from a lot, I could set higher inbound percentage for that peer specifically. 
```
{
    "include_private": false,
    "target": "NODE",
    "rule": {
        "type": "THRESHOLD",
        "minimum_inbound": 50,
        "minimum_outbound": 20,
    },
    "peer_rules": {
        "02b94e270e43efaa0dcff4d02dafd5551133afe9130e77b716c7df39ab59a24aac": {
            "type": "EXCLUDE",
        }
    },
    "channel_rules": {
    }
}
```

When we suggest swaps for this kind of setup, I think it's good to make it clear how we got to each group of swaps. 
(still working on the best way to do this)
```
{
    "suggestions": [
        {
            "target": "NODE",
            "rule": {
                "type": "THRESHOLD",
                "minimum_inbound": 20,
                "minimum_outbound": 0
            },
            "loop_out": [
                {
                    "amt": "26701",
                    "outgoing_chan_set": [
                        "215504279109632"
                    ],
                }
            ],
        },
        {
            "target": "02b94e270e43efaa0dcff4d02dafd5551133afe9130e77b716c7df39ab59a24aac",
            "rule": {
                "type": "EXCLUDE",
            },
        }
    ]
}
```


carlaKC:
Closing this for now, decision for MVP is to start with a very simple rule set and proceed to automation, coming back to more complex rule sets if/when we have demand for them.

PR #281, opened 2020-08-28
[2/3] loopd: add macaroon authentication to the loop RPC server

guggero:
Fixes https://github.com/lightninglabs/loop/issues/194.

Depends on https://github.com/lightningnetwork/lnd/pull/4464 (and indirectly on https://github.com/lightningnetwork/lnd/pull/4463).

With this PR we first secure the loop daemon's RPC connection by encrypting the transmitted data with TLS.
Next we add macaroon authentication to the RPC to be able to restrict access to authorized users only.
With these two features enabled by default, the loop RPC port can be opened to the internet because unauthorized access is no longer possible.

By default only one macaroon is created for loop and stored in a file called `~/.loop/<network>/loop.macaroon`.~To give users the ability to fully customize their access with custom macaroons, the two functionalities `BakeMacaroon` and `ListPermissions` are copied from `lnd` and added to both the RPC and command line.~ --> Split this part off into its own PR: https://github.com/lightninglabs/loop/pull/285


carlaKC:
> Depends on lightningnetwork/lnd#4464 (and indirectly on lightningnetwork/lnd#4463).

Just a procedural question about this dependency, is this PR now going to wait until we tag 0.12 to go in? Or are the lnd PRs going into `lnd v0.11.1`?

loopd/run.go
```diff
@@ -197,7 +197,7 @@ func Run(rpcCfg RPCConfig) error {
 	// we are running. When our command tries to get a lnd connection, it
 	// blocks until lnd is synced. We listen for interrupts so that we can
 	// shutdown the daemon while waiting for sync to complete.
-	signal.Intercept()
+	_ = signal.Intercept()
```
carlaKC, line 200: Check this error?

loopd/daemon.go
```diff
@@ -336,9 +358,14 @@ func (d *Daemon) initialize() error {
 	// Retrieve all currently existing swaps from the database.
 	swapsList, err := d.impl.FetchSwaps()
 	if err != nil {
-		// The client is the only thing we started yet, so if we clean
-		// up its connection now, nothing else needs to be shut down at
-		// this point.
+		// The client and the macaroon service are the only things we
+		// started yet, so if we clean that up now, nothing else needs
+		// to be shut down at this point.
+		err2 := d.stopMacaroonService()
```
carlaKC, line 388: nit: `err2`-> `macaroonErr`/`shutdownErr`?

loopd/macaroons.go
```diff
@@ -0,0 +1,189 @@
...
+const (
+	// loopMacaroonLocation is the value we use for the loopd macaroons'
+	// "Location" field when baking them.
+	loopMacaroonLocation = "loop"
+)
+
+var (
+	// RequiredPermissions is a map of all loop RPC methods and their
+	// required macaroon permissions to access loopd.
+	RequiredPermissions = map[string][]bakery.Op{
```
carlaKC, line 24: Nicee 🥇

loopd/macaroons.go
```diff
@@ -0,0 +1,189 @@
...
+			"%v", err)
+		log.Error(err)
+		return err
+	}
+
+	// Try to unlock the macaroon store with the private password.
+	err = d.macaroonService.CreateUnlock(&macDbDefaultPw)
+	if err != nil {
+		err := fmt.Errorf("unable to unlock macaroon DB: %v", err)
+		log.Error(err)
```
carlaKC, line 158: Could this logging happen when we call `startMacaroonService`?

guggero:
Yes, the dependent PRs are both candidates for `lnd v0.11.1`, so this PR has to wait for that to be released.
But it's a compile-time dependency only, not an RPC dependency.

loopd/daemon.go
```diff
@@ -205,14 +205,14 @@ func (d *Daemon) startWebServers() error {
 	serverOpts := []grpc.ServerOption{
 		grpc.Creds(credentials.NewTLS(serverTLSCfg)),
 	}
-	
+
```
carlaKC, line 208: move these to the previous commit where we check `!d.cfg.NoMacaroons`?

looprpc/client.proto
```diff
@@ -699,3 +710,19 @@ message BakeMacaroonResponse {
...
+message ListPermissionsRequest {
+}
+
+message ListPermissionsResponse {
+    /*
+    A map between all RPC method URIs and their required macaroon permissions to
+    access them.
+    */
+    map<string, MacaroonPermissionList> method_permissions = 1;
+}
```
carlaKC, line 728: nit: newline

carlaKC (commented):
Awesome to have thins functionality added to loop 🎉 
No major questions, since a lot of this is copied from lnd, just some nits from my first pass. Will do some testing for this as well. 

Perhaps also update the `README` since we're going to enable by default?

loopd/config.go
```diff
@@ -50,9 +81,20 @@ type Config struct {
...
+	DataDir    string `long:"datadir" description:"Directory for loopdb."`
+
+	TLSCertPath        string   `long:"tlscertpath" description:"Path to write the TLS certificate for loop's RPC and REST services"`
+	TLSKeyPath         string   `long:"tlskeypath" description:"Path to write the TLS private key for loop's RPC and REST services"`
+	TLSExtraIPs        []string `long:"tlsextraip" description:"Adds an extra IP to the generated certificate"`
+	TLSExtraDomains    []string `long:"tlsextradomain" description:"Adds an extra domain to the generated certificate"`
+	TLSAutoRefresh     bool     `long:"tlsautorefresh" description:"Re-generate TLS certificate and key if the IPs or domains are changed"`
+	TLSDisableAutofill bool     `long:"tlsdisableautofill" description:"Do not include the interface IPs or the system hostname in TLS certificate, use first --tlsextradomain as Common Name instead, if set"`
+
+	NoMacaroons bool   `long:"no-macaroons" description:"Disable macaroon authentication, dangerous if loop is reachable from the network"`
```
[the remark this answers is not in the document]
joostjager: Maybe we shouldn't even add the option `NoMacaroons`?

loopd/macaroons.go
```diff
@@ -0,0 +1,189 @@
...
+	}, {
+		Entity: "auth",
+		Action: "read",
+	}}
+
+	// macDbDefaultPw is the default encryption password used to encrypt the
+	// loop macaroon database.
+	//
+	// TODO(guggero): Allow the password to be specified by the user. Needs
+	// create/unlock calls in the RPC.
```
joostjager, line 129: Won't we end up in the same situation as with `lnd`? That people want to store the password on the box anyway, defeating the whole purpose?

joostjager (commented):
Just quickly skimmed through the PR. Is it worth to split this PR before continuing into the review cycle? For example in three parts:
* Add TLS
* Add (single) macaroon
* Add macaroon baking of rpc

joostjager (commented):
Tested and works. Was just wondering how I can verify that it is now actually using TLS?

loopd/config.go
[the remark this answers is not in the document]
guggero: Removed the flag. This makes it a bit harder to test when developing against the REST API for example as you can't just get some JSON in the browser. But if anyone needs this feature they can request it later on.

loopd/config.go
```diff
@@ -147,3 +205,81 @@ func Validate(cfg *Config) error {
...
+		certData, _, err = cert.LoadCert(
+			d.cfg.TLSCertPath, d.cfg.TLSKeyPath,
+		)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
+	tlsCfg := cert.TLSConfFromCert(certData)
+	restCreds, err := credentials.NewClientTLSFromFile(
```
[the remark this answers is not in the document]
guggero: These are just two different ways to hold the same certificate. Once we're using it as a server certificate (for gRPC) and once as a client connect CA certificate (for REST). Because the REST proxy connects to gRPC and is a client.

But you are right, something's missing. The REST server also needs to serve over a TLS connection!

loopd/macaroons.go
```diff
@@ -0,0 +1,189 @@
...
+		"/looprpc.SwapClient/LoopOut": {{
+			Entity: "swap",
+			Action: "execute",
+		}, {
+			Entity: "loop",
+			Action: "out",
+		}},
+		"/looprpc.SwapClient/LoopIn": {{
+			Entity: "swap",
+			Action: "execute",
```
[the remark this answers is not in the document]
guggero: Multiple permissions here mean that all of them are required. So it's an _AND_ operation, not _OR_. The idea is that (with the bakery) users could create macaroons that are working for loop in or loop out exclusively. I don't have an explicit use case in mind but that is something I think could come in handy for users.

cmd/loop/main.go
```diff
@@ -321,32 +405,23 @@ func getClientConn(address, tlsCertPath, macaroonPath string) (*grpc.ClientConn,
...
-
-		// Macaroons are only allowed to be transmitted over a TLS
-		// enabled connection.
-		if macaroonPath != "" {
-			opts = append(opts, readMacaroon(macaroonPath))
-		}
-
-		opts = append(opts, grpc.WithTransportCredentials(creds))
+	// TLS cannot be disabled, we'll always have a cert file to read.
+	creds, err := credentials.NewClientTLSFromFile(tlsCertPath, "")
```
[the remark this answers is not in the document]
guggero: I'm not exactly sure what you mean by this comment. But I did split the commit into the TLS part and the macaroon part. We now touch the same code twice but the commits are separated by the functionality they touch.

loopd/macaroons.go
guggero: I extended the TODO comment that if we want to add a password we should make it optional.

cmd/loop/main.go
```diff
@@ -136,6 +162,64 @@ func getMaxRoutingFee(amt btcutil.Amount) btcutil.Amount {
...
+		macPath = lncfg.CleanAndExpandPath(ctx.GlobalString(
+			macaroonPathFlag.Name,
+		))
+
+	default:
+		// Otherwise, we'll go into the path:
+		// loopdir/<network> in order to fetch the macaroon that we
+		// need.
+		macPath = filepath.Join(
+			loopDir, network, loopd.DefaultMacFilename,
```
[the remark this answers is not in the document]
guggero: It doesn't look like it. The `loop.db` is still in the network sub directory when a loop dir is set.

loopd/config.go
```diff
@@ -136,6 +180,20 @@ func Validate(cfg *Config) error {
 	cfg.DataDir = filepath.Join(cfg.DataDir, cfg.Network)
 	cfg.LogDir = filepath.Join(cfg.LogDir, cfg.Network)
 
+	// Prepare the other directories we need for TLS handling.
```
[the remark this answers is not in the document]
guggero: I added the restriction as suggested. But I don't want to change the behavior of the other flags in this PR, that's why the TLS and macaroon files are stored in the same place as `loop.db` at the moment (unless explicitly specified differently).

guggero:
I addressed all comments and also added a short section about TLS and macaroons to the README.

@joostjager you can verify if TLS is enabled by using `openssl s_client -connect <host>:<port>`. If you get presented a ceritifcate then TLS is turned on.

loopd/config.go
```diff
@@ -206,6 +212,12 @@ func Validate(cfg *Config) error {
 		)
 	}
 
+	// There currently is only one macaroon file and we want to store it in
+	// the namespaced network directory.
+	if cfg.MacaroonPath == "" {
```
joostjager, line 217: Doesn't it need a config default similar to tls path?

loopd/macaroons.go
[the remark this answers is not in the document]
joostjager: Ok, clear

loopd/macaroons.go
```diff
@@ -0,0 +1,177 @@
...
+		Entity: "auth",
+		Action: "read",
+	}}
+
+	// macDbDefaultPw is the default encryption password used to encrypt the
+	// loop macaroon database.
+	//
+	// TODO(guggero): Allow the password to be specified by the user. Needs
+	// create/unlock calls in the RPC. Using a password should be optional
+	// though.
```
joostjager, line 102: Does something still need to be done with this? I'd think that a default password in open source code provides a false sense of security?

cmd/loop/main.go
```diff
@@ -374,6 +392,9 @@ func getClientConn(address, tlsCertPath, macaroonPath string) (*grpc.ClientConn,
 
 	opts := []grpc.DialOption{
 		grpc.WithDefaultCallOptions(maxMsgRecvSize),
+
+		// We always need to send a macaroon.
+		readMacaroon(macaroonPath),
```
joostjager, line 397: `getClientConn` has an error return value, so you could pass back an error from `readMacaroon`

cmd/loop/main.go
```diff
@@ -170,9 +169,28 @@ func extractPathArgs(ctx *cli.Context) (string, string, error) {
...
+	if ctx.GlobalIsSet(macaroonPathFlag.Name) {
+		macPath = lncfg.CleanAndExpandPath(ctx.GlobalString(
+			macaroonPathFlag.Name,
+		))
+
+	} else {
+		// Otherwise, we'll go into the path:
+		// loopdir/<network> in order to fetch the macaroon that we
+		// need.
+		macPath = filepath.Join(
```
joostjager, line 189: This logic seems to be slightly different from what happens from tls path. I'd think these are two file paths and should be handled exactly the same? Perhaps even via a helper function.

README.md
```diff
@@ -333,6 +333,25 @@ pending swaps after a restart.
 Information about pending swaps is stored persistently in the swap database.
 Its location is `~/.loopd/<network>/loop.db`.
 
+## Authentication and transport security
+
+The gRPC and REST connections of `loopd` are encrypted with TLS by default and
```
joostjager, line 338: "by default" sounds as if it can be turned off?

README.md
```diff
@@ -333,6 +333,25 @@ pending swaps after a restart.
...
+secured with macaroon authentication the same way `lnd` is.
+
+If no custom loop data directory is set then the TLS certificate is stored in
```
joostjager, line 341: loop dir (`--loopdir`) or loop data dir (`--datadir`)? It is still a bit confusing which file is sought where.

joostjager: Perhaps the command line help should clarify that `loopdir` and the other path flags are mutually exclusive?

cmd/loop/main.go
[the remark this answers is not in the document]
joostjager: Discussed offline and want to keep in line with lnd.

loopd/run.go
```diff
@@ -51,22 +52,44 @@ type RPCConfig struct {
...
 
-			return net.Listen("tcp", config.RPCListen)
+			listener, err := net.Listen("tcp", config.RPCListen)
+			if err != nil {
+				return nil, err
+			}
+
+			// In case a TLS configuration was provided, we wrap the
+			// listener with a TLS one.
+			if tlsCfg != nil {
```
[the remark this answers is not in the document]
guggero: Yes. Removed the `if` and changed the comment. The TLS config can still be `nil` but that will result in an error from the TLS library itself.

loopd/run.go
```diff
@@ -220,7 +220,10 @@ func Run(rpcCfg RPCConfig) error {
 	// we are running. When our command tries to get a lnd connection, it
 	// blocks until lnd is synced. We listen for interrupts so that we can
 	// shutdown the daemon while waiting for sync to complete.
-	signal.Intercept()
+	err = signal.Intercept()
```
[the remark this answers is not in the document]
guggero: No, the update to the newest `lnd` version caused the linter to complain about this new, unhandled exception.

loopd/macaroons.go
```diff
@@ -0,0 +1,177 @@
...
+			Action: "read",
+		}},
+		"/looprpc.SwapClient/SwapInfo": {{
+			Entity: "swap",
+			Action: "read",
+		}},
+		"/looprpc.SwapClient/LoopOutTerms": {{
+			Entity: "terms",
+			Action: "read",
+		}},
```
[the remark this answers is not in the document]
guggero: My thought was that it's probably most important to be able to separate them for the write operations. But to be consequent we should probably use the same pattern for the read operations as well. Fixed.

loopd/macaroons.go
guggero: I changed the password to be an empty one and added some more information to the comment.

loopd/daemon.go
```diff
@@ -175,7 +178,7 @@ func (d *Daemon) startWebServers() error {
 
 	// Next, start the gRPC server listening for HTTP/2 connections.
 	log.Infof("Starting gRPC listener")
-	d.grpcListener, err = d.listenerCfg.grpcListener()
+	d.grpcListener, err = d.listenerCfg.grpcListener(nil)
```
joostjager: If it crashes with `nil`, this isn't a working commit and possibly a sign that it should be squashed with another one. But non-blocking.

joostjager (approved):

joostjager:
>But it's a compile-time dependency only, not an RPC dependency.

With grub, a comment like this isn't relevant anymore, because even for compile-time deps we need to wait for a release?

carlaKC (approved):
LGTM

loopd/daemon.go
```diff
@@ -324,6 +328,19 @@ func (d *Daemon) initialize() error {
 	// stop on main context cancel. So we create it early and pass it down.
 	d.mainCtx, d.mainCtxCancel = context.WithCancel(context.Background())
 
+	// Start the macaroon service and let it create its default macaroon in
+	// case it doesn't exist yet.
+	err = d.startMacaroonService()
+	if err != nil {
+		log.Errorf("Error starting macaroon service: %v", err)
```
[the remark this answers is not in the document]
guggero: Yeah, you're right. Fixed.

loopd/macaroons.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+			return err
+		}
+	}
+
+	return nil
+}
+
+// stopMacaroonService closes the macaroon database if it was started.
+func (d *Daemon) stopMacaroonService() error {
```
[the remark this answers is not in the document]
guggero: Ah, that comment was from when we still had the `NoMacaroons` flag. Removed the "if it was started" part.

guggero:
> With grub, a comment like this isn't relevant anymore, because even for compile-time deps we need to wait for a release?

We have to wait for a release (of lnd and loop that is) in LiT. But we can merge this PR as soon as the dependent PRs are merged to master. That's the main difference between an RPC and compile time dep IMO.

joostjager:
But won't your compile-time required code disappear when loop is embedded in grub and the dependency is overriden with the last release 0.11.0?

Kixunil:
Awesome! @guggero do you have a LN donation page? Would like to send you a tip for doing this. :)

guggero:
Thanks for the feedback, I'm happy to hear this is useful! No need for a tip though, I am getting paid to work on these projects ;-)

PR #287, opened 2020-09-04
Upgrade docker image to golang 1.13

tdickman:
Building the docker image currently fails due to a module that requires a minimum golang version of 1.13.


guggero (approved):
LGTM, thanks!

carlaKC (approved):
Thanks for the fix! 🚀

PR #289, opened 2020-09-08
suggestions: add fee and ongoing swap restrictions

carlaKC:
In preparation to adding automated dispatch of swaps, this PR adds some restrictions to our swap suggestions endpoint. The aim is for suggestions to act as a dry run for autoloop (with the only difference being dispatch of the actual swap), so we add all the restrictions that we would place on an autolooper. 
- Ongoing swaps: there is some time between creation of a swap and locking in htlcs, and shards can fail and be dispatched over the life of a swap. We restrict suggestions to swaps that are not currently being used by a swap, and if we have currently ongoing unrestricted swaps, we do not suggest anything.
- Failure backoff: if we're recently failed to perform a loop out on a channel, we exclude it from our suggestions until a configurable backoff has passed. 
- Fee limits: the fee limits that are available to us are added here. Without a prepay probe, we cannot set limits on off-chain-routing until we actually dispatch the swap, so these values will be added in a follow up.


liquidity/liquidity.go
```diff
@@ -26,6 +30,10 @@ type Config struct {
 	// to loop out swaps.
 	LoopOutRestrictions func(ctx context.Context) (*Restrictions, error)
 
+	// ListSwaps returns the set of swaps that loop has already created.
+	// These swaps may be in a final or pending state.
+	ListSwaps func(ctx context.Context) ([]ExistingSwap, error)
```
joostjager, line 88: What is the status on fixing this dependency problem?

liquidity/liquidity.go
```diff
@@ -78,6 +86,39 @@ func (p Parameters) validate() error {
...
+
+	// Type indicates the type of swap.
+	Type swap.Type
+
+	// Channels is the set of channels that a loop out swap is using.
+	Channels []lnwire.ShortChannelID
+
+	// Peer is the last hop set for loop in (if any).
+	Peer *route.Vertex
+}
```
joostjager, line 206: I would be careful mixing loop in and out again here, as we've been separating data structures elsewhere

liquidity/liquidity.go
```diff
@@ -155,19 +196,26 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+	// Get the current server side restrictions.
+	outRestrictions, err := m.cfg.LoopOutRestrictions(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	// Get the current server side restrictions.
-	outRestrictions, err := m.cfg.LoopOutRestrictions(ctx)
+	// List our current set of swaps so that we can determine which channels
+	// are already being utilized by swaps.
```
joostjager, line 313: Perhaps add comment explaining that there may be race conditions when at the same time swaps are initiated manually. I don't think we need to worry about that now, but is this something that needs to be addressed in the future? Or do we bet on full automation anyway?

liquidity/liquidity.go
```diff
@@ -246,9 +265,30 @@ func (m *Manager) getEligibleChannels(ctx context.Context,
...
+	// consider a channel eligible. Any channels involved in swaps that have
+	// failed since this point will not be considered.
+	failureCutoff := m.cfg.Clock.Now().Add(m.params.FailureBackOff * -1)
+
 	for _, s := range allSwaps {
+		// If a loop out swap failed due to off chain payment after our
+		// failure cutoff, we add all of its channels to a set of
+		// recently failed channels. It is possible that not all of
+		// these channels were used for the swap, but we play it safe
+		// and back off for all of them.
```
joostjager, line 393: Make note that we are (for now) ignoring temporary failures.

liquidity/liquidity.go
```diff
@@ -305,6 +345,15 @@ func (m *Manager) getEligibleChannels(ctx context.Context,
 	for _, channel := range channels {
 		shortID := lnwire.NewShortChanIDFromInt(channel.ChannelID)
 
+		lastFail, recentFail := failedOut[shortID]
+		if recentFail {
+			log.Infof("channel: %v not eligible for "+
```
joostjager, line 462: log at level debug?

liquidity/liquidity.go
```diff
@@ -56,6 +102,21 @@ type Parameters struct {
...
+	// MaximumSwapFeePPM is the maximum server fee we are willing to pay per
+	// swap expressed as parts per million of the swap volume.
+	MaximumSwapFeePPM int
+
+	// MaximumMinerFee is the maximum on chain fee we are willing to pay.
+	MaximumMinerFee btcutil.Amount
+
+	// ConfTarget is the number of blocks we aim to confirm our sweep
+	// transaction in. This value affects our miner fees.
+	ConfTarget int32
```
joostjager, line 118: I would expect the `ConfTarget` to be set on the suggested `LoopOutRequest` somewhere in this commit

cmd/loop/liquidity.go
```diff
@@ -158,13 +158,16 @@ func setParam(ctx *cli.Context) error {
 		)
 	}
 
+	// Just set the rules on our current set of parameters and leave the
+	// other values untouched.
+	// nolint: gocritic
+	params.Rules = append(otherRules, newRule)
```
joostjager, line 165: You could first append to `otherRules` and then assign `otherRules` to `param.Rules` to prevent the linter problem

cmd/loop/liquidity.go
```diff
@@ -174,6 +175,109 @@ func setRule(ctx *cli.Context) error {
...
+		return err
+	}
+
+	var flagCount int
+
+	if ctx.IsSet("maxswapfee") {
+		feePercentage := ctx.Float64("maxswapfee")
+
+		if feePercentage <= 0 || feePercentage >= 1 {
+			return fmt.Errorf("fee percentage must be in (0;1)")
```
joostjager, line 238: It is not a percentage if between 0 and 1. Fee rate?

cmd/loop/liquidity.go
```diff
@@ -174,6 +175,109 @@ func setRule(ctx *cli.Context) error {
...
+		params.SweepConfTarget = int32(ctx.Int("sweepconf"))
+		flagCount++
+	}
+
+	if ctx.IsSet("failurebackoff") {
+		params.FailureBackoff = ctx.Uint64("failurebackoff")
+		flagCount++
+	}
+
+	if flagCount == 0 {
```
joostjager, line 267: Could be a boolean

looprpc/client.proto
```diff
@@ -710,6 +710,38 @@ message LiquidityParameters{
...
+    as parts per million of the swap volume, swap suggestions with fees higher
+    than this will be excluded.
+    */
+    uint64 max_swap_fee_ppm = 2;
+
+    /*
+    The maximum no-show penalty in satoshis paid for a swap, swap suggestions
+    above this amount will be excluded.
+    */
+    uint64 max_prepay = 3;
```
joostjager, line 725: Add unit `_sat`?

looprpc/client.proto
```diff
@@ -710,6 +710,38 @@ message LiquidityParameters{
...
+    should be swept within.
+    */
+    int32 sweep_conf_target = 5;
+
+    /*
+    The amount of time we require pass since a channel was part of a failed
+    swap until it will be considered for swap suggestions again, expressed in
+    seconds.
+    */
+    uint64 failure_backoff = 6;
```
joostjager, line 744: Add unit `_secs`?

liquidity/liquidity.go
```diff
@@ -246,9 +265,30 @@ func (m *Manager) getEligibleChannels(ctx context.Context,
...
+	// failed since this point will not be considered.
+	failureCutoff := m.cfg.Clock.Now().Add(m.params.FailureBackOff * -1)
+
 	for _, s := range allSwaps {
+		// If a loop out swap failed due to off chain payment after our
+		// failure cutoff, we add all of its channels to a set of
+		// recently failed channels. It is possible that not all of
+		// these channels were used for the swap, but we play it safe
+		// and back off for all of them.
+		if s.State == loopdb.StateFailOffchainPayments &&
```
joostjager, line 394: Add comment why no back-off is needed for all other types of final failures?

joostjager (commented):
First pass looks good. I do think that the circular dependency issue needs to be addressed first, because this PR adds more code than just a passing on the call.

Interesting to see the kind of logic that emerges to properly auto-loop.

liquidity/liquidity.go
carlaKC: I think we will need to address it in future, we need to get quotes for each swap from the server which isn't a trivial amount of time. One solution could be to limit to only auto or manual (require disabling autoloop to dispatch a manual swap), or to add some locking which will block new swaps while we're figuring out autoloop?

joostjager: I thinking that blocking new swaps while auto-loop is running can be good enough. Maybe we can also spread out the auto-looping a bit, for example finding the most unbalanced channel, initiating the swap and then sleeping for some time before going on to the next.

liquidity/liquidity.go
```diff
@@ -187,3 +216,91 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+			chanSet = out.Contract.OutgoingChanSet
+		)
+
+		// Skip pending swaps, they can't affect our channel balances.
+		if state.Type() != loopdb.StateTypePending {
+			continue
+		}
+
+		if len(chanSet) == 0 {
+			log.Debugf("ongoing unrestricted loop out: "+
```
guggero, line 407: nit: capitalize log messages.

liquidity/liquidity.go
```diff
@@ -187,3 +216,91 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+		}
+
+		for _, id := range chanSet {
+			chanID := lnwire.NewShortChanIDFromInt(id)
+			existingOut[chanID] = true
+		}
+	}
+
+	for _, in := range loopIn {
+		// Skip pending swaps, they can't affect our channel balances.
```
guggero, line 420: Same here, comment is incorrect.

liquidity/liquidity.go
```diff
@@ -187,3 +216,91 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+	// Run through our set of channels and skip over any channels that
+	// are currently being utilized by a restricted swap (where restricted
+	// means that a loop out limited channels, or a loop in limited last
+	// hop).
+	var eligible []lndclient.ChannelInfo
+	for _, channel := range channels {
+		shortID := lnwire.NewShortChanIDFromInt(channel.ChannelID)
+
+		if existingOut[shortID] {
+			log.Infof("channel: %v not eligible for "+
```
guggero, line 458: Denote to `Debugf` as well? Not sure if this is more important than the above.

liquidity/liquidity.go
```diff
@@ -22,6 +23,12 @@ import (
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
+const (
+	// DefaultFailureBackOff is the default amount of time we wait after
+	// a swap has failed on a channel to recommend using it again.
+	DefaultFailureBackOff = time.Hour * 24 * 7
```
guggero, line 32: A week sounds like a very long time. Maybe the default should be a bit shorter, perhaps 1 day. Or is this to prevent suggesting the same bad channel every day?

liquidity/liquidity.go
```diff
@@ -27,11 +30,40 @@ const (
 	// DefaultFailureBackOff is the default amount of time we wait after
 	// a swap has failed on a channel to recommend using it again.
 	DefaultFailureBackOff = time.Hour * 24 * 7
+
+	// DefaultSwapFeePPM is the default limit we place of the fee paid to
+	// the server, expressed as parts per million of the swap volume, equal
+	// to 0.5 of a percent.
```
guggero, line 36: nit: "to 0.5 of a percent" sounds a bit weird to me. Why not just "to 0.5%"?

liquidity/liquidity.go
```diff
@@ -27,11 +30,40 @@ const (
...
+	DefaultSwapFeePPM = 5000
+
+	// DefaultPrepay is the default limit we place on prepayments.
+	DefaultPrepay btcutil.Amount = 20000
+
+	// DefaultMinerFee is the default limit we place on miner fees.
+	DefaultMinerFee btcutil.Amount = 15000
+
+	// FeeBase is the base that we use to express fees.
+	FeeBase = 1000000
```
guggero, line 46: nit: use `1e6` or `1_000_000` notation to make easier to read?

looprpc/client.proto
```diff
@@ -710,6 +710,38 @@ message LiquidityParameters{
     A set of liquidity rules that describe the desired liquidity balance.
     */
     repeated LiquidityRule rules = 1;
+
+    /*
+    The maximum fee in paid to the server for facilitating the swap, expressed
```
guggero, line 715: nit: s/fee in paid/fee paid/

cmd/loop/liquidity.go
```diff
@@ -174,6 +175,109 @@ func setRule(ctx *cli.Context) error {
...
+
+	if ctx.IsSet("maxswapfee") {
+		feeRate := ctx.Float64("maxswapfee")
+
+		if feeRate <= 0 || feeRate >= 1 {
+			return fmt.Errorf("fee rate must be in (0;1)")
+		}
+
+		params.MaxMinerFeeSat = uint64(
+			feeRate * float64(liquidity.FeeBase),
```
guggero, line 242: I don't think this is correct. Let's say I configure `--maxswapfee=0.5`, that would result in a PPM value of `500_000`, which is 50%. Also I'm not sure if we want to limit us to a maximum of 0.99% of fees that we can ever charge? Maybe that should also be a constant that we can change easily and hopefully not forget.

guggero (commented):
Nice and clean PR, very easy to follow :clap:

liquidity/liquidity.go
carlaKC: The thought here is that we don't want to get into a state where we have one really unbalanced channel, but it actually isn't possible to loop out through it. Thinking about it a bit more, I think we want some kind of exponential backoff here. WDYT about decreasing this to one day and adding a todo for more intelligent backoff?

guggero: Yes, that sounds good to me!

cmd/loop/liquidity.go
carlaKC: Setting `--maxswapfee=0.5` -> 50% is intentional, so we limit fees to somewhere between (0-100%) of the swap volume. I've updated the comment to be a bit clearer, hopefully that helps. 

We could just do PPM here, but I thought percentage was more user-friendly. Also could do a percentage integer, but I think setting fractions of a percent here is required.

carlaKC:
Changes since last review [here](https://github.com/lightninglabs/loop/compare/cc768db3cdd1af6b6ba8693abccc917d0756dea6..db09ec95d4ce0d43735b616c26cece4435aacef2), addresses feedback and added off chain fee restrictions (because we just do a flat estimate at present). [Second push](https://github.com/lightninglabs/loop/compare/db09ec95d4ce0d43735b616c26cece4435aacef2..aa225315e2b4076bf483950c6aa03c841b7e9b7d) has some slight refactoring of the fee code, nothing major.

liquidity/liquidity.go
carlaKC: Turns out that we are ok without a refactor, since we need to use the `Client` functions. There is some logic in the rpc that we will need to duplicate (eg, set sweep addr), but it's quite minimal. Not using the rpc because when we get to dispatching swaps, we need to be able to label them with reserved prefix (to identify autoloops), and that will be validated on rpc level (currently it's in the wrong place, going to move it up)

guggero (approved):
LGTM :taco:

looprpc/client.proto
```diff
@@ -710,6 +710,41 @@ message LiquidityParameters{
...
+    The maximum fee paid to the server for facilitating the swap, expressed
+    as parts per million of the swap volume.
+    */
+    uint64 max_swap_fee_ppm = 2;
+
+    /*
+    The maximum fee paid to route the swap prepay and invoice, expressed as
+    parts per million of the volume being routed.
+    */
+    uint64 max_routing_fee_ppm = 3;
```
joostjager, line 724: Isn't it that this value is only matched against a hard-coded constant now? Perhaps it is better to use `max_routing_fee_ppm` as the fee limit directly.

cmd/loop/liquidity.go
```diff
@@ -173,6 +174,131 @@ func setRule(ctx *cli.Context) error {
...
+var setParamsCommand = cli.Command{
+	Name:        "setparams",
+	Usage:       "update the parameters set for the liquidity manager",
+	Description: "Updates the parameters set for the liquidity manager.",
+	Flags: []cli.Flag{
+		cli.Float64Flag{
+			Name: "maxswapfee",
+			Usage: "the maximum ratio of swap volume we are " +
+				"willing to pay in server fees (eg, 0.01 = " +
+				"1%).",
```
joostjager, line 186: This explanation, doesn't this hint that we need to accept (float) percentages? It is easy to make a mistake with all these zeroes.

cmd/loop/liquidity.go
```diff
@@ -173,6 +174,131 @@ func setRule(ctx *cli.Context) error {
...
+		},
+		cli.Uint64Flag{
+			Name: "maxprepay",
+			Usage: "the maximum no-show (prepay) that swap " +
+				"suggestions should be limited to.",
+		},
+		cli.Uint64Flag{
+			Name: "maxminer",
+			Usage: "the maximum miner fee that swap " +
+				"suggestions should be limited to.",
```
joostjager, line 202: mention unit sat? also for prepay

release_notes.md
```diff
@@ -12,7 +12,10 @@ This file tracks release notes for the loop client.
 ## Next Release
 
 #### New Features
+- The swap suggestions endpoint has been updated to be fee-aware. Swaps that exceed the fee limits set by the liquidity manager will no longer be suggested (see `getParams` for the current limits, and use `setParams` to update these values). 
+- Swap suggestions are now aware of ongoing and previously failed swaps. They will not suggest swaps for channels that are currently being utilized for swaps, and will not suggest any swaps if a swap that is not limited to a specific peer or channel is ongoing. If a channel was part of a failed swap within the last 24H, it will be excluded from our swap suggestions (this value is configurable).
```
joostjager, line 16: Nice to have the release notes in the PR

liquidity/liquidity.go
```diff
@@ -223,10 +348,48 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+		if suggestion == nil {
+			continue
 		}
+
+		// Get a quote for a swap of this amount.
+		// TODO(carla): add support for non-fast swaps.
+		quote, err := m.cfg.LoopOutQuote(
+			ctx, &loop.LoopOutQuoteRequest{
+				Amount:                  suggestion.Amount,
+				SweepConfTarget:         m.params.ConfTarget,
```
joostjager, line 361: It seems that `SweepConfTarget` isn't set in `swapClientServer.SuggestSwaps` when mapping to `looprpc.LoopOutRequest`?

liquidity/liquidity.go
```diff
@@ -223,10 +348,48 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+			quote, suggestion.Amount, prePayFee, routeFee,
+		)
+		if err != nil {
+			log.Infof("suggestion: %v expected fees too high: %v",
+				suggestion, err)
+
+			continue
+		}
+
+		suggestions = append(suggestions, suggestion)
```
joostjager, line 392: Also these fee limits need to be set on the recommendation?

cmd/loop/liquidity.go
```diff
@@ -173,6 +174,131 @@ func setRule(ctx *cli.Context) error {
...
+			Name: "maxswapfee",
+			Usage: "the maximum ratio of swap volume we are " +
+				"willing to pay in server fees (eg, 0.01 = " +
+				"1%).",
+		},
+		cli.Float64Flag{
+			Name: "maxroutingfee",
+			Usage: "the maximum ratio of off-chain payment " +
+				"volume that are are willing to pay in " +
+				"routing fees (eg, 0.01 = 1%).",
```
joostjager, line 192: Does this also apply to the prepay? With small fractions, it may get problematic?

liquidity/liquidity.go
```diff
@@ -1,21 +1,90 @@
...
+	// DefaultRoutingFeePPM is the default limit we place on off chain
+	// routing fees (for our prepayment and swap invoice), expressed as
+	// parts per million of the amount being sent, equal to 1%.
+	DefaultRoutingFeePPM = 10000
+
+	// DefaultPrepay is the default limit we place on prepayments.
+	DefaultPrepay btcutil.Amount = 20000
+
+	// DefaultMinerFee is the default limit we place on miner fees.
+	DefaultMinerFee btcutil.Amount = 15000
```
joostjager, line 57: Maybe the defaults can be defined as a `Parameter` instance

liquidity/liquidity.go
```diff
@@ -175,11 +348,240 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+			return nil, err
+		}
+
+		// Our quote does not provide fee estimates for our off chain
+		// routing expenses, so we estimate them separately.
+		prePayFee, routeFee := m.estimateOffChain(
+			quote.PrepayAmount, suggestion.Amount,
+		)
+
+		log.Infof("quote for suggestion: %v, swap fee: %v, "+
```
joostjager, line 375: `Debugf`? Not sure

liquidity/liquidity.go
```diff
@@ -228,14 +243,49 @@ func (m *Manager) getEligibleChannels(ctx context.Context,
...
+		//
+		// We only backoff for off chain payment failures because this
+		// error indicates that our swap failed to route. We do not
+		// want to repeatedly try to route through bad channels which
+		// remain unbalanced because they cannot route a swap, so we
+		// backoff. Our other failure types do not represent temporary
+		// failures, so we do not backoff. We don't backoff on the
+		// temporary state itself because we require a restart to clear
+		// this state, which would prevent use of suggestions
+		// completely.
```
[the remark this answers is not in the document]
carlaKC: Wondering whether we shouldn't just backoff entirely if there's a temporary failure? In the case of a "persistent" temp fail (eg bitcoind being down?), we don't really address the issue by backing off for a particular channel, since temp failures aren't really swap specific?

liquidity/liquidity.go
carlaKC: Everything else needs debug, going to downgrade then make a note in the release notes recommended debug logging if you're using suggestions.

carlaKC:
Latest [changes](https://github.com/lightninglabs/loop/compare/aa225315e2b4076bf483950c6aa03c841b7e9b7d..088adb274387d373a740948bf7378a787adacfc9):
- Update suggestions to return `OutRequest`
- Separate out limits for prepay and invoice fees
- Some test refactoring, only test against `SuggestSwaps`, split across various tests for simplicity

cmd/loop/liquidity.go
carlaKC: Separated these out, but I think we should be ok now that prepay on mainnet is 20k? Not sure what to set the default as though.

liquidity/liquidity.go
carlaKC: Added this in the next commit, since it has the rpc changes for the new fields as well .

liquidity/liquidity.go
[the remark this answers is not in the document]
joostjager: Yes, maybe that is the safest thing to do. Because after a restart, those temporarily failed swaps may also again come to life.

loopd/swapclient_server.go
```diff
@@ -646,10 +646,8 @@ func (s *swapClientServer) SuggestSwaps(ctx context.Context,
 
 	for _, swap := range swaps {
 		loopOut = append(loopOut, &looprpc.LoopOutRequest{
-			Amt: int64(swap.Amount),
-			OutgoingChanSet: []uint64{
-				swap.Channel.ToUint64(),
-			},
+			Amt:             int64(swap.Amount),
+			OutgoingChanSet: swap.OutgoingChanSet,
```
joostjager: I now realize that the returned `LoopOutRequest` messages in master aren't directly usable because there are no fees set.

Maybe nice to do a quick test modification locally where you pass in the response of `SuggestSwaps` to `LoopOut` to see that it actually works.

In terms of commit structure, you could also choose to introduce the default values and just assign them here. Then the 'bug' would be fixed first. Further commits add new functionality.

liquidity/liquidity_test.go
```diff
@@ -140,21 +132,43 @@ func TestSuggestSwaps(t *testing.T) {
...
+	// them to use the rules set by the test.
+	manager := NewManager(cfg)
+
+	currentParams := manager.GetParameters()
+	currentParams.ChannelRules = rules
+
+	err := manager.SetParameters(currentParams)
+	require.NoError(t, err)
+
+	actual, err := manager.SuggestSwaps(context.Background())
```
joostjager, line 464: Nice black box testing

cmd/loop/liquidity.go
```diff
@@ -173,6 +174,149 @@ func setRule(ctx *cli.Context) error {
...
+				"volume that are are willing to pay in " +
+				"routing fees, expressed as a decimal (eg, " +
+				"0.01 = 1%).",
+		},
+		cli.Float64Flag{
+			Name: "maxprepayfee",
+			Usage: "the maximum percentage of off-chain prepay " +
+				"volume that are are willing to pay in " +
+				"routing fees, expressed as a decimal (eg, " +
+				"0.01 = 1%).",
```
joostjager, line 200: What do you think about my outstanding comment that a float may easily lead to the wrong number of zeroes? Isn't it better ux to accept a (float) percentage?

cmd/loop/liquidity.go
```diff
@@ -158,19 +158,165 @@ func setParam(ctx *cli.Context) error {
...
+			Usage: "the amount of time, in seconds, that " +
+				"should pass before a channel that " +
+				"previously had a failed swap will be " +
+				"included in suggestions.",
+		},
+	},
+	Action: setParams,
+}
+
+func setParams(ctx *cli.Context) error {
```
joostjager, line 229: the name of the command `getParams` isn't totally consistent, because it also returns the rules

joostjager (commented):
No serious comments remaining. Tested the PR and I am getting suggestions. Pretty comfortable UI, for a CLI.

cmd/loop/liquidity.go
carlaKC: indeed, couldn't really think of a better name for `setParams`, alternatively we could separate `getParams` into `getParams` and `getRule` but I don't think that adds much?

cmd/loop/liquidity.go
carlaKC: Ah yes definitely, I misunderstood the original comment. Updated to float percentage, much nicer!

carlaKC:
Last changes [here](https://github.com/lightninglabs/loop/compare/aa9deb1b1eb538d8707a5dfa1e04de90685b43d7..bbe4b437c653e02b1ef8e2fba4f643397936fd0f), some vars look like they've been moved, just I just reordered commits.

loopd/swapclient_server.go
carlaKC: Wasn't really thinking of swap suggestions as something that would be passed straight to the rpc tbh, was more thinking that people would use them for input into the cli. But nice idea, moved the defaults forward and added them to the output in the first commit.

cmd/loop/liquidity.go
joostjager: `getConfig`?

liquidity/liquidity.go
```diff
@@ -177,9 +207,40 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+		Amount: suggestion.Amount,
+		OutgoingChanSet: loopdb.ChannelSet{
+			suggestion.Channel.ToUint64(),
+		},
+		MaxPrepayRoutingFee: prepayMaxFee,
+		MaxSwapRoutingFee:   routeMaxFee,
+		MaxMinerFee:         defaultMaximumMinerFee,
+		MaxSwapFee:          maxSwapFee,
+		MaxPrepayAmount:     defaultMaximumPrepay,
+		SweepConfTarget:     loop.DefaultSweepConfTarget,
```
joostjager: :+1:

liquidity/liquidity.go
```diff
@@ -85,6 +91,12 @@ type Config struct {
 // Parameters is a set of parameters provided by the user which guide
 // how we assess liquidity.
 type Parameters struct {
+	// FailureBackOff is the amount of time that we require passes after a
+	// channel has been part of a failed loop out swap before we suggest
+	// using it again.
+	// TODO(carla): add exponential backoff
```
joostjager, line 135: Do you want to keep this comment here? There is so much to do still.

liquidity/liquidity.go
```diff
@@ -279,14 +292,50 @@ func (m *Manager) getEligibleChannels(ctx context.Context,
...
+		// We only backoff for off chain payment failures because this
+		// error indicates that our swap failed to route. We do not
+		// want to repeatedly try to route through bad channels which
+		// remain unbalanced because they cannot route a swap, so we
+		// backoff. Our other failure types do not represent temporary
+		// failures, so we do not backoff. We don't backoff on the
+		// temporary state itself because we require a restart to clear
+		// this state, which would prevent use of suggestions
+		// completely.
+		// TODO(carla): backoff entirely if we had a temporary failure
```
joostjager, line 463: Ah, this is a TODO because we don't have this information available easily now? Although I think that with temp failures, the swaps are still in a pending state, so we are already backing off entirely?

liquidity/liquidity.go
```diff
@@ -182,7 +279,13 @@ func (m *Manager) SetParameters(params Parameters) error {
...
+		FailureBackOff:             params.FailureBackOff,
+		MaximumPrepay:              params.MaximumPrepay,
+		MaximumMinerFee:            params.MaximumMinerFee,
+		MaximumSwapFeePPM:          params.MaximumSwapFeePPM,
+		MaximumRoutingFeePPM:       params.MaximumRoutingFeePPM,
+		MaximumPrepayRoutingFeePPM: params.MaximumPrepayRoutingFeePPM,
+		ConfTarget:                 params.ConfTarget,
 		ChannelRules: make(map[lnwire.ShortChannelID]*ThresholdRule,
 			len(params.ChannelRules)),
 	}
```
joostjager, line 291: You could also start off with `paramCopy := params` and then just clone the `ChannelRules`.

liquidity/liquidity.go
```diff
@@ -274,10 +413,10 @@ func makeLoopOutRequest(suggestion *LoopOutRecommendation) loop.OutRequest {
 		},
 		MaxPrepayRoutingFee: prepayMaxFee,
 		MaxSwapRoutingFee:   routeMaxFee,
-		MaxMinerFee:         defaultMaximumMinerFee,
-		MaxSwapFee:          maxSwapFee,
-		MaxPrepayAmount:     defaultMaximumPrepay,
-		SweepConfTarget:     loop.DefaultSweepConfTarget,
+		MaxMinerFee:         quote.MinerFee,
+		MaxSwapFee:          quote.SwapFee,
+		MaxPrepayAmount:     quote.PrepayAmount,
```
joostjager, line 418: Or should just our max values be used here (not the quote)? Especially with the `MinerFee` using the quote takes away all the leeway that we had with the max.

joostjager (approved):

cmd/loop/liquidity.go
carlaKC: thought about it, but it could be confused with the actual loop config?

liquidity/liquidity.go
carlaKC: It's my first TODO in the loop repo, so will leave it in. I do also think that un-swappable peers/channels will be important to address going forward with autoloop.

liquidity/liquidity.go
carlaKC: Makes sense for `MinerFee`, but I think using the quote value for swap fee/ prepay amount is ok? Values are unlikely to change

joostjager: Yes, those are ok, because the swap is only aborted if it can't happen for that price. The max miner fee is a security thing. We may not be able to get the htlc swept in time if chosen too low.

Is that factor 100x still somewhere for autoloop now?

liquidity/liquidity.go
joostjager: There are three in this PR :stuck_out_tongue:

carlaKC:
Latest [push](https://github.com/lightninglabs/loop/compare/ede428090294beef90649b8d156d6d0fe8231343..f6551a2918245014b3dc3069984b4e14ffa09695) separates out the sat/byte at which we will suggest swaps and the max miner fee (which now has a multiplier applied).

liquidity/liquidity.go
carlaKC: Split this out into sat/byte for our initial decision whether we want to even bother with suggestions and the max miner fee as discussed

cmd/loop/liquidity.go
```diff
@@ -158,19 +158,176 @@ func setParam(ctx *cli.Context) error {
...
 
+var setParamsCommand = cli.Command{
+	Name:        "setparams",
+	Usage:       "update the parameters set for the liquidity manager",
+	Description: "Updates the parameters set for the liquidity manager.",
+	Flags: []cli.Flag{
+		cli.IntFlag{
+			Name: "sweeplimit",
+			Usage: "the limit placed on our estimated sweep fee " +
+				"in sat/byte.",
```
guggero, line 185: nit: should probably say `sat/vbyte` as that's usually what we work with.

liquidity/liquidity.go
```diff
@@ -1,21 +1,112 @@
...
+	// we need to sweep the preimage.
+	defaultMaximumMinerFee = 15000 * 100
+
+	// defaultMaximumPrepay is the default limit we place on prepay
+	// invoices.
+	defaultMaximumPrepay = 30000
+
+	// defaultSweepFeeLimit is the default limit we place on estimated
+	// sweep fees, (750 *4 /1000 = 3 sat/byte).
+	defaultSweepFeeLimit = chainfee.SatPerKWeight(750)
```
guggero, line 71: nit: call this `defaultSweepFeeRateLimit` to distinguish from absolute fee limits like `defaultMaximumMinerFee`?
Also, it's `sat/vByte` here as well I think.

liquidity/liquidity.go
```diff
@@ -25,25 +116,78 @@ type Config struct {
...
+	// FailureBackOff is the amount of time that we require passes after a
+	// channel has been part of a failed loop out swap before we suggest
+	// using it again.
+	// TODO(carla): add exponential backoff
+	FailureBackOff time.Duration
+
+	// SweepFeeLimit is the limit that we place on our estimated sweep fee.
+	// A swap will not be suggested if estimated fee rate is above this
+	// value.
+	SweepFeeLimit chainfee.SatPerKWeight
```
guggero, line 153: Call it fee rate here as well?

liquidity/liquidity.go
```diff
@@ -71,6 +223,37 @@ func (p Parameters) validate() error {
 		}
 	}
 
+	if p.SweepFeeLimit <= 0 {
```
guggero, line 226: We need to make sure it's not below the `chainfee.FeePerKwFloor`.

guggero (commented):
Looks good, conversion check out. I mainly looked at https://github.com/lightninglabs/loop/compare/ede428090294beef90649b8d156d6d0fe8231343..f6551a2918245014b3dc3069984b4e14ffa09695.

The main comment is about consistent naming of `sat/vByte` as that's what we really use.

carlaKC:
Var renames + some extra comments in last [push](https://github.com/lightninglabs/loop/compare/33155ba0b9650304896b9b41ae4374db84d6bffe..8c5c671c8a2c025a7ffdd849396d1bc95a84738d) + rebase an change that adds macaroon permissions for suggestions.

PR #291, opened 2020-09-09
multi: move swapServerClient interface into loorpc package

carlaKC:
Move our wrapper for the swapServerClient into the looprpc package, since it is just a wrapper for our rpc. This is required to avoid circular dependencies when we have packages that the base loop package needs to import which also requie access to rpc calls using this wrapper (required in #289).

There are two naming collisions with the rpc:
`SwapServerClient` -> `SwapServer`
`LoopInRequest` => `InRequest` (to match the existing `OutRequest`)

We could arguably also pull this out into a separate `server` package, open to that direction.


carlaKC:
Ended up not needing this, shelving it for now.

PR #293, opened 2020-09-10
mutli: add release notes file to track in-repo

carlaKC:
This PR adds a file to track release notes in-repo. 
Open to suggestions to how we do this, but I think an append-only log will work?

For our next client release we will have a list like this:
```
# Next Release:
Feature 1
Feature 2
```

On release, we just shift the heading up:
```
# Next Release:

# v0.9.0
Feature 1
Feature 2
```

And just keep doing that. Decided on "Next Release" because we don't always know whether a change will go into a point release or a major one.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,0 +1,18 @@
+# Loop Client Release Notes
+This file tracks release notes for the loop client. 
+
+### Developers: 
+* When new features are added to the repo, a short description of the feature should be added under the "Next Release" heading.
+* This should be done in the same PR as the change so that our release notes stay in sync!
+
+### Release Manager: 
+* All of the items under the "Next Release" heading should be included in the release notes.
+* As part of the PR that bumps the client version, the "Next Release" heading should be replaced with the release version including the changes.
+
+## Next Release
+
+#### New Features
+
+#### Breaking Changes
+
+#### Bug Fixes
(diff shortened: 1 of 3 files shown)


carlaKC:
Requesting review from everybody, since the process affects us all. Feel free to remove yourself!

release_notes.md
```diff
@@ -0,0 +1,12 @@
+# Loop Client Release Notes
+This file tracks release notes for the loop client. 
+
+Developers: 
+* When new features are added to the repo, a short description of the feature should be added under the Next Release heading.
```
guggero, line 5: nit: put "Next Release" in quotes?

guggero (approved):
I like the proposed process! We should definitely try this and possibly also adopt in our other projects :100:

release_notes.md
```diff
@@ -0,0 +1,12 @@
+# Loop Client Release Notes
+This file tracks release notes for the loop client. 
+
+Developers:
```
bhandras, line 4: Consider making **Developers** and **Release Manager** bold to stand out better.

bhandras (approved):
Great idea!

joostjager:
You could also add a PR template to the repo to remind us about adding the release notes.

carlaKC:
Added a PR template (just with a reminder about notes, I don't think we need the full list that lnd has) and headings to the release notes (New Features/ Breaking Changes/ Bugfixes).

PR #294, opened 2020-09-11
release_notes: add macaroons

guggero:
Forgot to add the release notes for the macaroons in #281.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -15,4 +15,13 @@ This file tracks release notes for the loop client.
 
 #### Breaking Changes
 
+* Macaroon authentication has been enabled for the `loopd` gRPC and REST
+  connections. This makes it possible for the loop API to be exposed safely over
+  the internet as unauthorized access is now prevented.
+  
+  The daemon will write a default `loop.macaroon` in its main directory. For
+  mainnet this file will be picked up automatically by the `loop` CLI tool. For
+  testnet you need to specify the `--network=testnet` flag.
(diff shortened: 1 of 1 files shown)


release_notes.md
```diff
@@ -15,4 +15,8 @@ This file tracks release notes for the loop client.
 
 #### Breaking Changes
 
+* Macaroon authentication has been enabled for the `loopd` gRPC and REST
```
carlaKC, line 18: Perhaps we want to mention that it will pick up by default on mainnet, but testnet needs the flag?

carlaKC (approved):
LGTM! 

Not sure how much detail with these notes (can discuss as a team sometime), but I think it's worth pointing out that testnet now needs the network flag in this case.  

> Forgot to add the release notes for the macaroons in #281.

PR template will have your back in the future!

release_notes.md
joostjager, line 18: Do you want to add why this (in combination with tls) is good?

joostjager: I mean something about remote node support

guggero: Okay, I added the comment about the network. The rest (what directory things are located and so on) is described in the linked documentation so I think that should be fine.

guggero: Added a comment about exposing the port to the internet.

joostjager (approved):

release_notes.md
joostjager: I think this was already the case for 0.9.0.

PR #295, opened 2020-09-15
liquidity: add budget-limited autoloop

carlaKC:
Release notes tbd. 

Some open questions: 
- Do we want to disallow manual swaps completely if autoloop is enabled to reduce risk of manual swaps interacting with autoloop. It's still possible that lnd is dispatching payments manually anyway, so for this first version which is not "flow-aware", it may not make much of a difference

#### Pull Request Checklist


liquidity/liquidity.go
```diff
@@ -386,7 +450,33 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+	})
+
+	// Run through our suggested swaps in descending order of amount and
+	// return the first swap that will fall beneath our fee budget if it
+	// succeeds.
+	for _, swap := range suggestions {
+		successFees := swap.MaxMinerFee + swap.MaxSwapFee +
+			swap.MaxPrepayRoutingFee + swap.MaxSwapRoutingFee
+
+		if spent+successFees <= m.params.AutoFeeBudget {
```
guggero, line 551: I'm not sure I understand why only one swap is suggested here.
Wouldn't we want to suggest _all_ of the swaps that can fit in our remaining budget? Or is this a safety measure to not overshoot the budget by only suggesting one at a time?

loopd/swapclient_server.go
```diff
@@ -78,6 +79,11 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 		}
 	}
 
+	// Check that the label is valid.
+	if err := labels.Validate(in.Label); err != nil {
```
guggero, line 83: nit and pre-existing: most of the logic in Validate could be replaced by `strings.HasPrefix()`.

liquidity/liquidity.go
```diff
@@ -420,6 +510,71 @@ func (m *Manager) makeLoopOutRequest(suggestion *LoopOutRecommendation,
 	}
 }
 
+// calculateExpenditure calculates the total amount that has been spent by
+// automatically dispatched swaps. If there is an automatically dispatched swap
+// currently in progress, this function will fail with errAutoSwapInFlight
+// because we cannot accurately calculate off-chain fees for an in-flight
+// swap.
+func (m *Manager) calculateExpenditure(ctx context.Context,
```
guggero, line 614: nit: should be called `calculateOutExpenditure` or will it also return loop in fees in the future? If the latter is true, maybe we could already include the return parameter but always return it as 0.

liquidity/liquidity.go
```diff
@@ -220,6 +243,10 @@ func (p Parameters) validate(minConfs int32) error {
 		return ErrZeroMinerFee
 	}
 
+	if p.AutoFeeBudget < 0 {
```
guggero, line 268: Are there some sanity checks we can perform on the start date as well? Not sure if it makes sense or is even necessary...

liquidity/liquidity.go
```diff
@@ -420,6 +510,71 @@ func (m *Manager) makeLoopOutRequest(suggestion *LoopOutRecommendation,
...
+		// so we can just add our total cost.
+		case loopdb.StateTypeSuccess:
+			totalFees += out.LastUpdate().Cost.Total()
+
+		// If our swap failed, we will have have paid fees. However, if
+		// an automated swap failed but the prepay was already pulled,
+		// we add that amount to our fee total, because it is an expense
+		// that the auto-looper incurred.
+		case loopdb.StateTypeFail:
+			if out.State().State == loopdb.StateFailSweepTimeout {
```
guggero, line 653: Could make the `switch` statement on the `State` directly to avoid this nested `if`. Would almost look the same, except above it would be `loopdb.StateSuccess` instead of `loopdb.StateTypeSuccess`.

liquidity/liquidity.go
```diff
@@ -327,15 +368,47 @@ func cloneParameters(params Parameters) Parameters {
 	return paramCopy
 }
 
+// autoloop gets a set of suggested swaps and dispatches them automatically if
+// we have automated looping enabled.
+func (m *Manager) autoloop(ctx context.Context) error {
+	swaps, err := m.SuggestSwaps(ctx, true)
```
guggero, line 376: Not sure how "expensive" this call is, but maybe it makes sense to exit here already if auto loop is disabled?

guggero (commented):
Did a high-level pass. I really like the concept, it's simple but powerful. Good work on splitting up the PRs into easily digestible chunks! This reads very well.

liquidity/liquidity.go
```diff
@@ -420,6 +510,71 @@ func (m *Manager) makeLoopOutRequest(suggestion *LoopOutRecommendation,
...
+		// If we succeeded, we know that we have an non-nil last update
+		// so we can just add our total cost.
+		case loopdb.StateTypeSuccess:
+			totalFees += out.LastUpdate().Cost.Total()
+
+		// If our swap failed, we will have have paid fees. However, if
+		// an automated swap failed but the prepay was already pulled,
+		// we add that amount to our fee total, because it is an expense
+		// that the auto-looper incurred.
+		case loopdb.StateTypeFail:
```
joostjager, line 652: IIRC, `Cost.Total()` should always be correct if we are in a final fail state? It shouldn't be necessary to manually add the prepay amt

liquidity/liquidity.go
```diff
@@ -152,6 +165,14 @@ type Parameters struct {
 	// dispatched swaps in our current budget, inclusive.
 	AutoFeeStartDate time.Time
 
+	// AutoOut is a boolean that indicates whether the liquidity manager
+	// should dispatch loop out swaps automatically.
+	AutoOut bool
+
+	// MaxAutoInFlight is the maximum number of in-flight automatically
+	// dispatched swaps we allow.
+	MaxAutoInFlight int
```
joostjager, line 174: Unused in this commit. I also thought you would never do more than one, because of uncertainty about the final fee?

liquidity/liquidity.go
```diff
@@ -332,6 +361,41 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
 		return nil, err
 	}
 
+	// Calculate the total amount we have spent on automatically dispatched
+	// swaps in our relevant period. We catch the error thrown if we
+	// currently have a swap in flight and return no suggestions. This is
+	// because we cannot accurately worst-case off-chain fees for an in
```
joostjager, line 442: We do have a hard limit on all of the fees for in-flight swaps?

liquidity/liquidity.go
```diff
@@ -508,6 +590,18 @@ func (m *Manager) makeLoopOutRequest(suggestion *LoopOutRecommendation,
...
 	}
+
+	if autoOut {
+		request.Label = labels.AutoOutLabel()
+
+		addr, err := m.cfg.Lnd.WalletKit.NextAddr(ctx)
+		if err != nil {
+			return loop.OutRequest{}, err
+		}
+		request.DestAddr = addr
```
joostjager, line 603: Isn't it that no address defaults to generating an address when the swap is launched? Or otherwise the returned requests from `SuggestSwaps` won't work (no dest addr)

joostjager (commented):
Yes, I agree with @guggero, looks good.

liquidity/liquidity.go
carlaKC: Could add a check that the start date isn't in the future? Other than that, I do like being able to set a 0 value start date, which just means it's a budget for all time?

liquidity/liquidity.go
carlaKC: `SuggestSwaps` would be taken by the user and passed to the rpc server (which fills the addr), we pass directly to the swap client to bypass the label check (which exists on rpc server level) so we need to fill the addr for autoloop (but not for suggestions).

liquidity/liquidity.go
carlaKC: I was thinking that it's safer to have a single swap at a time, but have updated this to a `MaxInFlight` param so that it can be toggled as people like. Since we only do one swap per channel at a time, and use absolute worst case fees, I think it's safe to give people the option of multiple autoloops at once? Esp since sweeps could take a very long time for loop out.

liquidity/liquidity.go
carlaKC: Also been thinking about this a bit, as is I have the autoloop logging the swaps that it _would_ have dispatched, which I think is nice. But it does cost us 1x quote call to the server per suggested swap, which could get costly. 

Since this is quite a new feature and you have to explicitly set a per-channel rule (which would result in a quote), WDYT of leaving it as is now to give some visibility into the functionality and then turning it off at a later stage?

liquidity/liquidity.go
```diff
@@ -485,6 +571,92 @@ func (m *Manager) makeLoopOutRequest(suggestion *LoopOutRecommendation,
...
+
+	var summary existingAutoLoopSummary
+
+	for _, out := range loopOuts {
+		if out.Contract.Label != labels.AutoOutLabel() {
+			continue
+		}
+
+		// Skip over any swaps that completed before our start time.
+		if out.LastUpdateTime().Before(m.params.AutoFeeStartDate) {
```
bhandras, line 763: It could still be a pending swap right?

liquidity/liquidity.go
```diff
@@ -485,6 +571,92 @@ func (m *Manager) makeLoopOutRequest(suggestion *LoopOutRecommendation,
...
+			summary.pendingFees += worstCaseOutFees(
+				out.Contract.MaxPrepayRoutingFee,
+				out.Contract.MaxSwapRoutingFee,
+				out.Contract.MaxSwapFee,
+				out.Contract.MaxMinerFee,
+				mSatToSatoshis(prepay.Value),
+			)
+
+		} else {
+			summary.spentFees += out.State().Cost.Total()
```
bhandras, line 790: What if the state is `StateTypeFail`?

liquidity/liquidity.go
```diff
@@ -444,6 +462,14 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
 		return nil, nil
 	}
 
+	// If we have already reached our total allowed number of in flight
+	// swaps, we do not suggest any more at the moment.
+	allowedSwaps := m.params.MaxAutoInFlight - summary.inFlightCount
+	if allowedSwaps <= 0 {
+		log.Debugf("%v autoloops allowed, %v in flight")
```
bhandras, line 531: The format args are missing.

liquidity/liquidity.go
```diff
@@ -390,10 +424,36 @@ func cloneParameters(params Parameters) Parameters {
...
+// autoloop gets a set of suggested swaps and dispatches them automatically if
+// we have automated looping enabled.
+func (m *Manager) autoloop(ctx context.Context) error {
+	swaps, err := m.SuggestSwaps(ctx, true)
+	if err != nil {
+		return err
+	}
+
+	for _, swap := range swaps {
+		swap := swap
```
bhandras, line 438: nit: is this needed?

bhandras (commented):
Very nice change set, overall looks good, added a few comments and nits.

liquidity/liquidity.go
guggero: Yeah, I thought the same, maybe it is useful to turn it on later with a future date. Let's leave it the way it is then.

liquidity/liquidity.go
guggero: Sounds good to me :+1:

liquidity/liquidity.go
guggero: Yeah, makes sense :+1:

loopdb/loopin.go
```diff
@@ -125,8 +125,9 @@ func putLabel(bucket *bbolt.Bucket, label string) error {
 		return nil
 	}
 
-	if err := labels.Validate(label); err != nil {
-		return err
+	// Check that the label does not exceed our maximum length.
+	if len(label) > labels.MaxLength {
```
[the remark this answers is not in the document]
guggero: Yes, but `Validate()` is no longer called for swaps that are created by the autolooper. So this persistence level check probably makes sense.

liquidity/liquidity.go
```diff
@@ -88,9 +90,20 @@ const (
...
 var (
+	// defaultBudget is the default autoloop budget we set. This budget will
+	// only be used for automatically dispatched swaps if autoloop is
+	// explicitly enabled, so we are happy to set a non-zero value here. The
+	// amount chosen simply uses the current defaults to provide budget for
+	// a single swap. We don't have a swap amount to calculate our maximum
+	// routing fee, so we use 0.16 BTC for now.
+	defaultBudget = defaultSwapFeePPM + defaultMaximumMinerFee +
+		ppmToSat(defaultMaximumPrepay, defaultPrepayRoutingFeePPM) +
+		ppmToSat(1600000000, defaultRoutingFeePPM)
```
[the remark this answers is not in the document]
guggero: Or could use `lnd.MaxBtcFundingAmount`.

I also think the `defaultSwapFeePPM` needs to be wrapped in a `ppmToSat` call.

release_notes.md
```diff
@@ -31,6 +31,18 @@ This file tracks release notes for the loop client.
   value is configurable).
 * The `debug` logging level is recommended if using this feature.
 
+##### Introducing Autoloop
+* This release includes support for opt-in automatic dispatch of loop out swaps, 
+  based on the output of the `Suggestions` endpoint. 
+* To enable the autolooper, the following command can be used: 
+  `loop setParams --autoout=true --autobudget={budget in sats} --budgetstart={start time for budget}`
```
guggero, line 38: nit: s/loop setParams/loop setparams/

guggero (commented):
This is getting very close, nice work!

I think it would be good to also create an integration test case for this, to make sure everything's executed as expected. You can push the branch to the upstream repo so you can reference the commit in the itest setup.

loopdb/loopin.go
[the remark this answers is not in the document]
carlaKC: Yeah that was my thinking, it's pretty unlikely that we set a reserved label > max length, but I didn't think the check hurt.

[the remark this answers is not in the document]
bhandras: Ok, makes sense.

liquidity/liquidity.go
```diff
@@ -396,6 +427,23 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
+	// Get a summary of our existing swaps so that we can check our autoloop
+	// budget.
+	summary, err := m.checkExistingAutoLoops(ctx, loopOut)
+	if err != nil {
+		return nil, err
+	}
+
+	if summary.totalFees() >= m.params.AutoFeeBudget {
+		log.Debugf("autoloop budget: %v exhausted, %v spent on "+
+			"completed swaps, %v reserved for ongoing swaps, "+
```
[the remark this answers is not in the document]
carlaKC: indeed, removed!

liquidity/liquidity.go
```diff
@@ -396,6 +427,23 @@ func (m *Manager) SuggestSwaps(ctx context.Context) (
...
 
+	// Get a summary of our existing swaps so that we can check our autoloop
+	// budget.
+	summary, err := m.checkExistingAutoLoops(ctx, loopOut)
+	if err != nil {
+		return nil, err
+	}
+
+	if summary.totalFees() >= m.params.AutoFeeBudget {
+		log.Debugf("autoloop budget: %v exhausted, %v spent on "+
```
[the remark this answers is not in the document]
carlaKC: yeah that's nicer, done

liquidity/liquidity.go
carlaKC: The else will cover `StateTypeFail` and `StateTypeSuccess`, and our costs are reflected in state updates for both?

looprpc/client.proto
```diff
@@ -762,6 +762,33 @@ message LiquidityParameters{
...
+    be limited to the fee categories set by these parameters, and total
+    expenditure will be limited to the auto out budget.
+    */
+    bool auto_loop_out = 10;
+
+    /*
+    The total budget for automatically dispatched swaps since the budget start
+    time, expressed in satoshis.
+    */
+    uint64 auto_out_budget_sat = 11;
```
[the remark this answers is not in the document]
carlaKC: Yeah definitely, I think that being able to set min and max swap amount would be great. Since we're already bounded by the minimum swap amount that the server sets, I was going to leave that for a followup if that's ok?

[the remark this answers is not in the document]
bhandras: Yes indeed, just checked out that part of the code, the restrictions will do for now.

liquidity/liquidity.go
bhandras: Discussed offline. Please leave a comment as it's not trivial.

liquidity/liquidity.go
carlaKC: yeah indeed! moved the timestamp check to only apply to completed swaps, since anything that's pending is in our budget.

liquidity/liquidity.go
carlaKC: done

liquidity/liquidity.go
carlaKC: done, comment at top of block.

liquidity/liquidity.go
carlaKC: yeah, reference to a range variable: https://medium.com/@betable/3-go-gotchas-590b8c014e0a
added a comment since it's not very clear.

carlaKC:
Last set of changes [here](https://github.com/lightninglabs/loop/compare/a4a4f3c0b4643e4cc5bb731986e7de75357acfad..4debd622cc43250a55c5bff60b6671f8228745cb), but they included some incorrect changes to go.sum which were fixed in the [subsequent push](https://github.com/lightninglabs/loop/compare/4debd622cc43250a55c5bff60b6671f8228745cb..c0c0572571325ab88ef082906b020c47b07bacad).

carlaKC:
Working on an itest for this, feel free to hold off review till that's up!

guggero (approved):
LGTM, nice work :tada:

bhandras (approved):
LGTM, awesome that we have this! :1st_place_medal:

PR #298, opened 2020-09-18
mod+loopd: bump to lnd v0.11.1-beta.rc3

guggero:
To make sure we can use the latest version of loop in LiT, we need to
make sure loop compiles against the actual v0.11.1 branch of lnd. There
were some conflicting changes to the macaroon service that didn't make
it to the branch that we need to roll back. Those changes don't affect
loop at all as we were always using the default root key ID anyway.


carlaKC (approved):
Also LGTM!

joostjager (approved):

joostjager:
>Those changes don't affect loop at all as we were always using the default root key ID anyway.

I guess this is the risk of releasing a `loop` client build against non-released lnd, especially when something persistent is involved.

PR #303, opened 2020-10-12
go.sum: run go mod tidy to cleanup old dependencies

carlaKC:
`go mod tidy` (which our current modules check relies on) does not detect items that should be deleted in `go.sum`, this has recently been [addressed](https://github.com/golang/go/issues/33008) in go, so we can wait on the fix.  In the meantime, this PR cleans up our go.sum.


bhandras (approved):
LGTM

bhandras:
Out of curiosity, how did you find these old deps?

guggero (approved):
LGTM :zap:

carlaKC:
> Out of curiosity, how did you find these old deps?

When you make a change that does need to change `go.sum`, then modules pick up that there are stale entries. So I got a diff I wasn't expecting in #295.

PR #304, opened 2020-10-12
Send user agent string with each swap

guggero:
This PR adds the daemon's user agent string to each new swap.


version.go
```diff
@@ -29,8 +29,18 @@ const (
...
+	// defaultAgentName is the default name of the software that is added as
+	// the first part of the user agent string.
+	defaultAgentName = "loopd"
 )
 
+// AgentName stores the name of the software that is added as the first part of
+// the user agent string. This defaults to the value "loopd" when being run as
+// a standalone component but can be overwritten by LiT for example when loopd
+// is integrated into the UI.
+var AgentName = defaultAgentName
```
bhandras, line 42: When you say "can be overwritten..." how do you mean exactly? Should this be set trough the RPC interface when integrated?

guggero: We add `loopd` as a library in LiT and can simply set this global package-level variable. No need to add another RPC interface.

bhandras: I see, thanks for the explanation.

bhandras (approved):
LGTM :rocket:

version.go
```diff
@@ -52,6 +62,25 @@ func Version() string {
...
+// that is submitting swaps to the loop server.
+func UserAgent() string {
+	// Start with the major, minor, and patch versions.
+	version := fmt.Sprintf("%d.%d.%d", appMajor, appMinor, appPatch)
+
+	// Append pre-release version if there is one.  The hyphen called for
+	// by the semantic versioning spec is automatically appended and should
+	// not be contained in the pre-release string.  The pre-release version
+	// is not appended if it contains invalid characters.
+	preRelease := normalizeVerString(appPreRelease)
```
carlaKC, line 75: nit: put this in a versionStr() function rather than duplicate what we already have in `Version`?

carlaKC (approved):
LGTM! 

Only one nit + could possibly squash the rpc changes with the commit where we actually set user agent string (since it's only 2 lines)

carlaKC:
Hm just thinking, maybe also add this to `release_notes`? Makes it a little more likely that other devs building on loop will set the value?

alexbosworth (approved):

PR #305, opened 2020-10-13
liquidity: add force tick endpoint to trigger autoloop in tests

carlaKC:
This PR adds a force tick endpoint for testing purposes, to allow testing the autoloop functionality in itests without needing to wait for ticks (which only trigger every 10 minutes). The endpoint will not work on mainnet, since it is intended for testing only. 

Dependent on #295, only the last 2 commits need review. 

#### Pull Request Checklist


cmd/loop/liquidity.go
```diff
@@ -399,3 +399,24 @@ func suggestSwap(ctx *cli.Context) error {
 
 	return nil
 }
+
+var forceTickCmd = cli.Command{
```
guggero, line 403: Do we even need a CLI for this? I assume we'll want to trigger this over RPC in the itests only. If people shouldn't use it, why give them a command?

guggero (approved):
LGTM, barring the comment about the CLI addition :+1:

bhandras (approved):
LGTM :ocean:  One comment I have is that we could separate test endpoints in the future to a subserver guarded by build tag.

cmd/loop/liquidity.go
carlaKC: Included it because I used it a bit for testing, but agreed we don't need it. Removed.

carlaKC:
> One comment I have is that we could separate test endpoints in the future to a subserver guarded by build tag.

I think this is actually a really good idea, updated this PR to add a debug server behind a dev tag. Decided against the full subserver setup because it seems unlikely that we'll need multiple servers for loop, debug and swap client should be fine?

looprpc/debug.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package looprpc;
+
+/**
+DebugClient is a service that exposes endpoints intended for testing purposes.
+These endpoints should not operate on mainnet, and should only be included if
+loop is built with the dev build tag.
+*/
+service DebugClient {
```
bhandras, line 10: nit: (subjective): I'd call this simply `Debug`

bhandras (approved):
Very nice! LGTM

loopd/swapclient_debug_server.go
```diff
@@ -0,0 +1,25 @@
+// +build dev
```
guggero, line 1: I think the "idiomatic" naming of the file would be `swapclient_server_debug.go`, the build tag suffix usually comes last. Alternatively we could just add this method to `swapclient_server.go`. It wouldn't be used if not registered. But I guess that comes down to personal preference.

looprpc/debug.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package looprpc;
+
+/**
```
guggero, line 5: nit: double asterisk.

guggero (approved):
Awesome to have this functionality in a proper way, this should help us test more and more things in the integration tests. Only nits on my part, LGTM :100:

loopd/swapclient_debug_server.go
carlaKC: Renamed and moved function + macaroons into `swapclient_server_debug.go`.

PR #306, opened 2020-10-15
Update lndclient, add labels to on-chain transactions

guggero:
We update to the latest lndclient that now requires a label when
publishing an on-chain transaction. Instead of just adding an empty
string, we add proper labels.

#### Pull Request Checklist


labels/labels.go
```diff
@@ -26,6 +37,20 @@ var (
 	// ErrReservedPrefix is returned when a label contains the prefix
 	// which is reserved for internally produced labels.
 	ErrReservedPrefix = errors.New("label contains reserved prefix")
+
+	// loopOutSweepSuccess is the label used for loop out swaps to sweep
+	// the HTLC in the success case.
+	LoopOutSweepSuccess = fmt.Sprintf(
```
carlaKC, line 43: We have two kinds of labels overlapping here - labels for swaps and labels for individual transactions. 
* For swap labels, we use the reserved prefix to ensure that people don't label manual swaps as autoloops (which would mess up our budget). 
* For lnd labels, we do have internal labels (such as channel opens), but do not have a reserved prefix (perhaps we should have, but I didn't think of the requirement at the time, and if people want to intentinally mess up our labelling it's not _really_ our problem, vs autoloop where it could cause actual issues). 

So tldr, I don't think we need the reserved label on these?

labels/labels.go
```diff
@@ -17,6 +17,17 @@ const (
 	// autoOut is the label used for loop out swaps that are automatically
 	// dispatched.
 	autoOut = "autoloop-out"
+
+	// loopOutSweepSuccess is the label used for loop out swaps to sweep
+	// the HTLC in the success case.
+	loopOutSweepSuccess = "loop-out-sweep-success"
```
carlaKC, line 23: I'm currently working on custom categories for faraday reports (lightninglabs/faraday/pull/85) which allow you to create a new category in your statement by matching tx labels to a regex. Perhaps add an explicit prefix var (`loopd`) and then make all these labels build on that so it's clear that we need the prefix? Nice to have a consistent format across our services building on lnd.

release_notes.md
```diff
@@ -9,6 +9,13 @@ This file tracks release notes for the loop client.
 * All of the items under the "Next Release" heading should be included in the release notes.
 * As part of the PR that bumps the client version, the "Next Release" heading should be replaced with the release version including the changes.
 
+## Next release
+
+#### NewFeatures
```
carlaKC, line 14: nit/ask: do you mind copying the `Breaking Changes` and `Bug Fixes` headings in as well? Can see this getting inconsistent over time if we don't copy them across with each new release.

labels/labels.go
```diff
@@ -26,6 +37,20 @@ var (
 	// ErrReservedPrefix is returned when a label contains the prefix
 	// which is reserved for internally produced labels.
 	ErrReservedPrefix = errors.New("label contains reserved prefix")
+
+	// loopOutSweepSuccess is the label used for loop out swaps to sweep
```
bhandras, line 41: nit: capitalize

bhandras (approved):
LGTM, nice! :+1:

carlaKC (commented):
Really happy to see this change 🥇 :label_all_da_tings:

Some ponderings on label format from me. 

We also currently just add a memo saying "swap" to loop in invoices, which could probably be more descriptive, but that may break some existing checks/logic so perhaps leave for this PR.

labels/labels.go
guggero: Ah, I missed the different purposes of the labels. I moved the `lnd` level labels to its own file and added a more verbose pattern that contains some meta information as well.

release_notes.md
guggero: Sure thing!

carlaKC (approved):
magical ✨

PR #307, opened 2020-10-22
releasenotes: move autoloop and simplify instructions

carlaKC:
This PR moves the autoloop section of the release notes into the next release.

It also updates the instructions for the release to just clear everything that was there when we release, so that the notes are less ambiguous. These notes weren't really intended to replace the notes we add with tags, but rather to be a place where we can write updates, which can then be moved to the github tag and edited as we like once the release is ready.

Intended process is as follows:
- PRs include release notes under `Next Release`
- When it's time to release, bump the version and cut everything from `release_notes.md`
- Paste the release notes in the usual place

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -7,7 +7,10 @@ This file tracks release notes for the loop client.
 
 ### Release Manager: 
 * All of the items under the "Next Release" heading should be included in the release notes.
-* As part of the PR that bumps the client version, the "Next Release" heading should be replaced with the release version including the changes.
+* As part of the PR that bumps the client version, cut everything below the 'Next Release' heading. 
+* These notes can either be pasted in a temporary doc, or you can get them from the PR diff once it is merged. 
+* The notes are just a guideline as to the changes that have been made since the last release, they can be updated.
+* Once the version bump PR is merged and tagged, add the release notes to the tag on GitHub.
 
 ## Next release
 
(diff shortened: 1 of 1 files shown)


alexbosworth:
This doesn't really work logically because you can't paste into something that doesn't exist yet

carlaKC:
> This doesn't really work logically because you can't paste into something that doesn't exist yet

Could get it from the diff of the version bump PR? 

The motivation for trying this was to couple writing more detailed release notes with code changes. They can always be changed later for the release itself, but it's a starting point for release notes. I think it's worth trying for one more release, we can always revert if it's worse than no notes.

alexbosworth:
> > This doesn't really work logically because you can't paste into something that doesn't exist yet
> 
> Could get it from the diff of the version bump PR?
> 
> The motivation for trying this was to couple writing more detailed release notes with code changes. They can always be changed later for the release itself, but it's a starting point for release notes. I think it's worth trying for one more release, we can always revert if it's worse than no notes.

OK but then modify the instructions to indicate that process. I still don't think this is a good approach but I guess that is subjective

bhandras (approved):
LGTM :notebook:

alexbosworth:
Now there is an issue where the release notes for the last release are not present

carlaKC:
> Now there is an issue where the release notes for the last release are not present

Can I replace the link in the GitHub tag with the notes for 0.10?

alexbosworth:
> > Now there is an issue where the release notes for the last release are not present
> 
> Can I replace the link in the GitHub tag with the notes for 0.10?

Go for it

PR #311, opened 2020-10-30
docs: add autoloop documentation

carlaKC:
#### Pull Request Checklist

### DIFF
--- a/docs/autoloop.md
+++ b/docs/autoloop.md
@@ -1,0 +1,196 @@
+# Autoloop
+The loop client contains functionality to dispatch loop out swaps automatically,
+according to a set of rules configured for your node's channels, within a 
+budget of your choosing. 
+
+The autoloop functionality is disabled by default, and can be enabled using the 
+following command:
+```
+loop setparams --autoout=true
+```
+
+Swaps that are dispatched by the autolooper can be identified in the output of 
+`ListSwaps` by their label field, which will contain: `[reserved]: autoloop-out`.
+
+Even if you do not choose to enable the autolooper, we encourage you to 
+experiment with setting the parameters described in this document because the 
+client will log the actions that it would have taken to provide visibility into 
+its functionality. Alternatively, the `SuggestSwaps` rpc (`loop suggestswaps` 
+on the CLI) provides a set of swaps that the autolooper currently recommends, 
+which you can use to manually execute swaps if you'd like.
+
+Note that autoloop parameters and rules are not persisted, so must be set on 
+restart. We recommend running loopd with `--debuglevel=debug` when using this 
+feature.
+
+### Channel Thresholds 
+To setup the autolooper to dispatch swaps on your behalf, you need to tell it 
+which channels you would like it to perform swaps on, and the liquidity balance 
+you would like on each channel. Desired liqudity balance is expressed using 
+threshold incoming and outgoing percentages of channel capacity. The incoming 
+threshold you specify indicates the minimum percentage of your channel capacity
+that you would like in incoming capacity. The outgoing thresold allows you to 
+reserve a percentage of your balance for outgoing capacity, but may be set to 
+zero if you are only concerned with incoming capcity.
+
+The autolooper will perform swaps that push your incoming channel capacity to 
+at least the incoming threshold you specify, while reserving at least the 
+outgoing capacity threshold. Rules can be set as follows:
+
+```
+loop setrule {short channel id} --incoming_threshold={minimum % incoming} --outgoing_threshold={minimum % outgoing}
+```
+
+To remove a channel from consideration, its rule can simply be cleared:
+```
+loop setrule {short channel id} --clear
+```
+
+## Fees
+Fee control is one of the most important features of the autolooper, so we expose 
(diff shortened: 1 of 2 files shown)


docs/autoloop.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## Fees
+Fee control is one of the most important features of the autolooper, so we expose multiple fee related settings which can be used to tune the autolooper to your preference. Note that these fees are expressed on a per-swap basis, rather than as an overall budget. 
+
+### On-Chain Fees
+When performing a successful loop out swap, the loop client needs to sweep the on-chain HTLC sent by the server back into its own wallet. 
+
+#### Sweep Confirmation Target
+To estimate the amount that this sweep will cost, the client uses a confirmation target for the sweep - the number of blocks within which you would like this balance swept back to your wallet. The time to acquire your incoming liquidity is not dependent on sweep confirmation time, so we highly recommend setting a very large sweep confirmation target (up to 250 blocks), so that your sweep can go through with very low fees.
```
guggero, line 36: nit: s/To estimate the amount that/To estimate the amount of on-chain fees/ ?

docs/autoloop.md
```diff
@@ -0,0 +1,114 @@
...
+When performing a successful loop out swap, the loop client needs to sweep the on-chain HTLC sent by the server back into its own wallet. 
+
+#### Sweep Confirmation Target
+To estimate the amount that this sweep will cost, the client uses a confirmation target for the sweep - the number of blocks within which you would like this balance swept back to your wallet. The time to acquire your incoming liquidity is not dependent on sweep confirmation time, so we highly recommend setting a very large sweep confirmation target (up to 250 blocks), so that your sweep can go through with very low fees. 
+```
+loop setparams --sweepconf={target in blocks}
+```
+
+#### Fee Market Awareness
+The mempool often clears overnight, or on the weekends when fewer people are using chain space. This is an opportune time for the autolooper to dispatch a swap on your behalf while you sleep! Before dispatching a swap, the autolooper will get a fee estimate for you on-chain sweep transaction (using its `sweepconftarget`), and check it against the limit that has been configured. The `sweeplimit` parameter can be set to configure the autolooper to only dispatch in low-fee environments.
```
guggero, line 42: > The mempool often clears overnight, or on the weekends...

Those were good times :joy:

docs/autoloop.md
```diff
@@ -0,0 +1,114 @@
+# Autoloop
```
guggero, line 1: Would be nice to link to this document from the main README, otherwise people might not find it easily.

guggero (approved):
Great document, LGTM :scroll:

docs/autoloop.md
carlaKC: I 'member 😅

docs/autoloop.md
carlaKC: yeah good idea!

docs/autoloop.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## Fees
+Fee control is one of the most important features of the autolooper, so we expose multiple fee related settings which can be used to tune the autolooper to your preference. Note that these fees are expressed on a per-swap basis, rather than as an overall budget. 
+
+### On-Chain Fees
+When performing a successful loop out swap, the loop client needs to sweep the on-chain HTLC sent by the server back into its own wallet. 
+
+#### Sweep Confirmation Target
+To estimate the amount of on-chain fees that the swap wil require, the client uses a confirmation target for the sweep - the number of blocks within which you would like this balance swept back to your wallet. The time to acquire your incoming liquidity is not dependent on sweep confirmation time, so we highly recommend setting a very large sweep confirmation target (up to 250 blocks), so that your sweep can go through with very low fees.
```
guggero, line 36: nit: typo in 'wil'

docs/autoloop.md
```diff
@@ -0,0 +1,114 @@
...
+
+The default value for this parameter is 24hours, and it can be updated as follows:
+```
+loop setparams --failurebackoff={backoff in seconds}
+```
+
+## Manual Swap Interaction
+The autolooper will not dispatch swaps over channels that are already included in manually dispatched swaps - for loop out, this would mean the channel is specified in the outgoing channel swap, and for loop in the channel's peer is specified as the last hop for an ongoing swap. This check is put in place to prevent the autolooper from interfering with swaps you have created yourself. If there is an ongoing swap that does not have a restriction placed on it (no outgoing channel set, or last hop), then the autolooper will take no action until it has resolved, because it does not know how that swap will affect liquidity balances.
```
guggero, line 113: General nit (non-blocking): For further review and merging of PRs, it's generally useful to also try to adhere to the 80 character line length limit where possible (links and tables often don't fit but flow text looks nicer IMO).

guggero (approved):

docs/autoloop.md
carlaKC: tq!

docs/autoloop.md
carlaKC: makes sense, wrapped!

PR #312, opened 2020-10-30
mod: update to final lnd version

guggero:
To make it possible to remove the `replace` directive in LiT we update to the final `lnd` version.


carlaKC (approved):
🌷

bhandras (approved):
:+1:

PR #316, opened 2020-11-06
Add initiator field to user agent string

guggero:
When requesting a swap, a new `initiator` field can be set on the gRPC/REST
  interface that is appended to the user agent string that is sent to the server
  to give information about Loop usage. The initiator field is meant for user
  interfaces to add their name to give the full picture of the binary used
  (`loopd`, LiT) and the method/interface used for triggering the swap (`loop`
  CLI, autolooper, LiT UI, other 3rd party UI).

#### Pull Request Checklist


Makefile
```diff
@@ -10,7 +10,7 @@ GOINSTALL := GO111MODULE=on go install -v
 GOMOD := GO111MODULE=on go mod
 
 COMMIT := $(shell git describe --abbrev=40 --dirty)
-LDFLAGS := -ldflags "-X $(PKG)/build.Commit=$(COMMIT)"
+LDFLAGS := -ldflags "-X $(PKG).Commit=$(COMMIT)"
```
bhandras, line 13: :+1:

version.go
```diff
@@ -53,10 +53,21 @@ func Version() string {
 
 // UserAgent returns the full user agent string that identifies the software
 // that is submitting swaps to the loop server.
-func UserAgent() string {
+func UserAgent(initiator string) string {
+	// We'll only allow "safe" characters in the initiator portion of the
+	// user agent string and spaces only if surrounded by other characters.
+	initiatorAlphabet := semanticAlphabet + ". "
+	cleanInitiator := normalizeVerString(
```
bhandras, line 60: Could also have a max length for the `initiator` (but since it's enforced on the server side, it's not super important).

bhandras (approved):
Very clean PR, LGTM :1st_place_medal:

liquidity/autoloop_test.go
```diff
@@ -146,6 +146,7 @@ func TestAutoLoopEnabled(t *testing.T) {
 			SweepConfTarget: params.SweepConfTarget,
 			OutgoingChanSet: loopdb.ChannelSet{chanID1.ToUint64()},
 			Label:           labels.AutoOutLabel(),
+			Initiator:       autoloopSwapInitiator,
```
carlaKC, line 149: Really like being able to identify swaps from the autolooper! Wondering what the plan will be if we add autoloop to the LiT UI? Ideally we'd be able to easily update this to `LiT-autoloop`, will that be easy enough?

carlaKC (approved):
tACK 🦖

version.go
guggero: Yeah, makes sense. We don't want an UI to interfere with the success of submitting a swap. I limited the initiator part to 150 characters.

liquidity/autoloop_test.go
guggero: Good point, I haven't thought about that yet. My suggestion is to add it to the liquidity parameters of the autolooper. Then the UI can overwrite the default value similar to how it's done for normal swaps.

Issue #319, opened 2020-11-17
Autoloop V2 Features


carlaKC:
There are many directions we could take the autoloop feature in. This issue outlines a few features that could be implemented in upcoming iterations. 

Please comment if a specific feature would be helpful for your usecase, or you would like to request a feature that is not listed. 

### Liquidity Management
- Per-peer rules: rather than setting a liquidity thresholds on a per-channel basis, specify a desired liquidity balance per-peer
- Flow awareness: track lnd's sends/receives/forwards and inform autoloop's swap decisions based on htlc flows
- Custom swap amounts: allow user-set minimum and maximum swap amounts
- Auto-loop-in: add support for automated loop in

### Control
- Recurring budget: add a recurrence period for budget which will automatically roll over (eg, weekly/ monthly budget)
- Param handing: read parameters in from loop config, store on disk so they persist across restarts
- Batching: to save on fees, set autoloopers to execute at set times to take advantage of the server's ability to batch swaps

tdickman:
> Per-peer rules: rather than setting a liquidity thresholds on a per-channel basis, specify a desired liquidity balance per-peer

Taking this a step further, I would love to be able to set a default rule for all channels as in general, I want the same autoloop rules for all of my channels. In my case my desire is just to acquire more inbound liquidity.

> Batching: to save on fees, set autoloopers to execute at set times to take advantage of the server's ability to batch swaps

This would be awesome - I'd love to be able to batch loop operations across multiple channels / peers at once to save fees.

In addition, I'd love to just be able to set percentage fee limits rather than raw satoshis. If my node is doing a 2x larger loop, I'm okay with paying 2x more in fees. Normally the limits of what I'm willing to pay is percentage based.

Kixunil:
Wouldn't making loop scriptable be a better approach? Anyway, things that probably can't be done in scripts without direct support:

* Instead of executing `loop in` right away, somehow transfer the address to BTCPayServer and wait for payment, so that the next paying customer actually unknowingly pays for it, confusing chain analysis. Bonus points for doing PayJoin at the same time. Special prize if the PayJoin includes inputs from swap server.
* Optimize for UTXOs - if there's a UTXO with amount close to the desired amount, use that UTXO for `loop in`. Swap server could offer the same for `loop out` with discount.
* Automatically run the coins through some CoinJoin
[...]

PR #321, opened 2020-12-01
liquidity: add custom swap amounts to autoloop

carlaKC:
This PR adds the ability for clients to configure custom min/max swap amounts for autolooper. Before this PR, the minimum swap size was pegged to the server's minimum value, which could result in multiple small swaps for a single channel. Adding a configurable minimum allows clients to set their own minimum, tradeoff here is that a larger minimum means that we drop further below our target incoming %, but we make larger swaps so there is potential fee saving. 

High level tracking task: #319

#### Pull Request Checklist


docs/autoloop.md
```diff
@@ -191,6 +191,33 @@ The default value for this parameter is 24hours, and it can be updated as follow
...
 
+### Swap Size
+By default, the autolooper will execute a swap when the amount that needs to be
+rebalanced within a channel is equal to the swap server's minimum swap size. 
+This means that it will dispatch swaps more regularly, and ensure that channels 
+are not run down too far below their configured threshold. If you are willing 
+to allow your liquidity to drop further than the minimum swap amount below your 
+threshold, a custom minimum swap size can be set. If autolooper is configured 
+with a larger minimum swap size, it will allow channels to drop further below
+their target threshold, but will perform fewer swaps, potentially on saving
```
guggero, line 202: nit: s/on saving/saving on/

docs/autoloop.md
```diff
@@ -191,6 +191,33 @@ The default value for this parameter is 24hours, and it can be updated as follow
...
+
+Swaps are also limited to the maximum swap amount advertised by the server. If
+you would like to reduce the size of swap that autoloop created, this value can 
+also be configured. 
+
+```
+loop setparams --maxamt={amount in satoshis}
+```
+
+The server's current terms are provided by the `terms` cli command. Note that
```
guggero, line 217: nit: specify `loop terms` instead of just `terms`.

guggero (approved):
Nice and clean set of changes, LGTM :100:

liquidity/liquidity.go
```diff
@@ -347,6 +368,47 @@ func (p Parameters) validate(minConfs int32) error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateRestrictions checks that client restrictions fall within the server's
+// restrictions.
+func validateRestrictions(server, client *Restrictions) error {
```
bhandras, line 381: nit: we could also return the clipped boundaries in one go here.

bhandras (approved):
LGTM, sorry for the late review. Nice change! :1st_place_medal:

PR #331, opened 2021-02-02
autoloop: refactoring to prepare for loopin

carlaKC:
This PR contains some preparatory changes to generalize our autoloop code to add loop in functionality. Mostly renames, so pulled it out to simplify review.

#### Pull Request Checklist


cmd/loop/liquidity.go
```diff
@@ -338,18 +338,18 @@ func setParams(ctx *cli.Context) error {
 		flagSet = true
 	}
 
-	if ctx.IsSet("autoout") {
-		params.AutoLoopOut = ctx.Bool("autoout")
+	if ctx.IsSet("autoloop") {
```
guggero, line 341: nit: This parameter is referenced in `autoloop.md`, line 9.

looprpc/client.proto
```diff
@@ -783,25 +783,25 @@ message LiquidityParameters {
...
     /*
-    Set to true to enable automatic dispatch of loop out swaps. All swaps will
-    be limited to the fee categories set by these parameters, and total
-    expenditure will be limited to the auto out budget.
+    Set to true to enable automatic dispatch of swaps. All swaps will be limited 
+    to the fee categories set by these parameters, and total expenditure will 
+    be limited to the autoloop budget.
     */
-    bool auto_loop_out = 10;
+    bool autoloop = 10;
```
guggero, line 790: This will possibly break REST users if they referenced the parameter by name. But I guess not many people are using it yet anyway and we declared it as experimental so this is OK?

labels/labels.go
```diff
@@ -28,10 +32,14 @@ var (
 	ErrReservedPrefix = errors.New("label contains reserved prefix")
 )
 
-// AutoOutLabel returns a label with the reserved prefix that identifies
-// automatically dispatched loop outs.
-func AutoOutLabel() string {
-	return fmt.Sprintf("%v: %v", Reserved, autoOut)
+// AutoloopLabel returns a label with the reserved prefix that identifies
+// automatically dispatched swaps depending on the type of swap being executed.
+func AutoloopLabel(loopOut bool) string {
```
guggero, line 37: nit: use swap type instead of `bool`?

guggero (approved):
LGTM, nice refactor :tada:

looprpc/client.proto
bhandras: It is a breaking change, but since it's a rename only, I'd say it's less risky to do.

loopd/utils.go
```diff
@@ -38,16 +39,28 @@ func getLiquidityManager(client *loop.Client) *liquidity.Manager {
...
+			if swapType == swap.TypeOut {
+				outTerms, err := client.Server.GetLoopOutTerms(ctx)
+				if err != nil {
+					return nil, err
+				}
+
+				return liquidity.NewRestrictions(
+					outTerms.MinSwapAmount, outTerms.MaxSwapAmount,
+				), nil
+
```
bhandras, line 54: nit: extra nl

bhandras (approved):
LGTM :taco:

looprpc/client.proto
carlaKC: > I guess not many people are using it yet anyway and we declared it as experimental so this is OK?

This was my reasoning, and I don't think we will break anybody's scripts/code so I think it's worth the rename. 
Can also deprecate and add new fields if we're uncomfortable with it. 

@alexbosworth WDYT?

cmd/loop/liquidity.go
carlaKC: fixed!

loopd/utils.go
carlaKC: done!

looprpc/client.proto
carlaKC: Discussion in loop notes, ok with breaking change for experimental/low use API.

PR #332, opened 2021-02-02
autoloop: add reasons to explain no action

carlaKC:
One UX issue that people have run into using autoloop is that they don't know why no swaps are being executed, and need to go digging in debug logs to get an explanation. This PR surfaces a variety of reasons for not executing a swap to help better explain our logic, and updates our docs to provide recommendations for updating config to address these reasons.

This PR also relaxes our requirement that no "unrestricted" swaps are in flight, which was causing problems in practice where autoloop can't coexist with manually initiated swaps.

#### Pull Request Checklist


liquidity/liquidity.go
```diff
@@ -928,17 +935,13 @@ func (m *Manager) checkExistingAutoLoops(ctx context.Context,
...
 
-// getEligibleChannels takes lists of our existing loop out and in swaps, and
-// gets a list of channels that are not currently being utilized for a swap.
-func (m *Manager) getEligibleChannels(ctx context.Context,
-	loopOut []*loopdb.LoopOut, loopIn []*loopdb.LoopIn) (
-	[]lndclient.ChannelInfo, error) {
+// ineligibleChannels examines our existing swaps and returns a summary of the
+// current activity which can be used to determine whether we should perform
+// any swaps.
+func (m *Manager) ineligibleChannels(loopOut []*loopdb.LoopOut,
```
bhandras, line 1006: nit: imo it's a bit confusing to call this fn `ineligibleChannels`. How about simply calling it `currentSwapTraffic` since that's what it does?

bhandras (approved):
Very nice improvement! LGTM :1st_place_medal:

looprpc/client.proto
```diff
@@ -872,9 +872,102 @@ message SetLiquidityParamsResponse {
 message SuggestSwapsRequest {
 }
 
+enum AutoReason{
```
guggero, line 875: nit: formatting.

guggero (approved):
tACK, LGTM :100: 

This is extremely useful! Both the returned reason in the RPC as well as the log messages.

liquidity/liquidity.go
carlaKC: done!

PR #334, opened 2021-02-05
Support single macaroon for lnd

### LINKED ISSUE #299: Don't require all macaroons to use loopd
Currently users are required to specify a directory for LND's macaroons with --lnd.macaroondir. This can be problematic or annoying if you aren't running loopd on the same server that's running LND. I would suggest the ability to use a single macaroon (like admin.macaroon) to make it easy to use. Eventually once the macaroon baking changes land in LND we could even bake a macaroon specific for loop.

guggero:
Fixes #299 by allowing only one macaroon to be specified in the
--lnd.macaroonpath config option/command line flag.

#### Pull Request Checklist


DOCKER.md
```diff
@@ -28,7 +28,7 @@ The docker image contains:
 Docker is very flexible so you can use that information however you choose. This guide isn't meant to be prescriptive.
 
 
-### Example: Running loopd
```
carlaKC, line 31: 🤔

carlaKC (approved):
niceeee 🌈

DOCKER.md
guggero: It was a weird non-breaking space that probably got in there through copy/paste. My IDE showed it so I removed it.

loopd/config.go
```diff
@@ -235,6 +251,30 @@ func Validate(cfg *Config) error {
 		return err
 	}
 
+	// Make sure only one of the macaroon options is used.
+	switch {
+	case cfg.Lnd.MacaroonPath != "" && cfg.Lnd.MacaroonDir != "":
+		return fmt.Errorf("use --lnd.macaroonpath only")
```
bhandras, line 257: nit: Shouldn't this be either `macaroondir` or `macaroonpath`?

bhandras (approved):
LGTM

loopd/config.go
guggero: Yes. But since the `macaroondir` is deprecated, this error is a bit clearer what the action should be. And it's the exact same commit that I added to Pool, so at least the error messages will be the same :)

PR #340, opened 2021-02-19
liquidity: flat fee percentage for autoloop

carlaKC:
Need to add some unit tests.

#### Pull Request Checklist


liquidity/fees.go
```diff
@@ -0,0 +1,202 @@
...
+// FeeCategoryLimit is an implementation of the fee limit interface which sets
+// a specific fee limit per fee category.
+type FeeCategoryLimit struct {
+	// MaximumPrepay is the maximum prepay amount we are willing to pay per
+	// swap.
+	MaximumPrepay btcutil.Amount
+
+	// MaximumSwapFeePPM is the maximum server fee we are willing to pay per
+	// swap expressed as parts per million of the swap volume.
+	MaximumSwapFeePPM int
```
guggero, line 90: Define as `uint64` to make fewer conversions necessary? Same with the constants above?

liquidity/fees.go
```diff
@@ -200,3 +204,127 @@ func (f *FeeCategoryLimit) loopOutFees(amount btcutil.Amount,
...
+
+	prepay, route, miner := f.loopOutFees(swapAmt, quote)
+
+	// Calculate the worst case fees that we could pay for this swap,
+	// ensuring that we are within our fee limit even if the swap fails.
+	fees := worstCaseOutFees(
+		prepay, route, quote.SwapFee, miner, quote.PrepayAmount,
+	)
+
+	if feeLimit < fees {
```
guggero, line 296: nit: turn condition around to align with all above? I mean `fees >= feeLimit`.

liquidity/fees.go
```diff
@@ -200,3 +204,127 @@ func (f *FeeCategoryLimit) loopOutFees(amount btcutil.Amount,
...
+	if quote.PrepayAmount > feeLimit {
+		log.Debugf("prepay amount: %v greater than fee limit: %v, at "+
+			"%v ppm", quote.PrepayAmount, feeLimit, f.PartsPerMillion)
+
+		return newReasonError(ReasonPrepay)
+	}
+
+	// If our miner and swap fee equal our limit, we will have nothing left
+	// for off-chain fees, so we fail out early.
+	if quote.MinerFee+quote.SwapFee >= feeLimit {
```
guggero, line 280: Great approach of doing the checks incrementally, this should make it very clear to the user why a swap went through or not :clap:

liquidity/liquidity_test.go
```diff
@@ -1269,6 +1282,143 @@ func TestSizeRestrictions(t *testing.T) {
...
+			suggestions: &Suggestions{
+				OutSwaps: []loop.OutRequest{
+					rec,
+				},
+				DisqualifiedChans: noneDisqualified,
+				DisqualifiedPeers: noPeersDisqualified,
+			},
+		},
+		{
+			name:   "swap fee to high",
```
guggero, line 1371: nit: typo, and one below as well

loopd/swapclient_server.go
```diff
@@ -595,6 +590,19 @@ func (s *swapClientServer) GetLiquidityParams(_ context.Context,
 		MaxSwapAmount: uint64(cfg.ClientRestrictions.Maximum),
 	}
 
+	feeCategories, ok := cfg.FeeLimit.(*liquidity.FeeCategoryLimit)
```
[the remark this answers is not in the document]
guggero: Ah, I see why you did it that way. Never mind then :)

guggero (commented):
tACK, this is a great simplification for the user!

The only thing I ran into when testing this on regtest is that the check of the `m.params.SweepFeeRateLimit` complained when I only set the `--autoloop` param and left everything else at the defaults. Not sure if it would make sense to bypass that check if we're using the proportional fees? Since we look at the miner fees there too.

liquidity/fees.go
```diff
@@ -200,3 +204,127 @@ func (f *FeeCategoryLimit) loopOutFees(amount btcutil.Amount,
...
+	// off-chain fees.
+	feeLimit := ppmToSat(amount, f.PartsPerMillion)
+	available := feeLimit - quote.MinerFee - quote.SwapFee
+
+	prepayMaxFee, routeMaxFee := splitOffChain(
+		available, quote.PrepayAmount, amount,
+	)
+
+	return prepayMaxFee, routeMaxFee, quote.MinerFee
+
```
bhandras, line 326: nit: extra nl

loopd/swapclient_server.go
```diff
@@ -734,6 +735,39 @@ func (s *swapClientServer) SetLiquidityParams(ctx context.Context,
...
+
+	case isCategories:
+		return liquidity.NewFeeCategoryLimit(
+			int(req.MaxSwapFeePpm),
+			int(req.MaxRoutingFeePpm),
+			int(req.MaxPrepayRoutingFeePpm),
+			btcutil.Amount(req.MaxMinerFeeSat),
+			btcutil.Amount(req.MaxPrepaySat),
+		), nil
+	default:
```
bhandras, line 766: nit: nl above

liquidity/fees.go
bhandras: nit: just turn the order (`fees` > `feeLimit`) I think equality is fine.

liquidity/fees.go
```diff
@@ -200,3 +204,127 @@ func (f *FeeCategoryLimit) loopOutFees(amount btcutil.Amount,
...
+	if quote.MinerFee > feeLimit {
+		log.Debugf("miner fee: %v greater than fee limit: %v, at "+
+			"%v ppm", quote.MinerFee, feeLimit, f.PartsPerMillion)
+
+		return newReasonError(ReasonMinerFee)
+	}
+
+	if quote.SwapFee > feeLimit {
+		log.Debugf("swap fee: %v greater than fee limit: %v, at "+
```
bhandras, line 265: nit: Why not `Infof`?

bhandras (commented):
Super nice addition, only a few nits and ready!

liquidity/fees.go
carlaKC: This was debated in a previous PR and debug was decided on. I wanted info at the time, but now that we have reasons surfaced to the user, I think it's ok to leave "additional info" on a lower debug level.

liquidity/fees.go
```diff
@@ -0,0 +1,202 @@
...
+		"miner fee: %v, maximum swap fee ppm: %v, maximum "+
+		"routing fee ppm: %v, maximum prepay routing fee ppm: %v",
+		f.MaximumPrepay, f.MaximumMinerFee, f.MaximumSwapFeePPM,
+		f.MaximumRoutingFeePPM, f.MaximumPrepayRoutingFeePPM,
+	)
+}
+
+func (f *FeeCategoryLimit) validate() error {
+	// Check that we have non-zero fee limits.
+	if f.MaximumSwapFeePPM == 0 {
```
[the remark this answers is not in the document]
carlaKC: Switching over to unsigned!

liquidity/fees.go
```diff
@@ -200,3 +204,127 @@ func (f *FeeCategoryLimit) loopOutFees(amount btcutil.Amount,
 
 	return prepayMaxFee, routeMaxFee, f.MaximumMinerFee
 }
+
+// FeePortion is a fee limitation which limits fees to a set portion of
+// the swap amount.
+type FeePortion struct {
```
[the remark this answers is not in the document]
carlaKC: Just personal style preference,  I think that `f.PartsPerMillion` is more readable than `*f` when we're using the value in various funcs.

bhandras (approved):
LGTM :1st_place_medal:

guggero (approved):
LGTM, very nice :rocket: 

If you have any spare cycles left, maybe it would be nice to add a simple test case to the itests after this is merged? Something like: set low PPM, try small loop out, expect failure, try larger loop, expect success.

carlaKC:
> If you have any spare cycles left, maybe it would be nice to add a simple test case to the itests after this is merged? 

Yeah planning on updating the itest with the new client version anyway, so will add to the existing test!

carlaKC:
Tried this out on testnet and realized that we need to pad max miner fee so that we'll continue with swaps even if fees spike after initiating. Relatively small [diff](https://github.com/lightninglabs/loop/compare/9d845dda86f99837e79f99eb246cb5b7b15ff84c..34726a4130efc6e73c10a6dbf7478531ee1c3823), but just wanted to make reviewers aware @guggero @bhandras. This should be addressed when we figure out a better way of dealing with the max miner issue, current approach is just to do as we do for regular swaps.

guggero (approved):
Re-ACK, looks good to me. Even if it might be a bit harder to explain to the user why the percentage needs to be high-ish, it's better than forfeiting swaps because of the miner fee

PR #347, opened 2021-03-03
loopin: store loop in on-chain fees

### LINKED ISSUE #189: Implement loop in on-chain fee reporting
Currently the miner fee that was spent for the on-chain htlc isn't tracked. This is also visible in the client log when a swap is completed:

`Loop in swap completed: Success (final cost: server 0.00002505 BTC, onchain 0 BTC, offchain 0 BTC)`

`onchain` should never be zero.

carlaKC:
This PR updates the loop in swap execution logic to store our on-chain fees. 

To get our on-chain htlc's fee, we calculate our fee as lnd does using our chosen fee rate. The downside here is that we _could_ change fee estimation in lnd, but that seems unlikely at this stage. The alternative would be to move tx crafting into loop, which I decided against because we'd need to think about coin selection, change etc. While we may want to do this in the future, this PR is intended to unblock autoloop-in, so we take the easier route. 

One thing I noticed is that loop in records its server fee as a negative value, whereas loop out records fees as positive value. To me it's more intuitive to use a positive value. Open question as to whether we want to change the server fee to be a positive value? I worry it may break something user-side. 

Fixes #189

#### Pull Request Checklist


loopin.go
```diff
@@ -817,6 +839,9 @@ func (s *loopInSwap) processHtlcSpend(ctx context.Context,
 		// server cost balance.
 		s.cost.Server += htlcValue
 	} else {
+		// We needed another on chain tx to sweep the timeout clause,
+		// which we now include in our costs.
+		s.cost.Onchain += sweepFee
```
yyforyongyu, line 844: Is it possible that we move this line into `publishTimeoutTx`? That way it could save some changes made.

loopin.go
```diff
@@ -882,7 +908,7 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
 		s.log.Warnf("publish timeout: %v", err)
 	}
 
-	return nil
+	return fee, nil
```
yyforyongyu, line 911: Seems like we could update the cost here.

yyforyongyu (commented):
Still wrapping my head around the code. A few questions/comments below,


> downside here is that we could change fee estimation in lnd, but that seems unlikely at this stage. 

To me it seems better we have one place to unify the fee estimation logic. Why is this a downside? Would that break lots things here?


> One thing I noticed is that loop in records its server fee as a negative value, whereas loop out records fees as positive value. 

Yeah I think it's weird to have negative values here.

loopin.go
```diff
@@ -681,6 +691,16 @@ func (s *loopInSwap) publishOnChainHtlc(ctx context.Context) (bool, error) {
 
 }
 
+// getTxFee calculates our fee for a transaction that we have broadcast. We use
+// sat per kvbyte because this is what lnd uses, and we will run into rounding
+// issues if we do not use the same fee rate as lnd.
+func getTxFee(tx *wire.MsgTx, fee chainfee.SatPerKVByte) btcutil.Amount {
+	btcTx := btcutil.NewTx(tx)
+	vsize := mempool.GetTxVirtualSize(btcTx)
```
guggero, line 699: I know we can't get this any more accurately if we can't look at the spent UTXOs (since we don't have a chain backend to query). This might be slightly off from the actual fee though since the fee rate is not an exact outcome in the coin selection. For our use case that might be okay though.

guggero: Would be cool if we could start using the PSBT interfaces in the future so we could actually look at the UTXO information for stuff like this.

guggero (approved):
LGTM, nice fix :tada:

carlaKC:
> To me it seems better we have one place to unify the fee estimation logic. 

Yeah, definitely better to get the fee from LND rather than trying to calculate it after the fact like we do here. We'd need to add coin selection/reservation to loop to do this (using the PSBT api like @guggero recommends), which is a much bigger change, or update the api to return fees which will only happen in the next release. I think this approach will be _fine_ in the meantime, but certainly something we want to get to.

loopin.go
carlaKC: We call `publishTimeoutTx` every block, so we'd end up counting our fee on every republish. I thought just passing it out would be simpler than trying to add some once-only logic for storing fee, since we'd need to account for restarts as well.

yyforyongyu:
> > To me it seems better we have one place to unify the fee estimation logic.
> 
> Yeah, definitely better to get the fee from LND rather than trying to calculate it after the fact like we do here. We'd need to add coin selection/reservation to loop to do this (using the PSBT api like @guggero recommends), which is a much bigger change, or update the api to return fees which will only happen in the next release. I think this approach will be _fine_ in the meantime, but certainly something we want to get to.

Cool thanks for the clarification!

yyforyongyu (approved):
LGTM!

PR #356, opened 2021-03-10
[draft] loop-in: await probe from the server on loop-in quote

bhandras:
#### Pull Request Checklist


swap_server_client.go
```diff
@@ -202,16 +209,62 @@ func (s *grpcSwapServerClient) GetLoopInQuote(ctx context.Context,
...
+		// Generate random preimage.
+		var preimage lntypes.Preimage
+		if _, err := rand.Read(preimage[:]); err != nil {
+			log.Error("Cannot generate preimage")
+		}
+
+		probeHash := lntypes.Hash(sha256.Sum256(preimage[:]))
+
+		log.Infof("Creating quote probe invoice %v", probeHash)
+		probeInvoice, err := s.lnd.Invoices.AddHoldInvoice(
```
carlaKC, line 229: why hold invoice? Also is 3600 expiry enough? I think the server tries for longer than that, or does the probe still work even if the invoice is expired.

loopdb/protocol_version.go
```diff
@@ -39,13 +39,15 @@ const (
 	// invoice so that the server can perform a multi-path probe.
 	ProtocolVersionMultiLoopIn ProtocolVersion = 6
 
+	ProtocolVersionMultiLoopInV2 ProtocolVersion = 7
```
carlaKC, line 42: `ProtocolVersionLoopInProbe`?

swap_server_client.go
```diff
@@ -202,16 +209,62 @@ func (s *grpcSwapServerClient) GetLoopInQuote(ctx context.Context,
...
+	}
+
+	quoteResp, err := s.server.LoopInQuote(rpcCtx, quoteReq)
 	if err != nil {
 		return nil, err
 	}
 
+	if loopdb.CurrentRPCProtocolVersion == looprpc.ProtocolVersion_MULTI_LOOP_IN_V2 {
+		// Create a cancellable context that is used for monitoring the probe.
+		probeWaitCtx, probeWaitCancel := context.WithCancel(ctx)
```
carlaKC, line 251: Don't we need to create this with a timeout so that it'll cancel after some amount of time?

PR #357, opened 2021-03-12
loopd: verify loop out amount

### LINKED ISSUE #104: check amount is valid for loop out
if I requesting an amount above my outbound capacity, loop out will still proceed and fail the payment will "failed: unable to find a path to destination"

loop should check that the amount requested is equal or below my spending balance.

ellemouton:
This commit adds validation that checks if the loop out amount specified can
be satisfied given the nodes current channel balances. 

The total capacity is calculated as either the sum of the local balances of all active channels if no outgoing channel set is specified. Or if a set is specified then the total capacity is the sum of the local balances active channels in the specified set.

Fixes #104

### DIFF
--- a/config.go
+++ b/config.go
@@ -15,4 +15,5 @@ type clientConfig struct {
 	Store             loopdb.SwapStore
 	LsatStore         lsat.Store
 	CreateExpiryTimer func(expiry time.Duration) <-chan time.Time
+	LoopOutMaxParts   uint32
 }
--- a/loopd/swapclient_server.go
+++ b/loopd/swapclient_server.go
@@ -44,6 +44,13 @@ var (
 	// errConfTargetTooLow is returned when the chosen confirmation target
 	// is below the allowed minimum.
 	errConfTargetTooLow = errors.New("confirmation target too low")
+
+	// errBalanceTooLow is returned when the loop out amount can't be
+	// satisfied given total balance of the selection of channels to loop
+	// out on.
+	errBalanceTooLow = errors.New(
+		"channel balance too low for loop out amount",
+	)
 )
 
 // swapClientServer implements the grpc service exposed by loopd.
@@ -89,7 +96,8 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 	}
 
 	sweepConfTarget, err := validateLoopOutRequest(
-		s.lnd.ChainParams, in.SweepConfTarget, sweepAddr, in.Label,
+		ctx, s.lnd.Client, s.lnd.ChainParams, in, sweepAddr,
+		s.impl.LoopOutMaxParts,
 	)
 	if err != nil {
 		return nil, err
@@ -981,9 +989,12 @@ func validateLoopInRequest(htlcConfTarget int32, external bool) (int32, error) {
 }
 
 // validateLoopOutRequest validates the confirmation target, destination
-// address and label of the loop out request.
-func validateLoopOutRequest(chainParams *chaincfg.Params, confTarget int32,
-	sweepAddr btcutil.Address, label string) (int32, error) {
+// address and label of the loop out request. It also checks that the requested
+// loop amount is valid given the available balance.
+func validateLoopOutRequest(ctx context.Context, lnd lndclient.LightningClient,
+	chainParams *chaincfg.Params, req *looprpc.LoopOutRequest,
+	sweepAddr btcutil.Address, maxParts uint32) (int32, error) {
+
 	// Check that the provided destination address has the correct format
 	// for the active network.
 	if !sweepAddr.IsForNet(chainParams) {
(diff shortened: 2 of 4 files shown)


carlaKC (commented):
Thanks for another well tested improvement to our validation @ellemouton 🙌 Just some style nits from me that I think will improve readability a bit.

Tested for unrestricted swaps and a specific chanset and goods are as advertised! 

```
channel balance too low for loop out amount: Requested swap amount of 100000 sats along with the maximum routing fee of 2010 sats is more than the available outgoing capacity of 10950 sats
```

ellemouton:
Thanks for the review @carlaKC 🚀  updated 👍

loopd/swapclient_server.go
```diff
@@ -992,9 +1002,50 @@ func validateLoopOutRequest(chainParams *chaincfg.Params, confTarget int32,
...
+		if unlimitedChannels || outgoingChanSetMap[c.ChannelID] {
+			availabeBalance += c.LocalBalance
+		}
+	}
+
+	// Determine if the total outgoing capacity of the channel set is
+	// enough to satisfy the loop amount and the maximum possible
+	// routing fees that could be charged.
+	if req.Amt+req.MaxSwapRoutingFee >
+		int64(availabeBalance.ToUnit(btcutil.AmountSatoshi)) {
```
carlaKC, line 1040: Could still be a bit simpler: 
```
requiredBalance := btcutil.Amount(req.Amt + req.MaxSwapRoutingFee)
if requiredBalance > availabeBalance{
  ...
}
```

carlaKC (approved):
Looks good!

loopd/swapclient_server.go
ellemouton: oops sorry. I did update that but must have undone it by accident. Will update asap

loopd/swapclient_server.go
```diff
@@ -992,9 +1002,50 @@ func validateLoopOutRequest(chainParams *chaincfg.Params, confTarget int32,
...
+		if unlimitedChannels || outgoingChanSetMap[c.ChannelID] {
+			availabeBalance += c.LocalBalance
+		}
+	}
+
+	// Determine if the total outgoing capacity of the channel set is
+	// enough to satisfy the loop amount and the maximum possible
+	// routing fees that could be charged.
+	requiredBalance := btcutil.Amount(req.Amt + req.MaxSwapRoutingFee)
+	if requiredBalance > availabeBalance {
```
bhandras, line 1040: Note that the balance check is slightly more complicated than this. We do MPP payments when looping out and MPP (in LND) splits equally. So really we should check splitting up to the number of channels and then see if any split satisfies the required balance condition. This will also work if only one channel has enough balance for the payment or if we split maximally.

bhandras (changes requested):
Great work @ellemouton. My only comment is that due to how MPP works balance check is a bit more complicated (although very easy to fix).

loopd/swapclient_server.go
bhandras: Also when implementing, imo it's a good idea to separate this into it's own function and then it's easier to drop once MPP splits more intelligently.

ellemouton: ok cool! Didn't know this at all. Will update 👍

config.go
```diff
@@ -15,4 +15,5 @@ type clientConfig struct {
 	Store             loopdb.SwapStore
 	LsatStore         lsat.Store
 	CreateExpiryTimer func(expiry time.Duration) <-chan time.Time
+	LoopOutMaxParts   uint32
```
ellemouton, line 18: im not sure if this is the right place to put this. Let me know if there is a better way to give the swapclient_server access to this value

bhandras (changes requested):
Almost there, splitting algo needs a little correction.

loopd/swapclient_server.go
```diff
@@ -992,9 +1003,73 @@ func validateLoopOutRequest(chainParams *chaincfg.Params, confTarget int32,
...
+func isRoutable(channels []lndclient.ChannelInfo, amt btcutil.Amount,
+	maxParts uint32) bool {
+
+	var parts uint32
+	for parts = 1; parts <= maxParts; parts++ {
+		minHTLC := amt.MulF64(1 / float64(parts))
+
+		var totalRoutableParts uint32
+		for _, c := range channels {
+			routableParts := (c.LocalBalance / minHTLC).ToUnit(btcutil.AmountSatoshi)
```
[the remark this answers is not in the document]
ellemouton: Thanks so much for taking the time to create the example! Very very helpful. Yeah I didn't quite grasp the splitting in half logic before but I understand now 👍  
(The only thing is, I think line 38 should be `split /= 2` right? Otherwise an infinite loop can occur)

ellemouton:
Cool, I updated the algorithm @bhandras. Thanks again for clarifying 😊   The only thing I left out was the minAmt per shard because this amount can only be retrieved from LND using `DefaultShardMinAmt` and even then it could be incorrect cause the user may overwrite the default value. So im wondering if it is ok to assume that minLoopAmt/maxParts will pretty much never be < shardMinAmt since this is a best effort check?

loopd/swapclient_server.go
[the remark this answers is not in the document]
bhandras: Yes indeed, that was a mistake, I didn't test it well enough :)

bhandras:
> Cool, I updated the algorithm @bhandras. Thanks again for clarifying blush The only thing I left out was the minAmt per shard because this amount can only be retrieved from LND using `DefaultShardMinAmt` and even then it could be incorrect cause the user may overwrite the default value. So im wondering if it is ok to assume that minLoopAmt/maxParts will pretty much never be < shardMinAmt since this is a best effort check?

Yes it's always best effort since LND does actual path finding in the splitting algo (factoring in fees too), so there's no point in being super strict. More or less having an idea that a split is possibly valid is good enough imo.

bhandras (approved):
LGTM, thanks @ellemouton :1st_place_medal:

carlaKC:
Very nice, tACK @ 65fe06c!

PR #361, opened 2021-03-23
loopout: use max miner fee when deadline is close

yyforyongyu:
We can use the max miner fee to sweep when the deadline is close.

### DIFF
--- a/loopout.go
+++ b/loopout.go
@@ -922,12 +922,20 @@ func (s *loopOutSwap) sweep(ctx context.Context,
 		s.log.Warnf("Required fee %v exceeds max miner fee of %v",
 			fee, s.MaxMinerFee)
 
-		if s.state == loopdb.StatePreimageRevealed {
+		switch {
+		case s.state == loopdb.StatePreimageRevealed:
 			// The currently required fee exceeds the max, but we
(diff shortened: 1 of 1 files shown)


alexbosworth:
If a user specifies a low max miner fee it isn't safe from a non-trust perspective

yyforyongyu:
> If a user specifies a low max miner fee it isn't safe from a non-trust perspective

Not safe as in, when the sweep is not confirmed because the fee paid is insufficient, the `preimage` revealed in the transaction might be used by a non-trust server to collect the off-chain swap payment, then later timeout the on-chain HTLC?

alexbosworth:
> > If a user specifies a low max miner fee it isn't safe from a non-trust perspective
> 
> Not safe as in, when the sweep is not confirmed because the fee paid is insufficient, the `preimage` revealed in the transaction might be used by a non-trust server to collect the off-chain swap payment, then later timeout the on-chain HTLC?

not safe as in, they reveal the preimage, the server collects the off-chain component of the swap, the htlc on-chain times out, the server collects the on-chain component of the swap, client loses both sides of the swap

yyforyongyu:
Close as it's just a lame demo.

PR #366, opened 2021-04-07
loopout: do not publish preimage after on-chain HTLC has timed out

yyforyongyu:


loopout.go
```diff
@@ -844,6 +850,20 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
 		// timer.
 		case notification := <-s.blockEpochChan:
 			s.height = notification.(int32)
+
+			// Check if the on-chain HTLC's CLTV has been reached.
+			// If so, we will time out the swap because it's no
+			// longer safe for us to reveal the preimage.
+			if s.height >= s.CltvExpiry {
```
alexbosworth, line 857: Not that this situation is likely to happen due to timing but how about some buffer here? Like expiry height minus the final confirmation target so that you still have some time to sweep without the timeout path being active?

alexbosworth (changes requested):

loopout.go
yyforyongyu: Something like [this](https://github.com/lightninglabs/loop/blob/master/loopout.go#L647)?

alexbosworth: Yeah exactly

alexbosworth (approved):

loopout.go
```diff
@@ -844,6 +850,23 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
 		case notification := <-s.blockEpochChan:
 			s.height = notification.(int32)
+
+			maxPreimageRevealHeight := s.CltvExpiry -
+				MinLoopOutPreimageRevealDelta
+
+			// Check if the on-chain HTLC's CLTV has been reached.
+			// If so, we will time out the swap because it's no
+			// longer safe for us to reveal the preimage.
+			if s.height > maxPreimageRevealHeight {
```
carlaKC, line 860: I think you also need to check whether we've revealed the preimage here? 

If we just give up when we reach our expiry height, we could land ourselves in a situation where we've revealed the preimage by broadcasting our sweep, but we haven't got our funds on-chain yet. The server _could_ try to claim off-chain and timeout on-chain, it won't, but we want to be trustless so we have to keep going for the on-chain sweep.

The real solution is to ensure that we sweep on-chain before the expiry at all costs - insert deadline aware sweeper here :D - but in the meantime we need to keep going for the sweep if we've revealed preimage.

carlaKC (commented):
Great addition! Just need to add some state awareness so that we don't give up on swaps that have already revealed their preimage.

carlaKC:
Replaced by #372

PR #371, opened 2021-04-20
multi: bump minimum golang version to 1.15

ellemouton:
Bump the minimum golang version to 1.15 in the go.mod file. The go
version is also updated in the Dockerfile and travis.yml file. This is
required so that the correct `errors.Is` method is used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ git:
   depth: false
 
 go:
-  - "1.13.x"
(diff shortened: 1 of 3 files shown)


ellemouton:
fyi @carlaKC

carlaKC (approved):
Thanks @ellemouton!

bhandras (approved):

PR #372, opened 2021-04-21
loopout: do not reveal preimage too close to expiry

carlaKC:
If fees are too high for the client to reveal the preimage, and then suddenly drop as we reach our expiry, it is possible that the client reveals the preimage when we are undesirably close to the swap's expiry. This PR introduces a check that will abandon the swap if we are too close to expiry. This fee scenario is highly unlikely, but nonetheless we account for it just in case. It's important that we keep trying to sweep if the preimage has already been revealed (because the server can pull off-chain), so we only abandon swaps like this if we have not revealed the preimage yet. 

This PR introduces a check to ensure that we don't reveal our preimage too late, and adds a state that will indicate that the client chose not to reveal its preimage due to high fees. This will also cover the case where fees are persistently high after swap creation, and provide some visibility into this type of swap failure for clients.

Replaces #366, h/t @yyforyongyu for picking this one up!

### DIFF
--- a/loopout.go
+++ b/loopout.go
@@ -438,6 +438,12 @@ func (s *loopOutSwap) executeSwap(globalCtx context.Context) error {
 		return err
 	}
 
+	// If spend details are nil, we resolved the swap without waiting for
+	// its spend, so we can exit.
+	if spendDetails == nil {
+		return nil
+	}
+
 	// Inspect witness stack to see if it is a success transaction. We
 	// don't just try to match with the hash of our sweep tx, because it
 	// may be swept by a different (fee) sweep tx from a previous run.
@@ -854,6 +860,14 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
 				return nil, err
 			}
 
+			// If the result of our spend func was that the swap
+			// has reached a final state, then we return nil spend
+			// details, because there is no further action required
+			// for this swap.
+			if s.state.Type() != loopdb.StateTypePending {
+				return nil, nil
+			}
+
 			// If our off chain payment is not yet complete, we
 			// try to push our preimage to the server.
 			if !paymentComplete {
@@ -889,7 +903,9 @@ func (s *loopOutSwap) pushPreimage(ctx context.Context) {
 
 // sweep tries to sweep the given htlc to a destination address. It takes into
 // account the max miner fee and marks the preimage as revealed when it
-// published the tx.
+// published the tx. If the preimage has not yet been revealed, and the time
+// during which we can safely reveal it has passed, the swap will be marked
+// as failed, and the function will return.
 //
 // TODO: Use lnd sweeper?
 func (s *loopOutSwap) sweep(ctx context.Context,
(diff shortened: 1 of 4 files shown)


alexbosworth (approved):

loopout_test.go
```diff
@@ -683,17 +683,17 @@ func TestExpiryBeforeReveal(t *testing.T) {
...
-	// Tick our expiry chan again, this time we expect the swap to
-	// publish our sweep timeout, despite expiry having passed, and the
-	// potential for a race with the server.
+	// Tick our expiry channel again to trigger another sweep attempt.
 	expiryChan <- testTime
 
-	// Expect a signing request for the HTLC success transaction.
-	<-ctx.Lnd.SignOutputRawChannel
+	// We should see our swap marked as failed.
+	cfg.store.(*storeMock).assertLoopOutState(loopdb.StateFailPreimageNotRevealed)
```
yyforyongyu, line 692: nit: line too long

looprpc/client.proto
```diff
@@ -524,6 +524,13 @@ enum FailureReason {
     because the amount extended by an external loop in htlc is insufficient.
     */
     FAILURE_REASON_INCORRECT_AMOUNT = 6;
+
+    /* 
+    FAILURE_REASON_PREIMAGE_NOT_REVEALED indicates that a loop out swap 
+    intentionally did not reveal the preimage on-chain because sweep fees
+    were too high.
```
yyforyongyu, line 531: nit: because deadline is close...

yyforyongyu (approved):
LGTM! Very nice godocs! Just a few nits.

loopout.go
```diff
@@ -900,29 +914,47 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
 	if err != nil {
 		return err
 	}
 
+	// If we have not revealed our preimage, and we don't have time left
+	// to sweep the swap, we abandon the swap because we can no longer
+	// sweep the success path (without potentially having to compete with
+	// the server's timeout sweep), and we have not had any coins pulled
+	// off-chain.
+	if blocksToLastReveal <= 0 && !preimageRevealed {
```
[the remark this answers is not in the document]
carlaKC: Ah should have added an explanation here! In a follow up PR which adds reporting swap cancellation to the server, I want to report the miner fee and estimate to the server, so I put this below so that I don't need to move this in the next PR. 

Get fee calls aren't too expensive, and we'll only have one extra call before we exit for good if this condition is hit, so I thought it was ok?

loopout.go
```diff
@@ -900,29 +914,47 @@ func (s *loopOutSwap) sweep(ctx context.Context,
 		return s.htlc.GenSuccessWitness(sig, s.Preimage)
 	}
 
+	var (
+		remainingBlocks    = s.CltvExpiry - s.height
+		blocksToLastReveal = remainingBlocks - MinLoopOutPreimageRevealDelta
+		preimageRevealed   = s.state == loopdb.StatePreimageRevealed
```
guggero, line 920: micro nit: comes down to personal taste but for variables that have a value assigned, I prefer the `:=` syntax. Would also fix the slightly too long line.

guggero (commented):
Very nice fix!
Just one question about persisting the state, otherwise LGTM :tada:

loopout.go
```diff
@@ -900,29 +914,47 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
 	}
 
+	// If we have not revealed our preimage, and we don't have time left
+	// to sweep the swap, we abandon the swap because we can no longer
+	// sweep the success path (without potentially having to compete with
+	// the server's timeout sweep), and we have not had any coins pulled
+	// off-chain.
+	if blocksToLastReveal <= 0 && !preimageRevealed {
+		s.state = loopdb.StateFailPreimageNotRevealed
+		return nil
```
guggero, line 949: Do we need to call `s.persistState(ctx)` here?

carlaKC: We do! I thought that we were ok because we persist in `executeAndFinalize` but if we restart after this state, we'd set `FailedTimeout` when we resume the swap since it's still `Initiated`. I think that it's easiest to just save here than check for the htlc on restart to differentiate between timeout and not revealed.

carlaKC:
Updated this PR to remove `StateFailPreimageNotRevealed`. I think it's a nice state to have, but there's all sorts of complications when we restart because we can't distinguish between: 
- Server broadcast htlc, but we chose not to sweep due to fees
- Server htlc didn't confirm in time, so we never had a chance to sweep

I looked into this, but we'd need to introduce a htlc broadcast state to tell the difference between these two. Would rather get the change that prevents late reveal in, then we can come back and add a state properly. Mostly deleted code since last review, but I think it's changed significantly enough for another round of review.

guggero (approved):

yyforyongyu:
LGTM

yyforyongyu (approved):

PR #374, opened 2021-04-26
loopd+loopdb: add timeout to DB open

guggero:
To make sure we don't just block for forever if another Loop daemon
process is already running, we add a timeout and error out if obtaining
the unique file lock fails after 5 seconds.

#### Pull Request Checklist

### DIFF
--- a/loopd/macaroons.go
+++ b/loopd/macaroons.go
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"time"
 
+	"github.com/coreos/bbolt"
+	"github.com/lightninglabs/loop/loopdb"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/macaroons"
 	"google.golang.org/grpc"
@@ -17,10 +18,6 @@ const (
 	// loopMacaroonLocation is the value we use for the loopd macaroons'
 	// "Location" field when baking them.
 	loopMacaroonLocation = "loop"
-
-	// macDatabaseOpenTimeout is how long we wait for acquiring the lock on
-	// the macaroon database before we give up with an error.
-	macDatabaseOpenTimeout = time.Second * 5
 )
 
 var (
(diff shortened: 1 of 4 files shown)


loopdb/store.go
```diff
@@ -99,9 +99,18 @@ var (
...
+	// DefaultLoopOutHtlcConfirmations is the default number of
+	// confirmations we set for a loop out htlc.
+	DefaultLoopOutHtlcConfirmations uint32 = 1
+
+	// DefaultLoopDBTimeout is the default maximum time we wait for the
+	// Loop bbolt database to be opened. If the database is already opened
+	// by another process, the unique lock cannot be obtained. With the
+	// timeout we error out after the given time instead of just blocking
+	// for forever.
+	DefaultLoopDBTimeout = 5 * time.Second
```
yyforyongyu, line 112: there's also the `macDatabaseOpenTimeout`, should we unify them and use only `DefaultLoopDBTimeout` instead? Also, just a thought, maybe consider making this configurable in the future?

yyforyongyu (approved):
LGTM!

loopdb/store.go
guggero: Ah, good point. I removed the `macDatabaseOpenTimeout` and replaced it with `DefaultLoopDBTimeout`.
I'm not sure how much value customizing this timeout would give us at the moment, so I'm leaving it out of this PR. Should be pretty easy to add if we ever see a use case.

loopd/macaroons.go
```diff
@@ -150,8 +147,16 @@ func (d *Daemon) startMacaroonService() error {
 	var err error
 	d.macaroonService, err = macaroons.NewService(
 		d.cfg.DataDir, loopMacaroonLocation, false,
-		macDatabaseOpenTimeout, macaroons.IPLockChecker,
+		loopdb.DefaultLoopDBTimeout, macaroons.IPLockChecker,
 	)
+	if err == bbolt.ErrTimeout {
+		return fmt.Errorf("error while trying to open %s/%s: "+
```
carlaKC, line 153: Great explanatory error!

Just wondering where these errors should live. `startMacaroonService` and `NewBoltSwapStore` are both called from our `initialize` function for the daemon. WDYT about bubbling the bbolt.ErrTimeout all the way up (or wrap it so that we can also have the db name), and then we can have a single descriptive error in the [daemon](https://github.com/lightninglabs/loop/blob/master/loopd/daemon.go#L130)?

If we do it like that, the terminal-specific instructions could also be handled in terminal by checking the response for `StartAsSubserver`'s [response](https://github.com/lightninglabs/loop/blob/master/loopd/daemon.go#L171).

carlaKC (commented):
Great addition! Just some nittyness around where the user-instructive/ terminal-specific error should live.

loopd/macaroons.go
guggero: Okay, I've implemented your suggestion. What do you think?

carlaKC (approved):
Thanks for updating! LGTM

loopd/macaroons.go
carlaKC: Looks great!

PR #377, opened 2021-04-28
loop: wait for chain notifier server to start

bhandras:

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -3,6 +3,7 @@ package loop
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


rockstardev:
Thanks for quick turnaround on this @bhandras, now I only hope review from @guggero @carlaKC and merge + release will be as fast ;)

executor.go
```diff
@@ -52,10 +53,34 @@ func newExecutor(cfg *executorConfig) *executor {
...
+			s.lnd.ChainNotifier.RegisterBlockEpochNtfn(mainCtx)
+		if err != nil {
+			if strings.Contains(err.Error(),
+				"in the process of starting") {
+
+				// Give chain notifier some time to start and
+				// try to re-attempt block epoch subscription.
+				select {
+				case <-time.After(50 * time.Millisecond):
+					continue
```
carlaKC, line 73: Perhaps add a log here that we're waiting for the chain notifier? Otherwise loop will just be hanging and user may not know why?

executor.go
```diff
@@ -52,10 +53,34 @@ func newExecutor(cfg *executorConfig) *executor {
...
+
+				case <-mainCtx.Done():
+					return err
+				}
+			}
+
+			return err
+		}
+
+		break
```
carlaKC, line 83: Do we need to break? Couldn't we just return err here? Either it's nil or it's a non-"process of starting" error.

carlaKC (commented):
Just some nits!

executor.go
guggero: The function continues below. So getting here means we've succeeded with the connection.

executor.go
guggero: Logging sounds good to me. And maybe increase the delay to 500ms to not spam `lnd` too much? It's got enough on its plate at this moment of startup anyway :joy:

guggero (approved):
Thanks for the quick fix!
Just one nit, otherwise LGTM :100:

executor.go
bhandras: Yeah the 50ms was a bit overkill :) Log added.

executor.go
```diff
@@ -52,10 +53,37 @@ func newExecutor(cfg *executorConfig) *executor {
...
+	)
+
+	for {
+		blockEpochChan, blockErrorChan, err =
+			s.lnd.ChainNotifier.RegisterBlockEpochNtfn(mainCtx)
+		if err != nil {
+			if strings.Contains(err.Error(),
+				"in the process of starting") {
+
+				log.Warnf("LND chain notifier server not " +
```
carlaKC, line 69: nit: could just be `Warn`, but don't push again for this!

carlaKC (approved):
New record for fastest issue fix in loop! 🏁

LGTM

PR #378, opened 2021-04-29
loopout: Add swap cancelation for swaps that cannot route off-chain

carlaKC:
This PR adds communication between the server and client to explicitly cancel loop out swaps when the client cannot route to the server. This allows the server to release any htlcs that are being held (eg, the prepay since it's more likely to be routed) earlier, and decreases the amount of time that funds may be held off-chain.

#### Pull Request Checklist


looprpc/server.proto
```diff
@@ -309,3 +320,87 @@ message SubscribeLoopInUpdatesResponse{
...
+    */
+    FAILURE_REASON_INSUFFICIENT_BALANCE = 5;
+}
+
+message RouteCancel{
+    // The type of the payment that failed.
+    RoutePaymentType route_type = 1;
+
+    // The htlcs that the client tried to pay the server with, if any.
+    repeated HtlcAttempt attempts = 2;
```
alexbosworth, line 378: Why is this needed?

looprpc/server.proto
```diff
@@ -309,3 +312,88 @@ message SubscribeLoopInUpdatesResponse{
...
+    */
+    FAILURE_REASON_INCORRECT_PAYMENT_DETAILS = 4;
+
+    /*
+    Insufficient local balance.
+    */
+    FAILURE_REASON_INSUFFICIENT_BALANCE = 5;
+}
+
+message RouteCancel{
```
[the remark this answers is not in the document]
guggero: Oh, we don't have that yet... Should probably switch the protobuf generation over to Docker and add the clang-format template. I can add that as a separate PR.

swap_server_client.go
```diff
@@ -456,6 +461,104 @@ func (s *grpcSwapServerClient) makeServerUpdate(ctx context.Context,
...
+			// We can cast our lnd failure reason to a loop payment
+			// failure reason because these values are copied 1:1
+			// from lnd.
+			Failure: looprpc.PaymentFailureReason(
+				details.metadata.failureReason,
+			),
+		},
+	}
+
+	switch details.metadata.paymentType {
```
guggero, line 547: The `metadata` field needs a nil check. Or we should perhaps not use a pointer but the struct directly instead if the metadata always need to be set.

loopout.go
```diff
@@ -504,31 +533,64 @@ func (s *loopOutSwap) payInvoices(ctx context.Context) {
 	)
 }
 
+// paymentResult contains the response for a failed or settled payment, and
+// any errors that occurred if the payment unexpectedly failed.
+type paymentResult struct {
+	status *lndclient.PaymentStatus
```
guggero, line 558: Since we have the `Payment_UNKNOWN` state for value `0`, should we make this an explicit value instead of a pointer?

loopout.go
```diff
@@ -75,6 +75,24 @@ type executeConfig struct {
 	blockEpochChan  <-chan interface{}
 	timerFactory    func(d time.Duration) <-chan time.Time
 	loopOutMaxParts uint32
+	cancelSwap      func(context.Context, *outCancelDetails) error
+}
+
+func newExecuteConfig(sweeper *sweep.Sweeper, statusChan chan<- SwapInfo,
```
guggero, line 82: If we don't have any logic in here, why are we using a function to initialize the struct? I find it much more readable to have a struct initialization block than a long list of function arguments.

loopout.go
```diff
@@ -975,6 +983,92 @@ func (s *loopOutSwap) pushPreimage(ctx context.Context) {
...
+			continue
+		}
+
+		failureIdx := htlc.Failure.FailureSourceIndex
+		hops := uint32(len(htlc.Route.Hops))
+
+		// We really don't expect a failure index that is greater than
+		// our number of hops, but we don't want to underflow so we
+		// check and log a warning if this happens.
+		if failureIdx > hops {
```
guggero, line 1047: Greater or equal? Since `FailureSourceIndex` sounds zero-based while `hops` is a length? Or is this a special case for HTLC routes where index 0 is always the own node (which wouldn't be in the list of hops)?

guggero (commented):
Great UX improvement! Also very nice commit structure, small and easy to digest changes. Just a few minor comments, looks mostly good to me :tada:

looprpc/server.proto
```diff
@@ -309,3 +312,88 @@ message SubscribeLoopInUpdatesResponse{
...
+    // The swap hash.
+    bytes swap_hash = 2; 
+
+    // The payment address for the swap invoice, used to ensure that only the
+    // swap owner can cancel the payment.
+    bytes payment_address = 3;
+
+    // Additional information about the swap cancelation.
+    oneof cancel_info {
+        RouteCancel route_cancel = 5;
```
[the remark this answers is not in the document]
bhandras: +1

swap_server_client.go
```diff
@@ -456,6 +461,104 @@ func (s *grpcSwapServerClient) makeServerUpdate(ctx context.Context,
...
+	}
+
+	_, err = s.server.CancelLoopOutSwap(ctx, req)
+	return err
+}
+
+func rpcRouteCancel(details *outCancelDetails) (
+	*looprpc.CancelLoopOutSwapRequest_RouteCancel, error) {
+
+	attempts := make([]*looprpc.HtlcAttempt, len(details.metadata.attempts))
```
bhandras, line 528: Return early if `metadata` is nil.

bhandras (commented):
Apart from a few nits @guggero also suggested changes LGTM, very nice work.

looprpc/server.proto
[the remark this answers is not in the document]
carlaKC: Yeah, I was thinking about adding one for when the client gives up based on max miner fee, so wanted to leave this open. Can also leave it flat and then just set one or the other as we need?

loopout.go
carlaKC: I somewhat prefer the constructor because we use `executeConfig` in so many places, albeit mostly tests, easy way to enforce that all fields are set. Not overly attached, reverted.

swap_server_client.go
carlaKC: Removed pointer.

looprpc/server.proto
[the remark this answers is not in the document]
carlaKC: Manually fixed for now.

loopout.go
carlaKC: sgtm!

loopout.go
carlaKC: Yeah indeed, nice catch!

looprpc/server.proto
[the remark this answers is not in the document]
guggero: I'm okay with leaving this as is if there are potential other cancellation reasons.

loopout.go
guggero: Okay, cool. Good to know. Maybe we could even extend the comment somewhat with what you wrote above?

guggero (approved):
Nice, LGTM :100: 

Is there an easy way to test this e2e in the itest? Would probably be worth adding a simple case to the server side PR (if simulating the failure is not a huge pain to set up).

loopout.go
carlaKC: Updated comment!

carlaKC:
Pushed changes re htlc attempt [here](https://github.com/lightninglabs/loop/compare/01c46da993d179a39b8186076acebd0aaf02f43a..87c1aa4ccd2a69889da8582e2d4a51731c1f31ca).

bhandras (approved):
LGTM :icecream:

alexbosworth (approved):

carlaKC:
Server side deployed, client ok to merge.

Issue #390, opened 2021-05-24
Protobuf warnings for loop


carlaKC:
Now that we've bumped our grpc version, every interaction with the `loop` cli tool is prefixed with the following warnings: 
```
$ loop out --amt=100000
2021/05/24 09:35:16 WARNING: proto: file "server.proto" has a name conflict over looprpc.SUCCESS
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict

2021/05/24 09:35:16 WARNING: proto: file "server.proto" has a name conflict over looprpc.HTLC_PUBLISHED
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict

2021/05/24 09:35:16 WARNING: proto: file "server.proto" has a name conflict over looprpc.INITIATED
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict`

Send off-chain:                           1000000 sat
Receive on-chain:                          990488 sat
Estimated total fee:                         9512 sat

Regular swap speed requested, it might take up to 30m0s for the swap to be executed.
```

This is pretty ugly and/or scary for the user, although nothing is actually wrong right now. This will be an issue if we bump our proto version further, so we should address it anyway.

carlaKC:
Fixed by #395 + tagged in latest client.

PR #395, opened 2021-06-03
looprpc: prefix server and lnd enums to create unique names

carlaKC:
This is one of our options for debugging #390.

The issue is that we have duplicate names in `client.proto` and `server.proto` since the states that the server and the client have overlap (eg, both have an initiated state). Our options are: 
- Rename client enums: this would break any client/UI integrations that use the enums, not ideal
- Move server into different grpc package: complicated because the grpc package is part of the URI, so we'd need to serve on legacy and new package, which will get messy
- Rename server enums: chosen here, grpc uses the value of an enum so we can get away with the rename without breaking comms between server and client; this is provided we use the [binary grpc encoding](https://stackoverflow.com/questions/50857940/renaming-enum-value-in-protobuf-backward-compatibility), which I believe we do because we don't support REST

#### Pull Request Checklist


guggero (approved):
Great fix, thank you for taking it over and also running all the tests! LGTM :100:

bhandras (approved):
LGTM :100:

PR #396, opened 2021-06-07
loopin: handle SubscribeSingleInvoice termination

carlaKC:
In lnd 0.13 onwards, `SubscribeSingleInvoice` will terminate its stream once a final state is served (mirroring the behaviour of `TrackPayment`). 
* In regular cases where we're using the output of this stream, we are unaffected because we consume the final state then stop listening on the update channels. 
* When we're listening for invoice updates along with other cases (eg in `waitForSwapComplete` we're also waiting for the htlc spend to be confirmed) we need to handle the stream exit. 

This is mostly handled in lndclient, which closes the update channels, so this PR updates loop to use the new lndclient logic, and to test this functionality. We do a similar thing for loopout in `waitForHtlcSpendConfirmed`.

Depends on: lightninglabs/lndclient/pull/58

#### Pull Request Checklist

### DIFF
--- a/loopin_test.go
+++ b/loopin_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/loop/loopdb"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightninglabs/loop/test"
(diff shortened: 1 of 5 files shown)


loopin_test.go
```diff
@@ -89,17 +88,11 @@ func TestLoopInSuccess(t *testing.T) {
 	<-ctx.lnd.RegisterSpendChannel
 
 	// Client starts listening for swap invoice updates.
-	subscription := <-ctx.lnd.SingleInvoiceSubcribeChannel
-	if subscription.Hash != ctx.server.swapHash {
-		t.Fatal("client subscribing to wrong invoice")
-	}
+	ctx.assertSubscribeInvoice(ctx.server.swapHash)
```
yyforyongyu, line 91: 👍

yyforyongyu (approved):
LGTM👍 Nice small refactoring in the test!

loopin_test.go
```diff
@@ -257,10 +250,7 @@ func testLoopInTimeout(t *testing.T,
...
```
bhandras, line 253: nice to have these, much more readable!

bhandras (approved):
LGTM :racehorse:

PR #403, opened 2021-07-13
looprpc: generate protos with docker

guggero:
To get rid of the imported REST annotations in the proto files we extract those into their own `rest-annotations.yaml` file.
And to make it easier for developer to compile the protos with the correct version, we supply a docker image that does it.

I didn't update the release notes since this really only affects developers.

#### Pull Request Checklist


looprpc/server.proto
```diff
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package looprpc;
 
+option go_package = "github.com/lightninglabs/loop/looprpc";
```
carlaKC, line 5: Why do we need to add this, out of curiosity?

carlaKC (approved):

looprpc/server.proto
guggero: I think this is mandatory in the new version of protoc we use. But in general it's good practice to set.

looprpc/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.15.6-buster
+
+RUN apt-get update && apt-get install -y \
+  git \
+  protobuf-compiler='3.6*' \
+  clang-format='1:7.0*'
+
+# We don't want any default values for these variables to make sure they're
```
yyforyongyu, line 8: 👍

looprpc/client.proto
```diff
@@ -44,102 +34,63 @@ service SwapClient {
...
     GetLiquidityParams gets the parameters that the daemon's liquidity manager
     is currently configured with. This may be nil if nothing is configured.
     [EXPERIMENTAL]: endpoint is subject to change.
     */
-    rpc GetLiquidityParams (GetLiquidityParamsRequest) returns (LiquidityParameters) {
-        option (google.api.http) = {
-            get: "/v1/liquidity/params"
-        };
-    }
+    rpc GetLiquidityParams (GetLiquidityParamsRequest)
```
yyforyongyu, line 75: just curious why don't we have something like `loop: `listauth`` in the docs? Because it's experimental?

yyforyongyu (approved):
LGTM⚡️

looprpc/client.proto
guggero: Good question. I added them now. Those tags are for the API docs only so it can also add the output of the given command to the generated docs. @carlaKC any objections to adding them?

carlaKC: No objections!

guggero:
Had to fix some linter issues that were triggered by re-generating the protos. Re-requesting review for a sanity check :)

carlaKC (approved):

PR #405, opened 2021-07-16
looprpc: add swap multiple swap scripts server state to rpc enum

carlaKC:
Add a new failure state that the server may send to the client.
This state is sent for loop in swaps where the client has
erroneously published multiple outputs for a loop in swap (this
could happen if an end user created an external tx with too many
outputs). Failing the swap server side protects users from potential
footguns where they publish the hltc then the server can claim more
money than was paid out off chain. This can never happen when the
loop client publishes the htlc because it only ever contains one
output.

#### Pull Request Checklist


bhandras (approved):
LGTM :+1:

looprpc/server.proto
```diff
@@ -305,6 +305,12 @@ enum ServerSwapState {
     // The client canceled the swap because they could not route the swap
     // payment.
     SERVER_CLIENT_INVOICE_CANCEL = 14;
+
+    /*
+    A loop in swap was rejected because it contained multiple outputs for a
+    single swap.
+    */
+    SERVER_FAILED_MULTIPLE_SWAP_SCRIPTS = 15;
```
yyforyongyu, line 313: Q: When do we use the state?

yyforyongyu (approved):
LGTM⚡️

Issue #355, opened 2021-03-10
Loop In: real server error is swallowed


guggero:
Referencing https://github.com/lightninglabs/lightning-terminal/issues/198.

The user sees the error message:
```
2021-03-10 08:30:40.567 [ERR] LOOPD: Loop in: cannot initiate swap: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

Even though the server returns the following error code (payment probe failing due to timeout):
```
2021-03-10 08:30:41.367 [ERR] NAUT: Swap b044bd3xxxxx initiation error: FailedPrecondition (target unreachable)
```

It seems somewhere we swallow the actual error code.

andrerfneves:
@guggero in cases like this what is the best approach to resolve this? Smaller loop in amounts? Are these bad channels?

andrerfneves:
I've tried with quite a few diff channels and different amounts. All Loop Ins error out like this with the deadline exceeded call.

guggero:
That would've been my suggestion, to just try again a few times with different amounts and different channels. I just tried a loop in myself and it seems to work in general. Could be you have a peer between you and the pool server that's not properly forwarding the probe payment. Unfortunately this is quite hard to debug.

PR #407, opened 2021-07-19
multi: surface grpc error codes from server to client

carlaKC:
Formatting our error was stifling any grpc error returned by the
server. Instead, we bubble up our grpc error, setting an unknown
code if the server did not specifically return an error code.

Fixes: #355

#### Pull Request Checklist

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,6 +15,7 @@ import (
 	"github.com/lightninglabs/loop/loopdb"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightninglabs/loop/sweep"
+	"google.golang.org/grpc/status"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


guggero (approved):
tACK, LGTM :100: 

Before:
```
[loop] rpc error: code = Unknown desc = cannot initiate swap: rpc error: code = FailedPrecondition desc = target unreachable
```

After:
```
[loop] rpc error: code = FailedPrecondition desc = cannot initiate swap: target unreachable
```

client.go
```diff
@@ -610,3 +611,17 @@ func (s *Client) LoopInTerms(ctx context.Context) (
 
 	return s.Server.GetLoopInTerms(ctx)
 }
+
+// wrapGrpcError wraps the non-nil error provided with a message providing
+// additional context, preserving the grpc code returned with the original
+// error. If the original error has no grpc code, then codes.Unknown is used.
+func wrapGrpcError(message string, err error) error {
```
bhandras, line 618: Is there an advantage of wrapping unknown errors in the few places in the PR vs just returning `status.Error` there with proper error codes?

bhandras (approved):
LGTM :+1:

client.go
carlaKC: I think it's nice to be able to wrap with a bit more information and preserve the error? In the case of unknown errors, I think it's ok to add the unknown error code anyway, doesn't really take away much. We _could_ go through the whole api and add error codes everywhere, but I think surfacing the error codes returned by the server is a good start?

PR #413, opened 2021-08-04
Improve error handling of config file

### LINKED ISSUE #412: Settings in config file seem to be ignored
I have this config file:

```
network=regtest
datadir=/var/lib/loopd-system-regtest
rpclisten=0.0.0.0:11012
restlisten=0.0.0.0:11112
tlscertpath=/var/lib/loopd-system-regtest/loop.cert
tlskeypath=/var/lib/loopd-system-regtest/loop.key

lnd.host=127.0.0.1:10011

lnd.tlspath=/var/lib/lnd-system-regtest/public/tls.cert

lnd.macaroonpath=/var/lib/lnd-system-regtest/private/admin.macaroon
```

The `lnd.*` options seem to be ignored.

When I attempt to launch `loopd` 

**Expected behavior**

`loopd` reads certificate from `/var/lib/lnd-system-regtest/public/tls.cert`, macaron from `/var/lib/lnd-system-regtest/private/admin.macaroon` and connects to `127.0.0.1:10011`

**Actual behavior**

`loopd` attempts to connect to `127.0.0.1:10009` and read `/var/lib/loopd-system-regtest/.lnd/tls.cert`

Log output:
```
2021-08-03 14:29:21.840 [INF] LOOPD: Version: 0.14.2-beta commit=
2021-08-03 14:29:21.840 [INF] LNDC: Creating lnd connection to localhost:10009
loopd exited with an error: open /var/lib/loopd-system-regtest/.lnd/tls.cert: no such file or directory
```

**To reproduce**

Create the configuration file above and launch `loopd` with argument `--configfile /etc/loopd-system-regtest/loopd.conf`

**System information**

* Standalone binary
* Interaction with loop impossible because it can't start
* Loopd version: 0.14.2 - downloaded from releases
* Machine: Debian 10, x86_64, NVMe SSD
* OS: Debian 10

Kixunil:
This makes two changes to handling of errors when configuration file
could not be loaded:

1. Only NotFound errors are considered OK - access errors and other FS
   issues are now treated as fatal.
2. Failing to load config file specified explicitly via `--configfile`
   option is alway a fatal error.

Rationale: If the configfile was specified explicitly then it indicates
the user really wishes to load it. While the user could want it to be
optionally loaded for extra configuration options, this can be
accomplished using an empty file.

If the config file was not specified explicitly then its' path was
computed from loop directory. If the file is inaccessible due to
permissions or other FS errors it's nearly certain other following
operations will fail as well. Failing early with a clear message is thus
beneficial.

This still leaves room for uncaught user error (e.g. mistakenly naming
config file inside loop dir as `loop.conf` instead of `loopd.conf`) but
it's greatly reduced and such error should be easier to identify.

(Indirectly) closes #412

#### Pull Request Checklist

### DIFF
--- a/loopd/run.go
+++ b/loopd/run.go
@@ -250,7 +250,8 @@ func Run(rpcCfg RPCConfig) error {
 }
 
 // getConfigPath gets our config path based on the values that are set in our
-// config.
+// config. The returned bool is set to true if the config file path was set
+// explicitly by the user and thus should not be ignored if it doesn't exist.
 func getConfigPath(cfg Config, loopDir string) (string, bool) {
 	// If the config file path provided by the user is set, then we just
 	// use this value.
--- a/loopd/run.go
+++ b/loopd/run.go
@@ -153,7 +153,7 @@ func Run(rpcCfg RPCConfig) error {
 
 	if err := flags.IniParse(configFile, &config); err != nil {
 		// File not existing is OK as long as it wasn't specified
-		// explicitly.  All other errors (parsing, EACCESS...) indicate
+		// explicitly. All other errors (parsing, EACCESS...) indicate
 		// misconfiguration and need to be reported. In case of
 		// non-not-found FS errors there's high likelihood that other
 		// operations in data directory would also fail so we treat it
--- a/loopd/run.go
+++ b/loopd/run.go
@@ -149,7 +149,7 @@ func Run(rpcCfg RPCConfig) error {
 
 	// Parse ini file.
 	loopDir := lncfg.CleanAndExpandPath(config.LoopDir)
-	configFile, explicitConfig := getConfigPath(config, loopDir)
+	configFile, hasExplicitConfig := getConfigPath(config, loopDir)
 
 	if err := flags.IniParse(configFile, &config); err != nil {
 		// File not existing is OK as long as it wasn't specified
@@ -158,7 +158,7 @@ func Run(rpcCfg RPCConfig) error {
 		// non-not-found FS errors there's high likelihood that other
 		// operations in data directory would also fail so we treat it
 		// as early detection of a problem.
-		if explicitConfig || !os.IsNotExist(err) {
+		if hasExplicitConfig || !os.IsNotExist(err) {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


release_notes.md
```diff
@@ -18,6 +18,12 @@ This file tracks release notes for the loop client.
 
 #### Breaking Changes
 
+* Failing to load configuration file specified by `--configfile` for any reason if now hard error.
```
carlaKC, line 21: nit: wrap at 80

loopd/run.go
```diff
@@ -149,13 +149,16 @@ func Run(rpcCfg RPCConfig) error {
 
 	// Parse ini file.
 	loopDir := lncfg.CleanAndExpandPath(config.LoopDir)
-	configFile := getConfigPath(config, loopDir)
+	configFile, explicitConfig := getConfigPath(config, loopDir)
```
carlaKC, line 152: nit: `customConfig` makes a little more sense than `explicitConfig` to me, but nonblocking

carlaKC (approved):
tACK! Thank you for the fix, great improvement :)

release_notes.md
Kixunil: Hmm, previous lines weren't 80-wrapped so I thought the limit doesn't apply. Will fix.

loopd/run.go
Kixunil: I'm not really sure. Each config is custom. :) Maybe `nonDefaultConfigPath`?

release_notes.md
Kixunil: Done.

loopd/run.go
yyforyongyu: maybe an action in the name? something like `hasExplicitConfig`,`hasConfigSet`, etc...

yyforyongyu (approved):
Thanks for the PR @Kixunil ! Changes look good to me, just a few nits.

Kixunil:
Addressed the nits.

carlaKC:
Thanks for this one, great improvement! 

Tested: 
1. `--configfile` exists -> ok
2. `--configfile` doesn't exist -> exits
3. No `--configfile`, default location cfg file does not exist -> ok

PR #415, opened 2021-08-24
loopin: Add --private parameter

arshbot:
Passthrough for route hints to lnd

#### Pull Request Checklist


carlaKC (commented):
Great start! 

It takes a while to get familiar with all the places that changes need to be made (too many things called client + server 😅), so would suggest taking a look in the following spots: 
1. Add hop hints to [LoopInRequest](https://github.com/lightninglabs/loop/blob/master/looprpc/client.proto#L212) and the [loopd grpc server to use the hints.](https://github.com/lightninglabs/loop/blob/master/loopd/swapclient_server.go#L598)
2. The server has a [client impl](https://github.com/lightninglabs/loop/blob/master/loopd/swapclient_server.go#L632) that contains our actual swap logic, which [creates the swap](https://github.com/lightninglabs/loop/blob/master/client.go#L523) and [executes it](https://github.com/lightninglabs/loop/blob/master/client.go#L532).
3. We need to pass our hints through when we [get a quote](https://github.com/lightninglabs/loop/blob/7ac6e26e90c4f5e086cfce356121da03c8bf0cdb/loopin.go#L92) and for our [swap](https://github.com/lightninglabs/loop/blob/7ac6e26e90c4f5e086cfce356121da03c8bf0cdb/loopin.go#L130) and [probe](https://github.com/lightninglabs/loop/blob/7ac6e26e90c4f5e086cfce356121da03c8bf0cdb/loopin.go#L149) invoices.

cmd/loop/loopin.go
```diff
@@ -120,12 +126,20 @@ func loopIn(ctx *cli.Context) error {
 
 		lastHop = lastHopVertex[:]
 	}
+	var hints []*looprpc.RouteHint
```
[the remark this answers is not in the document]
bhandras: +1 for the `--private` flag. Power users will probably use the API.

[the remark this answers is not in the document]
arshbot: > Power users will probably use the API.

Is it worth nixing `--route_hints` in this case?

loopd/swapclient_server.go
```diff
@@ -494,8 +494,13 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 
 		lastHop = &lastHopVertex
 	}
-
-	routeHints, err := unmarshallRouteHints(req.LoopInRouteHints)
+	var routeHints [][]zpay32.HopHint
```
carlaKC, line 497: probably want some validation here that the client doesn't set both? Doesn't really make sense for caller to set both

loopin.go
```diff
@@ -89,10 +100,10 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
 	// route hints.
 	quote, err := cfg.server.GetLoopInQuote(
 		globalCtx, request.Amount, cfg.lnd.NodePubkey, request.LastHop,
-		nil,
+		routeHints,
```
carlaKC, line 103: In addition to adding these hints to the quote, we also need to add them to the swap and probe invoices below so that we use the same set of hints for all operations. 

You'll need to make another lndclient PR to add these params, then update `go.mod` to point to the new version.

loopin.go
```diff
@@ -80,6 +81,16 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
 	currentHeight int32, request *LoopInRequest) (*loopInInitResult,
 	error) {
 
+	var routeHints [][]zpay32.HopHint
+	var err error
+	if request.Private {
+		routeHints, err = SelectHopHints(globalCtx, cfg.lnd,
```
carlaKC, line 87: One consideration here is whether we have a `request.LastHop` set. If we're only targetting one peer, it makes sense to only include hops for that peer (minimize privacy leaks).

carlaKC (commented):
Looking good! Def think that this approach is the right way to go :D 

Just need to thread those hops through to our swap invoice + probe and we're good!

loopd/swapclient_server.go
arshbot: I added validation in upstream in the cli tool, however here I'm not it's possible to tell if a the `private` flag is set or not set since it _should_ default to false ([looking at this stackoverflow post](https://stackoverflow.com/a/43351386/5838056)). I'm open to suggestions here though if the cli validation isn't enough. As it is, while it won't complain for both parameters being set, it will prioritize the `private` flag over `routehints`

carlaKC: Generally my preference is to keep validation in on the server side so that all calls are subject to the same validation. If we just keep the validation in cli then we might miss adding it if we add a new call to this server endpoint. 

In terms of validation here, can't you just have: 
```
haveRouteHints := len(req.LoopInRequestHints) !=0

switch{
    case haveRouteHints && req.Private: 
         return "err can't set both"
         
     case haveRouteHints: 
        routeHints, err = unmarshallRouteHints(req.LoopInRouteHints)
        
     case: req.Private: 
         routeHints, err = loop.SelectHopHints(ctx, s.lnd,...
}
```

> however here I'm not it's possible to tell if a the private flag is set or not set since it should default to false

I believe that the only option that the client has is `--private` in our flags library, they can't set `--private=false` so we don't need to worry about default values.

cmd/loop/loopin.go
```diff
@@ -31,6 +32,10 @@ var (
 			"with our reserved prefix: %v.",
 			labels.MaxLength, labels.Reserved),
 	}
+	routeHints = cli.StringFlag{
```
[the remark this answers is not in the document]
arshbot: Great idea, done

cmd/loop/loopin.go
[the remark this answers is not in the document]
arshbot: `--private` implemented

loopin.go
arshbot: I didn't end up needing to update lndclient for some reason, tested with a private channel and was able to pay? In any case, passed these values to where they needed to go

loopin.go
arshbot: That's a good point, but makes me also wonder. What if the user specifies some combination of `lasthop`/`routehints`? I'm going to error out in that case

arshbot: Actually nvm, it seems like this case should be totally fine since it's handled downstream

loopd/swapclient_server.go
arshbot: > In terms of validation here, can't you just have:

Yup that worked wonders! Thank you :) 

Implemented, tested, and pushed :D

loopin.go
arshbot: we now respect `LastHop`!

cmd/loop/loopin.go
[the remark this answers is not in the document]
carlaKC: Can you set a `StringFlag` multiple times? I think it's different to `StringSlice`?

cmd/loop/loopin.go
```diff
@@ -36,6 +36,10 @@ var (
 		Name:  "route_hints",
 		Usage: "",
 	}
+	privateFlag = cli.BoolFlag{
+		Name:  "private",
+		Usage: "",
```
carlaKC: Add usage here to explain to the user what this will do?

loopin.go
carlaKC: Are you sure the swap was actually paid on a private channel? If you have private and public channels then the swap will succeed on the public channel. I can't get a swap to succeed here with a single private channel.

utils.go
```diff
@@ -87,7 +87,13 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices, amtMSat btc
 			continue
 		}
 
-		// TODO: HARSHA cache this
+		// If include_nodes is set, we'll only add channels with peers in include_node.
+		// This is done to respect the last_hop parameter
+		if _, ok := include_nodes[channel.PubKeyBytes]; include_nodes != nil && !ok {
```
carlaKC, line 92: Since we're either going to have to keep this up to date with lnd or update to point to lnd, I think it would make sense to do our custom processing outside of this function.

release_notes.md
```diff
@@ -16,6 +16,16 @@ This file tracks release notes for the loop client.
 
 #### New Features
 
+* `--private` flag is now available on the `loop in` command, meaning users can
+  now loop in to private nodes! This was implemented in [#415](https://github.com/lightninglabs/loop/pull/415)
+  and has an implementation of LND's hop hints creation feature for the time
+  being, with the benefit of not creating an extra new invoice for each loop in
```
carlaKC, line 22: Nice update! I think we can cut the part of "not creating an extra invoice" because that will be confusing to the end user (they don't know our discussions about the dev of this feature). These notes are more targeted at explaining how to use new features in loop than a record for developers (since we can go back and look at the original prs).

utils.go
```diff
@@ -0,0 +1,180 @@
...
+	if bytes.Equal(channel.PubKeyBytes[:], serializedPubKey) {
+		remotePolicy = p1
+	} else {
+		remotePolicy = p2
+	}
+
+	return remotePolicy, true
+}
+
+func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices, amtMSat btcutil.Amount,
```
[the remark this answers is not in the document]
bhandras: nit: godoc

loopd/swapclient_server.go
```diff
@@ -483,26 +483,41 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 		return nil, err
 	}
 
+	var routeHints [][]zpay32.HopHint
+	haveRouteHints := len(req.LoopInRouteHints) != 0
+	haveLastHop := req.LoopInLastHop != nil
 	var lastHop *route.Vertex
-	if req.LoopInLastHop != nil {
+
+	switch {
```
bhandras, line 491: nit: for `switch` we have a code style we use all across the codebase: https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#code-spacing

loopd/swapclient_server.go
```diff
@@ -483,26 +483,41 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
+	// private and routehints are mutually exclusive as setting private
+	// means we retrieve our own routehints from the connected node
+	case haveRouteHints && req.Private:
+		return nil, fmt.Errorf("private and route_hints both set")
+	case haveLastHop:
 		lastHopVertex, err := route.NewVertexFromBytes(
 			req.LoopInLastHop,
 		)
+		lastHop = &lastHopVertex
```
bhandras, line 500: nit: change the order of this line and the err check below

loopd/swapclient_server.go
```diff
@@ -483,26 +483,41 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
 		}
-
-		lastHop = &lastHopVertex
-	}
-	var routeHints [][]zpay32.HopHint
-	if req.Private {
+	case req.Private:
+		// If last_hop is set, we'll only add channels with peers
+		// set to the last_hop parameter
+		var include_nodes map[route.Vertex]struct{}
```
bhandras, line 507: nit:  use `camelCase` for local variables too

loopd/swapclient_server.go
```diff
@@ -483,26 +483,41 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
+		if req.LoopInLastHop != nil {
+			include_nodes = make(map[route.Vertex]struct{})
+		}
 		routeHints, err = loop.SelectHopHints(ctx, s.lnd,
-			btcutil.Amount(req.Amt), 20)
-	} else {
+			btcutil.Amount(req.Amt), 20, include_nodes)
```
bhandras, line 512: nit: use named constant

loopd/swapclient_server.go
```diff
@@ -483,26 +483,41 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
-		lastHop = &lastHopVertex
-	}
-	var routeHints [][]zpay32.HopHint
-	if req.Private {
+	case req.Private:
+		// If last_hop is set, we'll only add channels with peers
+		// set to the last_hop parameter
+		var include_nodes map[route.Vertex]struct{}
+		if req.LoopInLastHop != nil {
+			include_nodes = make(map[route.Vertex]struct{})
```
bhandras, line 509: This part doesn't seem to be finished?

utils.go
```diff
@@ -66,9 +67,8 @@ func chanCanBeHopHint(ctx context.Context, lnd *lndclient.LndServices, channels
 }
 
 func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices, amtMSat btcutil.Amount,
-	numMaxHophints int) ([][]zpay32.HopHint, error) {
+	numMaxHophints int, include_nodes map[route.Vertex]struct{}) ([][]zpay32.HopHint, error) {
```
bhandras, line 70: nit: camelCase

bhandras (changes requested):
This is going to be an awesome addition to the loop-in experience! Tried to add comments also focusing on general stlye and readability. Since the changes added introduce some non-trivial behavior, it'd be nice to add some unit tests too where applicable.

cmd/loop/loopin.go
```diff
@@ -120,12 +126,20 @@ func loopIn(ctx *cli.Context) error {
 
 		lastHop = lastHopVertex[:]
 	}
+	var hints []*looprpc.RouteHint
+	if ctx.IsSet(routeHints.Name) {
+		err = json.Unmarshal([]byte(ctx.String(routeHints.Name)), &hints)
+		if err != nil {
+			fmt.Errorf("unable to parse json: %v", err)
```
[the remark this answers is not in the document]
arshbot: Done

loopin.go
arshbot: confirmed it works on my end, @carlaKC could you confirm on your end?

arshbot: Okay, I can confirm it works on my end with a single channel -- however I did run into a single instance of probe failure and I'm not sure what the cause was. Gonna throw some unit tests and ensure we have parity with the lnd instance

utils.go
arshbot: not sure i follow, i really just replaced a lot of the processing

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/urfave/cli v1.20.0
 	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4
+	google.golang.org/appengine v1.6.6 // indirect
```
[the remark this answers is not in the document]
carlaKC: Delete then `go mod tidy` and see how you go?

looprpc/client.proto
```diff
@@ -264,6 +264,8 @@ message LoopInRequest {
     triggering the swap (loop CLI, autolooper, LiT UI, other 3rd party UI).
     */
     string initiator = 8;
+
```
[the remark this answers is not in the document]
carlaKC: What if you're a user who wants to use the api but wants auto-populated route hints?

loopin.go
carlaKC: Do you have more than 1 hop between primary and client? Because I'm working with the following setup and I can't get a success:

```
alice --(private)-- gateway --- primary
```

I don't see how the primary can make a payment for a swap to alice without hop hints, eg even if I try: 
```
(alice) lncli addinvoice --amt=10 --private=false
(primary) lncli sendpayment --pay_req={alice pay req}
```
(here, I set `--private=false` because the cli defaults to true, but the direct api calls that loop makes won't have that default)

But that same invoice can be paid by alice's direct peer because it knows about the channel. This is def something we should have client side itests for :/ much easier to try things out with that structure.

utils.go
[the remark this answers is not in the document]
carlaKC: Have you pushed since these comments? Noticed a few godocs are still missing, might need a repush?

loopin.go
carlaKC: @arshbot any input here?

arshbot: Took me some time to figure out what was wrong (mainly due to another bug that got in the way), but it's because the current implementation calls listChannels with `publicOnly` set to true, which returns 0 channels from C in the case of a simple `A ---public--- B ---private--- C` setup.

For some reason I assumed this parameter returned public nodes only, but of course it returns public channels only. The alternative is to call all channels, but then somehow determine which peers are fully private (defined by only having private channels).

Looking at [how we currently do this in LND](https://github.com/lightningnetwork/lnd/blob/00af978bf24eb1ce7181c9e05f197211741284ba/channeldb/graph.go#L2494), it seems the logic is "if you have ANY channels with anyone that isn't us... you're public".

By that logic, it seems per node for a private channel I have, I can use `getnodeInfo` and check all listed channels to check if the peer has any channels with anyone that isn't me. 

Let me know if you have any suggestions on this current plan

looprpc/client.proto
```diff
@@ -580,6 +582,8 @@ message QuoteRequest {
     Optional route hints to reach the destination through private channels.
     */
     repeated RouteHint loop_in_route_hints = 6;
+
+    bool private = 7;
```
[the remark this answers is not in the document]
carlaKC: Still no comments.

loopin.go
carlaKC: Why can't we just use lnd's existing logic and provide it with private channels as input? Since you've added that parameter in lightninglabs/lndclient/pull/66. It's ok if this doesn't succeed for +0 depth private channels, we can't route to that anyway.

> mainly due to another bug that got in the way

Please will you open an issue with steps to reproduce + description of problem if there's a bug?

loopd/swapclient_server.go
```diff
@@ -483,21 +483,45 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
 	var lastHop *route.Vertex
-	if req.LoopInLastHop != nil {
+
+	switch {
+	// Private and routehints are mutually exclusive as setting private
+	// means we retrieve our own routehints from the connected node.
+	case haveRouteHints && req.Private:
+		return nil, fmt.Errorf("private and route_hints both set")
+
+	case haveLastHop:
```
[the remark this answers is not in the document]
arshbot: ah interesting, I thought this case statement would cascade since there isn't a break or return here, but it sounds like that isn't the case.

loopin.go
arshbot: Becausee that `privateOnly` flag doesn't apply to private channels! It actually means private nodes!! This is the same mistake I made :see_no_evil:

arshbot: > Ok, needed to try this out for myself to figure out whether it's possible.
>
> We can use the existing API to get hop hints rather than needing those internals: https://github.com/carlaKC/loop/tree/550-privatechannels. Nice thing here is that by looking up nodes in our public graph, a NotFound error tells us that they aren't public, and finding them tells us that they are, so we don't need to worry about a lot of the other internals.

I tried this out for myself, and I didn't come to the same conclusion, in my tests it appears that in the following graph:

```
A ---public--- B ---private--- C
```
When `B.getNodeInfo(C)` is run, it does return node info. However, because I had my doubts and started testing this out, I did discover that with slight modifications to your idea, we can achieve the same results! Because while the aforementioned command returns node info, it _doesn't_ return channel info as the channel is private. I played around with this extensively and it appears that `getNodeInfo`

-- does not return private channels (even channels that should exist in the node graph due to consumed hophints)
-- returns "private" nodes

therefore, if ANY channels are returned for a given node, we can assume that node is public. This strategy is much less complex than my previously mentioned strategy. Thank you for pointing me towards `getNodeInfo` more :)

arshbot: Current implementation does not leak private nodes, tested.

loopin.go
```diff
@@ -80,6 +81,16 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
 	currentHeight int32, request *LoopInRequest) (*loopInInitResult,
 	error) {
 
+	var err error
+	var routeHints [][]zpay32.HopHint
+	if request.Private {
+		routeHints, err = SelectHopHints(globalCtx, cfg.lnd,
+			request.Amount, 20)
```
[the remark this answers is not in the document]
arshbot: Done

utils.go
```diff
@@ -12,80 +11,44 @@ import (
...
-		edgePolicy, canBeHopHint := chanCanBeHopHint(openChannels, channel, chanInfo)
-		if edgePolicy == nil || !canBeHopHint {
+		// Cache the GetChanInfo result since it might be useful
+		hopHintChans[i] = chanInfo
+		if channel.RemoteBalance < amtMSat {
+			continue
+		}
+		// If includeNodes is set, we'll only add channels with peers in
+		// include_node. This is done to respect the last_hop parameter
+		if _, ok := includeNodes[channel.PubKeyBytes]; includeNodes == nil && len(includeNodes) != 0 && !ok {
```
carlaKC, line 56: I think that we can squash this with the original commit? Unrelated to performance enhancements.

looprpc/client.proto
```diff
@@ -269,6 +269,8 @@ message LoopInRequest {
     Optional route hints to reach the destination through private channels.
     */
     repeated RouteHint route_hints = 9;
+
+    bool private = 10;
```
carlaKC, line 278: LIkewise can swash + needs comment

cmd/loop/loopin.go
```diff
@@ -32,9 +32,16 @@ var (
 			"with our reserved prefix: %v.",
 			labels.MaxLength, labels.Reserved),
 	}
-	routeHints = cli.StringFlag{
-		Name:  "route_hints",
-		Usage: "",
+	routeHintsFlag = cli.StringSliceFlag{
```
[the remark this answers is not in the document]
arshbot: Done

arshbot (commented):
Squashed!

utils.go
arshbot: done

utils.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/lightningnetwork/lnd/zpay32"
+)
+
+var (
+	// Default MaxHopHints is set to 20 as that is the default set in LND
```
carlaKC: s/Default MaxHopHints/DefaultMaxHopHints

utils.go
```diff
@@ -0,0 +1,150 @@
...
+	DefaultMaxHopHints = 20
+)
+
+// SelectHopHints is a direct port of the SelectHopHints found in lnd. It
+// iterates through all the active and public channels available and returns
+// eligible channels. Eligibility requirements are simple: does the channel
+// have enough liquidity to fulfill the request and is the node whitelisted (if // specified)
```
carlaKC, line 22: (if // specified) on next line?

utils.go
```diff
@@ -0,0 +1,150 @@
...
+		chanInfo, err := lnd.Client.GetChanInfo(ctx, channel.ChannelID)
+		if err != nil {
+			return nil, err
+		}
+		// Cache the GetChanInfo result since it might be useful
+		hopHintChans[i] = chanInfo
+		if channel.RemoteBalance < amtMSat {
+			continue
+		}
+		// If includeNodes is set, we'll only add channels with peers in
```
carlaKC, line 54: Would simplify this to just always pass a non-nil `includeNodes` in here, or break the line: 
```
if includeNodes!=nil{
  if _, ok := includeNodes[channel.PubKeyBytes]; !ok{
    continue
  }
}
```

utils.go
```diff
@@ -0,0 +1,150 @@
...
+		if _, ok := includeNodes[channel.PubKeyBytes]; includeNodes == nil && len(includeNodes) != 0 && !ok {
+			continue
+		}
+
+		// We want to prevent leaking private nodes, which we define as
+		// nodes with only private channels
+		//
+		// GetNodeInfo will never return private channels, even if they're
+		// somehow known to us. If there are any channels returned, we can
+		// consider the node to be public.
```
carlaKC, line 65: Great comment!

utils.go
```diff
@@ -0,0 +1,150 @@
...
+		nodeChannels, err := lnd.Client.GetNodeInfo(
+			ctx, channel.PubKeyBytes, true,
+		)
+		if err != nil {
+			return nil, err
+		}
+		if len(nodeChannels.Channels) == 0 {
+			continue
```
carlaKC, line 73: nit: add a log here to explain why we skipped this node? Similar to how we log in lnd.

utils.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/lightningnetwork/lnd/zpay32"
+)
+
+var (
+	// Default MaxHopHints is set to 20 as that is the default set in LND
+	DefaultMaxHopHints = 20
+)
+
+// SelectHopHints is a direct port of the SelectHopHints found in lnd. It
```
carlaKC, line 19: Make a note of why we've changed this a bit? ie, as in in lnd it requires internals that we don't have access to.

cmd/loop/loopin.go
```diff
@@ -32,9 +32,16 @@ var (
...
+		Name: "route_hints",
+		Usage: "Route hints that can each be individually used " +
+			"to assist in reaching the invoice's destination.",
+	}
+	privateFlag = cli.BoolFlag{
+		Name: "private",
+		Usage: "Indicates the offchain destination is private, " +
+			"or only reachable via private channels. Route hints " +
+			"will be automatically included",
```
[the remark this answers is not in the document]
carlaKC: Comment still applies? I don't think end-users need to understand the concept of "private destination", just that this will add hop hints + expose their private chans

utils.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	// Default MaxHopHints is set to 20 as that is the default set in LND
+	// DefaultMaxHopHints is set to 20 as that is the default set in LND
```
carlaKC, line 15: 😬 fixup to wrong commit?

cmd/loop/loopin.go
```diff
@@ -68,6 +74,7 @@ var (
 			labelFlag,
 			verboseFlag,
 			routeHintsFlag,
+			privateFlag,
```
carlaKC, line 77: General style nit here, **not for this PR** but just generally in loop dev. 

I find that it makes sense to start with the changes _within_ loop, then fan them out. Eg in this case: 
- Add a `Private` bool to [LoopInRequest](https://github.com/lightninglabs/loop/blob/4c58d0c5e8eac5506e083caa1d84ad79d9b407bf/interface.go#L171) and add `SelectHopHints` if `Private` is true in the `Client`'s [LoopIn](https://github.com/lightninglabs/loop/blob/master/client.go#L508) method
- At this point, this logic is never triggered, since the `Private` bool is always false
- Next, update `client.proto` to add a `Private` bool to the [LoopInRequest](https://github.com/lightninglabs/loop/blob/master/looprpc/client.proto#L212) message and update `swapClientServer`'s [LoopIn](https://github.com/lightninglabs/loop/blob/4c58d0c5e8eac5506e083caa1d84ad79d9b407bf/loopd/swapclient_server.go#L598) method to set the value on the [LoopInRequest](https://github.com/lightninglabs/loop/blob/4c58d0c5e8eac5506e083caa1d84ad79d9b407bf/loopd/swapclient_server.go#L623) that you pass to the client
- At this point, the `loopd` API now supports the `Private` bool, although it is never set on the cli
- Finally, we surface this value on the cli, by adding a flag and setting it on the [LoopInRequest](https://github.com/lightninglabs/loop/blob/master/cmd/loop/loopin.go#L164) that the cli passes to `loopd`

This "fan out" means that once the flag is added on the cli, it's already implemented in the loop client. If you start with the cli, then checking out this commit actually has a meaningless flag, since the functionality hasn't been included yet.

loopd/swapclient_server.go
```diff
@@ -1111,7 +1111,7 @@ func validateLoopOutRequest(ctx context.Context, lnd lndclient.LightningClient,
 		return 0, err
 	}
 
-	channels, err := lnd.ListChannels(ctx)
+	channels, err := lnd.ListChannels(ctx, false, false)
```
carlaKC, line 1151: Do we need to bump lndclient version if we're always passing in false,false here?

cmd/loop/quote.go
```diff
@@ -53,9 +53,27 @@ func quoteIn(ctx *cli.Context) error {
...
+	if ctx.IsSet(routeHintsFlag.Name) {
+		if ctx.IsSet(privateFlag.Name) {
+			return fmt.Errorf("private and route_hints both set")
+		}
+
+		err = json.Unmarshal([]byte(ctx.String(routeHintsFlag.Name)), &hints)
+		if err != nil {
+			return fmt.Errorf("unable to parse json: %v", err)
+		}
+
```
carlaKC, line 72: unnecessary newline

loopin.go
```diff
@@ -80,6 +82,28 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
 	currentHeight int32, request *LoopInRequest) (*loopInInitResult,
 	error) {
 
+	var err error
+	var routeHints [][]zpay32.HopHint
```
carlaKC, line 86: Should we set this to `routeHints = request.RouteHints`? So that if we didn't set `Private` but did set `RouteHints` they'll be used? We've already validated that we don't have both set at this point, but seems we're not actually using `RouteHints`?

loopd/swapclient_server.go
```diff
@@ -483,21 +483,50 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
+		// If last_hop is set, we'll only add channels with peers
+		// set to the last_hop parameter
+		includeNodes := make(map[route.Vertex]struct{})
+		if haveLastHop {
+			includeNodes[lastHopVertex] = struct{}{}
+		}
+
+		// Because the Private flag is set, we'll generate our own
+		// set of hop hints
+		routeHints, err = loop.SelectHopHints(
```
carlaKC, line 518: Got a slightly different implementation here: 
* GetLoopInQuote: selects hop hints and passes them to `impl`
* LoopIn: passes a `Private` bool through to `impl`

Since `swapClientServer` is really just a wrapper for grpc stuff which passes through to our `impl`, I think it would be more consitent to add a `Private` bool to `loop.LoopInQuoteRequest` and pass it through like you do for `LoopIn`?

release_notes.md
```diff
@@ -16,6 +16,15 @@ This file tracks release notes for the loop client.
 
 #### New Features
 
+* `--private` flag is now available on the `loop in` command, meaning users can
```
carlaKC, line 19: Very useful release notes!

swap_server_client.go
```diff
@@ -214,8 +212,9 @@ func (s *grpcSwapServerClient) GetLoopInQuote(ctx context.Context,
 	routeHints [][]zpay32.HopHint) (*LoopInQuote, error) {
 
 	err := s.Probe(ctx, amt, pubKey, lastHop, routeHints)
-	if err != nil && status.Code(err) != codes.Unavailable {
+	if err != nil {
```
carlaKC, line 215: Can we address this server side as discussed?

carlaKC:
Needs rebase + linter failing.

utils.go
```diff
@@ -0,0 +1,160 @@
...
+		// We want to prevent leaking private nodes, which we define as
+		// nodes with only private channels
+		//
+		// GetNodeInfo will never return private channels, even if they're
+		// somehow known to us. If there are any channels returned, we can
+		// consider the node to be public.
+		nodeChannels, err := lnd.Client.GetNodeInfo(
+			ctx, channel.PubKeyBytes, true,
+		)
+		if err != nil {
```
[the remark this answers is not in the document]
carlaKC: > If we have a [private/nonprivate] channel with a peer, we'll have it in our graph

Are you certain this is true? Sometimes lnd's graph can lag a bit, so I think it's possible that we have a channel with a peer but they're not yet in our graph. Don't want the swap to error out just because we don't know about a channel yet, when we could just skip it?

utils.go
```diff
@@ -0,0 +1,160 @@
...
+		// isolation can't satisfy this payment.
+
+		// Retrieve extra info for each channel not available in listChannels
+		chanInfo, err := lnd.Client.GetChanInfo(ctx, channel.ChannelID)
+		if err != nil {
+			return nil, err
+		}
+		// Cache the GetChanInfo result since it might be useful
+		hopHintChans[i] = chanInfo
+		if channel.RemoteBalance < amtMSat {
```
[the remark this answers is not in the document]
carlaKC: I'd _probably_ go with `Debug`, since it may be a bit spammy for nodes with lots of channels. Non-blocking though, choice of logging level is pretty subjective.

cmd/loop/utils.go
```diff
@@ -0,0 +1,33 @@
...
+		}
+
+		stringHints := []byte(ctx.String(routeHintsFlag.Name))
+		err := json.Unmarshal(stringHints, &hints)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse json: %v", err)
+		}
+	}
+
+	return nil, nil
```
carlaKC, line 32: linter complaining here, we don't return hints?

carlaKC (approved):
One linter fix away from done! Grats on getting this beast through, it was *not* as simple as I thought it would be :')

I think that we should take a look at improvements we can make on the server side's probing logic/ potentially write an itest for this before merge?

utils.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+
+		// We want to prevent leaking private nodes, which we define as
+		// nodes with only private channels
+		//
+		// GetNodeInfo will never return private channels, even if
+		// they're somehow known to us. If there are any channels
+		// returned, we can consider the node to be public.
+		nodeInfo, err := lnd.Client.GetNodeInfo(
```
bhandras, line 79: nitty: This may potentially result in a lot of `GetNodeInfo` calls. How about just filtering the `openChannels` so it doesn't include channels to private nodes?

utils.go
```diff
@@ -0,0 +1,168 @@
...
+
+			break
+		}
+
+		// Skip the channel if we already selected it.
+		if _, ok := hopHintChans[i]; ok {
+			continue
+		}
+		channel := openChannels[i]
+		chanInfo := hopHintChans[i]
```
bhandras, line 141: This isn't working (see where we should add to `hopHintChans`). It's only containing the channel info because we add all channels to the `hopHintChans` (which is why we never reach this far).

utils.go
bhandras: Also just a thought: if a node only has private channels, we won't be able to loopin there. I'd say we also need to support that? wdyt @carlaKC ?

loopd/swapclient_server.go
```diff
@@ -621,6 +633,8 @@ func (s *swapClientServer) LoopIn(ctx context.Context,
 		ExternalHtlc:   in.ExternalHtlc,
 		Label:          in.Label,
 		Initiator:      in.Initiator,
+		Private:        in.Private,
```
bhandras, line 636: Shouldn't private and routehints be mutually exclusive here too?

client.go
```diff
@@ -563,6 +564,30 @@ func (s *Client) LoopInQuote(ctx context.Context,
 		return nil, ErrSwapAmountTooHigh
 	}
 
+	// Private and routehints are mutually exclusive as setting private
+	// means we retrieve our own routehints from the connected node.
+	if len(request.RouteHints) != 0 && request.Private {
```
bhandras, line 569: We want this for `Client.LoopIn` too right?

bhandras (changes requested):
Great job on this @arshbot! :tada:  I think the PR is in a fairly good shape, but there're a few issues that need fixing.

utils.go
carlaKC: The issue here is that we're trying to mimic lnd's "private node" behaviour, where if we're connected to a node by a private channel which has no other public channels, then we're nice and we don't dox them by adding a hop hint. In any case, adding that hop hint wouldn't be useful, because there are no channels to route to the private node and then to us. 

Agree on lots of `GetNodeInfo` calls, since we're probably checking our same set of peers over and over. Can probably add come kind of cache which lives for 10 mins (block time)? Not a disaster if it's a bit stale. Can probably be in a followup though, @bhandras?

bhandras: > The issue here is that we're trying to mimic lnd's "private node" behaviour, where if we're connected to a node by a private channel which has no other public channels, then we're nice and we don't dox them by adding a hop hint. In any case, adding that hop hint wouldn't be useful, because there are no channels to route to the private node and then to us.

Yeah, indeed!

> Agree on lots of GetNodeInfo calls, since we're probably checking our same set of peers over and over. Can probably add come kind of cache which lives for 10 mins (block time)? Not a disaster if it's a bit stale. 

Now that I thought this through again, I was wrong here. I meant to just use the `openChannels` but indeed we need to know if the peer is fully private so we need to query. I think the caching is not that important in this case since we'd only spare a few calls if we have multiple channels to fully private nodes.

client.go
arshbot: not `Client.LoopIn` exactly, but yes the other code flow that starts at `swapClientServer.LoopIn`. Done

cmd/loop/utils.go
arshbot: That's my bad. Fixed

loopd/swapclient_server.go
arshbot: Not at this point in the flow, we process `Private` later on in `newLoopInSwap`, and then enforce the rule

arshbot (commented):
Fixed the linter issues, updated the tests, and fixed the bugs @bhandras and @carlaKC pointed out!

utils.go
```diff
@@ -0,0 +1,193 @@
...
+			log.Debugf(
+				"Skipping ChannelID: %v for hints as "+
+					"remote balance (%v sats) "+
+					"insufficientappears to be private",
+				channel.ChannelID, channel.RemoteBalance,
+			)
+			continue
+		}
+		// If includeNodes is set, we'll only add channels with peers in
+		// include_node. This is done to respect the last_hop parameter
```
bhandras, line 76: nit: s/include_node/includeNodes.

utils.go
```diff
@@ -0,0 +1,193 @@
...
+		if len(includeNodes) > 0 {
+			if _, ok := includeNodes[channel.PubKeyBytes]; !ok {
+				continue
+			}
+		}
+
+		// Mark the index to skip so we can skip it on the next
+		// iteration if needed. We'll skip all channels that make
+		// it past this point as they'll likely belong to private
+		// nodes or be selected
```
bhandras, line 86: nit: missing . here and below in the next comment.

cmd/loop/utils.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/lightninglabs/loop/looprpc"
+	"github.com/urfave/cli"
+)
+
+// validateRouteHints ensures that the Private flag isn't set along with
+// the RouteHints flag. We don't allow both options to be set as these options
+// are alternatives to each other. Private autogenerates hopHints while
+// RouteHints are manually passed
```
bhandras, line 14: nit: missing . at the end.

test/lightning_client_mock.go
```diff
@@ -180,9 +181,54 @@ func (h *mockLightningClient) ListTransactions(
...
+					}
+				}
+			}
+		}
+	}
+
+	nodeInfo.ChannelCount = len(nodeInfo.Channels)
+
+	return &nodeInfo, nil
+
```
bhandras, line 216: nit: extra nl

test/lightning_client_mock.go
```diff
@@ -180,9 +181,54 @@ func (h *mockLightningClient) ListTransactions(
...
 }
 
+// GetNodeInfo retrieves info on the node, and if includeChannels is True,
+// will return other channels the node may have with other peers
+func (h *mockLightningClient) GetNodeInfo(ctx context.Context,
+	pubKeyBytes route.Vertex, includeChannels bool) (*lndclient.NodeInfo, error) {
+
+	nodeInfo := lndclient.NodeInfo{}
+
+	if includeChannels {
```
bhandras, line 191: nit: could return here if `!includechannels` to reduce nesting.

bhandras (approved):
utACK, great work @arshbot :zap: ! Will do some manual testing before we merge!

utils.go
```diff
@@ -0,0 +1,193 @@
...
+		// nodes with only private channels
+		//
+		// GetNodeInfo will never return private channels, even if
+		// they're somehow known to us. If there are any channels
+		// returned, we can consider the node to be public.
+		nodeInfo, err := lnd.Client.GetNodeInfo(
+			ctx, channel.PubKeyBytes, true,
+		)
+		if err != nil {
+			log.Warnf("Skipping ChannelID: %v for hints as peer "+
```
carlaKC, line 99: Rather than all errors, could check grpc error code:

```
status, ok := status.FromError(err)
if ok && status.Code == code.NotFound{
     log warning
}else if err!=nil{
   return err
}
```

This way we only log a warning if we specifically get a not found error, and will fail out for other errors.

utils.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// We'll add our hop hints in two passes, first we'll add all channels
+	// that are eligible to be hop hints, and also have a local balance
+	// above the payment amount.
+	var totalHintBandwidth btcutil.Amount
+
+	// chanInfoCache is a simple cache for any information we retrieve
+	// through GetChanInfo
+	chanInfoCache := make(map[int]*lndclient.ChannelEdge)
```
carlaKC, line 43: Generally we use short channel ID rather than index in maps like this, reads a bit easier IMO.

utils.go
```diff
@@ -0,0 +1,193 @@
...
+			if _, ok := includeNodes[channel.PubKeyBytes]; !ok {
+				continue
+			}
+		}
+
+		// Mark the index to skip so we can skip it on the next
+		// iteration if needed. We'll skip all channels that make
+		// it past this point as they'll likely belong to private
+		// nodes or be selected
+		skipCache[i] = struct{}{}
```
carlaKC, line 87: Why do we set this here rather than above where we actually add the hop hint?

utils.go
```diff
@@ -0,0 +1,193 @@
...
+			log.Infof(
+				"Skipping ChannelID: %v for hints as peer "+
+					"(NodeID: %v) appears to be private",
+				channel.ChannelID, channel.PubKeyBytes.String(),
+			)
+			continue
+		}
+
+		nodeID, err := btcec.ParsePubKey(
+			channel.PubKeyBytes[:], btcec.S256())
```
carlaKC, line 115: nit: spacing
```
nodeID, err := btcec.ParsePubKey(
     channel.PubKeyBytes[:], btcec.S256(),
)
```

client.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/lightninglabs/loop/loopdb"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightninglabs/loop/sweep"
+	"github.com/lightningnetwork/lnd/routing/route"
```
carlaKC, line 18: commit message: non-blocking, but generally when we touch multiple packages, can just prefix commit name with `multi:`, saves some time :)

looprpc/client.proto
```diff
@@ -580,6 +592,13 @@ message QuoteRequest {
     Optional route hints to reach the destination through private channels.
     */
     repeated RouteHint loop_in_route_hints = 6;
+
+    /*
+    Private indicates whether the destination node should be considered
+    private. In which case, loop will generate hophints to assist with
```
carlaKC, line 598: Still not super happy with the wording here. "destination node should be considered private" likely doesn't make a lot of sense to the end-user. "Whether to include hop hints" would read better imo.

looprpc/server.proto
```diff
@@ -206,9 +206,12 @@ message ServerLoopInResponse {
 }
 
 message ServerLoopInQuoteRequest {
-    /// The swap amount. If zero, a quote for a maximum amt swap will be given.
+    // The swap amount. If zero, a quote for a maximum amt swap will be given.
```
carlaKC, line 209: We copy this file across from the server (to keep in sync), so I wouldn't advise making aesthetic changes here unless they're done server-side as well. Can be pulled out into a separate PR if we want to clean things up?

utils_test.go
```diff
@@ -0,0 +1,293 @@
+package loop
```
carlaKC, line 1: nit: squash with commit that adds select hop hints?

carlaKC (approved):
Only nits left! Tested locally and works for very basic setup:
 ```
A (client) --- (private) ---- B ------ C (server)
```

File: utils_test.go (hunk 59 above)
Comment (arshbot) [reply utils_test.go:1]: interesting. i buy that

**Review by arshbot — COMMENTED** [review]
Addressed all(most all) nits!

File: utils_test.go [hunk 60]
```diff
@@ -0,0 +1,293 @@
...
+		Active:        true,
+		ChannelID:     chanID4.ToUint64(),
+		PubKeyBytes:   peer2,
+		LocalBalance:  10000,
+		RemoteBalance: 0,
+		Capacity:      chan4Capacity,
+	}
+	channelEdge4 = lndclient.ChannelEdge{
+		ChannelID:    chanID4.ToUint64(),
+		ChannelPoint: "6fe4408bba52c0a0ee15365e107105de" + "fabfc70c497556af69351c4cfbc167b:0",
```
carlaKC, line 182: why the ""+"" on the same line?

client.go
carlaKC: I usually cut it at 2 fwiw, but no hard rule

carlaKC (approved):
LGTM! 

Last step before merge would be to do some more expansive manual testing + post the outcomes here? I've done the bare minimum here (A --- private --- B ---- C).

bhandras (approved):
LGTM :rocket:

utils_test.go
arshbot: ah, my editor hit this from me bc of force wrap. fixed

carlaKC:
> Did some testing with some basic goals

Thanks for laying these out! Great to know it works as expected. 

> the tests do cover these scenarios as well

Do you mean they cover hop hint selection? 
fwiw, the intention of manual testing was more to test the whole flow than the workings of an individual function. Ideally we add an itest for this one day so that it's fully covered (not required for merge).

Just needs rebase then this can land!

carlaKC:
Can probably also remove WIP from the title now :')

carlaKC (approved):

PR #418, opened 2021-09-07
liquidity: add swap builder interface

carlaKC:
#### Pull Request Checklist


liquidity/loopout.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/lightninglabs/loop"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// Compile-time assertion that loopOutSwapSuggestion satisfies the
+// swapSuggestion interface.
+var _ swapSuggestion = (*loopOutSwapSuggestion)(nil)
+
+type loopOutSwapSuggestion struct {
```
arshbot, line 14: nit: comment the usecase for this struct, and all of it's functions

liquidity/loopout.go
```diff
@@ -0,0 +1,60 @@
...
+	channels := make([]lnwire.ShortChannelID, len(l.OutgoingChanSet))
+
+	for i, id := range l.OutgoingChanSet {
+		channels[i] = lnwire.NewShortChanIDFromInt(id)
+	}
+
+	return channels
+}
+
+// peers returns the set of peers that the loop out swap is restricted to.
```
arshbot, line 39: Why would the swap have restrictions?

liquidity/loopout.go
```diff
@@ -0,0 +1,60 @@
...
+func (l *loopOutSwapSuggestion) peers(
+	knownChans map[uint64]route.Vertex) []route.Vertex {
```
arshbot, line 41: is `knownChans` ever different from the output of `ListChannels`?

liquidity/loopout_builder.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// maySwap checks whether we can currently execute a swap, examining the
+// current on-chain fee conditions against relevant to our swap type against
+// our fee restrictions.
+//
+// For loop out, we check whether the fees required for our on-chain sweep
+// transaction exceed our fee limits.
+func (b *loopOutBuilder) maySwap(ctx context.Context, params Parameters) error {
+	estimate, err := b.cfg.Lnd.WalletKit.EstimateFee(
```
arshbot, line 43: huh, it is deeply unfortunate that `EstimateFee` doesn't take into account tx size, it would greatly increase the accuracy of the fee

liquidity/loopout_builder.go
```diff
@@ -0,0 +1,160 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	return params.FeeLimit.mayLoopOut(estimate)
+}
+
+// inUse examines our current swap traffic to determine whether we should
+// suggest the builder's type of swap for the peer and channels suggested.
```
arshbot, line 54: lost; what is the 'builder's type of swap' and what is the alternative for a loopout?

liquidity/loopout_builder.go
```diff
@@ -0,0 +1,160 @@
...
+		if recentFail {
+			log.Debugf("Channel: %v not eligible for suggestions, "+
+				"was part of a failed swap at: %v", chanID,
+				lastFail)
+
+			return newReasonError(ReasonFailureBackoff)
+		}
+
+		if traffic.ongoingLoopOut[chanID] {
+			log.Debugf("Channel: %v not eligible for suggestions, "+
```
arshbot, line 69: interesting, you can only have one loop out in a given channel? How does this work?

liquidity/loopout_builder.go
```diff
@@ -0,0 +1,160 @@
...
+
+		return newReasonError(ReasonLoopIn)
+	}
+
+	return nil
+}
+
+// buildSwap creates a swap for the target peer/channels provided. The autoloop
+// boolean indicates whether this swap will actually be executed, because there
+// are some calls we can leave out if this swap is just for a dry run.
```
arshbot, line 88: if that's the case, would `dryrun` be a better param name?

arshbot (commented):
Looks good! Need to review the upstream ticket/pr to understand the larger ramifications of this ticket

liquidity/loopout.go
carlaKC: Because people only want to target liquidity for a certain peer/ channel.

liquidity/loopout.go
carlaKC: nope, we just pass a map in here for ease of use

liquidity/liquidity.go
```diff
@@ -846,65 +844,24 @@ func (m *Manager) suggestSwap(ctx context.Context, traffic *swapTraffic,
...
+	// for our swap.
+	err := m.builder.inUse(traffic, balance.pubkey, balance.channels)
 	if err != nil {
 		return nil, err
 	}
 
-	// We can have nil suggestions in the case where no action is
-	// required, so we skip over them.
+	// Next, get the amount that we need to swap for this entity, skipping
+	// over it if no change in liquidity is required.
```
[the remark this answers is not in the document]
carlaKC: Happy it's not too verbose! :')

liquidity/loopout_builder.go
carlaKC: We can have more than one loop out, we just configure autoloop to be pretty conservative. If there are many loop outs going on in a channel at the same time it's more difficult for it to keep track of its own state, so we just limit it.

liquidity/loopout_builder.go
carlaKC: Dryrun would flip the meaning of the bool, and since we have it in other parts of the `liquidity` package as `autoloop` I'm going to leave it? Reasoning for `autoloop` is that it's opt-in, so made sense to have a bool that defaulted to false.

liquidity/loopout_builder.go
carlaKC: The "type of swap" refers to loop out or loop in. This comment is a bit forward looking to the time that we have a `loopInBuilder` (see followup PR). Will rephrase, clunky comment fosho.

liquidity/loopout_builder.go
carlaKC: I think it's ok in this case, we can always calculate the fee with known tx weight. Here we're looking at a parameter which says "I only want to autoloop if it's going to cost me less than x sat/byte". Easier way to frame it to the user since they may not have a good idea of the size of the swap tx.

liquidity/interface.go
```diff
@@ -31,6 +34,32 @@ type FeeLimit interface {
...
+	// should suggest the builder's type of swap for the peer and channels
+	// suggested.
+	inUse(traffic *swapTraffic, peer route.Vertex,
+		channels []lnwire.ShortChannelID) error
+
+	// buildSwap creates a swap for the target peer/channels provided. The
+	// autoloop boolean indicates whether this swap will actually be
+	// executed, because there are some calls we can leave out if this swap
+	// is just for a dry run.
+	buildSwap(ctx context.Context, peer route.Vertex,
```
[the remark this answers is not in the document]
carlaKC: yeah, see calling code. It's broken down into these steps so that we can make some lnd calls inbetween.

carlaKC:
thx for the review @arshbot! Great to have some fresh eyes on this, gives you a bit of perspective on some of you're junky comments :') 

GH seems to be double-posting all of my comments, so ignore any repeatrepeatrepeat responses 😬

bhandras (approved):
LGTM, interface looks good! :cake:

liquidity/loopout_builder.go
bhandras: It just returns a sats/KWeight value that can later on be used with the actual tx.

liquidity/loopout_builder.go
bhandras: Agree, though I think the returned suggestion is kind of a dry run depending on use. We could avoid generating a new wallet though. Maybe some more clarification in the godoc can help understanding what dry-run means in this context.

liquidity/loopout_builder.go
carlaKC: The end user actually specifies a fee rate limit here in sat/byte so this is actually what we want.

liquidity/loopout_builder.go
carlaKC: cri, I responded to all of these a while ago and GH didn't post them. 

This just means a loop out or loop in, see follow up PR for more context.

liquidity/loopout_builder.go
carlaKC: This is just a restriction to make autoloop more conservative, we don't want to be performing more autoloops if the channel is busy being rebalanced with another swap. Just simplifies life for us, in the regular case you can have many swaps per channel.

liquidity/interface.go
```diff
@@ -31,6 +34,32 @@ type FeeLimit interface {
...
 }
 
+// swapBuilder is an interface used to build our different swap types.
+type swapBuilder interface {
+	// swapType returns the swap type that the builder is responsible for
+	// creating.
+	swapType() swap.Type
+
+	// maySwap checks whether we can currently execute a swap, examining
+	// the current on-chain fee conditions against relevant to our swap
```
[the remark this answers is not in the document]
carlaKC: Had a short crisis thinking I didn't know how to spell "against" here till I realized it's out of place :')

liquidity/interface.go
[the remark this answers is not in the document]
carlaKC: Yeah, they're just generic steps in the process of deciding whether we can perform swaps. Separated out here to allow the main `SuggestSwaps` method to exit early if we don't want to autoloop rn.

carlaKC:
@arshbot PTAL

PR #419, opened 2021-09-07
liquidity: add loopin to autoloop

carlaKC:
This PR updates autoloop to add support for loop in swaps. 

#### Pull Request Checklist


carlaKC:
There are some tests needed here still, but wanted to get a first round of review before I fill everything out. 

Thinking of changing the style of test in the `liquidity` subsystem. In the past we've gone for black box tests, driving everything through `SuggestSwaps` rather than more traditional unit tests of each components. IMO that has made unit testing quite "heavy", because we have to set up channels/ parameters/ expected outcomes for each unit test. Would be interested in thoughts here.

carlaKC:
@arshbot @bhandras PTAL 🙏

carlaKC:
❌  hold off on review, want to push a few more changes!

carlaKC:
Rebased on dependent PR. I'm going to add a bunch more tests here to make sure there's no freaky interaction when we have autoloop-out and autoloop-in both enabled. 

Would appreciate a first pass on approach before I delve too deep @bhandras, @arshbot 🙏

looprpc/client.proto
```diff
@@ -829,6 +829,11 @@ message LiquidityParameters {
     specified by the LoopOutTerms endpoint.
     */
     uint64 max_swap_amount = 15;
+
+    /*
+    The confirmation target for loop in on-chain htlcs.
+    */
+    int32 htlc_conf_target = 17;
```
bhandras, line 836: nit: 16?

cmd/loop/liquidity.go
```diff
@@ -48,6 +48,14 @@ var setLiquidityRuleCommand = cli.Command{
 	Description: "Update or remove the liquidity rule for a channel/peer.",
 	ArgsUsage:   "{shortchanid |  peerpubkey}",
 	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name: "type",
+			Usage: "the type of swap to perform, set to 'out' for" +
```
bhandras, line 53: nit: needs space at the end

bhandras (commented):
Looks really good!

looprpc/client.proto
carlaKC: 16 is taken by `fee_ppm`, just higher up in the struct for nicer ordering on the cli

carlaKC:
Diff from last review [here](https://github.com/lightninglabs/loop/compare/81eaede1dfeaeb37af572b371a7151534c6cbd2b..f40d138c13188a28ecf78d2b04e2266ce62e4566), majority of change is more tests.

bhandras (approved):
LGTM, not even a nit :1st_place_medal:

liquidity/autoloop_testcontext_test.go
```diff
@@ -177,31 +217,70 @@ type loopOutRequestResp struct {
...
 	// Send a mocked response from the server with the swap size limits.
-	c.loopOutRestrictions <- NewRestrictions(minAmt, maxAmt)
+	c.loopOutRestrictions <- NewRestrictions(step.minAmt, step.maxAmt)
+	c.loopInRestrictions <- NewRestrictions(step.minAmt, step.maxAmt)
 
 	// Provide the liquidity manager with our desired existing set of swaps.
-	c.loopOuts <- existingOut
-	c.loopIns <- nil
+	c.loopOuts <- step.existingOut
+	c.loopIns <- step.existingIn
```
arshbot, line 263: ha, seems like implementing loop ins was always plart of the plan! That's always nice

liquidity/liquidity.go
```diff
@@ -544,6 +553,27 @@ func (m *Manager) autoloop(ctx context.Context) error {
...
+		// If we don't actually have dispatch of swaps enabled, log
+		// suggestions.
+		if !m.params.Autoloop {
+			log.Debugf("recommended autoloop in: %v sats over "+
+				"%v", in.Amount, in.LastHop)
+
+			continue
+		}
+
+		in := in
```
arshbot, line 566: What is this go trickery?

arshbot (changes requested):
Looks great! I left _one_ comment that I'm just not sure on, but other than that -- echo what @bhandras said :)

liquidity/autoloop_testcontext_test.go
carlaKC: yeah! Original intention was to try make this as "swap type" so that we can slot in loop in when we're ready :)

liquidity/liquidity.go
carlaKC: If we take the straight reference to `in` on 556, then the value will be updated next time we loop around the for loop (thus updating the value we sent in our `LoopIn` call). Declaring a var in scope of the loop ensures that we don't accidentally overwrite our own stuff. 

Difficult to explain, see goplay example: https://go.dev/play/p/heo4uvfawrc

arshbot: Makes sense, thank you for the explanation!

arshbot (approved):

PR #433, opened 2021-11-15
liquidity: add swap type and loop in fee logic

carlaKC:
This PR does some preliminary work for adding autoloopin support. Primarily adding a swap type to our liquidity rules, and implementing some of the fees + budgeting plumbing we need. 

#### Pull Request Checklist

### DIFF
--- a/liquidity/fees.go
+++ b/liquidity/fees.go
@@ -223,6 +223,27 @@ func (f *FeeCategoryLimit) loopOutLimits(amount btcutil.Amount,
 	return nil
 }
 
+func (f *FeeCategoryLimit) loopInLimits(amount btcutil.Amount,
+	quote *loop.LoopInQuote) error {
+
+	maxServerFee := ppmToSat(amount, f.MaximumSwapFeePPM)
+	if quote.SwapFee > maxServerFee {
+		log.Debugf("quoted swap fee: %v > maximum swap fee: %v",
+			quote.SwapFee, maxServerFee)
+
+		return newReasonError(ReasonSwapFee)
+	}
+
+	if quote.MinerFee > f.MaximumMinerFee {
+		log.Debugf("quoted miner fee: %v > maximum miner "+
+			"fee: %v", quote.MinerFee, f.MaximumMinerFee)
+
+		return newReasonError(ReasonMinerFee)
+	}
+
+	return nil
+}
+
 // loopOutFees returns the prepay and routing and miner fees we are willing to
 // pay for a loop out swap.
 func (f *FeeCategoryLimit) loopOutFees(amount btcutil.Amount,
@@ -383,4 +404,43 @@ func splitOffChain(available, prepayAmt,
 // scaleMinerFee scales our miner fee by our constant multiplier.
 func scaleMinerFee(estimate btcutil.Amount) btcutil.Amount {
 	return estimate * btcutil.Amount(minerMultiplier)
+}
+
+func (f *FeePortion) loopInLimits(amount btcutil.Amount,
+	quote *loop.LoopInQuote) error {
+
+	// Calculate the total amount that this swap may spend in fees, as a
+	// portion of the swap amount.
+	totalFeeSpend := ppmToSat(amount, f.PartsPerMillion)
+
+	// Check individual fee components so that we can give more specific
+	// feedback.
+	if quote.MinerFee > totalFeeSpend {
+		log.Debugf("miner fee: %v greater than fee limit: %v, at "+
+			"%v ppm", quote.MinerFee, totalFeeSpend,
+			f.PartsPerMillion)
+
+		return newReasonError(ReasonMinerFee)
+	}
+
+	if quote.SwapFee > totalFeeSpend {
+		log.Debugf("swap fee: %v greater than fee limit: %v, at "+
+			"%v ppm", quote.SwapFee, totalFeeSpend,
+			f.PartsPerMillion)
+
+		return newReasonError(ReasonSwapFee)
+	}
+
+	fees := worstCaseInFees(
+		quote.MinerFee, quote.SwapFee, defaultLoopInSweepFee,
+	)
+
+	if fees > totalFeeSpend {
+		log.Debugf("total fees for swap: %v > fee limit: %v, at "+
+			"%v ppm", fees, totalFeeSpend, f.PartsPerMillion)
+
+		return newReasonError(ReasonFeePPMInsufficient)
+	}
+
+	return nil
 }
--- a/liquidity/liquidity.go
+++ b/liquidity/liquidity.go
@@ -97,8 +97,8 @@ var (
 	defaultParameters = Parameters{
 		AutoFeeBudget:   defaultBudget,
 		MaxAutoInFlight: defaultMaxInFlight,
-		ChannelRules:    make(map[lnwire.ShortChannelID]*ThresholdRule),
-		PeerRules:       make(map[route.Vertex]*ThresholdRule),
+		ChannelRules:    make(map[lnwire.ShortChannelID]*SwapRule),
+		PeerRules:       make(map[route.Vertex]*SwapRule),
 		FailureBackOff:  defaultFailureBackoff,
 		SweepConfTarget: defaultConfTarget,
 		FeeLimit:        defaultFeePortion(),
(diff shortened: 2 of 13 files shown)


carlaKC:
Split up #419 into 2 parts for easier review. Follow up PR is up (just one unit test needs fixing) for reference of how this will be used. 

I intend to merge them at the same time, since this doesn't really add much functionality, just easier to split early IMO.

loopd/swapclient_server.go
```diff
@@ -752,13 +752,20 @@ func (s *swapClientServer) GetLiquidityParams(_ context.Context,
...
+	rpcRule := &looprpc.LiquidityRule{
 		ChannelId:         channelID,
 		Pubkey:            peer,
 		Type:              looprpc.LiquidityRuleType_THRESHOLD,
 		IncomingThreshold: uint32(rule.MinimumIncoming),
 		OutgoingThreshold: uint32(rule.MinimumOutgoing),
+		SwapType:          looprpc.SwapType_LOOP_OUT,
 	}
+
+	if rule.Type == swap.TypeIn {
```
arshbot, line 764: `swap.TypeIn` == 0, but `looprpc.SwapType_LOOP_OUT` == 1

Is there any value in making these match, and simplifying the logic here so we can cast instead of having if else statements?

liquidity/loopin.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return successFee
+}
+
+// loopInSweepFee provides an estimated fee for our sweep transaction, based
+// on the fee rate provided. We can calculate our fees for htlcv2 and p2wkh
+// timeout addresses because automated loop ins will be handled entirely by the
+// client, so we know what types will be used.
+func loopInSweepFee(fee chainfee.SatPerKWeight) btcutil.Amount {
```
arshbot, line 67: context q: does autoloop sweep for the user on a loopin? a sweep in?

liquidity/fees.go
```diff
@@ -384,3 +405,38 @@ func splitOffChain(available, prepayAmt,
 func scaleMinerFee(estimate btcutil.Amount) btcutil.Amount {
 	return estimate * btcutil.Amount(minerMultiplier)
 }
+
+func (f *FeePortion) loopInLimits(amount btcutil.Amount,
+	quote *loop.LoopInQuote) error {
+
+	feeLimit := ppmToSat(amount, f.PartsPerMillion)
```
arshbot, line 412: nit: more documentation; what defines this limit?

liquidity/fees.go
```diff
@@ -223,6 +223,27 @@ func (f *FeeCategoryLimit) loopOutLimits(amount btcutil.Amount,
 	return nil
 }
 
+func (f *FeeCategoryLimit) loopInLimits(amount btcutil.Amount,
+	quote *loop.LoopInQuote) error {
+
+	maxFee := ppmToSat(amount, f.MaximumSwapFeePPM)
```
arshbot, line 229: nit: difference between `maxFee := ppmToSat(amount, f.MaximumSwapFeePPM)` and  `feeLimit := ppmToSat(amount, f.PartsPerMillion)`. They both kinda seem like different ways to say the same thing

arshbot (changes requested):
A necessary PR, love the inclusion of loop in's to the accounting stuff, that'll be useful to a lot of people. Left some minor nits and context qs

liquidity/loopin.go
carlaKC: If the loop in fails, the client needs to sweep the htlc back to its wallet on-chain. This is part of a regular failed loop-in, autoloop just needs to be aware that we might face that cost.

liquidity/fees.go
carlaKC: They are, we either allow people to set a `feeLimit` which is just a total % of the swap amount that they're willing to pay in fees, or break it down into individual fee categories (one of those being the `maxFee` that the server charges). 

These are two different implementations of the same interface.

liquidity/fees.go
carlaKC: comment added!

carlaKC:
@bhandras PTAL!

liquidity/liquidity.go
```diff
@@ -841,12 +823,34 @@ func (m *Manager) SuggestSwaps(ctx context.Context, autoloop bool) (
...
 
+	var (
+		builder      swapBuilder
+		restrictions *Restrictions
+	)
+
+	switch rule.Type {
+	case swap.TypeOut:
+		builder = newLoopOutBuilder(m.cfg)
+		restrictions = outRestrictions
```
bhandras, line 843: nit: I'd keep the incoming `outRestrictions` arg with it's old name (`restrictions`) instead.

liquidity/fees.go
```diff
@@ -384,3 +405,40 @@ func splitOffChain(available, prepayAmt,
...
+func (f *FeePortion) loopInLimits(amount btcutil.Amount,
+	quote *loop.LoopInQuote) error {
+
+	// Calculate the total amount that this swap may spend in fees, as a
+	// portion of the swap amount.
+	feeLimit := ppmToSat(amount, f.PartsPerMillion)
+
+	// Check individual fee components so that we can give more specific
+	// feedback.
+	if quote.MinerFee >= feeLimit {
```
bhandras, line 418: nit: why >= ?

bhandras (approved):
LGTM :100:

liquidity/threshold_rule.go
```diff
@@ -65,73 +65,88 @@ func (r *ThresholdRule) validate() error {
...
-	outgoingThresholdPercent int) btcutil.Amount {
-
-	minimumIncoming := btcutil.Amount(uint64(
-		balances.capacity) *
-		uint64(incomingThresholdPercent) / 100,
+// calculateSwapAmount calculates amount for a swap based on thresholds.
+// This function can be used for loop out or loop in, but the concept is the
+// same - we want liquidity in one (target) direction, while preserving some
+// minimum in the other (reserve) direction.
+// * target: this is the side of the channel(s) where we want to acquire some
```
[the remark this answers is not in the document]
carlaKC: How does the existing comment for the block need to be expanded?
```
	// For loop out, we want to acquire liquidity in the incoming
	// direction, so this is our target, while preserving some outgoing
	// balance, which is our reserve.
```

liquidity/threshold_rule.go
```diff
@@ -19,6 +20,12 @@ var (
 		"percentages must be < 100")
 )
 
+// TypedRule is a liquidity rule with a specific swap type.
+type TypedRule struct {
```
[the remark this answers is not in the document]
carlaKC: 🙏 thank you for releasing me from naming purgatory.

loopd/swapclient_server.go
carlaKC: Would be nice (why oh why do they not match 😭 ), but I think both need migrations? Either db migration or rpc enum (since rpc relies on numerical value), I think it's too messy to justify the small win.

liquidity/fees.go
carlaKC: Reasoning was that if we use the whole fee limit for our miner fee, we won't have anything left for our swap fee. 

That doesn't really come across very clearly, and these checks are mostly aesthetic, so will change to just `>`, since we have the full amount check later anyway

bhandras: Yeah I thought the same just wasn't sure. Thanks for explaining :+1:

loopd/swapclient_server.go
arshbot: db migrations are indeed the big sad, was afraid it'd be something like that

liquidity/threshold_rule.go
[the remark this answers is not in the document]
arshbot: Oh my bad, I meant we should have this kind of variable level description for each variable like so

```
                // targetBalance represents the amount we want on the side of the channel
                // where we want to acquire some liquidity.
		targetBalance     = channel.incoming

                // targetPercentage is the liquidity threshold we aim for on the side of the
                // channel where we want to acquire some liquidity.
		targetPercentage  = uint64(r.MinimumIncoming)

                // reserveBalance is the value that does the thing in the opposite direction.
		reserveBalance    = channel.outgoing

                // reservePercentage is like targetPercentage but in the opposite direction.
		reservePercentage = uint64(r.MinimumOutgoing)
```

basically, per variable comments because it wasn't immediately clear what each one was for

liquidity/fees.go
arshbot: ah gotcha! an absolute vs relative limit. hmm would it make sense to replace `feeLimit` with `percentageLimit` or something that clarifies the this distinction?

carlaKC: done

liquidity/threshold_rule.go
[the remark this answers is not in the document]
carlaKC: done

arshbot (approved):
LGTM!!

PR #436, opened 2021-11-16
main: add --force to loop in/loop out

arshbot:
--allow-yes skips the confirmation prompt when performing loops. This is
super userful for scripts and negates the use of gnu yes

Fixes: 435

### DIFF
--- a/cmd/loop/loopout.go
+++ b/cmd/loop/loopout.go
@@ -73,6 +73,7 @@ var loopOutCommand = cli.Command{
 				"Not setting this flag therefore might " +
 				"result in a lower swap fee.",
 		},
+		forceFlag,
 		labelFlag,
 		verboseFlag,
 	},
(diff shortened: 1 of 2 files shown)


cmd/loop/loopout.go
```diff
@@ -73,6 +73,10 @@ var loopOutCommand = cli.Command{
 				"Not setting this flag therefore might " +
 				"result in a lower swap fee.",
 		},
+		cli.BoolFlag{
+			Name:  "force, f",
+			Usage: "Assumes yes during confirmation. Using this " + "option will result in an immediate loop in",
```
carlaKC, line 78: s/"immediate loop in"/ "immediate loop out"

Could also just say "immediate swap", declare the flag as a var then reuse across commands.

carlaKC (approved):
LGTM, just a copy-fu nit on the flag description

cmd/loop/loopout.go
arshbot: ah because they're in the same package. Was looking for a way to do that but forgot in go it's by package. Done!

carlaKC (approved):
⏩

yyforyongyu (approved):
LGTM⚡️

PR #439, opened 2021-11-29
loopout: using server recommended routing plugins + low/high router

bhandras:
#### Pull Request Checklist


routing_plugin.go
```diff
@@ -104,3 +115,322 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	// Walk until the first fork (if there's any). This first
+	// fork will be where we're going to try to manipulate success
+	// probabilities to increasingly prefer more expensive edges.
+
+	// We track all visited peers, so we won't end up walking
+	// back and forth on graphs that have no forks.
+	visited := make(map[route.Vertex]struct{})
+	for {
+		targetNodeInfo, err := r.lnd.Client.GetNodeInfo(
+			ctx, target, true,
```
carlaKC, line 163: Is target the primary? Are we going to be able to find it in the graph if it's all private channels?

carlaKC: Ah nvm, we have hop hints so it should be there, obviously 🤦‍♀️

routing_plugin.go
```diff
@@ -0,0 +1,106 @@
...
+// able to acquire a RoutingPlugin instance. If someone is already holding the
+// instance a nil is returned.
+func AcquireRoutingPlugin(ctx context.Context, pluginType RoutingPluginType,
+	lnd lndclient.LndServices, target route.Vertex, amt btcutil.Amount) (
+	RoutingPlugin, error) {
+
+	routingPluginMx.Lock()
+	defer routingPluginMx.Unlock()
+
+	// Another swap is already using the routing plugin.
```
carlaKC, line 72: Wondering whether we'll want this for all routing plugins will want to be exclusively for a single swap at a time, or that's pretty specific to the high-to-low strategy?

carlaKC (commented):
Approach looks good! Just one q about whether we want to limit to using a plugin for a single swap at a time for _all_ plugins or just this one?

loopout.go
```diff
@@ -537,16 +542,28 @@ func (s *loopOutSwap) payInvoices(ctx context.Context) {
...
+
+	// Use the recommended routing plugin.
 	s.swapPaymentChan = s.payInvoice(
 		ctx, s.SwapInvoice, s.MaxSwapRoutingFee,
-		s.LoopOutContract.OutgoingChanSet,
+		s.LoopOutContract.OutgoingChanSet, plugin,
 	)
 
-	// Pay the prepay invoice.
+	// Pay the prepay invoice. Won't use the routing plugin here.
```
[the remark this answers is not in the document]
bhandras: done

routing_plugin.go
bhandras: Realized that indeed the original solution wouldn't work with private channels. With the latest version edges from the hop hints in the invoice are used too.

routing_plugin.go
bhandras: Good question. I think it's worth having just one active routing plugin since as long as it manipulates the MC.

routing_plugin.go
```diff
@@ -104,3 +120,419 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+func (r *lowToHighRoutingPlugin) Done(ctx context.Context) error {
+	if r.mcState == nil {
+		return nil
+	}
+
+	var err error
+	if len(r.mcState) > 0 {
+		now := r.clock.Now()
+		// Roll the entry times forward (to be able to override recent
+		// updates). TODO: maybe we should reset here first instead.
```
carlaKC, line 512: (re TODO): I think that resetting first might be a bit risky, because an inconvenient restart could completely wipe MC (and we lose data).

routing_plugin.go
```diff
@@ -104,3 +120,419 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+		for i := range r.mcState {
+			if !r.mcState[i].FailTime.IsZero() {
+				r.mcState[i].FailTime = now
+			}
+
+			if !r.mcState[i].SuccessTime.IsZero() {
```
carlaKC, line 518: Do we need to roll our whole mission control state forward here? Could be a less invasive approach to keep track of the pairs that we've modified and only reset them to their previous state (+ rolled forward time).

Motivation here is that unaffected results will still be "cooled down" based on their timestamp, rather than being brought forward to the present (which could make routing worse even? although idk)

bhandras: Yes thought about this myself too. Actually waited for some feedback from reviewers before refining what we store and how we restore.

routing_plugin.go
bhandras: Yes good point.

client.go
```diff
@@ -107,6 +107,14 @@ type ClientConfig struct {
 	// for a loop out swap. When greater than one, a multi-part payment may
 	// be attempted.
 	LoopOutMaxParts uint32
+
+	// TotalPaymentTimeout is the total amount of time until we time out
+	// off-chain payments (used in loop out).
+	TotalPaymentTimeout time.Duration
```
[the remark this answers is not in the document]
bhandras: I'd imagine most users wouldn't want to tune these things but of course open to making it per swap configurable.

loopout.go
```diff
@@ -638,50 +664,114 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
-		case payState := <-payStatusChan:
-			s.log.Infof("Payment %v: %v", hash, payState)
-
-			switch payState.State {
-			case lnrpc.Payment_SUCCEEDED:
-				return &payState, nil
-
-			case lnrpc.Payment_FAILED:
-				return &payState, nil
+	var (
```
yyforyongyu, line 673: Could we pull out the following logic into a function? Also makes testing easier.

swap_server_client.go
```diff
@@ -611,6 +645,60 @@ func (s *grpcSwapServerClient) CancelLoopOutSwap(ctx context.Context,
...
+	req := &looprpc.ReportRoutingResultReq{
+		ProtocolVersion: loopdb.CurrentRPCProtocolVersion,
+		SwapHash:        swapHash[:],
+		Plugin:          rpcRoutingPlugin,
+		Success:         success,
+		Attempts:        attempts,
+		TotalTime:       totalTime,
+	}
+
+	_, err := s.server.ReportRoutingResult(ctx, req)
```
yyforyongyu, line 698: maybe use a timeout context here in case the call hangs?

routing_plugin.go
```diff
@@ -0,0 +1,106 @@
...
+// RoutingPlugin is a generic interface for off-chain payment helpers.
+type RoutingPlugin interface {
+	// Init initializes the routing plugin.
+	Init(ctx context.Context, target route.Vertex, amt btcutil.Amount) error
+
+	// Done deinitializes the routing plugin (restoring any state the
+	// plugin might have changed).
+	Done(ctx context.Context) error
+
+	// BeforePayment is is called before each payment. Attempt counter
```
yyforyongyu, line 46: typo: `is is` -> `is`.

routing_plugin.go
```diff
@@ -104,3 +120,419 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	}
+	r.mcState = entries
+	log.Debugf("Saved MC state: %v", spew.Sdump(r.mcState))
+
+	// maxFeePerNode assigns the maximum fees that would be paid through
+	// selected nodes to the target node for the target amount.
+	maxFeePerNode := make(map[route.Vertex]int64)
+	totalCapacityPerNode := make(map[route.Vertex]btcutil.Amount)
+
+	amtMsat := int64(amt * 1000)
```
yyforyongyu, line 353: nit: can make into its own function

routing_plugin.go
```diff
@@ -104,3 +120,419 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+				fee:      fee,
+			},
+		)
+	}
+
+	sort.Slice(r.nodesByMaxFee, func(i, j int) bool {
+		return r.nodesByMaxFee[i].fee < r.nodesByMaxFee[j].fee
+	})
+
+	if log.Level() == btclog.LevelTrace {
```
yyforyongyu, line 402: Why this check? To avoid going through the slice below?

routing_plugin.go
```diff
@@ -104,3 +120,419 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+		}
+
+		// If there are no more peers to visit then we can't use
+		// this routing plugin.
+		if len(peers) == 0 {
+			return ErrRoutingPluginNotApplicable
+		}
+
+		// Found the first fork to our target.
+		break
```
yyforyongyu, line 301: So I guess the idea here is to find the first node which has multiple channels, which is named a "fork" here, and try all its channels?

routing_plugin.go
```diff
@@ -104,3 +120,419 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+
+	log.Debugf("Routing plugin target: '%v' %v", nodeInfo.Alias,
+		nodeInfo.PubKey.String())
+
+	// Gather node info (including channels) for the nodes we're
+	// interested in.
+	targetChanged := exit != target
+	nodes := make(map[route.Vertex]*lndclient.NodeInfo)
+	for _, edge := range nodeInfo.Channels {
+		// Skip edges to the exit node.
```
yyforyongyu, line 317: Not sure that I understand this, could you add more docs to explain why we continue?

yyforyongyu (commented):
Still reviewing the graph algorithm, nice approach so far! And what's the plan on itest? Guess it'd be challenging to write.

client.go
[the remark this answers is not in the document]
carlaKC: Yep, perhaps a one day feature!

loopout.go
```diff
@@ -53,9 +53,14 @@ var (
...
-	// paymentTimeout is the timeout for the loop out payment loop as
+	// totalPaymentTimeout is the timeout for the loop out payment loop as
 	// communicated to lnd.
-	paymentTimeout = time.Minute * 30
+	totalPaymentTimeout = time.Minute * 30
+
+	// maxPaymentRetries is the maximum number of times the client will
+	// attempt to pay the invoice before failing the swap. This retry limit
+	// only applies to when the client uses a routing helper plugin.
+	maxPaymentRetries = 6
```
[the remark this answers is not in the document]
bhandras: It's meant to timeout after 5 minutes after each attempt.

loopout.go
```diff
@@ -616,16 +633,25 @@ func (s *loopOutSwap) payInvoice(ctx context.Context, invoice string,
...
 	if err != nil {
 		return nil, err
 	}
 
+	maxRetries := 1
+	if pluginType != RoutingPluginNone {
+		maxRetries = maxPaymentRetries
+	}
+
+	paymentTimeout := totalPaymentTimeout / time.Duration(maxRetries)
```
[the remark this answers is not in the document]
bhandras: Exactly.

loopout.go
```diff
@@ -638,50 +664,114 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
+	}
 
-			case lnrpc.Payment_IN_FLIGHT:
-				// Continue waiting for final state.
+	// canRetry returns true if the payment timed out and the
+	// we haven't attempted the payment too many times.
+	canRetry := func(tryCount int, status lndclient.PaymentStatus) bool {
+		return (tryCount <= maxRetries &&
+			status.FailureReason ==
+				lnrpc.PaymentFailureReason_FAILURE_REASON_TIMEOUT)
```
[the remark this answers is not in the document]
bhandras: Yes, PTAL at the itests (added you to the PR).

executor.go
```diff
@@ -141,12 +145,14 @@ func (s *executor) run(mainCtx context.Context,
...
-					timerFactory:    s.executorConfig.createExpiryTimer,
-					loopOutMaxParts: s.executorConfig.loopOutMaxParts,
-					cancelSwap:      s.executorConfig.cancelSwap,
+					statusChan:         statusChan,
+					sweeper:            s.sweeper,
+					blockEpochChan:     queue.ChanOut(),
+					timerFactory:       s.executorConfig.createExpiryTimer,
+					loopOutMaxParts:    s.executorConfig.loopOutMaxParts,
+					totalPaymentTimout: s.executorConfig.totalPaymentTimeout,
+					maxPaymentRetries:  s.executorConfig.maxPaymentRetries,
```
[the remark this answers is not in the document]
bhandras: done

loopout.go
```diff
@@ -647,11 +640,13 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
 	}
 
 	maxRetries := 1
+	paymentTimeout := s.executeConfig.totalPaymentTimout
```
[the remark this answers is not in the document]
bhandras: Total payment timeout is now maximized in 1 hour (30 minutes by default).

routing_plugin.go
bhandras: Yes exactly. We walk back from the invoice target until we find the first node with channels to multiple peers.

routing_plugin.go
bhandras: PTAL

routing_plugin.go
bhandras: Removed, yeah just to not iterate the slice but it's unimportant really.

routing_plugin.go
bhandras: Not sure which block you mean: the conversion to msat or the rest below?

bhandras (commented):
Thank you both for the thorough review. With the server component and the itest it should be more clear now hopefully.

looprpc/server.proto
```diff
@@ -454,3 +464,45 @@ message ServerProbeRequest {
...
+
+    // The hash of the swap requesting a routing plugin.
+    bytes swap_hash = 2;
+}
+
+enum RoutingPlugin {
+    /// Client won't use any plugins to help with payment routing.
+    NONE = 0;
+
+    /// Client to try increasingly more expensive routes for off-chain payments.
```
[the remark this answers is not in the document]
bhandras: done

looprpc/server.proto
```diff
@@ -454,3 +464,45 @@ message ServerProbeRequest {
...
+
+message RecommendRoutingPluginReq {
+    ProtocolVersion protocol_version = 1;
+
+    // The hash of the swap requesting a routing plugin.
+    bytes swap_hash = 2;
+}
+
+enum RoutingPlugin {
+    /// Client won't use any plugins to help with payment routing.
```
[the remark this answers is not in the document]
bhandras: done

bhandras:
I don't know why github duplicates some of my comments... mistery

routing_plugin.go
```diff
@@ -0,0 +1,106 @@
...
+	routingPluginMx.Lock()
+	defer routingPluginMx.Unlock()
+
+	// Another swap is already using the routing plugin.
+	if routingPluginInstance != nil {
+		return nil, nil
+	}
+
+	instance := makeRoutingPlugin(pluginType, lnd)
+	if instance == nil {
```
carlaKC, line 83: When would `instance !=nil` and we hit `ErrRoutingPluginNotApplicable` below? Surely if it's not available we'd get a nil `instance` from `makeRoutingPlugin`?

routing_plugin.go
```diff
@@ -29,7 +36,8 @@ var (
 // RoutingPlugin is a generic interface for off-chain payment helpers.
 type RoutingPlugin interface {
 	// Init initializes the routing plugin.
-	Init(ctx context.Context, target route.Vertex, amt btcutil.Amount) error
+	Init(ctx context.Context, target route.Vertex,
```
carlaKC, line 39: nit: move to original commit?

routing_plugin.go
```diff
@@ -104,3 +120,433 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	if err != nil {
+		// It's still possible that we should know this node from
+		// the provided route hints even though it is not part of the
+		// public graph. If we don't know any private channels then
+		// just return the error.
+		if _, ok := privateEdges[nodeID]; !ok {
+			return nil, err
+		}
+
+		err = nil
```
carlaKC, line 235: This is a bit of an odd way to handle this to me. Could we rather match on `codes.NotFound` + check private edges else return error regardless?

routing_plugin.go
```diff
@@ -104,3 +120,433 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+		nodeInfo.PubKey.String())
+
+	// Gather node info (including channels) for the nodes we're
+	// interested in.
+	targetChanged := exit != target
+	nodes := make(map[route.Vertex]*lndclient.NodeInfo)
+	for _, edge := range nodeInfo.Channels {
+		// Skip edges to the exit node since from there the route to
+		// the invoice target is always constructed from the same hops.
+		if targetChanged && (edge.Node1 == exit ||
```
carlaKC, line 334: Why do we need `targetChanged` in here? Wouldn't we always skip edges to the exit node?

routing_plugin.go
```diff
@@ -104,3 +120,433 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+			edge.Node2 == exit) {
+			continue
+		}
+
+		var node route.Vertex
```
carlaKC, line 339: nit: `node` -> `peer` here?

routing_plugin.go
```diff
@@ -104,3 +120,433 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+
+	queryRoutesReq := lndclient.QueryRoutesRequest{
+		Source:            &r.lnd.NodePubkey,
+		PubKey:            r.target,
+		AmtMsat:           lnwire.MilliSatoshi(r.amount * 1000),
+		FeeLimitMsat:      lnwire.MilliSatoshi(r.amount * 1000),
+		UseMissionControl: true,
+	}
+
+	if log.Level() == btclog.LevelTrace {
```
carlaKC, line 441: Why do we need to check log level? Shouldn't `log.Tracef` only log if we're in trace?

routing_plugin.go
```diff
@@ -104,3 +120,433 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	}
+
+	// Do not do anything unless we tried to route the payment at least
+	// once.
+	if currAttempt < 2 {
+		return nil
+	}
+
+	// Calculate the limit until we'll disable edges. Anything above this
+	// limit should be encouraged.
```
carlaKC, line 454: Expand comment to explain how we're scaling this a bit? I had to manually calculate a few values here to understand how we're stepping the limit up per attempt.

routing_plugin.go
carlaKC: We've got our full mission control state in `r.mcstate`, yes? 

Since the idea here was to minimally disrupt mission control, I think that a good reset option would be to track the edges that we actually modified in `BeforePayment` and only reset those by rolling them forward.

carlaKC (commented):
Looking good! I _finally_ grok high to low :')

Two major comments are the resetting of MC + providing payment address with requests to limit use of these APIs to swap initiators (I know it's a haul to add, but if we want to enforce it rather do it now then later when we'll have to add a new protocol version because old clients don't set the field).

loopout.go
```diff
@@ -638,50 +664,116 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
+	canRetry := func(tryCount int, status lndclient.PaymentStatus) bool {
+		return (tryCount <= maxRetries &&
+			status.FailureReason ==
+				lnrpc.PaymentFailureReason_FAILURE_REASON_TIMEOUT)
+	}
 
-			default:
-				return nil, errors.New("unknown payment state")
-			}
+	start = time.Now()
```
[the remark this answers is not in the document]
alexbosworth: Updating to make it slower and doing 3x 20 sounds ok to me.

There are a lot of variables so I think the benefit of extending pathfinding time is to make it be at least no worse than the existing pathfinding at finding a route since the new heuristics would only start working as a last resort.

routing_plugin_test.go
```diff
@@ -0,0 +1,537 @@
...
+			)
+			require.NotNil(t, plugin)
+
+			// Set start state for MC.
+			mockLnd.MissionControlState = tc.missionControlState[0]
+
+			// Initialize the routing plugin.
+			require.Equal(
+				t, tc.initError,
+				plugin.Init(
```
yyforyongyu, line 420: Do you think it's possible to write tests for the method `Init` only? The current test looks good, though it's more like a functional test than a unit test. Since the `Init` method has quite some complexity in it, maybe dedicated unit tests could help others with understanding the algorithm.

loopout.go
yyforyongyu: Cool. Yeah usually I'd prefer small functions, sort like sticking to the SOLID principle. This type of decision is hard though, as it quickly becomes vague about what's the industry's best practice vs what's a personal taste😂 My take is that it's better to orginize code to reduce cognitive load. For instance the `StatusLoop` below has a simple task, to track the payment state and report the result, seems can be abstracted into a function like `processPayStatus`? Also I see we have two important logical branches here, that is whether `routingPlugin` is nil or not, which determines whether we want to have a `maxRetries` of 1 or more. It probably could be better if we could organize the branches like,

```
if routingPlugin == nil {
	sendPaymentOnce()
} else {
	sendPaymentMultipleTimes()
}
```
But in the end, what do I know. Since I'm not the author and do not know the details as comprehensively, it's up to you to make the decisions:)

yyforyongyu (changes requested):
Nice work on the graph traversal algo and the tests look good! Still got some questions re the algo, will check the itest to help myself understand more!

routing_plugin_test.go
```diff
@@ -0,0 +1,537 @@
...
+	privFrank, _ = btcec.NewPrivateKey(btcec.S256())
+	frankPubKey  = privFrank.PubKey()
+	frank        = route.NewVertex(frankPubKey)
+
+	privGeorge, _ = btcec.NewPrivateKey(btcec.S256())
+	georgePubKey  = privGeorge.PubKey()
+	george        = route.NewVertex(georgePubKey)
+)
+
+// testChan is holds simplified test data for channels.
```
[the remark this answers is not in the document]
bhandras: done

routing_plugin_test.go
```diff
@@ -0,0 +1,537 @@
...
+		name                string
+		channels            []testChan
+		routeHints          [][]zpay32.HopHint
+		initError           error
+		missionControlState [][]lndclient.MissionControlEntry
+	}{
+		{
+			name: "degenerate network 1",
+			//
+			// Alice --- Loop
```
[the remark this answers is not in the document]
bhandras: Done, this is basically the same as a simple degenerate case A->B->C->D

bhandras:
Thank you all for the thorough reviews :pray:! Hopefully the client side is now ready apart from maybe some code structuring and additional test clarity. The server side needs some changes to incorporate the payment addr.

routing_plugin.go
```diff
@@ -0,0 +1,109 @@
...
+
+	routingPluginMx.Lock()
+	defer routingPluginMx.Unlock()
+
+	// Another swap is already using the routing plugin.
+	if routingPluginInstance != nil {
+		return nil, nil
+	}
+
+	instance := makeRoutingPlugin(pluginType, lnd)
```
carlaKC: Do we need a new var here? Since we've got the lock, can we use `routingPluginInstance` throughout?

routing_plugin.go
```diff
@@ -0,0 +1,109 @@
+package loop
```
carlaKC, line 1: nitty, but this commit is pretty big. Would be happy to pull `routing_plugin.go` out into its own commit with some very rudimental unit tests to reduce size a bit.

loopd/config.go
```diff
@@ -299,6 +306,16 @@ func Validate(cfg *Config) error {
 		return fmt.Errorf("must specify --lnd.macaroonpath")
 	}
 
+	// Allow at most 1 hour timeout.
```
carlaKC, line 309: nit: comment + error out of date now that we've bumped the default, I'd just go with "at most double our sane default" + fill the value in the error

loopout.go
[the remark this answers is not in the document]
carlaKC: > is our htlc deadline measured in timestamp or in blocks

Timestamp, I believe the reasoning is that people in LN world think "I need inbound liquidity in 5 / 20 / 30 mins" rather than in blocks.

routing_plugin.go
```diff
@@ -107,3 +120,518 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	target route.Vertex) map[route.Vertex][]*lndclient.ChannelEdge {
+
+	edges := make(map[route.Vertex][]*lndclient.ChannelEdge)
+
+	// Traverse through all of the available hop hints and include them in
+	// our edges map.
+	for _, routeHint := range routeHints {
+		// If multiple hop hints are provided within a single route
+		// hint, we'll assume they must be chained together and sorted
+		// in forward order in order to reach the target successfully.
```
carlaKC, line 168: By chained together do you mean this layout?
```
Client ---- N1 ---(private)--- Gateway  ---(private)--- Server
```

I see that our docs currently say "chained together to make a route", but I'm pretty sure we don't support beyond depth 1 hints. If we have multiple hints it'll be this sort of setup:
```
                     ---(private)--- 
Client ----- Gateway ---(private)--- Server
                     ---(private)--- 
```

routing_plugin.go
```diff
@@ -107,3 +120,518 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+		for i, hopHint := range routeHint {
+			// In order to determine the end node of this hint,
+			// we'll need to look at the next hint's start node. If
+			// we've reached the end of the hints list, we can
+			// assume we've reached the target.
```
carlaKC, line 173: Not sure that this holds, I think we might need to use the dest from the invoice and assume all hop hints are depth 1 away from the destination.

routing_plugin.go
carlaKC: We _really_ shouldn't change error codes returned by lnd, so I think we're safe in this case. Also a lot of other naut/operator stuff will bork if we remove the `NotFound` status from `GetNodeInfo` 😅

routing_plugin.go
carlaKC: Ah right.

carlaKC (commented):
Still need to dig into the MC restore logic (agree that we're probably in need of a `force` option). Not quite understanding hop hints logic, pretty sure we don't chain hop hints (could be wrong!)

bhandras:
Hopefully I managed to fix all remaining issues. PTAL

loopout.go
```diff
@@ -635,12 +709,75 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
+	}
+
+	return paymentStatus, err
+}
+
+// sendPaymentWithRetry will send the payment, optionally with the passed
+// routing plugin retrying at most maxRetries times.
+func (s *loopOutSwap) sendPaymentWithRetry(ctx context.Context,
+	hash lntypes.Hash, req *lndclient.SendPaymentRequest, maxRetries int,
+	routingPlugin RoutingPlugin, pluginType RoutingPluginType) (
```
yyforyongyu, line 736: I think the `routingPlugin` already contains the `pluginType` info since that is how it's initialized. We can add a method to that interface returning us the plugin type to remove the redundancy here.

loopout.go
```diff
@@ -649,7 +786,7 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
 		select {
 		// Payment advanced to the next state.
 		case payState := <-payStatusChan:
-			s.log.Infof("Payment %v: %v", hash, payState)
+			s.log.Infof("Payment %v: %v", req.PaymentHash, payState)
```
yyforyongyu, line 787: Where did we set the `req.PaymentHash`? And is it different from `hash`?

loopout.go
```diff
@@ -635,12 +709,75 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
 	// Lookup state of the swap payment.
-	paymentStateCtx, cancel := context.WithCancel(ctx)
+	payCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	payStatusChan, payErrChan, err := s.lnd.Router.SendPayment(
-		paymentStateCtx, req,
+	start := time.Now()
+	paymentStatus, attempts, err := s.sendPaymentWithRetry(
+		payCtx, hash, &req, maxRetries, routingPlugin, pluginType,
```
yyforyongyu, line 715: I don't think we need to pass `maxRetries` if it's a config value within the scope of `loopOutSwap`?

yyforyongyu (commented):
Halfway through. Will come back to review the low/high routing plugin part.

routing_plugin.go
carlaKC: Ah I see, I was looking at invoice creation where we only add depth 1. 

Could we import `RouteHintsToEdges` so that we don't diverge with lnd? We can just convert that output to `lndclient.ChannelEdge`.

routing_plugin.go
```diff
@@ -106,3 +123,522 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+
+// BeforePayment will reconfigure the mission control on each payment attempt.
+func (r *lowToHighRoutingPlugin) BeforePayment(ctx context.Context,
+	currAttempt int, maxAttempts int) error {
+
+	queryRoutesReq := lndclient.QueryRoutesRequest{
+		Source:            &r.lnd.NodePubkey,
+		PubKey:            r.target,
+		AmtMsat:           lnwire.MilliSatoshi(r.amount * 1000),
+		FeeLimitMsat:      lnwire.MilliSatoshi(r.amount * 1000),
```
carlaKC, line 485: Why are we setting the fee limit to the full amount? Don't swaps have their own restrictions?

routing_plugin_test.go
```diff
@@ -71,6 +74,552 @@ func makeTestNetwork(channels []testChan) ([]lndclient.ChannelInfo,
...
+					},
+					{
+						NodeFrom:    charlie,
+						NodeTo:      dave,
+						SuccessTime: testTimeMc,
+						SuccessAmt:  999000,
+					},
+				},
+			},
+			// We don't expect to res
```
carlaKC, line 297: s/res/restore?

carlaKC (approved):
No major comments left!

Would like to possibly follow up with some more itest coverage, but I think we're good to land this as is.

routing_plugin.go
bhandras: Yes, you're correct. It's just a debug query so I felt it's not super important to be fully pedantic here, nor is it correct in the sense that we use the full amount as normally we may split too. I can remove these debug logs too, I just found them useful for understanding what's happening.

routing_plugin.go
```diff
@@ -106,3 +123,522 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+		return nil
+	}
+
+	// Calculate the limit until we'll disable edges. The way we calculate
+	// this limit is that we take the minimum and maximum fee peers which
+	// define our fee range. Within this fee range we'll scale linearly
+	// where each step euqals to the range divided by maxAttempts.
+	minFee := r.nodesByMaxFee[0].fee
+	maxFee := r.nodesByMaxFee[len(r.nodesByMaxFee)-1].fee
+	limit := minFee + ((maxFee-minFee)/int64(maxAttempts))*int64(currAttempt)
```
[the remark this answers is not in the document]
bhandras: done

routing_plugin.go
```diff
@@ -106,3 +123,522 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	// Create a timestamp just slightly in the future as Mission Control
+	// stores timestamps with sub second precision where as we send a unix
+	// timestamp it may occur that we can't override the last entries as
+	// they have the same unix timestamp.
+	// TODO(bhandras): not very reliable, ideally we'd need a force import
+	// for MC.
+	now := r.clock.Now().Add(time.Second)
+
+	allowed := 0
+	entries := make([]lndclient.MissionControlEntry, 0, len(r.nodesByMaxFee))
```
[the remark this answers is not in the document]
bhandras: done

routing_plugin_test.go
bhandras: sorry was a leftover,  removed

routing_plugin.go
```diff
@@ -106,3 +123,522 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+	fee      int64
+}
+
+// Enforce that lowToHighRoutingPlugin implements the RoutingPlugin interface.
+var _ RoutingPlugin = (*lowToHighRoutingPlugin)(nil)
+
+// buildPrivateChannels creates the private channel map from the passed route
+// hints. The code is taken and adapted from LND. Original source:
+// lnd/routing/payment_session_source.go.
+func buildPrivateChannels(routeHints [][]zpay32.HopHint,
```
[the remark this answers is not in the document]
yyforyongyu: So I guess the key difference here is that you want to know all the edges by given a node, rather than having only the directional edges where the node is pointing to next.

routing_plugin.go
```diff
@@ -106,3 +123,522 @@ func ReleaseRoutingPlugin(ctx context.Context) {
...
+			if fee > maxFeePerNode[nodeID] {
+				maxFeePerNode[nodeID] = fee
+			}
+		}
+		totalCapacityPerNode[nodeID] = totalCapacity
+	}
+
+	nodesByMaxFee := make([]nodeFeeInfo, 0, len(maxFeePerNode))
+
+	// Sort peers by maximum fee, so that we can later or disable edges
```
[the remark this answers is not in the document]
yyforyongyu: sorry i meant the typo,
```suggestion
	// Sort peers by maximum fee, so that we can later disable edges
```

yyforyongyu (approved):
Great work! Ideally we'd have more unit tests, but I get it that the PR is already large so maybe followups. LGTM!

PR #444, opened 2021-12-09
multi: remove shadowed label variable in loopin swaps

carlaKC:
We already have a Label field in the embedded SwapContract
field for loop in swaps. This commit removes an erroneously
added Label field in LoopInContract which may be a cause of
ambiguity when referencing this field.


bhandras (approved):
Ouch, nice catch!

LGTM :+1:

yyforyongyu (approved):
LGTM💡

PR #446, opened 2021-12-13
multi: always alias swapserverrpc import for use as a package

carlaKC:
The `looprpc` package is used for client and server protos, despite their being in different directories (`github.com/lightninglabs/loop/looprpc` and `github.com/lightninglabs/loop/swapserverrpc`). This is required to maintain backwards compatibility, because the package is used as part of the URI in grpc. 

[1] Use of mismatched package path/names (eg, import path `github.com/lightninglabs/loop/swapserverrpc` for `looprpc` package) causes issues when loop is imported as a package (as is the case in the server's itests) - the import can't be found. 

[2] Likewise, the presence of two `looprpc` packages (as is the case in `loopd/swapclient_server.go`) also confuses modules when we import `loopd` (to run `loopd` as a package, rather than a standalone binary in itests).

This PR updates loop to alias `swapserverrpc` with `looprpc` everywhere to address [1] and uses distinct aliases in `loopd` to address [2].

#### Pull Request Checklist

### DIFF
--- a/loopd/swapclient_server.go
+++ b/loopd/swapclient_server.go
@@ -17,7 +17,7 @@ import (
 	"github.com/lightninglabs/loop/labels"
 	"github.com/lightninglabs/loop/liquidity"
 	"github.com/lightninglabs/loop/loopdb"
-	"github.com/lightninglabs/loop/looprpc"
+	clientrpc "github.com/lightninglabs/loop/looprpc"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightninglabs/loop/swapserverrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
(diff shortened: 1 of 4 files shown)


loopd/swapclient_server.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/lightninglabs/loop/loopdb"
 	clientrpc "github.com/lightninglabs/loop/looprpc"
 	"github.com/lightninglabs/loop/swap"
-	"github.com/lightninglabs/loop/swapserverrpc"
+	serverrpc "github.com/lightninglabs/loop/swapserverrpc"
```
guggero, line 22: Should we use `looprpc` here to look the same as in the first two commits? Or use `serverrpc` there as well?

guggero (approved):
LGTM :tada:

bhandras (approved):
LGTM :+1:

loopd/swapclient_server.go
carlaKC: I thought the server import might get upset with 2x `looprpc` packages and one aliased with `looprpc`, but just tried it out and it's fine. Will update!

PR #453, opened 2022-01-13
multi: use lndclient MacaroonService

ellemouton:
Since the code for creating and using a macaroon service is the same for
multiple projects (pool, loop, litd etc), the code has been unified in
lndclient. So this commit removes the macaroon service code and instead
uses the lndclient code.

Depends on https://github.com/lightninglabs/lndclient/pull/86


carlaKC (approved):
tACK

guggero (approved):
utACK :tada:

PR #454, opened 2022-01-14
utils: fix multiple issues in SelectHopHints

bhandras:
#### Pull Request Checklist

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -18,6 +18,23 @@ var (
 	DefaultMaxHopHints = 20
 )
 
+// chanCanBeHopHint checks whether the passed channel could be used as a private
+// hophint.
+func chanCanBeHopHint(chanInfo *lndclient.ChannelInfo) bool {
+	return chanInfo.Private && chanInfo.Active
+}
+
+// chanRemotePolicy selectes the correct remote routing policy.
+func chanRemotePolicy(remotePub route.Vertex,
+	edgeInfo *lndclient.ChannelEdge) *lndclient.RoutingPolicy {
+
+	if remotePub == edgeInfo.Node1 {
+		return edgeInfo.Node1Policy
+	}
+
+	return edgeInfo.Node2Policy
+}
+
 // SelectHopHints is a direct port of the SelectHopHints found in lnd. It was
 // reimplemented because the current implementation in LND relies on internals
 // not externalized through the API. Hopefully in the future SelectHopHints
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -44,49 +61,22 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices,
 	// through GetChanInfo
 	chanInfoCache := make(map[uint64]*lndclient.ChannelEdge)
 
-	// skipCache is a simple cache which holds the indice of any
-	// channel we've added to final hopHints
+	// skipCache is a simple cache which holds the indices of any
+	// channel we've added to final hopHints.
```
carlaKC, line 65: Does this comment still hold? Seems that we also add private nodes to `skipCache`, and we're not using them for final hop hints?

utils.go
```diff
@@ -44,49 +61,22 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices,
...
-		// Skip if channel can't forward payment
-		if channel.RemoteBalance < amtMSat {
-			log.Debugf(
-				"Skipping ChannelID: %v for hints as "+
-					"remote balance (%v sats) "+
-					"insufficient appears to be private",
-				channel.ChannelID, channel.RemoteBalance,
-			)
+		// Skip public or inactive channels.
+		if !chanCanBeHopHint(&channel) {
```
carlaKC, line 77: Following on from comment above, would it make sense to also use `skipCache` here (since we're already using it for private nodes)? Saves us the duplicate check below.

carlaKC: Would also perhaps add some debug logging, I'm sure somebody will wonder why an inactive private channel wasn't added so nice to be able to see that in the logs

utils.go
```diff
@@ -130,17 +157,20 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices,
 		// Now that we now this channel use usable, add it as a hop
 		// hint and the indexes we'll use later.
 		hopHints = append(hopHints, []zpay32.HopHint{{
-			NodeID:      nodeID,
+			NodeID:      nodePubKey,
 			ChannelID:   channel.ChannelID,
-			FeeBaseMSat: uint32(chanInfo.Node2Policy.FeeBaseMsat),
+			FeeBaseMSat: uint32(policy.FeeBaseMsat),
```
carlaKC, line 162: Ah damn, so we were using node 2 policy but it's lexicographically ordered so sometimes it was the wrong policy?

carlaKC (commented):
Thanks for the fix! Just one question about use of `skipCache`

utils.go
bhandras: Yeah, it's a bit tricky due to the ordering.

utils.go
bhandras: > Following on from comment above, would it make sense to also use skipCache here (since we're already using it for private nodes)? Saves us the duplicate check below.

Good idea, done.

> Would also perhaps add some debug logging, I'm sure somebody will wonder why an inactive private channel wasn't added so nice to be able to see that in the logs

PTAL

utils.go
bhandras: Yes the comment was a bit outdated, PTAL

utils.go
```diff
@@ -165,14 +198,23 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices,
 			break
 		}
 
-		// Skip the channel if we already selected it.
+		// Channels of private nodes, inacitve, or public channels or
+		// those that have already been selected can be skipped in this
+		// iteration.
```
carlaKC, line 203: Are we obeying the set of peers in `includeNodes` here? Seems like in this second pass we could end up adding channels with peers that we didn't intend to provide hints for? Either need to add them to skipcache or re-check here, I think? 

Test case for `includeNodes` adherence would be good as well, since restricting loop in to a subset of peers is an important feature for lots of people.

utils_test.go
```diff
@@ -211,42 +248,37 @@ func TestSelectHopHints(t *testing.T) {
...
 		expectedError    error
 	}{
-		// 3 inputs set assumes the host node has 3 channels to chose
-		// from. Only channel 2 with peer 2 is ideal, however we should
-		// still include the other 2 after in the order they were
-		// provided just in case
+		// Chan2: private, active, remote balance OK
+		// Chan3: private, !active
+		// Chan4: private, active, small remote balance (second round).
+		// Chan5: public, active => makes peer2 public.
```
carlaKC, line 254: Nice!

carlaKC (commented):
Thanks for all the logging/ comment clarifications, great improvement to readability! 

Just one questions remaining about limiting to a set of peers using `includeNodes`, I think that needs a test case as well since its an important feature.

utils.go
```diff
@@ -113,14 +106,51 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices,
...
+			)
+			continue
+		}
+		// If includeNodes is set, we'll only add channels with peers in
+		// includeNodes. This is done to respect the last_hop parameter.
+		if len(includeNodes) > 0 {
+			if _, ok := includeNodes[channel.PubKeyBytes]; !ok {
+				continue
+			}
+		}
```
arshbot, line 145: maybe else add to skipcache if include nodes is set and channel isn't part of the set?

utils_test.go
arshbot: :pray: this is a much better way of explaining it!

arshbot (changes requested):
Appreciate the second pair of eyes on this! Mostly spelling nits (most of which are actually mine) and I think also suggested a solution to @carlaKC's main q

bhandras:
> Just one questions remaining about limiting to a set of peers using `includeNodes`, I think that needs a test case as well since its an important feature.

Added a test for `includeNodes` too.

carlaKC (commented):
Holding off on review till we can refactor to use lnd with: lightningnetwork/lnd/pull/6182.

bhandras:
Closing this in favor of https://github.com/lightninglabs/loop/pull/455

PR #462, opened 2022-02-17
loopout: do not report prepays and fix plugin type if acquire fails

bhandras:
PR changes two things:

- We don't want to report routing results for prepays as those never use plugins. 
- If a plugin was requested but wasn't acquired then report result with the corrected label.

### DIFF
--- a/loopout.go
+++ b/loopout.go
@@ -587,7 +587,7 @@ func (s *loopOutSwap) payInvoices(ctx context.Context) {
 	// Use the recommended routing plugin.
 	s.swapPaymentChan = s.payInvoice(
 		ctx, s.SwapInvoice, s.MaxSwapRoutingFee,
-		s.LoopOutContract.OutgoingChanSet, pluginType,
+		s.LoopOutContract.OutgoingChanSet, pluginType, true,
 	)
 
 	// Pay the prepay invoice. Won't use the routing plugin here as the
@@ -595,7 +595,7 @@ func (s *loopOutSwap) payInvoices(ctx context.Context) {
 	s.log.Infof("Sending prepayment %v", s.PrepayInvoice)
 	s.prePaymentChan = s.payInvoice(
 		ctx, s.PrepayInvoice, s.MaxPrepayRoutingFee,
-		nil, RoutingPluginNone,
+		nil, RoutingPluginNone, false,
 	)
 }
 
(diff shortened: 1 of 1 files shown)


loopout.go
```diff
@@ -719,11 +719,23 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
 	paymentSuccess := err == nil &&
 		paymentStatus.State == lnrpc.Payment_SUCCEEDED
 
-	if err := s.swapKit.server.ReportRoutingResult(
-		ctx, s.swapInfo().SwapHash, s.swapInvoicePaymentAddr, pluginType,
-		paymentSuccess, int32(attempts), dt.Milliseconds(),
-	); err != nil {
-		s.log.Warnf("Failed to report routing result: %v", err)
+	// RoutingPluginNone is requested for prepays which we don't want to
```
carlaKC, line 722: It's possible that we also have `RoutingPluginNone` for swap payments if we've got multiple payments going off at the same time, right?

loopout.go
```diff
@@ -719,11 +719,23 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
+	// report.
+	if pluginType != RoutingPluginNone {
```
carlaKC, line 724: Is there any point in reporting swap payments that use `RoutingPluginNone` so that we can compare them to ones that use the plugin? I'd guess not because they're also affected by the mission control changes? 

If yes, I think we should pass in a `reportPlugin` bool ?

bhandras: The passed in bool is solves this nicely.

loopout.go
bhandras: Using bool flag now.

yyforyongyu (approved):

loopout.go
```diff
@@ -719,11 +721,21 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
-		paymentSuccess, int32(attempts), dt.Milliseconds(),
-	); err != nil {
-		s.log.Warnf("Failed to report routing result: %v", err)
+	if reportPluginResult {
+		// If the plugin couldn't be acquired then override the reported
+		// plugin type to RoutingPluginNone.
+		reportType := pluginType
+		if routingPlugin == nil {
+			reportType = RoutingPluginNone
+		}
```
carlaKC, line 730: Don't need this? The enum will default to `RoutingPluginNone` ?

carlaKC (approved):

loopout.go
bhandras: The reason for this is because routingPlugin can be nil if the acquire failed (even though plugin type wasn't none).

carlaKC: Right, thought that `pluginType` would always be the zero value, got it!

PR #465, opened 2022-03-01
main: fix feepercent from intflags to float

arshbot:
Previously, the feepercent flag on autoloop was set to an IntFlag and
later converted to Float, leading to the issue where users couldn't
specify decimal rates even though we allowed it.

#### Pull Request Checklist


carlaKC (approved):
Nice!

PR #466, opened 2022-03-11
routing: integrate with the forced MC import to eliminate the "1 sec hack" when importing pair history

bhandras:
Note: this PR requires loop clients to use v0.14.2-beta.

With the changes in this PR we fix the "1 sec hack" when setting MC pair state through RPC. Previously MC required newly imported pairs to have a newer timestamp vs the current pair state which we satisfied by pushing  timestamps 1 sec into the future. This was a clunky way to force successful import. With the latest forced import feature integrated this isn't needed anymore.


yyforyongyu:
test case `TestAutoLoopInEnabled` failed.

go.mod
```diff
@@ -11,10 +11,10 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.5.0
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/lightninglabs/aperture v0.1.6-beta
-	github.com/lightninglabs/lndclient v0.14.0-8
+	github.com/lightninglabs/lndclient v0.14.2-3
```
carlaKC, line 14: We need to update `LoopMinRequiredLndVersion` since we're using a new api param. 

We could remove it completely so that we just use the lndclient default every time we bump. I'm somewhat in favor of this, since it's easy to forget to bump, but the downside is more upgrade pressure for clients.

carlaKC (commented):
Great to come back and address the workaround! 

Just needs min version fix.

bhandras:
> test case `TestAutoLoopInEnabled` failed.

It looks like a different flake not related to this PR.

carlaKC (approved):
LGTM! 

Just needs a release notes entry noting the update + need to bump to 14.2.

bhandras:
> Just needs a release notes entry noting the update + need to bump to 14.2.

done

PR #469, opened 2022-03-21
sweep: always send signed output's pk script

### LINKED ISSUE #457: sweeper: `SignOutputRaw` called without UTXO pk script
Found in combination with remote signing, see https://github.com/lightningnetwork/lnd/issues/6227.

In https://github.com/lightninglabs/loop/blob/master/sweep/sweeper.go#L64 we create the sign descriptor for calling `SignOutputRaw` but we don't include the UTXOs pk script, only the value.
This works fine in the normal (full private keys available) wallet but presents a problem in a remote-signing setup where the signing information needs to be packaged in a PSBT to be sent over to the signing instance. An empty pk script field in the `WitnessUTXO` field of a PSBT is invalid (apparently).

I checked the code and we have the HTLC pk script in all cases where we use the sweeper, so it should be relatively easy to pass it into the call.

guggero:
Fixes #457.
The SignOutputRaw RPC is normally too lenient when it comes to fields
not being set, which is why this used to work previously. But with lnd
in remote signing mode, we need all the extra meta information available
on an input being signed for, so we actually need the pk script being
set.
Loop fully works with a remote signing lnd backend pair after applying
this fix.

#### Pull Request Checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -25,4 +25,7 @@ This file tracks release notes for the loop client.
 
 #### Bug Fixes
 
+* Loop now supports being hooked up to a remote signing pair of `lnd` nodes,
+  as long as `lnd` is `v0.14.3-beta` or later.
+
 #### Maintenance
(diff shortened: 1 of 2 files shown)


arshbot (approved):
LGTM!

PR #471, opened 2022-03-24
WIP: Htlc v3 taproot

arshbot:
In this PR we introduce V3 loop out and loop in submarine swaps via taproot. The general approach has been to implement v3 htlcs in `htlc.go` and patch the plumbing from loop -> loopserver to accommodate the extra values. In some cases we default to creating taproot htlcs instead of traditional p2sh, which may be something a keen reviewer might want to look into, should we decide not to go all in on taproot for the next release. 

TODOs
- [x] loopd: btcec version bump
- [x] loopserver: btcec version bump
- [x] loopd: loopin 
- [x] loopserver: loopin 
- [x] loopd: loopout 
- [x] loopserver: loopout
- [ ] tidying up: docs

Known Issues:
- [ ] validation on sweep fails
- [ ] de/serialization to etcd fails due to different key size


swap/htlc.go
```diff
@@ -512,3 +569,147 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	// TODO: Unsilence errors
+	controlBlockBytes, _ := h.genControlBlock(h.timeoutScript)
+	return wire.TxWitness{
+		senderSig,
+		h.timeoutScript,
+		controlBlockBytes,
+	}
+}
+
+func (h *HtlcScriptV3) IsSuccessWitness(witness wire.TxWitness) bool {
```
carlaKC, line 674: Needs values?

swap/htlc_test.go
```diff
@@ -319,3 +322,230 @@ func TestHtlcV2(t *testing.T) {
 		})
 	}
 }
+
+/*
+	CLAIM PATH
+
+	<reciever_key> OP_CHECKSIGVERIFY OP_SIZE 20 OP_EQUALVERIFY OP_RIPEMD160 <hash> OP_EQUALVERIFY 1 OP_CHECKSEQUENCEVERIFY
+*/
+func createClaimPathLeaf(t *testing.T, recieverHtlcKey [32]byte, swapHash lntypes.Hash) (txscript.TapLeaf, []byte) {
```
carlaKC: Rather than storing `claimScript`/`timeoutScript` in `HtlcScriptV3`, move these functions into `htlc.go` and then we can just recreate them using our receiver/sender key and cltv when we need these scripts?

swap/htlc.go
```diff
@@ -512,3 +569,147 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	builder.AddOp(txscript.OP_CHECKSIGVERIFY)
+	builder.AddInt64(int64(cltvExpiry))
+	builder.AddOp(txscript.OP_CHECKLOCKTIMEVERIFY)
+	timeoutPathScript, err := builder.Script()
+
+	if err != nil {
+		return nil, err
+	}
+
+	randomPub, _ := hex.DecodeString(
```
carlaKC, line 598: If we're going to deploy these chagnes incrementally (ie, do tapscript then musig keyspend) this is going to need to be deterministic / communicated between client/server so that each party knows the other doesn't know the privkey (and can just claim the htlc). 

Probably not worth the effort, so would just add a clarifying comment that this is a dummy value that *should not* be used in prod.

swap/htlc.go
```diff
@@ -671,13 +674,13 @@ func (h *HtlcScriptV3) genControlBlock(leafScript []byte) ([]byte, error) {
 	return controlBlockBytes, nil
 }
 
-func (h *HtlcScriptV3) genSuccessWitness(receiverSig []byte, preimage lntypes.Preimage) wire.TxWitness {
+func (h *HtlcScriptV3) genSuccessWitness(signature []byte, preimage lntypes.Preimage) wire.TxWitness {
 
 	// TODO: Unsilence errors
-	controlBlockBytes, _ := h.genControlBlock(h.claimScript)
+	controlBlockBytes, _ := h.genControlBlock(h.timeoutScript)
```
carlaKC: is this correct?

swap/htlc.go
```diff
@@ -686,7 +689,7 @@ func (h *HtlcScriptV3) genSuccessWitness(receiverSig []byte, preimage lntypes.Pr
 func (h *HtlcScriptV3) GenTimeoutWitness(senderSig []byte) wire.TxWitness {
 
 	// TODO: Unsilence errors
-	controlBlockBytes, _ := h.genControlBlock(h.timeoutScript)
+	controlBlockBytes, _ := h.genControlBlock(h.claimScript)
```
carlaKC: likewise, seems like these have switched?

loopdb/loop.go
```diff
@@ -20,11 +20,11 @@ type SwapContract struct {
 
 	// SenderKey is the key of the sender that will be used in the on-chain
 	// HTLC.
-	SenderKey [33]byte
+	SenderKey []byte
```
carlaKC, line 23: move this commit to before we add `HtlcScriptV3` + don't intermingle refactors with logical changes (eg `SuccessSequence` returns 10 now?)

looprpc/client.proto
```diff
@@ -317,8 +317,10 @@ message SwapResponse {
     */
     string htlc_address_p2wsh = 5;
 
+    string htlc_address_p2tr = 6;
+
     // A human-readable message received from the loop server.
-    string server_message = 6;
+    string server_message = 7;
```
carlaKC, line 323: changing the proto number of `server_message` will break the API for old clients.

swap/htlc.go
```diff
@@ -42,7 +42,7 @@ const (
 	// HtlcV2 refers to the improved version of the HTLC script.
 	HtlcV2
 
-	HtlcV3
+	HtlcV3 = 10
```
carlaKC, line 45: why not use the existing iota?

carlaKC (commented):
Tapscript changes looking good! However the structure of this PR makes it quite difficult to review in depth, needs some cleanup/ refactors before next review please.

PR #473, opened 2022-03-28
swap: HTLCV3 added

arshbot:
In this commit we add the version 3 htlc, which is implemented with
taproot script spending the two payment paths: the claim path case, and
the timeout case.

#### Pull Request Checklist


swap/htlc.go
```diff
@@ -512,3 +555,220 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	// - witness_script: len(script) bytes
+	// - control_block_length: 1 byte
+	// - control_block: 4129 bytes
+	return 1 + 1 + 73 + 1 + len(h.TimeoutScript) + 1 + 4129
+}
+
+// SuccessSequence returns the sequence to spend this htlc in the
+// success case.
+func (h *HtlcScriptV3) SuccessSequence() uint32 {
+	return 10
```
carlaKC, line 773: Why is this 10?

swap/htlc.go
```diff
@@ -512,3 +555,220 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
 func (h *HtlcScriptV2) SuccessSequence() uint32 {
 	return 1
 }
+
+// HtlcScriptV3 encapsulates the htlc v3 script.
+type HtlcScriptV3 struct {
+	TimeoutScript  []byte
+	ClaimScript    []byte
+	TaprootKey     *secp.PublicKey
+	InternalPubKey *secp.PublicKey
```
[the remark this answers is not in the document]
arshbot: Storing as byte slice adds a lot of conversions all over the place, but better here than throughout the codebase. Storing as array will only add another conversion without reason imo. 

(also, I think `secp.PublicKey` are 33 bytes)

swap/htlc.go
arshbot: Not sure. Changed to 1 to match htlcv2, pinged @alexbosworth offline to make sure these values are correct and maybe shed light on why we have 1 for htlcv2 and 0 for htlcv1

arshbot: 1 because we force the redeemer to wait for the htlc to conf before allowing spend

swap/htlc.go
```diff
@@ -132,6 +145,10 @@ func NewHtlc(version ScriptVersion, cltvExpiry int32,
 		htlc, err = newHTLCScriptV2(
 			cltvExpiry, senderKey, receiverKey, hash,
 		)
+	case HtlcV3:
+		htlc, err = newHTLCScriptV3(
+			cltvExpiry, senderKey, receiverKey, hash,
```
carlaKC, line 150: Ordering here is sender/receiver but functoin has receiver/sender.

carlaKC, line 150: Ordering here is sender/receiver but functoin has receiver/sender.

swap/htlc.go
```diff
@@ -512,3 +561,227 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	ClaimScript    []byte
+	TaprootKey     []byte
+	InternalPubKey []byte
+}
+
+// newHTLCScriptV3 constructs a HtlcScipt with the HTLC V3 taproot script.
+func newHTLCScriptV3(cltvExpiry int32, receiverHtlcKey, senderHtlcKey [33]byte,
+	swapHash lntypes.Hash) (*HtlcScriptV3, error) {
+
+	_, receiverPubKey := btcec.PrivKeyFromBytes(
```
carlaKC, line 577: Pubkey from privkeyfrombytes? 


```
receiverPubkey, err:= btcec.ParsePubkey(receiverHtlcKey[:])
copy(schnorrReceiverKey, schnorr.SerializePubkey(receiverPubkey)
```

carlaKC, line 577: Pubkey from privkeyfrombytes? 


```
receiverPubkey, err:= btcec.ParsePubkey(receiverHtlcKey[:])
copy(schnorrReceiverKey, schnorr.SerializePubkey(receiverPubkey)
```

carlaKC (changes requested):

swap/htlc.go
arshbot: From `NewHtlc`: 

> // NewHtlc returns a new instance. For V1 and V2 scripts, receiver and sender
> // keys are expected to be in ecdsa compressed format. For V3 scripts, keys are
> // expected to be raw private key bytes.

This was feedback I received from laolu and oli in regards to dealing with the serializeLoopContract issue which requires the keys stored to be 33 bytes (which ecdsa compressed are), however schnorr keys are 32 bytes, which throws the serializer off. To mitigate this issue, we store raw private key bytes in the LoopContract in memory and feed those raw bytes to `NewHtlc` _only_ in the htlcv3 case.

arshbot: See https://github.com/lightninglabs/loop/pull/473#discussion_r838793222

carlaKC: ugh duplicate GH comment post.

swap/htlc.go
```diff
@@ -512,3 +555,220 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	//
+	// - number_of_witness_elements: 1 byte
+	// - sig_length: 1 byte
+	// - sig: 73 bytes
+	// - preimage_length: 1 byte
+	// - preimage: 32 bytes
+	// - witness_script_length: 1 byte
+	// - witness_script: len(script) bytes
+	// - control_block_length: 1 byte
+	// - control_block: 4129 bytes
```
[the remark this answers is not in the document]
carlaKC: I _think_ this should be based on the worst-case size of our control block so that we can get as close to accurate as possible.

swap/htlc.go
carlaKC: Why can't we just store `[33]byte` then `schnorr.SerializePubkey` when we need to use the key in a `HtlcV3` context? 

Do you have a link to that convo so I can understand the context of that discussion?

swap/htlc.go
```diff
@@ -111,7 +122,9 @@ func (h HtlcOutputType) String() string {
 	}
 }
 
-// NewHtlc returns a new instance.
+// NewHtlc returns a new instance. For V1 and V2 scripts, receiver and sender
+// keys are expected to be in ecdsa compressed format. For V3 scripts, keys are
+// expected to be raw private key bytes.
```
carlaKC, line 127: I do not think it is good design to expect different parameter values for receiver/sender key based on the value of the `version` parmater. Footguns, footguns everywhere. 

Additionally, we're not going to have access to the raw private key bytes (this is all managed in lnd for loop) - when we derive our keys, we only get the [pubkey](https://github.com/carlaKC/loop/blob/master/loopout.go#L110). We don't want to add privkey handling to loop.

swap/htlc.go
```diff
@@ -171,29 +188,59 @@ func NewHtlc(version ScriptVersion, cltvExpiry int32,
...
 		)
 		if err != nil {
 			return nil, err
 		}
+	case HtlcP2TR:
+		// Confirm we have a v3 htlc.
+		var trHtlc *HtlcScriptV3
+		trHtlc, ok := htlc.(*HtlcScriptV3)
+		if !ok {
+			return nil, errors.New("taproot output selected for nontaproot htlc")
```
bhandras, line 222: nit: `s/nontaproot/non taproot`

swap/htlc.go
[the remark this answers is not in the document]
bhandras: I agree with Carla, since our taptree is shallow, our control block is also small compared to the worst theoretical case, so we should calculate the max for our case imo. (Even though we don't use this function).

swap/htlc.go
```diff
@@ -512,3 +561,227 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	// Calculate maximum timeout witness size
+	//
+	// - number_of_witness_elements: 1 byte
+	// - sigLength: 1 byte
+	// - sig: 64 bytes
+	// - witness_script_length: 1 byte
+	// - witness_script: len(script) bytes
+	// - control_block_length: 1 byte
+	// - control_block: 4129 bytes
+	return 1 + 1 + 64 + 1 + len(h.TimeoutScript) + 1 + 4129
```
bhandras, line 780: Similarly here, we should probably calculate the worst control block size for our use case (even though we don't use this function).

swap/htlc_test.go
```diff
@@ -306,8 +322,275 @@ func TestHtlcV2(t *testing.T) {
...
+	receiverPrivKey := [33]byte{
+		0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, byte(2),
+	}
+	cltvExpiry := int32(10)
+
+	htlc, err := NewHtlc(
+		HtlcV3, cltvExpiry, senderPrivKey, receiverPrivKey, hashedPreimage, HtlcP2TR, &chaincfg.MainNetParams,
```
bhandras, line 356: nit: line too long

swap/htlc_test.go
```diff
@@ -306,8 +322,275 @@ func TestHtlcV2(t *testing.T) {
...
+
+	tx := wire.NewMsgTx(2)
+	tx.TxIn = []*wire.TxIn{{
+		PreviousOutPoint: wire.OutPoint{
+			Hash:  chainhash.Hash(sha256.Sum256([]byte{1, 2, 3})),
+			Index: 50,
+		},
+	}}
+	tx.TxOut = []*wire.TxOut{{
+		PkScript: []byte{0, 20, 2, 141, 221, 230, 144, 171, 89, 230, 219, 198, 90, 157, 110, 89, 89, 67, 128, 16, 150, 186},
```
bhandras, line 375: nit: formatting

swap/htlc_test.go
```diff
@@ -306,8 +322,275 @@ func TestHtlcV2(t *testing.T) {
...
+				var trHtlc *HtlcScriptV3
+				trHtlc, ok := htlc.HtlcScript.(*HtlcScriptV3)
+				require.True(t, ok)
+
+				senderPrivKey, _ := btcec.PrivKeyFromBytes(
+					senderPrivKey[:],
+				)
+
+				sig := signTx(
+					tx, senderPrivKey, txscript.NewBaseTapLeaf(trHtlc.ClaimScript),
```
bhandras, line 419: The testcase descriptions and the actual testcases are inverted.

swap/htlc_test.go
```diff
@@ -306,8 +322,275 @@ func TestHtlcV2(t *testing.T) {
...
+					senderPrivKey[:],
+				)
+				var shnorrSenderKey [32]byte
+				copy(
+					shnorrSenderKey[:],
+					schnorr.SerializePubKey(senderPubKey),
+				)
+
+				htlc, err := NewHtlc(
+					HtlcV3, cltvExpiry, bogusKey, receiverPrivKey, hashedPreimage, HtlcP2TR, &chaincfg.MainNetParams,
```
bhandras, line 547: nit: formatting

swap/htlc_test.go
```diff
@@ -306,8 +322,275 @@ func TestHtlcV2(t *testing.T) {
...
+			"timeout case receiver cannot spend",
+			func(t *testing.T) wire.TxWitness {
+				tx.TxIn[0].Sequence = htlc.SuccessSequence()
+				tx.LockTime = uint32(cltvExpiry)
+
+				var trHtlc *HtlcScriptV3
+				trHtlc, ok := htlc.HtlcScript.(*HtlcScriptV3)
+				require.True(t, ok)
+
+				senderPrivKey, _ := btcec.PrivKeyFromBytes(
```
bhandras, line 516: nit: the `senderPrivKey` and `receiverPrivKey` could be precalculated outside of the witness function bodies.

bhandras (commented):
Great PR, almost there just a few comments remain.

swap/htlc.go
```diff
@@ -512,3 +561,227 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+		TaprootKey:     taprootKey.SerializeCompressed(),
+		InternalPubKey: internalPubKey.SerializeCompressed(),
+	}, nil
+}
+
+// GenTimeoutPathScript constructs an HtlcScript for the timeout payment path.
+//
+//	<senderHtlcKey> OP_CHECKSIGVERIFY <cltvExpiry> OP_CHECKLOCKTIMEVERIFY
+func GenTimeoutPathScript(
+	receiverHtlcKey [32]byte, cltvExpiry int64) ([]byte, error) {
```
bhandras, line 639: Seems like there's a mismatch here too: sender/receiver

swap/htlc.go
```diff
@@ -512,3 +561,227 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+}
+
+// GenClaimPathScript constructs an HtlcScript for the claim payment path.
+//
+//	<receiverHtlcKey> OP_CHECKSIGVERIFY
+//	OP_SIZE 32 OP_EQUALVERIFY
+//	OP_HASH160 <ripemd160h(swapHash)> OP_EQUALVERIFY
+//	1 OP_CHECKSEQUENCEVERIFY
+func GenClaimPathScript(
+	senderHtlcKey [32]byte, swapHash lntypes.Hash) ([]byte, error) {
```
bhandras, line 656: Mixup here too (sender/receiver)

swap/htlc.go
```diff
@@ -512,3 +561,227 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	}
+
+	controlBlock := txscript.ControlBlock{
+		InternalKey:     internalPubKey,
+		OutputKeyYIsOdd: outputKeyYIsOdd,
+		LeafVersion:     txscript.BaseLeafVersion,
+		InclusionProof:  proof[:],
+	}
+	controlBlockBytes, err := controlBlock.ToBytes()
+
```
bhandras, line 700: nit: extra nl

arshbot (commented):
Made the requested changes! Ready for another pass

swap/htlc.go
arshbot: @bhandras I agree, however that should be included in a seperate PR

guggero (commented):
Looks way better now, but still a lot of nits. We want to be extra careful, explicit and verbose with anything that touches on-chain scripts.

swap/htlc_test.go
```diff
@@ -306,8 +322,289 @@ func TestHtlcV2(t *testing.T) {
...
+
+	var (
+		receiverKey [33]byte
+		senderKey   [33]byte
+	)
+
+	preimage := [32]byte{1, 2, 3}
+	p := lntypes.Preimage(preimage)
+	hashedPreimage := sha256.Sum256(p[:])
+	value := int64(800_000 - 500) // TODO(guggero): Calculate actual fee.
```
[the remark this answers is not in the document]
arshbot: Removed fee because previous tests don't include and it's a distraction from what this is testing

swap/htlc.go
```diff
@@ -512,3 +563,244 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	// Create our timeout path leaf, we'll use this separately
+	// later on to generate the timeout path leaf.
+	timeoutPathScript, err := GenTimeoutPathScript(
+		schnorrReceiverKey, int64(cltvExpiry),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	randomPub, _ := hex.DecodeString(
```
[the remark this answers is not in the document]
bhandras: Don't think it's a good idea, perhaps better to use a real random pubkey instead?

swap/htlc.go
```diff
@@ -141,18 +158,18 @@ func NewHtlc(version ScriptVersion, cltvExpiry int32,
...
-	if err != nil {
-		return nil, err
-	}
-
 	var pkScript, sigScript []byte
 	var address btcutil.Address
 
 	switch outputType {
 	case HtlcNP2WSH:
+		pkScript, err := input.WitnessScriptHash(htlc.Script())
```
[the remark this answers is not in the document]
arshbot: Added this change in `c9b6b9600b6aecedbad084c647dd6b23fadce349`

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	builder.AddOp(txscript.OP_CHECKSIGVERIFY)
+	builder.AddOp(txscript.OP_SIZE)
+	builder.AddInt64(32)
+	builder.AddOp(txscript.OP_EQUALVERIFY)
+	builder.AddOp(txscript.OP_HASH160)
+	builder.AddData(input.Ripemd160H(swapHash[:]))
+	builder.AddOp(txscript.OP_EQUALVERIFY)
+	builder.AddInt64(1)
+	builder.AddOp(txscript.OP_CHECKSEQUENCEVERIFY)
+	a, err := builder.Script()
```
bhandras, line 676: nit: just `return builder.Script()`

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+}
+
+// genControlBlock constructs the control block with the depth 1 leaf of the
+// unused path to compute the proof. For example if spending path a of (root ->
+// a, root -> b), genControlBlock(b.Script) would be used to create the
+// controlBlock for a.
+func (h *HtlcScriptV3) genControlBlock(leafScript []byte) ([]byte, error) {
+	var outputKeyYIsOdd bool
+
+	// Check for odd bit
```
bhandras, line 687: nit: . here and below

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+		return nil, err
+	}
+
+	controlBlock := txscript.ControlBlock{
+		InternalKey:     internalPubKey,
+		OutputKeyYIsOdd: outputKeyYIsOdd,
+		LeafVersion:     txscript.BaseLeafVersion,
+		InclusionProof:  proof[:],
+	}
+	controlBlockBytes, err := controlBlock.ToBytes()
```
bhandras, line 707: nit: we could just return here `return controlBlock.ToBytes()`

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+// genSuccessWitness returns the success script to spend this htlc with
+// the preimage.
+func (h *HtlcScriptV3) genSuccessWitness(
+	receiverSig []byte, preimage lntypes.Preimage) (wire.TxWitness, error) {
+
+	controlBlockBytes, err := h.genControlBlock(h.TimeoutScript)
+	if err != nil {
+		return nil, err
+	}
+	return wire.TxWitness{
```
bhandras, line 724: nit: add a nl

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+// GenTimeoutWitness returns the timeout script to spend this htlc after
+// timeout.
+func (h *HtlcScriptV3) GenTimeoutWitness(
+	senderSig []byte) (wire.TxWitness, error) {
+
+	controlBlockBytes, err := h.genControlBlock(h.SuccessScript)
+	if err != nil {
+		return nil, err
+	}
+	return wire.TxWitness{
```
bhandras, line 741: nit: add a nl

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+
+// GenSuccessPathScript constructs an HtlcScript for the claim payment path.
+// Largest possible bytesize of the script is 32 + 5*1 + 20 + 3*1 = 60
+//
+//	<senderHtlcKey> OP_CHECKSIGVERIFY
+//	OP_SIZE 32 OP_EQUALVERIFY
+//	OP_HASH160 <ripemd160h(swapHash)> OP_EQUALVERIFY
+//	1 OP_CHECKSEQUENCEVERIFY
+func GenSuccessPathScript(
+	senderHtlcKey [32]byte, swapHash lntypes.Hash) ([]byte, error) {
```
bhandras, line 663: I'm getting a bit confused but this should be the `receiverHtlcKey` right?

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	}, nil
+}
+
+// GenTimeoutPathScript constructs an HtlcScript for the timeout payment path.
+// Largest possible bytesize of the script is 32 + 1 + 2(normally 1, but max
+// size is 400) + 1 = 36
+//
+//	<receiverHtlcKey> OP_CHECKSIGVERIFY <cltvExpiry> OP_CHECKLOCKTIMEVERIFY
+func GenTimeoutPathScript(
+	receiverHtlcKey [32]byte, cltvExpiry int64) ([]byte, error) {
```
bhandras, line 645: Similarly this should be the `senderHtlcKey` if i'm not mistaken?

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	// later on to generate the claim path leaf.
+	claimPathScript, err := GenSuccessPathScript(schnorrSenderKey, swapHash)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create our timeout path leaf, we'll use this separately
+	// later on to generate the timeout path leaf.
+	timeoutPathScript, err := GenTimeoutPathScript(
+		schnorrReceiverKey, int64(cltvExpiry),
```
bhandras, line 608: And this should be the sender key.

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+		TimeoutScript:  timeoutPathScript,
+		SuccessScript:  claimPathScript,
+		TaprootKey:     taprootKey.SerializeCompressed(),
+		InternalPubKey: sharedKey.SerializeCompressed(),
+	}, nil
+}
+
+// GenTimeoutPathScript constructs an HtlcScript for the timeout payment path.
+// Largest possible bytesize of the script is 32 + 1 + 2(normally 1, but max
+// size is 400) + 1 = 36
```
bhandras, line 641: not sure I understand the 400?

swap/htlc_test.go
```diff
@@ -306,8 +323,298 @@ func TestHtlcV2(t *testing.T) {
...
+			}
+
+			assertEngineExecution(t, testCase.valid, newEngine)
+		})
+	}
+}
+
+// TestHtlcV3 tests the HTLC V3 script success and timeout spend cases.
+func TestHtlcV3(t *testing.T) {
+
```
bhandras, line 337: nit: extra nl

swap/htlc_test.go
```diff
@@ -306,8 +323,298 @@ func TestHtlcV2(t *testing.T) {
...
+			"success case no spend with valid preimage and zero sequence",
+			func(t *testing.T) wire.TxWitness {
+				tx.TxIn[0].Sequence = 0
+
+				var trHtlc *HtlcScriptV3
+				trHtlc, ok := htlc.HtlcScript.(*HtlcScriptV3)
+				require.True(t, ok)
+
+				sig := signTx(
+					tx, senderPrivKey,
```
bhandras, line 455: this should be the receiver's priv key?

swap/htlc_test.go
```diff
@@ -306,8 +323,298 @@ func TestHtlcV2(t *testing.T) {
...
+			func(t *testing.T) wire.TxWitness {
+				tx.TxIn[0].Sequence = htlc.SuccessSequence()
+				tx.LockTime = uint32(cltvExpiry) - 1
+
+				var trHtlc *HtlcScriptV3
+				trHtlc, ok := htlc.HtlcScript.(*HtlcScriptV3)
+				require.True(t, ok)
+
+				sig := signTx(
+					tx, receiverPrivKey,
```
bhandras, line 480: this should be the sender priv key?

swap/htlc_test.go
```diff
@@ -306,8 +323,298 @@ func TestHtlcV2(t *testing.T) {
...
+			func(t *testing.T) wire.TxWitness {
+				tx.TxIn[0].Sequence = htlc.SuccessSequence()
+				tx.LockTime = uint32(cltvExpiry)
+
+				var trHtlc *HtlcScriptV3
+				trHtlc, ok := htlc.HtlcScript.(*HtlcScriptV3)
+				require.True(t, ok)
+
+				sig := signTx(
+					tx, receiverPrivKey,
```
bhandras, line 504: same as above

swap/htlc_test.go
```diff
@@ -306,8 +323,298 @@ func TestHtlcV2(t *testing.T) {
...
+			func(t *testing.T) wire.TxWitness {
+				tx.TxIn[0].Sequence = htlc.SuccessSequence()
+				tx.LockTime = uint32(cltvExpiry)
+
+				var trHtlc *HtlcScriptV3
+				trHtlc, ok := htlc.HtlcScript.(*HtlcScriptV3)
+				require.True(t, ok)
+
+				sig := signTx(
+					tx, senderPrivKey,
```
bhandras, line 528: same

bhandras (commented):
Tried to find everything, hopefully we're now just one round away from being ready.

swap/htlc.go
```diff
@@ -141,18 +141,18 @@ func NewHtlc(version ScriptVersion, cltvExpiry int32,
...
-	if err != nil {
-		return nil, err
-	}
-
 	var pkScript, sigScript []byte
 	var address btcutil.Address
 
 	switch outputType {
 	case HtlcNP2WSH:
+		pkScript, err := input.WitnessScriptHash(htlc.Script())
```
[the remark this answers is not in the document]
arshbot: Good catch, done. Since `err` is already declared earlier, the redeclaration shouldn't be necessary

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	// Create our timeout path leaf, we'll use this separately
+	// later on to generate the timeout path leaf.
+	timeoutPathScript, err := GenTimeoutPathScript(
+		schnorrReceiverKey, int64(cltvExpiry),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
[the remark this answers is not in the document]
arshbot: can't be too careful

swap/htlc.go
```diff
@@ -111,9 +122,10 @@ func (h HtlcOutputType) String() string {
 	}
 }
 
-// NewHtlc returns a new instance.
+// NewHtlc returns a new instance. For V1 and V2 scripts, receiver and sender
+// keys are expected to be in compressed format.
 func NewHtlc(version ScriptVersion, cltvExpiry int32,
-	senderKey, receiverKey [33]byte,
+	senderKey, receiverKey [33]byte, sharedKey *btcec.PublicKey,
```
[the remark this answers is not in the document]
arshbot: Over the pointer? Why?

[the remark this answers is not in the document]
arshbot: Oh I think you meant to be consistent about which `Publickey` class is used - done

swap/htlc.go
arshbot: I removed the explanation, it seems needlessly verbose. Basically cltvExpiry is going to take up 1 byte in most cases as the value should be under uint 128 (1 byte) but it appears we have a max allowance of uint 400 which will require 2 bytes.

swap/htlc.go
arshbot: Yes, I went through the code and corrected it fully. Shouldn't be an issue anymore

swap/htlc.go
```diff
@@ -515,3 +563,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+	TaprootKey     []byte
+	InternalPubKey []byte
+}
+
+// newHTLCScriptV3 constructs a HtlcScipt with the HTLC V3 taproot script.
+func newHTLCScriptV3(cltvExpiry int32, senderHtlcKey,
+	receiverHtlcKey [33]byte, sharedKey *secp.PublicKey,
+	swapHash lntypes.Hash) (*HtlcScriptV3, error) {
+
+	receiverPubKey, err := btcec.ParsePubKey(
```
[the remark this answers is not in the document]
arshbot: I think I prefer the protection of serializing and deserializing since this is supposed to be a standalone module - though I see your point

swap/htlc.go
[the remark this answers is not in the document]
bhandras: It's still wrong?

swap/htlc.go
```diff
@@ -133,6 +145,12 @@ func NewHtlc(version ScriptVersion, cltvExpiry int32,
 			cltvExpiry, senderKey, receiverKey, hash,
 		)
 
+	case HtlcV3:
+		htlc, err = newHTLCScriptV3(
+			cltvExpiry, senderKey, receiverKey,
+			sharedKey, hash,
+		)
```
bhandras, line 152: Check `err != nil` is needed.

swap/htlc.go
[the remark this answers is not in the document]
bhandras: edit: I see you fixed it in the other commit. Please fix it in this commit.

swap/htlc.go
```diff
@@ -515,3 +559,239 @@ func (h *HtlcScriptV2) MaxTimeoutWitnessSize() int {
...
+
+	// Calculate top level taproot key.
+	taprootKey := txscript.ComputeTaprootOutputKey(
+		sharedKey, rootHash[:],
+	)
+
+	return &HtlcScriptV3{
+		TimeoutScript:  timeoutPathScript,
+		SuccessScript:  successPathScript,
+		TaprootKey:     taprootKey,
```
bhandras, line 638: nit: switch the last two

bhandras (commented):
Almost ready :)

swap/htlc.go
arshbot: Isn't there one on [line 140](https://github.com/lightninglabs/loop/pull/473/files/c9b6b9600b6aecedbad084c647dd6b23fadce349..2135bcd5676f3f327c9fbf5aaa015d8a83c57e0e#diff-9e0e39ac7c78e0cda4bdd24a48f9c3e41f9c98aa27519c8822dfe5c869e0fec5L140)?

swap/htlc.go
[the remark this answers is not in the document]
arshbot: done

swap/htlc.go
bhandras: Ah right, missed it.

bhandras (approved):
LGTM :tada:

swap/htlc.go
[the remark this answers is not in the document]
bhandras: This Is still wrong in the first commit.

bhandras:
Uhm, these PRs need two approvals..

PR #475, opened 2022-03-30
multi: upgrade to P2TR htlcs without keyspend path

carlaKC:
This PR upgrades loop to use P2TR htlcs, with a tree containing a tapleaf for our unchanged preimage reveal and timeout paths. It is a untested, and code structure could use improvement, but it runs. 

Depends on #473. 

Issues I have not addressed: 
* Code Health: introducing `P2TR` in the existing `swap/htlc.go` code is a bit messy, because we can't use all output types with 
* Resume swap: as is the code fails when we restart + have a `P2TR` swap ongoing, fixing the above code health issuewill likely help with this
* Loopin: TBD, doesn't work rn
* Keyspend: this code negotiates a combined musig key for our keyspend, but does not implement the spend path (currently waiting on lnd code to allow tweaked keyspend)


carlaKC:
Closing for takeover.

Issue #460, opened 2022-02-13
test: autoloop test flake


bhandras:
Found this flake using the CI build.

```=== RUN   TestAutoLoopInEnabled
    autoloop_testcontext_test.go:271: 
        	Error Trace:	autoloop_testcontext_test.go:271
        	            				autoloop_test.go:601
        	Error:      	Not equal: 
        	            	expected: 80000
        	            	actual  : 160000
        	Test:       	TestAutoLoopInEnabled
    autoloop_testcontext_test.go:271: 
        	Error Trace:	autoloop_testcontext_test.go:271
        	            				autoloop_test.go:601
        	Error:      	Not equal: 
        	            	expected: 160000
        	            	actual  : 80000
        	Test:       	TestAutoLoopInEnabled
    autoloop_testcontext_test.go:310: 
        	Error Trace:	autoloop_testcontext_test.go:310
        	            				autoloop_test.go:601
        	Error:      	Not equal: 
        	            	expected: &loop.LoopInRequest{Amount:80000, MaxSwapFee:1000, MaxMinerFee:500, HtlcConfTarget:10, LastHop:(*route.Vertex)(0x1c522e0), ExternalHtlc:false, Label:"[reserved]: autoloop-in", Initiator:"autoloop", Private:false, RouteHints:[][]zpay32.HopHint(nil)}
        	            	actual  : &loop.LoopInRequest{Amount:160000, MaxSwapFee:1000, MaxMinerFee:500, HtlcConfTarget:10, LastHop:(*route.Vertex)(0xc000044e40), ExternalHtlc:false, Label:"[reserved]: autoloop-in", Initiator:"autoloop", Private:false, RouteHints:[][]zpay32.HopHint(nil)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 (*loop.LoopInRequest)({
        	            	- Amount: (btcutil.Amount) 80000,
        	            	+ Amount: (btcutil.Amount) 160000,
        	            	  MaxSwapFee: (btcutil.Amount) 1000,
        	            	@@ -6,3 +6,3 @@
        	            	  LastHop: (*route.Vertex)((len=33) {
        	            	-  00000000  01 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
        	            	+  00000000  02 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
[...]

PR #483, opened 2022-04-22
liquidity: fix TestAutoLoopInEnabled flakes

arshbot:
In this commit we fix the flakes in TestAutoLoopInEnabled by removing the expectation that quotes are returned in a particular order. 

Reviewers, this is a draft and not mergable code. Rather I request that the approach of removing the order expectation be reviewed (mergable code will include a better solution than O(n^2) and will ensure that duplicate value bugs will not be introduced).

Solves https://github.com/lightninglabs/loop/issues/460

### DIFF
--- a/liquidity/autoloop_testcontext_test.go
+++ b/liquidity/autoloop_testcontext_test.go
@@ -2,6 +2,7 @@ package liquidity
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	"github.com/btcsuite/btcd/btcutil"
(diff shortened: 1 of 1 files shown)


liquidity/autoloop_testcontext_test.go
```diff
@@ -300,14 +320,27 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
...
 		actual := <-c.inRequest
 
-		assert.Equal(c.t, expected.request, actual)
+		check = false
+		for _, expected := range step.expectedIn {
+			if reflect.DeepEqual(expected.request, actual) {
+				check = true
+			}
+		}
+		assert.Truef(c.t, check, "Expected:\n%v\n\nBut got:\n%v\n", expected.request, actual)
```
arshbot, line 343: I was unable to get rid of the flake here, kept getting 

```
--- FAIL: TestAutoLoopInEnabled (0.00s)
    autoloop_testcontext_test.go:343: 
                Error Trace:    autoloop_testcontext_test.go:343
                                                        autoloop_test.go:601
                Error:          Should be true
                Test:           TestAutoLoopInEnabled
                Messages:       Expected:
                                &{0.0008 BTC 0.00001 BTC 0.000005 BTC 10 010000000000000000000000000000000000000000000000000000000000000000 false [reserved]: autoloop-in autoloop false []}
                            
                                But got:
                                &{0.0016 BTC 0.00001 BTC 0.000005 BTC 10 020000000000000000000000000000000000000000000000000000000000000000 false [reserved]: autoloop-in autoloop false []}
```

liquidity/autoloop_testcontext_test.go
```diff
@@ -266,29 +267,48 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
...
 		request := <-c.quoteRequestIn
-		assert.Equal(
-			c.t, expected.request.Amount, request.Amount,
-		)
 
-		assert.Equal(
-			c.t, expected.request.HtlcConfTarget,
-			request.HtlcConfTarget,
-		)
+		check = false
```
yyforyongyu, line 274: Maybe we could cache the `request` received, order them and compare them against `step.quotesIn`?

liquidity/autoloop_testcontext_test.go
yyforyongyu: Maybe check what are the returned values from `<-c.inRequest`?

bhandras:
@arshbot do you plan to continue this PR?

PR #495, opened 2022-05-19
loopout: reject P2PK addresses

sputn1ck:
This PR adds a check to reject unsupported address formats such as P2PK (e.g. node pubkeys) or P2TR addresses

Should I add the fix to the `release_notes.md`?

references https://github.com/lightninglabs/lightning-terminal/issues/368

### DIFF
--- a/loopd/swapclient_server.go
+++ b/loopd/swapclient_server.go
@@ -55,6 +55,12 @@ var (
 	errBalanceTooLow = errors.New(
 		"channel balance too low for loop out amount",
 	)
+
+	// errInvalidAddress is returned when the destination address is of
+	// an unsupported format such as P2PK or P2TR addresses.
+	errInvalidAddress = errors.New(
+		"invalid or unsupported address",
+	)
 )
 
 // swapClientServer implements the grpc service exposed by loopd.
@@ -1152,6 +1158,18 @@ func validateLoopOutRequest(ctx context.Context, lnd lndclient.LightningClient,
 	if !sweepAddr.IsForNet(chainParams) {
 		return 0, fmt.Errorf("%w: Current active network is %s",
 			errIncorrectChain, chainParams.Name)
+	}
+
+	// Check that the provided destination address is a supported
+	// address format.
+	switch sweepAddr.(type) {
+	case *btcutil.AddressWitnessScriptHash,
+		*btcutil.AddressWitnessPubKeyHash,
+		*btcutil.AddressScriptHash,
+		*btcutil.AddressPubKeyHash:
+
+	default:
+		return 0, errInvalidAddress
 	}
 
 	// Check that the label is valid.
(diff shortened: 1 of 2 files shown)


loopd/swapclient_server.go
```diff
@@ -1154,6 +1160,12 @@ func validateLoopOutRequest(ctx context.Context, lnd lndclient.LightningClient,
 			errIncorrectChain, chainParams.Name)
 	}
 
+	// Check that the provided destination address is not of type P2PK.
+	switch sweepAddr.(type) {
+	case *btcutil.AddressPubKey:
```
guggero, line 1165: Should we instead have a `switch` with all allowed cases (that don't do anything) and then a `default` case that returns a more generic error?
That way we'd also disallow legacy (1xxx) addresses.

sputn1ck: What are the allowed cases? Everything besides `p2pk`? So `p2pkh` `p2wpkh` `p2wsh` `p2tr`?

guggero: It looks like we only allow `p2wsh`, `p2wpkh`, `p2sh` and `p2pkh`: https://github.com/lightninglabs/loop/blob/master/sweep/sweeper.go#L108

We'll probably need to add `p2tr` at some point later. We have to match the switch linked to above, otherwise we get the error as seen in the issue.

sputn1ck: Pushed the changes

loopd/swapclient_server.go
```diff
@@ -1154,6 +1160,21 @@ func validateLoopOutRequest(ctx context.Context, lnd lndclient.LightningClient,
 			errIncorrectChain, chainParams.Name)
 	}
 
+	// Check that the provided destination address is a supported
+	// address format.
+	switch sweepAddr.(type) {
+	case *btcutil.AddressWitnessScriptHash:
```
guggero, line 1166: nit: the `break` in Golang is implicit. So they aren't needed. And you can use an enumeration of values in a `case` statement, which I think is pretty handy too. And we usually add an extra line after a `case` body (except for the last). So this could look just like:

```
switch sweepAddr.(type) {
	case *btcutil.AddressWitnessScriptHash,
		*btcutil.AddressWitnessPubKeyHash,
		*btcutil.AddressScriptHash,
		*btcutil.AddressPubKeyHash,

	default:
		return 0, errInvalidAddress
	}
```

guggero (approved):
Only a style nit left, everything else looks good to me. Thanks for the quick fix!

guggero (approved):

loopd/swapclient_server.go
```diff
@@ -1154,6 +1160,18 @@ func validateLoopOutRequest(ctx context.Context, lnd lndclient.LightningClient,
 			errIncorrectChain, chainParams.Name)
 	}
 
+	// Check that the provided destination address is a supported
+	// address format.
+	switch sweepAddr.(type) {
```
arshbot, line 1165: Very clean!

arshbot (approved):
LGTM!

PR #496, opened 2022-05-20
Dockerize linter, add gosimports

sputn1ck:
This PR dockerizes the golangci-lint setup and adds `gosimports` to `make fmt`


.golangci.yml
```diff
@@ -2,45 +2,136 @@ run:
...
+  - forcetypeassert
+  - nilerr
+  - nilnil
+  - stylecheck
+  - thelper
+
+  # Additions compared to LND
+  - gocritic
+  - exhaustruct
+  - nonamedreturns
```
guggero, line 113: Are there many named returns? I personally don't like them, so it would be also an option to fix those instead of disabling the linter. But that's just my opinion :sweat_smile:

tools/tools.go
```diff
@@ -0,0 +1,8 @@
+//go:build tools
+// +build tools
+
+package nautilus
```
guggero, line 4: nope ;-)

tools/go.mod
```diff
@@ -2,4 +2,7 @@ module github.com/lightninglabs/loop/tools
 
 go 1.16
 
-require github.com/golangci/golangci-lint v1.46.2 // indirect
+require (
+	github.com/golangci/golangci-lint v1.46.2 // indirect
+	github.com/rinchsan/gosimports v0.1.5 // indirect
```
guggero, line 7: Need to add `gosimports` to `tools.go

guggero (approved):
tACK, LGTM :tada:

tools/tools.go
bhandras: pls rename

bhandras (approved):
LGTM :tada:

.golangci.yml
sputn1ck: I copied `.golangci.yml` from another project, but forgot to check if the linters have issues. There are no named returns. Exhauststruct throws a lot of complaints though. `gocritic` only has an issue with https://github.com/lightninglabs/loop/blob/1741896e336213e9b219d47fa0b095296e6ad45d/loopd/swapclient_server.go#L317 citing `appendAssign: append result not assigned to the same slice (gocritic)` Should I change the code and leave the `gocritic` linter in?

guggero: I think `gocritic` finds good things in general. So if this is the only issue, I think a `//nolint` comment would be preferable if it helps us catch future `append()` related bugs.

sputn1ck:
I added the requested changes

PR #477, opened 2022-04-03
multi: changes to the taproot HTLC  required for both client and server

bhandras:
This PR adds a few changes to the already existing taproot HTLC code to make it simpler to integrate.
- Refactor locking conditions.
- Derive the shared internal public key when creating the HTLC and export the root hash.

PR #497, opened 2022-05-20
multi: upgrade to using P2TR htlcs and added support for MuSig2 loopout sweep

bhandras:
Adds support for P2TR loops and an optional MuSig2 loopout sweep.

Depends on: 
- ~https://github.com/lightninglabs/loop/pull/477~
- https://github.com/lightninglabs/loop/pull/506

#### Pull Request Checklist


loopd/swapclient_server.go
```diff
@@ -665,17 +681,20 @@ func (s *swapClientServer) LoopIn(ctx context.Context,
...
-		ServerMessage:    swapInfo.ServerMessage,
+		Id:            swapInfo.SwapHash.String(),
+		IdBytes:       swapInfo.SwapHash[:],
+		ServerMessage: swapInfo.ServerMessage,
 	}
 
-	if req.ExternalHtlc {
+	if loopdb.CurrentProtocolVersion() < loopdb.ProtocolVersionHtlcV3 {
+		p2wshAddr := swapInfo.HtlcAddressP2WSH.String()
+		response.HtlcAddress = p2wshAddr // nolint
```
yyforyongyu, line 699: Curious, what did the linter say?

loopdb/protocol_version.go
```diff
@@ -59,18 +59,45 @@ const (
...
+
+// EnableNextProtocolVersion sets the current protocol version to the next one.
+func EnableNextProtocolVersion() {
+	currentRPCProtocolVersion = nextRPCProtocolVersion
+}
+
+// ResetCurrentProtocolVersion resets the current protocol version to the one
+// before the next version.
+// Note: Test only!
+func ResetCurrentProtocolVersion() {
```
yyforyongyu, line 94: If only used in test, maybe we could make it a private method?

loopd/swapclient_server.go
yyforyongyu: Also do we need to check `req.ExternalHtlc` here?

loopin.go
```diff
@@ -605,16 +633,29 @@ func (s *loopInSwap) waitForHtlcConf(globalCtx context.Context) (
 
 	notifier := s.lnd.ChainNotifier
 
-	confChanP2WSH, confErrP2WSH, err := notifier.RegisterConfirmationsNtfn(
-		ctx, s.htlcTxHash, s.htlcP2WSH.PkScript, 1, s.InitiationHeight,
-	)
+	notifyConfirmation := func(htlc *swap.Htlc) (
+		chan *chainntnfs.TxConfirmation, chan error, error) {
+
+		if htlc == nil {
```
yyforyongyu, line 639: Do we allow all three htlcs to be nil here? If not, I think we could do switch with cases `htlc != nil` and a default case to error out.

loopin_test.go
```diff
@@ -87,10 +101,15 @@ func TestLoopInSuccess(t *testing.T) {
 	require.NotNil(t, state.HtlcTxHash)
 	require.Equal(t, cost, state.Cost)
 
-	// Expect register for htlc conf.
-	<-ctx.lnd.RegisterConfChannel
+	// Expect register for htlc conf (only one, the since the htlc is p2tr).
```
yyforyongyu, line 104: nit: typo

loopin_test.go
```diff
@@ -87,10 +101,15 @@ func TestLoopInSuccess(t *testing.T) {
...
 	<-ctx.lnd.RegisterConfChannel
 
+	// If the swap is legacy, then we'll register two confirmation
+	// notifications.
+	if !s.isTaproot() {
+		<-ctx.lnd.RegisterConfChannel
```
yyforyongyu, line 110: hmm..why do we need this extra conf? I think when `conf != nil` we'd exit the for loop in `waitForHtlcConf` right?

loopin_test.go
```diff
@@ -220,11 +263,12 @@ func testLoopInTimeout(t *testing.T,
 	} else {
 		// Create an external htlc publish tx.
 		var pkScript []byte
-		if outputType == swap.HtlcNP2WSH {
+		if !s.isTaproot() {
```
yyforyongyu, line 266: What about `htlcNP2WSH`?

loopd/swapclient_server.go
```diff
@@ -217,20 +217,36 @@ func (s *swapClientServer) marshallSwap(loopSwap *loop.SwapInfo) (
...
+		htlcAddressNP2WSH string
+	)
 	var outGoingChanSet []uint64
 	var lastHop []byte
 
 	switch loopSwap.SwapType {
 	case swap.TypeIn:
 		swapType = clientrpc.SwapType_LOOP_IN
-		htlcAddressP2WSH = loopSwap.HtlcAddressP2WSH.EncodeAddress()
+		scriptVersion := loop.GetHtlcScriptVersion(
```
yyforyongyu, line 240: One thing I've noticed here is the version info is implicitly encoded in the htlc address already. By checking `loopSwap.HtlcAddressP2TR` against nil we'd know the current version. This could clear the logical branches a bit, here and a few other places. However it's not as explicit, so no strong opinions here.

client.go
```diff
@@ -499,7 +499,23 @@ func (s *Client) LoopOutQuote(ctx context.Context,
...
+		SwapFee:         quote.SwapFee,
+		MinerFee:        minerFee,
+		PrepayAmount:    quote.PrepayAmount,
+		SwapPaymentDest: quote.SwapPaymentDest,
+	}, nil
+}
+
+// getLoopOutSweepFee is a helper method to estimate the loop out htlc sweep
+// fee to a p2wsh address.
+func (s *Client) getLoopOutSweepFee(ctx context.Context, confTarget int32) (
```
yyforyongyu, line 517: 👍

loopdb/codec.go
```diff
@@ -40,3 +42,31 @@ func MarshalProtocolVersion(version ProtocolVersion) []byte {
 
 	return versionBytes[:]
 }
+
+// MarshalKeyLocator marshals a keychain.KeyLocator to a byte slice.
+func MarshalKeyLocator(keyLocator keychain.KeyLocator) []byte {
```
yyforyongyu, line 47: I think we could use `channeldb`'s `EKeyLocator` and `DKeyLocator` here.

loopdb/loop.go
```diff
@@ -26,6 +27,11 @@ type SwapContract struct {
 	// HTLC.
 	ReceiverKey [33]byte
 
+	// KeyLocator is the key locator (family and index) for the receiver
+	// key if this is a loop out contract, or the sender key if this is a
+	// loop in contract.
+	KeyLocator keychain.KeyLocator
```
yyforyongyu, line 33: nit: maybe name this to be remote/server/peer key to help us understand that this is used for key aggregation.

loopout_test.go
```diff
@@ -581,6 +581,11 @@ func testPreimagePush(t *testing.T) {
 	// preimage is not revealed, we also do not expect a preimage push.
 	expiryChan <- testTime
 
+	// Since we don't have a reliable mechanism to non-intursively avoid
```
yyforyongyu, line 638: typo: intursively -> intrusively

yyforyongyu: Also what happens when it's too soon?

swap_server_client.go
```diff
@@ -121,6 +121,13 @@ type swapServerClient interface {
 		swapHash lntypes.Hash, paymentAddr [32]byte,
 		plugin RoutingPluginType, success bool, attempts int32,
 		totalTime int64) error
+
+	// MuSig2SignSweep calls the server for to copperatively sign the MuSig2
```
yyforyongyu, line 125: typo: copperatively

swap_server_client.go
```diff
@@ -720,6 +727,32 @@ func (s *grpcSwapServerClient) ReportRoutingResult(ctx context.Context,
 	return err
 }
 
+// MuSig2SignSweep calls the server for to copperatively sign the MuSig2 htlc
```
yyforyongyu, line 730: same typo

yyforyongyu (changes requested):
Great job! This is surely a lot of work! Down to last commit, no major comments from me. Also do we have docs re the new flow under musig2 where I could grasp the full picture?

loopd/daemon.go
```diff
@@ -158,6 +158,12 @@ func (d *Daemon) Start() error {
 	return nil
 }
 
+// ProtocolVersion is a helper to be able to retrieve the protocol version of
+// the daemon in tests.
```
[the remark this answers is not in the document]
bhandras: It's more about the integration tests on the server side.

loopd/run.go
```diff
@@ -191,6 +192,11 @@ func Run(rpcCfg RPCConfig) error {
 		return err
 	}
 
+	// Enable the next protocol version if requested.
+	if config.EnableNextProtocol {
```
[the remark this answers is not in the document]
arshbot: I _believe_ this is necessary for the loop cli, seems like this function is only called from `cmd/loopd`

loopin.go
arshbot: What is meant by `Do we allow all three htlcs to be nil here?`?

loopin_test.go
```diff
@@ -27,6 +28,19 @@ var (
 // TestLoopInSuccess tests the success scenario where the swap completes the
 // happy flow.
 func TestLoopInSuccess(t *testing.T) {
+	t.Run("legacy protocol", func(t *testing.T) {
+		testLoopInSuccess(t)
+	})
+
+	t.Run("next protocol", func(t *testing.T) {
```
arshbot, line 35: should this be in a separate test?

loopin_test.go
```diff
@@ -47,13 +61,13 @@ func TestLoopInSuccess(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	swap := initResult.swap
+	s := initResult.swap
```
arshbot, line 64: nit: shouldn't have single char variables

loopin_test.go
arshbot: perhaps the first conf is for the htlc, and the second is for the htlc redemption conf? also curious, but the previous version of this code has two confs. taproot coop redemption wouldn't need two though which is my guess as to why it's purposefully excluded

loopin_test.go
```diff
@@ -411,7 +486,7 @@ func testLoopInResume(t *testing.T, state loopdb.SwapState, expired bool,
 		t.Fatal(err)
 	}
 
-	swap, err := resumeLoopInSwap(
+	s, err := resumeLoopInSwap(
```
arshbot, line 489: nit: single char variable

loopin.go
yyforyongyu: The three types, `s.htlcP2WSH`, `s.htlcNP2WSH` and `s.htlcP2TR`.

loopdb/codec.go
```diff
@@ -40,3 +45,41 @@ func MarshalProtocolVersion(version ProtocolVersion) []byte {
 
 	return versionBytes[:]
 }
+
+// MarshalKeyLocator marshals a keychain.KeyLocator to a byte slice.
+func MarshalKeyLocator(keyLocator keychain.KeyLocator) ([]byte, error) {
+	var (
+		scratch [8]byte
```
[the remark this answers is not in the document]
yyforyongyu: This is used as a buffer in the encoding below, which is then used by tlv encoding.

loopdb/store.go
```diff
@@ -102,6 +102,16 @@ var (
 	// parameters.
 	liquidtyParamsKey = []byte("params")
 
+	// keyLocatorKey is the key that stores the receiver key's locator info
+	// for loop outs or the sender key's locator infor for loop ins. This is
```
yyforyongyu, line 106: typo: infor -> info

swapserverrpc/server.proto
```diff
@@ -526,3 +528,21 @@ message ReportRoutingResultReq {
 
 message ReportRoutingResultRes {
 }
+
+message MuSig2SignSweepReq {
+    ProtocolVersion protocol_version = 1;
```
yyforyongyu, line 533: missing docs.

loopout.go
```diff
@@ -1113,23 +1135,92 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+					// invoice. So that we can continue
+					// with the MuSig2 sweep afterwards.
+					s.pushPreimage(ctx)
+				}
+
+				// Now attempt to publish a MuSig2 sweep txn.
+				// Only attempt at most maxMusigSweepRetires
+				// times to still leave time for an emergency
+				// script path sweep.
+				if musigSweepTryCount < maxMusigSweepRetires {
```
yyforyongyu, line 1164: Compared to the old flow, this does give the server extra 100s to cheat us. Since mining 20 blocks within that time is extremely unlikely so it doesn't matter. Only something to be noted if we decide one day to change. the value of `MinLoopOutPreimageRevealDelta`.

yyforyongyu (changes requested):
Finished another round. I think lots of the complexity comes from the fact that we are making the new version optional. It'd also be nice to have a doc explaining the current flow to address the extra communication round brought by musig. I notice there's the `architecture.md` doc, which is a bit outdated. Given how large this PR is, maybe a new issue and a future PR?

loopout.go
bhandras: Was curious about this so I looked it up. So the probability of finding k blocks in a row follows a Poisson[1] distribution, assuming 10 minute block time: `P(k, t) = (t/10)^k * e^(-t/10) / k!)`, where `t` is minutes, in this case it's 1.5 (100/60) and `k` is 20 (blocks in a row), which gives us `P(20, 1.5) = 0.15^20 * e^-0.15 / 20!`, so in Python `math.pow(0.15, 20) * math.pow(math.e, -0.15) / math.factorial(20)` which gives us  `1.176403727573749e-35`. Hope I'm not misunderstanding how to calculate :)

[1] https://en.wikipedia.org/wiki/Poisson_distribution

loopout.go
```diff
@@ -1238,24 +1329,115 @@ func (s *loopOutSwap) failOffChain(ctx context.Context, paymentType paymentType,
...
-// account the max miner fee and marks the preimage as revealed when it
-// published the tx. If the preimage has not yet been revealed, and the time
-// during which we can safely reveal it has passed, the swap will be marked
-// as failed, and the function will return.
-//
-// TODO: Use lnd sweeper?
-func (s *loopOutSwap) sweep(ctx context.Context,
-	htlcOutpoint wire.OutPoint,
-	htlcValue btcutil.Amount) error {
+// createMuSig2SweepTxn creats a taproot keyspend sweep transaction and attempts
```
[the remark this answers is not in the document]
bhandras: done

loopout.go
```diff
@@ -1286,42 +1468,79 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
-	)
-	if err != nil {
-		return err
-	}
+	return confTarget, true
+}
 
+// clampSweepFee will clamp the passed in sweep fee to the maximum configured
+// miner fee.
+func (s *loopOutSwap) clampSweepFee(fee btcutil.Amount) (btcutil.Amount, bool) {
```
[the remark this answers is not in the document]
bhandras: Added some more comments to clarify, ptal.

loopout.go
```diff
@@ -1286,42 +1468,79 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
-		if preimageRevealed {
+		if s.state == loopdb.StatePreimageRevealed {
 			// The currently required fee exceeds the max, but we
 			// already revealed the preimage. The best we can do now
 			// is to republish with the max fee.
 			fee = s.MaxMinerFee
 		} else {
 			s.log.Warnf("Not revealing preimage")
-			return nil
+			return 0, false
```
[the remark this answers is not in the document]
bhandras: It's just the default value that's returned and a `false` meaning that the sweep shouldn't be continued as fees are too high. Note that this only applies to the p2wsh sweep as in the musig case we reveal the preimage the server in order to gain trust for the cooperative signing process.

loopout.go
```diff
@@ -1286,42 +1468,79 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
-	// Create sweep tx.
-	sweepTx, err := s.sweeper.CreateSweepTx(
-		ctx, s.height, s.htlc.SuccessSequence(), s.htlc, htlcOutpoint,
-		s.ReceiverKey, redeemScript, witnessFunc, htlcValue, fee,
-		s.DestAddr,
+	return fee, true
+}
+
+// sweepMuSig2 attempts to sweep the on-chain HTLC using MuSig2. If anything
+// fails, we'll log it but will simply return to allow retries or an
```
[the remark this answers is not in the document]
bhandras: ptal, hope it's cleaner now

loopout.go
```diff
@@ -1286,42 +1468,79 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
+// uncooperative sweep if everything fails.
+func (s *loopOutSwap) sweepMuSig2(ctx context.Context,
+	htlcOutpoint wire.OutPoint, htlcValue btcutil.Amount) bool {
+
+	addInputToEstimator := func(e *input.TxWeightEstimator) error {
+		e.AddTaprootKeySpendInput(txscript.SigHashDefault)
+		return nil
+	}
+
+	confTarget, _ := s.sweepConfTarget()
```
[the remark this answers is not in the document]
bhandras: Because we need to continue sweeping regardless as the preimage is revealed to the server at this point.

loopout.go
```diff
@@ -1286,42 +1468,79 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
 	if err != nil {
-		return err
+		s.log.Warnf("Failed to estimate fee MuSig2 sweep txn: %v", err)
+		return false
 	}
 
-	// Before publishing the tx, already mark the preimage as revealed. This
-	// is a precaution in case the publish call never returns and would
-	// leave us thinking we didn't reveal yet.
+	fee, _ = s.clampSweepFee(fee)
```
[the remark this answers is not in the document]
bhandras: No, again this is because the client revealed the preimage to the server to make the server trust our musig sweep attempts and sign for us.

loopout.go
```diff
@@ -1331,6 +1550,59 @@ func (s *loopOutSwap) sweep(ctx context.Context,
 		}
 	}
 
+	return nil
+}
+
+// sweep tries to sweep the given htlc to a destination address. It takes into
+// account the max miner fee and marks the preimage as revealed when it
```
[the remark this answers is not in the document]
bhandras: ptal

loopout.go
```diff
@@ -1331,6 +1550,59 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
+// published the tx. If the preimage has not yet been revealed, and the time
+// during which we can safely reveal it has passed, the swap will be marked
+// as failed, and the function will return.
+func (s *loopOutSwap) sweep(ctx context.Context, htlcOutpoint wire.OutPoint,
+	htlcValue btcutil.Amount) error {
+
+	confTarget, canSweep := s.sweepConfTarget()
+	if !canSweep {
```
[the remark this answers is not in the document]
bhandras: The flow is a bit hard to understand but in the normal sweep case (no musig), `sweepConfTarget` will mark the swap as failed if the deadline is exceeded and the preimage is not revealed and we'll at that point fail the swap. In the musig case this function is called if the coop signing fails and at that point we must we go on since the preimage is revealed so we'll try to sweep until it succeeds.

bhandras (commented):
Thank you both for the thorough reviews. Tried to address all comments and did some cleanup regarding the experimental flag. PTAL.

loopin_test.go
arshbot: This code doesn't account for the htlcV1 path

loopout.go
```diff
@@ -1113,23 +1139,92 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
+				// swap has reached a final state, then we
+				// return nil spend details, because there is no
+				// further action required for this swap.
+				if s.state.Type() != loopdb.StateTypePending {
+					return nil, nil
+				}
+
+				// If our off chain payment is not yet complete,
+				// we try to push our preimage to the server.
+				if !paymentComplete {
```
arshbot, line 1225: In the case where the fees are more than our max fee limit, it seems like we would push the preimage to the server over and over again without even trying to sweep.

loopout.go
```diff
@@ -1273,42 +1472,84 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
 			fee, s.MaxMinerFee)
 
-		if preimageRevealed {
+		if s.state == loopdb.StatePreimageRevealed {
 			// The currently required fee exceeds the max, but we
 			// already revealed the preimage. The best we can do now
 			// is to republish with the max fee.
 			fee = s.MaxMinerFee
 		} else {
 			s.log.Warnf("Not revealing preimage")
```
arshbot, line 1494: I don't think this would be the case actually (at least in the below flow). Might need to be amended to "Not revealing preimage onchain"

loopout.go: 1210
```
err := s.sweep(ctx, htlcOutpoint, htlcValue)
 └─►	fee, canSweep = s.clampSweepFee(fee)
 
if s.state.Type() != loopdb.StateTypePending {
	return nil, nil
}

if !paymentComplete {
	s.pushPreimage(ctx)   ◄──── [we reveal preimage]
}
```

loopout.go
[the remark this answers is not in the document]
arshbot: can't return nil here because of `btcutil.Amount`

sweep/sweeper.go
```diff
@@ -19,6 +19,66 @@ type Sweeper struct {
...
+		return nil, nil, err
+	}
+
+	sweepTx.AddTxOut(&wire.TxOut{
+		PkScript: sweepPkScript,
+		Value:    int64(amount - fee),
+	})
+
+	// We need our previous outputs for taproot spends, and there's no
+	// harm including them for segwit v0, so we always inlucde our prevOut.
```
arshbot, line 56: nit: typo

sweep/sweeper.go
```diff
@@ -19,6 +19,66 @@ type Sweeper struct {
 	Lnd *lndclient.LndServices
 }
 
+// CreateUnsignedTaprootKeySpendSweepTx creates a taproot htlc sweep tx using
+// keyspend. Returns the raw unsigned txn, the sighash or an error.
```
arshbot, line 23: Thought it was saying it returned txn or sighash or error

```suggestion
// keyspend. Returns the raw unsigned txn and the sighash or an error.
```

loopout.go
```diff
@@ -1225,24 +1333,115 @@ func (s *loopOutSwap) failOffChain(ctx context.Context, paymentType paymentType,
...
+	if err != nil {
+		return nil, err
+	}
+
+	var serverPublicNonce [musig2.PubNonceSize]byte
+	copy(serverPublicNonce[:], serverNonce)
+
+	// Register the server's nonce before attempting to create our partial
+	// signature.
+	_, err = s.lnd.Signer.MuSig2RegisterNonces(
```
arshbot, line 1389: we should check if this is true. If it isn't something unexpected has happened, and we are not in session with only two participants (assuming the server has already registered it's nonce)

loopout.go
```diff
@@ -1225,24 +1333,115 @@ func (s *loopOutSwap) failOffChain(ctx context.Context, paymentType paymentType,
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !haveAllSigs {
+		return nil, fmt.Errorf("failed to combine signatures")
+	}
+
+	// To be sure that we're good, parse and validate that the combined
+	// signature is indeed valid for the sig hash and the internal pubkey.
```
arshbot, line 1423: :+1:

swap/htlc.go
```diff
@@ -862,7 +862,7 @@ func (h *HtlcScriptV3) GenTimeoutWitness(
 // IsSuccessWitness checks whether the given stack is valid for
 // redeeming the htlc.
 func (h *HtlcScriptV3) IsSuccessWitness(witness wire.TxWitness) bool {
-	return len(witness) == 4
+	return len(witness) == 4 || len(witness) == 1
```
arshbot, line 865: bless :relieved:

arshbot (commented):
I think i just finished my first round lmao, this was a chunko - good shit @bhandras, looks good, left all the nits in the world and some comments on possibly preemptively pushing preimage (alliteration ftw)

loopout.go
bhandras: Fortunately this is handled as `clampSwapFee` will simply clamp the fee at the maximum requested miner fee and the sweep will be published as the preimage is revealed. Maybe a bit hard to see from the diff but if you check out the whole branch it's more visible.

loopout.go
bhandras: I think this confusion comes from the old flow being somewhat convoluted. The `pushPreimage()` function doesn't actually push the preimage to the server unless the swap state is `StatePreimageRevealed` which in the non-musig case happens only after the sweep has txn has been sent (not necessarily confirmed yet).

loopout.go
```diff
@@ -1273,42 +1468,79 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
+	fee, err := s.sweeper.GetSweepFee(
+		ctx, addInputToEstimator, s.DestAddr, confTarget,
 	)
 	if err != nil {
-		return err
+		s.log.Warnf("Failed to estimate fee MuSig2 sweep txn: %v", err)
+		return false
 	}
 
-	// Before publishing the tx, already mark the preimage as revealed. This
```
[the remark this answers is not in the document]
bhandras: It's still there just the diff is a bit scattered.

sweep/sweeper.go
bhandras: done

sweep/sweeper.go
bhandras: done

swap/htlc.go
bhandras: Added some comments to clarify too.

loopout.go
```diff
@@ -1113,23 +1135,92 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmed(globalCtx context.Context,
...
-			// for this swap.
-			if s.state.Type() != loopdb.StateTypePending {
-				return nil, nil
-			}
+				if !paymentComplete {
+					// Push the preimage for as long as the
+					// server is able to settle the swap
+					// invoice. So that we can continue
+					// with the MuSig2 sweep afterwards.
+					s.pushPreimage(ctx)
```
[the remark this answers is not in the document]
bhandras: I think this is a fair point as if the client was stopped for a long time but we didn't reveal the preimage yet and we're also less than 20 blocks from reveal we can abort the swap here. In this case we won't reveal the preimage to the server nor will we attempt to sweep (the off-chain payments will time out too).

Added logic above to cover this case, ptal.

bhandras (commented):
Thanks for the additional comments @arshbot, great review!

yyforyongyu (commented):
Pretty close! Just finished another round. I think we should ditch the thought of maintaining another branch as it would cause much pain. And if the current approach is determined, no major comments and just a few nits/typos.

An alternative, tho, is to make two versions here. 
- `ProtocolVersionHtlcV3`, where we upgrade our htlc to use taproot. For loopout, the flow is unchanged as the server will be still extracting preimage from the witness stack. The key spend path is still musig, but not allowed.
- `ProtocolVersionMusig2`, an experimental version where we upgrade our flow for loopout.

By upgrading our current protocol to `ProtocolVersionHtlcV3`, we'd be able to remove several places where we determine the htlc script based on the current version and should be able to alleviate some pain from both server and client sides.  By only making the musig part optional, I think the only place we need to create logical branches is when we `waitForHtlcSpendConfirmed`.

loopd/daemon.go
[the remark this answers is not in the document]
bhandras: Different address space for a package variable than the same variable that's used as part of the loop daemon.

loopin_test.go
```diff
@@ -220,11 +265,12 @@ func testLoopInTimeout(t *testing.T,
 	} else {
 		// Create an external htlc publish tx.
 		var pkScript []byte
-		if outputType == swap.HtlcNP2WSH {
-			pkScript = s.htlcNP2WSH.PkScript
+		if !inSwap.isTaproot() {
```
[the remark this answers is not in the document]
bhandras: No, let me clarify. When taproot and musig is active we'll use p2tr for both external and internal case. Otherwise p2wsh is internal and np2wsh is external so no changes.

loopout_test.go
```diff
@@ -24,6 +24,22 @@ import (
...
 func TestLoopOutPaymentParameters(t *testing.T) {
+	t.Run("legacy protocol", func(t *testing.T) {
+		testLoopOutPaymentParameters(t)
+	})
+
+	t.Run("next protocol", func(t *testing.T) {
+		loopdb.EnableNextProtocolVersion()
+		defer loopdb.ResetCurrentProtocolVersion()
+
+		testLoopOutPaymentParameters(t)
```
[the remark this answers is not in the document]
bhandras: I think it's necessary as the underlying variable tracking the current protocol version is global (in the `loopdb`) package.

loopdb/codec_test.go
```diff
@@ -51,3 +53,40 @@ func TestProtocolVersionMarshalUnMarshal(t *testing.T) {
...
+
+// TestKeyLocatorMarshalUnMarshal tests that marshalling and unmarshalling
+// keychain.KeyLocator works correctly.
+func TestKeyLocatorMarshalUnMarshal(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		keyLoc keychain.KeyLocator
+	}{
+		{
```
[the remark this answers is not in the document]
bhandras: done

loopout.go
```diff
@@ -1286,42 +1497,84 @@ func (s *loopOutSwap) sweep(ctx context.Context,
...
-		ctx, s.htlc.AddSuccessToEstimator, s.DestAddr, confTarget,
-	)
-	if err != nil {
-		return err
-	}
+	return confTarget, true
+}
 
+// clampSweepFee will clamp the passed in sweep fee to the maximum configured
+// miner fee. Returns fals if sweeping should not continue. Note that in the
```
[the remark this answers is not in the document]
bhandras: fixed

bhandras:
@yyforyongyu thanks for your quick review. Answered all your questions except one where I think there's some discussion ongoing offline. Since github is mixing things up, it's best to check comments per commit as the main view seems to have lost track of threads.

swap/htlc.go
```diff
@@ -108,17 +108,26 @@ type Htlc struct {
 }
 
 var (
-	quoteKey [33]byte
-
+	dummyPubKey = [33]byte{
```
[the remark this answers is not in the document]
bhandras: Oh sorry, I don't know why I made that comment. It's actually used for the pre-constructed quote htlcs. We need an actual pubkey there since it's parsed when the htlc is constructed. Sorry again for the confusion.

loopout_test.go
```diff
@@ -604,6 +733,18 @@ func TestPreimagePush(t *testing.T) {
 // we have revealed our preimage, demonstrating that we do not reveal our
 // preimage once we've reached our expiry height.
 func TestExpiryBeforeReveal(t *testing.T) {
+	t.Run("stable protocol", func(t *testing.T) {
+		testExpiryBeforeReveal(t)
+	})
+
+	// Note that there's no point of testing this case with the new
+	// protocol where we use taproot htlc and attempt MuSig2 sweep. The
+	// reason being is that the preimage is revealed to the server once the
```
yyforyongyu, line 742:
```suggestion
	// reason is that the preimage is revealed to the server once the
```

sweep/sweeper.go
```diff
@@ -19,6 +19,66 @@ type Sweeper struct {
 	Lnd *lndclient.LndServices
 }
 
+// CreateUnsignedTaprootKeySpendSweepTx creates a taproot htlc sweep tx using
+// keyspend. Returns the raw unsigned txn and the sighash or an error.
+func (s *Sweeper) CreateUnsignedTaprootKeySpendSweepTx(
```
yyforyongyu, line 24: Ideally for new functions we should always add explicit unit tests.

yyforyongyu (approved):
LGTM🎉 We may need to start focusing on code health for this repo. For instance we have `isTaproot` implemented in three places to do the same thing, while I think that's what `swapKit` or `swapContract` are meant to do. Aslo we are missing quite some unit tests. However this PR is already large, so it may be something to do in the future.

arshbot (approved):
LGTM! left a few nits on comments, but ndb.

interface.go
```diff
@@ -351,6 +354,10 @@ type SwapInfo struct {
 	// swap htlc. This is only used for external loop-in.
 	HtlcAddressNP2WSH btcutil.Address
 
+	// HtlcAddressP2TR stores the address of the P2TR (taproot) swap htlc.
+	// This is used for both internal and external loop-in.
```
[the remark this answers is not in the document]
bhandras: done

loopin_test.go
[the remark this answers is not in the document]
bhandras: ah ok, fixed

loopout_test.go
bhandras: done

swap/htlc.go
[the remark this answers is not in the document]
bhandras: done

loopout.go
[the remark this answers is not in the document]
bhandras: Yeah it's returning  `0, false` where the number is automatically casted to `btcutil.Amount`

sweep/sweeper.go
bhandras: Added unit test coverage (separate commit).

bhandras:
> For instance we have `isTaproot` implemented in three places to do the same thing, while I think that's what `swapKit` or `swapContract` are meant to do.

Fixed.

swap.go
```diff
@@ -73,6 +73,12 @@ func GetHtlcScriptVersion(
 	return swap.HtlcV3
 }
 
+// IsTaproot returns true if the swap referenced by the passed swap contract
+// uses the v3 (taproot) htlc.
+func IsTaprootSwap(swapContract *loopdb.SwapContract) bool {
```
yyforyongyu, line 78: nit: is it possible to make this a method on `swapContract`? imo it seems to be a bit cleaner.

loopout_test.go
```diff
@@ -1011,3 +1018,174 @@ func testFailedOffChainCancelation(t *testing.T) {
...
+	server := newServerMock(lnd)
+
+	testReq := *testRequest
+	testReq.SweepConfTarget = 10
+	testReq.Expiry = ctx.Lnd.Height + testLoopOutMinOnChainCltvDelta
+
+	// We set our mock fee estimate for our target sweep confs to be our
+	// max miner fee * 2. With MuSig2 we still expect that the client will
+	// publish the sweep but with the fee clamped to the maximum allowed
+	// miner fee as the preimage is revealed befor the sweep txn is
```
yyforyongyu, line 1042: typo: before

loopout_test.go
```diff
@@ -1011,3 +1018,174 @@ func testFailedOffChainCancelation(t *testing.T) {
...
+	// to reflect that the server received our preimage push and settled off
+	// chain.
+	trackPayment.Updates <- lndclient.PaymentStatus{
+		State: lnrpc.Payment_SUCCEEDED,
+	}
+
+	// Make sure our sweep tx has a single witness indicating keyspend.
+	require.Len(t, sweepTx.TxIn[0].Witness, 1)
+
+	// Finally, we put this swap out of its misery and notify a successful
```
yyforyongyu, line 1183: lol, guess the swap is finally happy hahah

client.go
```diff
@@ -158,6 +160,18 @@ func NewClient(dbDir string, cfg *ClientConfig) (*Client, func(), error) {
 		totalPaymentTimeout: cfg.TotalPaymentTimeout,
 		maxPaymentRetries:   cfg.MaxPaymentRetries,
 		cancelSwap:          swapServerClient.CancelLoopOutSwap,
+		verifySchnorrSig: func(pubKey *btcec.PublicKey, hash, sig []byte) error {
```
yyforyongyu, line 163: Cool so this is made so the test can mock?

yyforyongyu (commented):
Cool thanks for the latest update!

client.go
bhandras: yep

loopout_test.go
bhandras: done

PR #500, opened 2022-05-30
Include provided last_hop and outgoing_chan_set to listswaps rpc

### LINKED ISSUE #499: Include last_hop and outgoing_chan_set for each swap in ListSwaps RPC
**Is your feature request related to a problem? Please describe.**
In Terminal, we display a list of all peers & channels of the connected node. When a swap is performed on a specific channel/peer, it is displayed directly below the channel/peer to make it easy for the user to identify where liquidity is being adjusted. 

Currently, when we make the RPC request to initiate a loop, we provide the `last_hop` parameter for `LoopIn` or `outgoing_chan_set` for `LoopOut`. After the request is made, these two values are not returned in the `ListSwaps` response. When the page is reloaded, it is no longer possible to know which swaps are associated with which channel or peer. 

The workaround we've been using has been to persist the mapping of `swap_id` -> `outgoing_chan_set`/`last_hop` in the browser's local storage. The app can then know which swap should be displayed alongside which channel or peer. The drawback of this workaround is that the association is lost if the user visits Terminal from a different device/browser because the mapping is stored in the browser that initiated the swap.

Now that I am working on integrating autoloop into Terminal, I've realized that this workaround is no longer possible. When i add an autoloop rule, the swap is initiated asynchronously.  The app never receives a `swap_id`. It is not possible to know which channel/peer was used from the `ListSwaps` response.

**Describe the solution you'd like**
Update the `ListSwaps` RPC to include `last_hop` for loop ins and `outgoing_chan_set` for for loop outs. This would allow the app to know which channel/peer the autoloop swaps are associated with. This would also eliminate the need for our workaround, therefore resolving the multi-browser issue.

sputn1ck:
This PR adds the initially provided 'last_hop' to loop ins and 'outgoing_chan_set' to loop outs in the 'listswaps' rpc.

I updated the unit tests to include a check for the values and tested with a local regtest setup.

closes #499


interface.go
```diff
@@ -353,6 +353,15 @@ type SwapInfo struct {
 
 	// ExternalHtlc is set to true for external loop-in swaps.
 	ExternalHtlc bool
+
+	// LastHop optionally specifies the last hop to use for the loop in
+	// payment. On a loop ou this field is nil.
```
bhandras, line 358: nit: loop ou => loop out

loopin_test.go
```diff
@@ -54,7 +60,23 @@ func TestLoopInSuccess(t *testing.T) {
 		errChan <- err
 	}()
 
-	ctx.assertState(loopdb.StateInitiated)
+	swapInfo := <-ctx.statusChan
+	if swapInfo.State != loopdb.StateInitiated {
```
bhandras, line 64: nit: why not just use `require.Equal` here and below?

loopout.go
```diff
@@ -1320,6 +1320,13 @@ func (s *loopOutSwap) sweep(ctx context.Context,
 	return nil
 }
 
+func (s *loopOutSwap) swapInfo() *SwapInfo {
+	swapInfo := s.swapKit.swapInfo()
+	swapInfo.OutgoingChanSet = s.OutgoingChanSet
```
bhandras, line 1325: Here the `OutgoungChanSet` is passed by reference, which may be dangerous as ownership is shared.

loopout_test.go
```diff
@@ -90,6 +91,17 @@ func TestLoopOutPaymentParameters(t *testing.T) {
 		t.Fatal("unexpected state")
 	}
 
+	// check that the SwapInfo contains the outgoing chan set
+	if !reflect.DeepEqual(state.OutgoingChanSet, chanSet) {
```
bhandras, line 95: nit: use `require.Equal`

loopout_test.go
```diff
@@ -90,6 +91,17 @@ func TestLoopOutPaymentParameters(t *testing.T) {
...
+		t.Fatal("expected chan set")
+	}
+
+	// check that the SwapInfo does not contain a last hop
+	if state.LastHop != nil {
```
bhandras, line 100: nit: `require.Nil`

loopin.go
```diff
@@ -994,3 +994,10 @@ func (s *loopInSwap) setState(state loopdb.SwapState) {
 	s.lastUpdateTime = time.Now()
 	s.state = state
 }
+
+func (s *loopInSwap) swapInfo() *SwapInfo {
+	swapInfo := s.swapKit.swapInfo()
+	swapInfo.LastHop = s.LastHop
```
bhandras, line 1000: Could be potentially dangerous as the ownership of `LastHop` is shared.

looprpc/client.proto
```diff
@@ -396,6 +396,12 @@ message SwapStatus {
     // Off-chain routing fees
     int64 cost_offchain = 10;
 
+    // Provided last hop
+    bytes provided_last_hop = 16;
```
bhandras, line 400: wdyt about just calling these `last_hop` and `outgoing_chan_set`?

loopin.go
yyforyongyu: I think we should do this in `sendUpdate` where we construct the `info`.

yyforyongyu (changes requested):

looprpc/client.proto
sputn1ck: I was unsure what would be expected as an api user, but I can see that removing `provided` is more consistent with the rest of the api

loopin.go
sputn1ck: I must have somehow missed `sendUpdate`. Makes a lot of sense to move this there

loopout.go
bhandras: Yes exactly. By sharing ownership I meant that the owner of `OutgoingChanSet` is the `loopOutSwap` object. By passing it as reference without first copying it we also share the ownership if that slice with the newly created `SwapInfo` object.

looprpc/client.proto
bhandras: Yes was my thinking exactly, just to be consistent with naming.

sputn1ck:
I force pushed the requested changes

loopin.go
```diff
@@ -431,6 +431,14 @@ func (s *loopInSwap) sendUpdate(ctx context.Context) error {
...
 	info.ExternalHtlc = s.ExternalHtlc
 
+	// In order to avoid potentially dangerous ownership sharing
+	// we copy the last hop vertex.
+	if s.LastHop != nil {
+		lastHop := &route.Vertex{}
+		copy(lastHop[:], s.LastHop[:])
+		info.LastHop = lastHop
+	}
+
```
sputn1ck, line 441: This could also be done like this:
```go
if s.LastHop != nil {
  lastHop, err := route.NewVertexFromBytes(s.LastHop[:])
  if err != nil {
	  return err
  }
  
  info.LastHop = &lastHop
}
```
any preferences?

loopout.go
```diff
@@ -306,6 +306,15 @@ func (s *loopOutSwap) sendUpdate(ctx context.Context) error {
...
 
+	// In order to avoid potentially dangerous ownership sharing
+	// we copy the outgoing channel set.
+	if s.OutgoingChanSet != nil {
+		outgoingChanSet := make(loopdb.ChannelSet, len(s.OutgoingChanSet))
+		copy(outgoingChanSet[:], s.OutgoingChanSet[:])
+
+		info.OutgoingChanSet = outgoingChanSet
+	}
+
```
sputn1ck, line 317: This could also be done like this:
```go
if s.LastHop != nil {
    lastHop, err := route.NewVertexFromBytes(s.LastHop[:])
    if err != nil {
	    return err
    }
    
    info.LastHop = &lastHop
}
```
any preferences?

yyforyongyu: same logic above. We only do this to copy values, not check their validity.

yyforyongyu (commented):
Almost good, just a few nits!

loopin_test.go
```diff
@@ -54,7 +61,18 @@ func TestLoopInSuccess(t *testing.T) {
...
-	ctx.assertState(loopdb.StateInitiated)
+	swapInfo := <-ctx.statusChan
+	require.Equal(t, loopdb.StateInitiated, swapInfo.State)
+
+	// check that the SwapInfo contains the provided last hop
+	require.Equal(t, expectedLastHop, swapInfo.LastHop)
+
+	//
+	swapInfo.LastHop = &route.Vertex{0x03}
+	assert.Equal(t, &route.Vertex{0x02}, swap.LastHop)
```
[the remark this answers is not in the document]
sputn1ck: I removed the 2 following lines, as they were to test whether changing `swapInfo.LastHop`  would also change `swap.LastHop`

yyforyongyu (approved):
LGTM🎉

loopout_test.go
```diff
@@ -90,6 +91,12 @@ func TestLoopOutPaymentParameters(t *testing.T) {
 		t.Fatal("unexpected state")
 	}
 
+	// check that the SwapInfo contains the outgoing chan set
```
bhandras, line 94: style nit: comments should be sentences (starting capitalized, ending in punctuation).

bhandras (approved):
LGTM :heavy_check_mark:

PR #509, opened 2022-07-08
loopout: use psbt when obtaining the MuSig2 signature

bhandras:
This PR extends the `swapsserverrpc.MuSig2SignSweepReq` with the sweep txn's psbt to make server side validation of the sighash possible.


sputn1ck (approved):
Useful change! LGTM 🚀

sweep/sweeper.go
```diff
@@ -20,51 +22,53 @@ type Sweeper struct {
...
-	sweepTx.AddTxOut(&wire.TxOut{
-		PkScript: sweepPkScript,
-		Value:    int64(amount - fee),
-	})
+	// Compose tx.
+	sweepTx := &wire.MsgTx{
+		Version: 2,
+		TxIn: []*wire.TxIn{{
+			PreviousOutPoint: htlcOutpoint,
+			SignatureScript:  htlc.SigScript,
```
guggero, line 48: Are np2wkh HTLCs still supported? Otherwise we could scrap the `SignatureScript`.

sweep/sweeper.go
```diff
@@ -20,51 +22,53 @@ type Sweeper struct {
...
+	}
 
-	// We need our previous outputs for taproot spends, and there's no
-	// harm including them for segwit v0, so we always include our prevOut.
-	prevOut := []*wire.TxOut{
-		{
-			Value:    int64(amount),
-			PkScript: htlc.PkScript,
-		},
+	packet, err := psbt.NewFromUnsignedTx(sweepTx)
```
guggero, line 57: Should we set the `WitnessUtxo` field on the input as well?

guggero (approved):
LGTM :tada: 

I think just for completeness' sake we should add the UTXO information to the packet, even though it isn't strictly needed.

sweep/sweeper.go
bhandras: done

sweep/sweeper.go
bhandras: removed

PR #533, opened 2022-10-20
utils: fix SelectHopHints for lnd0.15.3

sputn1ck:
This commit adds the `SelectHopHints` from pre `lnd v0.15.3-beta`logic to utils in order to allow loop to compile with `lnd v0.15.3-beta`.


bhandras:
Changes lgtm, just need to fix go.sum first to make CI happy.

bhandras (approved):
LGTM 🎉

utils.go
```diff
@@ -161,9 +166,310 @@ func SelectHopHints(ctx context.Context, lnd *lndclient.LndServices,
...
+
+	// FetchChannelEdgesByID attempts to lookup the two directed edges for
+	// the channel identified by the channel ID.
+	FetchChannelEdgesByID func(chanID uint64) (*channeldb.ChannelEdgeInfo,
+		*channeldb.ChannelEdgePolicy, *channeldb.ChannelEdgePolicy,
+		error)
+
+	// GetAlias allows the peer's alias SCID to be retrieved for private
+	// option_scid_alias channels.
+	GetAlias func(lnwire.ChannelID) (lnwire.ShortChannelID, error)
```
guggero, line 316: This function needs to be set, otherwise this will crash with a nil pointer. And I think because fetching the alias was very slow in lnd for some reason, that was the main reason for the whole refactor change. So maybe we need to do some caching? Or just make sure we read all the aliases properly when fetching the channel info.

guggero (changes requested):
Pretty sure this crashes with a nil pointer panic. Can we please hook the PR up on the server side and run a full itest suite to make sure we don't break loop ins?

utils.go
sputn1ck: Hm this function was previously also not set, weird that it didn't crash

sputn1ck:
Added a fixup commit https://github.com/lightninglabs/loop/commit/64f8b7b602db585ddef9194be71d64f21f835f12 with the `GetAlias` function and am currently running the server itests locally.

guggero (approved):
Cool, I tested this locally with a zero-conf channel and the SCID alias was encoded correctly in the hop hints :tada:

bhandras (approved):
LGTM :100:

PR #542, opened 2022-12-06
liquidity: add `destaddr` parameter for autoloop loop out

### LINKED ISSUE #540: Request: Autoloop to accept `addr` parameter
Loop out supports the --addr parameter to directly loop out to another address.  This is very useful for our specific use case and I've been using loop out very happily on the CLI

I was really hoping to use autoloop, but unfortunately since it doesn't support a destination address, it would require two on-chain transactions to get to the right destination.

Is it possible to enable autoloop/liquidity manager to accept the following: `loop setparams --addr=XXX` ?

Two comments:

1. This would be a game changer for my use case (currently requires manual admin every day), and I would definitely use autoloop if this is enabled
2. I don't need to change the address often (possibly at all)

Or is there some alternative/workaround that I have not thought about?

GeorgeTsagk:
# Description

This PR introduces a parameter for autoloop that lets the user define a custom address to be used for loop out.

Closes #540


ckritzinger (approved):

cmd/loop/liquidity.go
```diff
@@ -286,6 +286,11 @@ var setParamsCommand = cli.Command{
 				"of swaps, limited to the budget set by " +
 				"autobudget",
 		},
+		cli.StringFlag{
+			Name: "outaddr",
```
ckritzinger, line 290: nit: for the loop command, this same parameter is "--addr", I don't know how much the consistency matters to you

looprpc/client.proto
```diff
@@ -854,6 +854,11 @@ message LiquidityParameters {
     The confirmation target for loop in on-chain htlcs.
     */
     int32 htlc_conf_target = 17;
+
+    /*
+    The address to use for autoloop loop out HTLCs.
```
bhandras, line 859: nit: do you mean for the the destination address?

bhandras (approved):
Changes LGTM :tada: 

Although the changes are trivial, some basic test coverage would still be nice.

looprpc/client.proto
GeorgeTsagk: Yes, the address where the funds are going to after we sweep the HTLC

sputn1ck (commented):
LGTM, but would also vouch for more consistent naming. Do you plan on adding tests?

GeorgeTsagk:
> LGTM, but would also vouch for more consistent naming. Do you plan on adding tests?

Yes, will drop them later today.

bhandras (approved):
Re-reviewed with the added test coverage, LGTM :tada:

sputn1ck (approved):
LGTM. I'm still of the opinion for the consistent naming though 😅

GeorgeTsagk:
Changed parameter name to `destaddr` and changed codebase variables accordingly.

GeorgeTsagk:
Rebased on master following @sputn1ck 's CI fix

PR #548, opened 2023-02-03
Autoloop amount backoff

GeorgeTsagk:
## Description

Introduces the required logic to perform amount based back off on autoloop's loop out swap payments.

When consuming a swap suggestion, instead of dispatching a loop out directly through the client we instead call the `dispatchStickyLoopOut` wrapper. This wrapper dispatches a loop out with the full amount through the client, then monitors the swap status and waits for possible `StateFailOffchainPayments`. If such an update is delivered then the wrapper performs the amount back off and dispatches the loop out again.

## TODOs

- [ ] ~~Enhance autoloop tracking to track backoffs (?)~~
- [x] Fix tests


GeorgeTsagk:
Pushed an extra commit that fixes the unit tests, the `dispatchStickyLoopOut` call was hanging forever causing the guard to timeout all of the autoloop tests.

Also the commit that added an extra `persistState` call to the loop out state machine was dropped as that proved unnecessary and also caused some loop out tests to fail too.

GeorgeTsagk:
If there are more efficient / clean ways of tracking the state of a swap in order to perform the back off I'm open to re-writing that part of logic.

alexbosworth:
a possible future iteration would be to use the MPP failure records as a guide, LND internally will already be doing backoff in shard size by cutting them in half itself, the shards that succeed within 2 minutes will show up differently in the failed payment as MPP failure, the shards that fail will show as pathfinding timeout failure. Summing the shards that succeeded could be a good guide for a lower bound size that should work

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,124 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+				// better to return and not attempt a retry.
+
+				log.Infof("No update received for swap")
+
+				return nil
+			}
+
+			if *update == loopdb.StateFailOffchainPayments {
+				// If we failed to pay the server, we will
+				// increase the amount of the swap and try
```
sputn1ck, line 1092: nit: decrease

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,124 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+		case <-time.After(interval):
+		}
+
+		swap, err = m.cfg.GetLoopOut(swapHash)
+		if err != nil {
+			log.Errorf("Error getting swap: %v", err)
+			return
+		}
+
+		if swap == nil {
```
bhandras, line 1143: IUUC we never get here since if the swap is not found `GetLoopOut` will return with an error.

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,124 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+		if update == nil {
+			continue
+		}
+
+		// If the swap is still pending, we will continue to wait.
+		switch update.State {
+		case loopdb.StateFailOffchainPayments:
+			updateChan <- &update.State
+			return
+		case loopdb.StateHtlcPublished:
```
bhandras, line 1159: nit: add nl between cases

loopdb/store_test.go
```diff
@@ -129,24 +131,29 @@ func testLoopOutStore(t *testing.T, pendingSwap *LoopOutContract) {
...
-		if !reflect.DeepEqual(swap, pendingSwap) {
+		swap, err := store.FetchLoopOutSwap(hash)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		require.Equal(t, swap.Hash, swaps[0].Hash)
+
+		swapContract := swap.Contract
+		if !reflect.DeepEqual(swapContract, pendingSwap) {
```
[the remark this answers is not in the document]
GeorgeTsagk: For this particular line you commented on, what would the `require` equivalent be for a `reflect.DeepEqual` ?

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,124 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+
+			if *update == loopdb.StateFailOffchainPayments {
+				// If we failed to pay the server, we will
+				// increase the amount of the swap and try
+				// again.
+				out.Amount -= btcutil.Amount(
+					float64(out.Amount) * amountBackoff,
+				)
+
+				log.Infof("swap %v: amount backoff new amount="+
```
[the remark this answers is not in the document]
GeorgeTsagk: X and Y in your example are source and destination of payment?

loopdb/store.go
```diff
@@ -371,6 +371,134 @@ func (s *boltSwapStore) FetchLoopOutSwaps() ([]*LoopOut, error) {
...
 
+// FetchLoopOutSwap returns the loop out swap with the given hash.
+//
+// NOTE: Part of the loopdb.SwapStore interface.
+func (s *boltSwapStore) FetchLoopOutSwap(hash lntypes.Hash) (*LoopOut, error) {
+	var swap *LoopOut
+
+	err := s.db.View(func(tx *bbolt.Tx) error {
+		// First, we'll grab our main loop out bucket key.
+		rootBucket := tx.Bucket(loopOutBucketKey)
```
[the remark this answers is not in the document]
GeorgeTsagk: Created 2 deserialize funcs for LoopInContract and LoopOutContract

loopdb/store_test.go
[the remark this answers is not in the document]
bhandras: `require.Equal`

liquidity/liquidity.go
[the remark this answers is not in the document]
bhandras: No, sorry maybe I was a bit vague with my wording. What I meant is that we could log the old amount along with the new amount so we have immediate information about how much the swap amount was reduced by.

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,123 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+	for time.Since(startTime) < timeout {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(interval):
+		}
+
+		swap, err = m.cfg.GetLoopOut(swapHash)
+		if err != nil {
+			log.Errorf("Error getting swap: %v", err)
```
bhandras, line 1142: nit: log the swap hash too.

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,123 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+			return
```
bhandras, line 1143: Wouldn't this return cause `dispatchStickyLoopOut` to hang?

liquidity/liquidity.go
```diff
@@ -1020,6 +1050,123 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+			continue
+		}
+
+		// If the swap is still pending, we will continue to wait.
+		switch update.State {
+		case loopdb.StateFailOffchainPayments:
+			updateChan <- &update.State
+			return
+
+		case loopdb.StateHtlcPublished:
```
bhandras, line 1158: Loop out swaps won't reach this state as it's used for loop ins.

bhandras (changes requested):

loopdb/store.go
```diff
@@ -371,53 +282,41 @@ func (s *boltSwapStore) FetchLoopOutSwaps() ([]*LoopOut, error) {
...
+		if rootBucket == nil {
+			return errors.New("bucket does not exist")
 		}
 
-		basicState := updateBucket.Get(basicStateKey)
-		if basicState == nil {
-			return errors.New("no basic state for update")
+		// We'll now traverse the root bucket for all active swaps. The
+		// primary key is the swap hash itself.
+		swapBucket := rootBucket.Bucket(hash[:])
```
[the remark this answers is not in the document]
GeorgeTsagk: Good idea

loopdb/store_test.go
```diff
@@ -96,68 +96,57 @@ func TestLoopOutStore(t *testing.T) {
...
-		if swaps[0].State().State != expectedState {
-			t.Fatalf("expected state %v, but got %v",
-				expectedState, swaps[0].State(),
-			)
-		}
+		require.Equal(t, swap.Hash, swaps[0].Hash)
+
+		swapContract := swap.Contract
+
+		require.True(t, reflect.DeepEqual(swapContract, pendingSwap))
```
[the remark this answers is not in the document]
GeorgeTsagk: Oh okay, so it will check for **deep** equality

liquidity/liquidity.go
```diff
@@ -156,9 +176,17 @@ type Config struct {
...
+	// GetLoopOut returns a single loop out swap based on the provided swap
+	// hash.
+	GetLoopOut func(hash lntypes.Hash) (*loopdb.LoopOut, error)
+
 	// ListLoopIn returns all of the loop in swaps stored on disk.
 	ListLoopIn func() ([]*loopdb.LoopIn, error)
 
+	// GetLoopIn returns a single loop in swap based on the provided swap
+	// hash.
+	GetLoopIn func(hash lntypes.Hash) (*loopdb.LoopIn, error)
```
sputn1ck, line 188: GetLoopIn is unused, I think we can lower the diff of this PR if loop in amount backoff is not in the scope here.

liquidity/liquidity.go
```diff
@@ -387,13 +415,15 @@ func (m *Manager) autoloop(ctx context.Context) error {
 			swap.DestAddr = m.params.DestAddr
 		}
 
-		loopOut, err := m.cfg.LoopOut(ctx, &swap)
+		err := m.dispatchStickyLoopOut(
```
sputn1ck, line 418: IMO the backoff logic should either be configurable / and or be able to be disabled, as the backoff amount might collide with the rules we have set. I'm not sure If a 3 times reduced swap would be favorable

liquidity/liquidity.go
GeorgeTsagk: Yes amount back-off is only meant to be applied on the payment part of a loop-out. Any reason we should keep it for consistency?

liquidity/liquidity.go
GeorgeTsagk: Right, was confused with the `SERVER_HTLC_PUBLISHED` state, which is used for loop out.

liquidity/liquidity.go
GeorgeTsagk: The previous behavior is a sub-set to this PR. With no failures present, the behavior will be identical whether we have backoff implemented or not.

> I'm not sure If a 3 times reduced swap would be favorable

What would "favorable" mean here?

GeorgeTsagk: The user defined restrictions for swaps (w.r.t to fee percentages etc) are checked for every individual loop that we dispatch, regardless of if we have applied an amount backoff to it.

liquidity/liquidity.go
bhandras: Given it's not used, I think it's safe to remove.

GeorgeTsagk:
Removed the unused store method `FetchLoopInSwap`

store_mock_test.go
```diff
@@ -70,6 +70,39 @@ func (s *storeMock) FetchLoopOutSwaps() ([]*loopdb.LoopOut, error) {
 	return result, nil
 }
 
+// FetchLoopOutSwaps returns all swaps currently in the store.
+//
+// NOTE: Part of the loopdb.SwapStore interface.
+func (s *storeMock) FetchLoopOutSwap(
+	hash lntypes.Hash) (*loopdb.LoopOut, error) {
+
+	var result *loopdb.LoopOut
```
bhandras, line 79: nit: not needed

store_mock_test.go
```diff
@@ -70,6 +70,39 @@ func (s *storeMock) FetchLoopOutSwaps() ([]*loopdb.LoopOut, error) {
...
+
+	updates := s.loopOutUpdates[hash]
+	events := make([]*loopdb.LoopEvent, len(updates))
+	for i, u := range updates {
+		events[i] = &loopdb.LoopEvent{
+			SwapStateData: u,
+		}
+	}
+
+	swap := &loopdb.LoopOut{
```
bhandras, line 94: nit: can just return the swap here + nil

bhandras (approved):
LGTM, great work @GeorgeTsagk 🚀

PR #551, opened 2023-02-08
multi: add recurring autoloop budget

GeorgeTsagk:
## Description

This PR introduces the required changes to convert our current autoloop budget into a recurring budget.

This PR replaces the old `budgetstart` cli parameter with `autobudgetrefreshperiod`, which now indicates how often the budget should be refreshed.

Within the codebase what allowed minimal changes was replacing the `AutoloopBudgetStartSec` with `AutoloopBudgetRefreshTimestamp`. The second one is now set to current time whenever `autobudgetrefreshperiod` elapses. This way we won't change the way we calculate our budget for loops which used the `AutoloopBudgetStartSec` based date.

## Todo

- [ ] Update autoloop documentation
- [ ] Tests
- [ ] Persist on restart (?)


version.go
```diff
@@ -27,7 +27,7 @@ const (
 	// Note: please update release_notes.md when you change these values.
 	appMajor uint = 0
 	appMinor uint = 21
-	appPatch uint = 0
+	appPatch uint = 1
```
bhandras, line 30: nit: no need to bump the version here, as we cut releases when we tag the `master` branch.

liquidity/liquidity.go
```diff
@@ -902,7 +913,10 @@ func (m *Manager) checkExistingAutoLoops(ctx context.Context,
 				out.Contract.MaxMinerFee,
 				mSatToSatoshis(prepay.Value),
 			)
-		} else if !out.LastUpdateTime().Before(m.params.AutoFeeStartDate) {
+		} else if out.LastUpdateTime().After(
```
sputn1ck, line 916: Doesn't this mean that even if a swap was started in the last budget period, it will be counted for the current period?

liquidity/liquidity.go
```diff
@@ -506,16 +521,6 @@ func (m *Manager) SuggestSwaps(ctx context.Context, autoloop bool) (
 		return nil, ErrNoRules
 	}
 
-	// If our start date is in the future, we interpret this as meaning that
-	// we should start using our budget at this date. This means that we
-	// have no budget for the present, so we just return.
-	if m.params.AutoFeeStartDate.After(m.cfg.Clock.Now()) {
```
sputn1ck, line 512: I'm not sure I that the start date should be replaced with the budget, IMO they serve different purposes.

liquidity/liquidity.go
```diff
@@ -1020,6 +1035,20 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
 	return traffic
 }
 
+// refreshAutoloopBudget checks whether the elapsed time since our last autoloop
+// budget refresh is greater than our configured refresh period. If so, the last
+// refresh timestamp.
+func (m *Manager) refreshAutoloopBudget(ctx context.Context) error {
```
sputn1ck, line 1041: I'm not sure if I'm missing something, but I don't find where the actual budget is refreshed

sputn1ck (commented):
Nice and easy to folow PR, though I'm not sure if the current logic works as intended. AFAICT the refreshing logic isn't tested anywhere, so adding a test might be nice.

liquidity/liquidity.go
GeorgeTsagk: Yes, if a swap is still active in any way within our current period. regardless of when it started, it will count in the current period's budget.

liquidity/liquidity.go
GeorgeTsagk: You mean the budget timestamp?

liquidity/autoloop_test.go
```diff
@@ -966,6 +966,207 @@ func TestAutoloopBothTypes(t *testing.T) {
 	c.stop()
 }
 
+func TestAutoLoopRecurringBudget(t *testing.T) {
+	// defer test.Guard(t)()
```
sputn1ck, line 970: nit: either uncomment or remove

version.go
sputn1ck: Commit still needs to be removed

sputn1ck (approved):
Changes LGTM! 🚀

liquidity/autoloop_test.go
GeorgeTsagk: woops

liquidity/liquidity.go
```diff
@@ -1020,6 +1035,20 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+	if time.Since(m.cfg.AutoloopBudgetRefreshTimestamp) >
+		m.params.AutoFeeRefreshPeriod {
+
+		log.Debug("Refreshing autoloop budget")
+		m.cfg.AutoloopBudgetRefreshTimestamp = time.Now()
+	}
+
+	return nil
```
bhandras, line 1049: Could just not return anything and then there's no need to check where used.

looprpc/client.proto
```diff
@@ -823,12 +823,10 @@ message LiquidityParameters {
...
     /*
-    The start time for autoloop budget, expressed as a unix timestamp in
-    seconds. If this value is 0, the budget will be applied for all
-    automatically dispatched swaps. Swaps that were completed before this date
-    will not be included in budget calculations.
+    The period over which the autoloop budget is refreshed, expressed in
+    seconds.
     */
-    uint64 autoloop_budget_start_sec = 12;
+    uint64 autoloop_budget_refresh_period_sec = 12;
```
bhandras, line 829: This is bad practice. Generally we want to deprecate or delete a field and add a new field instead to no break compatibility.
See: https://protobuf.dev/programming-guides/dos-donts/#dont-re-use-a-tag-number

loopd/utils.go
```diff
@@ -39,9 +40,10 @@ func getClient(config *Config, lnd *lndclient.LndServices) (*loop.Client,
 
 func getLiquidityManager(client *loop.Client) *liquidity.Manager {
 	mngrCfg := &liquidity.Config{
-		AutoloopTicker: ticker.NewForce(liquidity.DefaultAutoloopTicker),
-		LoopOut:        client.LoopOut,
-		LoopIn:         client.LoopIn,
+		AutoloopTicker:                 ticker.NewForce(liquidity.DefaultAutoloopTicker),
+		AutoloopBudgetRefreshTimestamp: time.Now(),
```
bhandras, line 44: Starting with `Now()` would mean that we need to wait a whole refresh period right?

liquidity/autoloop_test.go
```diff
@@ -966,6 +966,208 @@ func TestAutoloopBothTypes(t *testing.T) {
 	c.stop()
 }
 
+func TestAutoLoopRecurringBudget(t *testing.T) {
```
bhandras, line 969: nit: add godoc.

loopd/utils.go
GeorgeTsagk: You don't need to wait for it to perform loops, it means that a new budget-cycle has just started.

cmd/loop/liquidity.go
```diff
@@ -298,11 +298,11 @@ var setParamsCommand = cli.Command{
...
-		cli.Uint64Flag{
-			Name: "budgetstart",
-			Usage: "the start time for the automated loop " +
-				"out budget, expressed as a unix timestamp " +
-				"in seconds",
+		cli.DurationFlag{
+			Name: "autobudgetrefreshperiod",
+			Usage: "the time period over which the automated " +
+				"loop budget is refreshed, expressed in " +
+				"seconds",
```
bhandras, line 305: nit: not necessarily in seconds.

liquidity/liquidity.go
```diff
@@ -1020,6 +1032,18 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
 
+// refreshAutoloopBudget checks whether the elapsed time since our last autoloop
+// budget refresh is greater than our configured refresh period. If so, the last
+// refresh timestamp.
+func (m *Manager) refreshAutoloopBudget(ctx context.Context) {
+	if time.Since(m.cfg.AutoloopBudgetLastRefresh) >
+		m.params.AutoFeeRefreshPeriod {
+
+		log.Debug("Refreshing autoloop budget")
+		m.cfg.AutoloopBudgetLastRefresh = time.Now()
```
bhandras, line 1043: nit: instead of `time.Now()` it's better to use `m.cfg.Clock.Now()` (the `clock` package is a decorator around the the usual `time` functions to allow seamless test flow).

bhandras (approved):
LGTM 🎉

PR #558, opened 2023-03-15
Fix macaroon store

sputn1ck:
This PR fixes a macaroons store typo

### DIFF
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -390,7 +390,7 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 	}
 
 	rks, db, err := lndclient.NewBoltMacaroonStore(
-		d.cfg.DataDir, "macarooons.db", loopdb.DefaultLoopDBTimeout,
+		d.cfg.DataDir, "macaroons.db", loopdb.DefaultLoopDBTimeout,
 	)
 	if err != nil {
 		return err


go.mod
```diff
@@ -7,25 +7,25 @@ require (
...
+	github.com/btcsuite/btcwallet v0.16.7
 	github.com/btcsuite/btcwallet/wtxmgr v1.5.0
 	github.com/coreos/bbolt v1.3.3
 	github.com/davecgh/go-spew v1.1.1
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.0.1
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.5.0
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/lightninglabs/aperture v0.1.18-beta
+	github.com/lightninglabs/aperture v0.1.19-beta.0.20230227124349-b952e4ee6ff4
```
bhandras, line 18: Maybe we can also tag `aperture` so we have a proper tag here.

bhandras (approved):

go.mod
guggero: My plan was to tag aperture as soon as we bumped it to `lnd v0.16.0-beta` final. Otherwise if we have the RC in all the `go.mod` files, go mod will want to use that instead of the final one in litd. So as soon as the final release is out, we'll bump to that version in all projects and then tag for inclusion in litd.

loopd/daemon.go
```diff
@@ -390,7 +390,7 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 	}
 
 	rks, db, err := lndclient.NewBoltMacaroonStore(
-		d.cfg.DataDir, "macarooons.db", loopdb.DefaultLoopDBTimeout,
+		d.cfg.DataDir, "macaroons.db", loopdb.DefaultLoopDBTimeout,
```
GeorgeTsagk, line 393: Is this going to cause any issues with existing `macarooons.db` files?

GeorgeTsagk (approved):
LGTM

guggero (approved):

PR #559, opened 2023-03-17
Autoloop parse old budget to new parameters

GeorgeTsagk:
Depends on:
- https://github.com/lightninglabs/loop/pull/556

## Description

If an old client updates then we parse the old style budget to our new parameters, where we define an near-infinite interval and assign the last budget refresh to the (now deprecated) budget start timestamp. This way if a client had an autoloop budget in-flight it will remain as is until the moment it is updated.


bhandras (approved):
LGTM 🎉

PR #560, opened 2023-03-24
loopd: nil-pointer bug when showing subsystems

ziggie1984:
Just a short drive-bug fix: loopd crashes when you wanna show the logging subsystems with `loopd --debuglevel show`

This fix just copies the rogue code to another place, but I think a better fix would be to get rid of the global pointer and maybe just return the logger when setting it up ?

#### Pull Request Checklist

### DIFF
--- a/loopd/run.go
+++ b/loopd/run.go
@@ -180,13 +180,6 @@ func Run(rpcCfg RPCConfig) error {
 		os.Exit(0)
 	}
 
-	// Special show command to list supported subsystems and exit.
-	if config.DebugLevel == "show" {
-		fmt.Printf("Supported subsystems: %v\n",
-			logWriter.SupportedSubsystems())
-		os.Exit(0)
-	}
-
 	// Validate our config before we proceed.
 	if err := Validate(&config); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


loopd/run.go
```diff
@@ -204,6 +197,13 @@ func Run(rpcCfg RPCConfig) error {
 	logWriter := build.NewRotatingLogWriter()
 	SetupLoggers(logWriter, shutdownInterceptor)
 
+	// Special show command to list supported subsystems and exit.
+	if config.DebugLevel == "show" {
```
GeorgeTsagk, line 201: Is this fix effective because `if err := Validate(&config)` precedes now?

ziggie1984: hmm let me check

ziggie1984: yeah so debuglevel validation seems not part and not necessary to validate so it should be fixed

hieblmi (approved):
Rebased this PR and tested, as @ziggie1984 mentioned the validation doesn't check the debuglevel, so moving the code is fine. LGTM 💾

✅ tAck

PR #563, opened 2023-03-29
Remove swapserverrpc replace

guggero:
This was discovered when building litd with the latest loop version.
A new `swapserverrpc/v1.0.4` tag was pushed to fix it, but in order to not forget this in the future, we remove the local replace directive. This does make it a bit harder to change things in that RPC package but will make it less likely that we forget to tag things.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -171,9 +171,6 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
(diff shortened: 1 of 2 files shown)


bhandras (approved):
LGTM, thanks! 🎉

GeorgeTsagk (approved):

PR #567, opened 2023-04-11
Autoloop rework -- easy autoloop

GeorgeTsagk:
### Description

Adds a new type of autoloop which is easier to setup. The only option is the target hot wallet size which is the maximum amount of funds we want to keep on the lightning node. If that target is exceeded a loop-out will automatically move those funds to an on-chain address.

### TODOs
- [ ] ~~Add inactive peer auto-closer~~
- [x] Add tests


cmd/loop/liquidity.go
```diff
@@ -324,6 +324,17 @@ var setParamsCommand = cli.Command{
...
 				"htlcs",
 		},
+		cli.BoolFlag{
+			Name: "easyautoloop",
+			Usage: "set to true to enable easy autoloop, which " +
+				"will automatically dispatch swaps in order to meet " +
+				"the target hot wallet size",
+		},
+		cli.Uint64Flag{
+			Name: "hotwalletsizesat",
```
bhandras, line 334: If agree with above, maybe we could call this param `offchainliquiditytargetsats` or just simply `offchaintargetsats`.

liquidity/liquidity.go
```diff
@@ -457,6 +499,160 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+	}
+
+	// Get a summary of our existing swaps so that we can check our autoloop
+	// budget.
+	summary, err := m.checkExistingAutoLoops(ctx, loopOut, loopIn)
+	if err != nil {
+		return err
+	}
+
+	if summary.totalFees() >= m.params.AutoFeeBudget {
```
bhandras, line 528: Given these checks are the same we already do in `SuggestSwaps` perhaps worth creating helper functions for them to not duplicate code and logs.

liquidity/liquidity.go
```diff
@@ -457,6 +499,160 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+		amount = btcutil.Amount(terms.MaxSwapAmount)
+	}
+
+	log.Debugf("easy autoloop: hot_wallet=%v, target=%v, "+
+		"attempting to loop out %v", hotWalletBalance,
+		m.params.EasyAutoloopTarget, amount)
+
+	// Sort the candidate channels based on descending local balance.
+	sort.Slice(channels, func(i, j int) bool {
+		max1 := math.Min(
```
bhandras, line 585: This tricky sort deserves a bit more explanation.

liquidity/liquidity.go
```diff
@@ -457,6 +499,160 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+
+	switch t := suggestion.(type) {
+	case *loopOutSwapSuggestion:
+		swap = t.OutRequest
+	default:
+		return fmt.Errorf("unexpected swap suggestion type: %T", t)
+	}
+
+	// Dispatch a sticky loop out.
+	go m.dispatchStickyLoopOut(
```
bhandras, line 640: What if a previous call of `dispatchBestEasyAutoloopSwap` already dispatched a sticky loopout for a channel which backoffs? Wouldn't we happily do start a competing `dispatchStickyLoopOut`?

bhandras (commented):
Great start but needs a bit more explanation on the semantics of the new easy mode in the PR description as well as more discussion may be needed whether we should also support easy auto loopins?

More test coverage is also very welcome!

liquidity/liquidity.go
```diff
@@ -457,6 +499,160 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	hotWalletBalance := btcutil.Amount(0)
+	for _, channel := range channels {
+		hotWalletBalance += channel.LocalBalance
+	}
+
+	// Since we're only autolooping-out we need to check if we are below
```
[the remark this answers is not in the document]
GeorgeTsagk: We decided to keep it simple for this first iteration and only introduce loop-out.

liquidity/liquidity.go
GeorgeTsagk: a) If we hit our max inflight swap limit (default=`1`) we don't dispatch new swaps.

b) Check `pickEasyAutoloopChannel`: If there is an on-going swap on a channel we avoid picking that channel again for another easy-autoloop. Also if there was a failed swap on that channel we avoid picking it for some interval.

> Wouldn't we happily do start a competing `dispatchStickyLoopOut`?

So yes, we **could** happily start a new sticky loop out, but
- over a different outgoing channel
- over a channel that hasn't failed recently
- only if user has specified a `MaxAutoInFlight` > `1`

liquidity/liquidity.go
```diff
@@ -457,6 +505,142 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+	suggestion, err := builder.buildSwap(
+		ctx, channel.PubKeyBytes, outgoing, swapAmt, true, easyParams,
+	)
+	if err != nil {
+		return err
+	}
+
+	swap := loop.OutRequest{}
+
+	switch t := suggestion.(type) {
```
bhandras, line 620: I think this `switch` is unnecessary given `builder` returns a `loopOutSwapSuggestion`.

liquidity/autoloop_test.go
```diff
@@ -207,7 +207,7 @@ func TestAutoLoopEnabled(t *testing.T) {
 				Events: []*loopdb.LoopEvent{
 					{
 						SwapStateData: loopdb.SwapStateData{
-							State: loopdb.StateInitiated,
+							State: loopdb.StateSuccess,
```
bhandras, line 210: Why change the expected change? Did existing autoloop semantics also change?

liquidity/liquidity.go
```diff
@@ -259,15 +271,28 @@ func (m *Manager) Run(ctx context.Context) error {
 	for {
 		select {
 		case <-m.cfg.AutoloopTicker.Ticks():
-			err := m.autoloop(ctx)
-			switch err {
-			case ErrNoRules:
-				log.Debugf("No rules configured for autoloop")
+			switch m.params.EasyAutoloop {
+			case true:
+				err := m.easyAutoLoop(ctx)
```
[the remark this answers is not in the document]
GeorgeTsagk: I changed the doc string for `easyautoloop`

liquidity/liquidity.go
```diff
@@ -457,6 +505,142 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+
+	restrictions, err := m.cfg.Restrictions(ctx, swap.TypeOut)
+	if err != nil {
+		return err
+	}
+
+	// Calculate the amount that we want to loop out. Check against the max
+	// swap amount of terms.
+	amount := hotWalletBalance - m.params.EasyAutoloopTarget
+	if amount > restrictions.Maximum {
```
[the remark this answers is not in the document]
GeorgeTsagk: We're attempting swaps from biggest to smallest, so got to check that our highest values respect the max limit, because those are most likely to be picked next.

If we pick a channel for a swap and the amount is <= 250k that means it also was our best choice so lets let the server decide

liquidity/liquidity.go
```diff
@@ -457,6 +505,142 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+		channels, restrictions, loopOut, loopIn, amount,
+	)
+	if channel == nil {
+		return fmt.Errorf("no eligible channel for easy autoloop")
+	}
+
+	log.Debugf("easy autoloop: picked channel %v with local balance %v",
+		channel.ChannelID, channel.LocalBalance)
+
+	swapAmt, err := btcutil.NewAmount(
```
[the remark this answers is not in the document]
GeorgeTsagk: Because it could also be beyond the max allowed, so we don't want it to overflow that value and get the swap rejected

liquidity/liquidity.go
```diff
@@ -1190,6 +1383,98 @@ func (m *Manager) waitForSwapPayment(ctx context.Context, swapHash lntypes.Hash,
...
+func (m *Manager) pickEasyAutoloopChannel(channels []lndclient.ChannelInfo,
+	restrictions *Restrictions, loopOut []*loopdb.LoopOut,
+	loopIn []*loopdb.LoopIn, amount btcutil.Amount) *lndclient.ChannelInfo {
+
+	traffic := m.currentSwapTraffic(loopOut, loopIn)
+
+	// Sort the candidate channels based on descending local balance. We
+	// want to prioritize picking a channel with the highest possible local
+	// balance.
+	sort.Slice(channels, func(i, j int) bool {
```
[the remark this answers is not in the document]
GeorgeTsagk: Right, once picked I then also check against max amount so it was redundant

liquidity/liquidity.go
```diff
@@ -1190,6 +1383,98 @@ func (m *Manager) waitForSwapPayment(ctx context.Context, swapHash lntypes.Hash,
...
+			log.Debugf("Channel %v cannot be used for easy "+
+				"autoloop: ongoing swap", channel.ChannelID)
+			continue
+		}
+
+		if channel.LocalBalance < restrictions.Minimum {
+			log.Debugf("Channel %v cannot be used for easy "+
+				"autoloop: insufficient local balance",
+				channel.ChannelID)
+			continue
```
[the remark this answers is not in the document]
GeorgeTsagk: Right, all the following channels will also have less or equal local balance. good catch ty

liquidity/autoloop_test.go
GeorgeTsagk: No, first time I introduced `stickyLoopOut` we used to check the `OutRequests` that the mock received, but didn't care about the outcome of the swap itself.

Now I changed the `autoloop` step function to also check that `activeStickyLoops` is `0` before returning. A sticky loop is fetching the swap info from db with the `GetSIngleLoopOut` call which return the value defined in the line above.

`StateInitiated` doesn't really provide any meaningful feedback on the state of the swap so sticky loop out wouldn't return, so I changed it to be `StateSuccess` so that sticky loop outs would immediately return, thus making the `autoloop` step function return with success.

liquidity/liquidity.go
```diff
@@ -457,6 +504,136 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+
+	suggestion, err := builder.buildSwap(
+		ctx, channel.PubKeyBytes, outgoing, swapAmt, true, easyParams,
+	)
+	if err != nil {
+		return err
+	}
+
+	swap := loop.OutRequest{}
+
```
bhandras, line 613: nit: extra nl

liquidity/liquidity.go
```diff
@@ -457,6 +504,136 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+	return nil
+}
+
+// autoCloseInactivePeers automatically closes channels with peers that are
+// considered inactive. This will check channel opens & closes, payment
+// forwards, and payments sent or received. This is mainly executed as a way to
+// free up liquidity that is locked towards a peer that is not good enough to
+// handle it.
+func (m *Manager) autoCloseInactivePeers(ctx context.Context) {
```
bhandras, line 633: Wrong PR?

liquidity/liquidity.go
```diff
@@ -563,23 +740,13 @@ func (m *Manager) SuggestSwaps(ctx context.Context, autoloop bool) (
...
 	}
 
-	if summary.totalFees() >= m.params.AutoFeeBudget {
-		log.Debugf("autoloop fee budget: %v exhausted, %v spent on "+
-			"completed swaps, %v reserved for ongoing swaps "+
-			"(upper limit)",
-			m.params.AutoFeeBudget, summary.spentFees,
-			summary.pendingFees)
-
+	err = m.checkSummaryBudget(summary)
```
bhandras, line 743: So I mentioned before that easy autoloop's documentation should mention which params are in use. Budget and "max in flight" are two params that are surely used. Maybe could add them to the godocs etc?

liquidity/autoloop_test.go
bhandras: Yes I see now, we check `numActiveStickyLoops()` in the test context. Thanks for the explanation!

liquidity/autoloop_test.go
```diff
@@ -1235,6 +1250,175 @@ func TestAutoLoopRecurringBudget(t *testing.T) {
...
+	c.easyautoloop(step, false)
+
+	// In order to reflect the change on the channel balances we create a
+	// new context and restart the autolooper.
+	easyChannel1.LocalBalance -= chan1Swap.Amount
+	channels = []lndclient.ChannelInfo{
+		easyChannel1, easyChannel2,
+	}
+
+	c = newAutoloopTestCtx(t, params, channels, testRestrictions)
```
bhandras, line 1354: Shouldn't we first stop the existing context? (so we also check that no unexpected error happened)

liquidity/autoloop_testcontext_test.go
```diff
@@ -325,6 +334,50 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
 
 	require.True(c.t, c.matchLoopOuts(step.expectedOut, step.keepDestAddr))
 	require.True(c.t, c.matchLoopIns(step.expectedIn))
+
+	require.Eventuallyf(c.t, func() bool {
+		return c.manager.numActiveStickyLoops() == 0
+	}, time.Second*45, time.Millisecond*100, "failed to wait for sticky loop"+
```
bhandras, line 340: nit: these values could be extracted as constants.

liquidity/autoloop_testcontext_test.go
```diff
@@ -325,6 +334,50 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
...
+		"counter")
```
bhandras, line 341: nit: needs a space.

bhandras (commented):
Almost ready, just a few remaining nits! 🎉

liquidity/liquidity.go
GeorgeTsagk: Left over from when this was within this PR's scope

liquidity/liquidity.go
```diff
@@ -457,6 +505,135 @@ func (m *Manager) ForceAutoLoop(ctx context.Context) error {
...
+	// allowed clamp it to max.
+	amount := localTotal - m.params.EasyAutoloopTarget
+	if amount > restrictions.Maximum {
+		amount = btcutil.Amount(restrictions.Maximum)
+	}
+
+	// If the amount we want to loop out is less than the minimum we can't
+	// proceed with a swap, so we return early.
+	if amount < restrictions.Minimum {
+		log.Debugf("easy autoloop: total local balance is below "+
```
bhandras, line 573: nit: I think you meant to write "swap amount is below"

liquidity/autoloop_test.go
```diff
@@ -1235,6 +1250,178 @@ func TestAutoLoopRecurringBudget(t *testing.T) {
...
+	// We have met the target of 75_000 so we don't expect any action from
+	// easy autoloop. That's why we set noop to true in the call below.
+	step = &easyAutoloopStep{
+		minAmt: 1,
+		maxAmt: 50000,
+	}
+
+	c.easyautoloop(step, true)
+	c.stop()
+
```
bhandras, line 1422: nit: extra nl

bhandras (approved):
Nice work @GeorgeTsagk! LGTM 🎉

PR #568, opened 2023-04-12
liquidity: exclude prepay from total swap fees

GeorgeTsagk:
### Description

This PR removes the `PrepayAmount` from fee calculations for the `FeePortion` policy (percentage fee). This will allow smaller swaps to be dispatched and not get disqualified due to very high fees.

### DIFF
--- a/liquidity/liquidity_test.go
+++ b/liquidity/liquidity_test.go
@@ -1564,21 +1564,6 @@ func TestFeePercentage(t *testing.T) {
 				DisqualifiedPeers: noPeersDisqualified,
 			},
 		},
-		{
-			name:   "prepay too high",
-			feePPM: 30000,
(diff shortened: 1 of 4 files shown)


liquidity/liquidity_test.go
```diff
@@ -1564,21 +1564,6 @@ func TestFeePercentage(t *testing.T) {
 				DisqualifiedPeers: noPeersDisqualified,
 			},
 		},
-		{
-			name:   "prepay too high",
-			feePPM: 30000,
-			quote: &loop.LoopOutQuote{
```
bhandras, line 1570: I wonder if instead of deleting just fixing the outcome and adding a comment why this will result in a suggestion is better.

bhandras (approved):
LGTM 🌮

PR #574, opened 2023-04-25
protocol: set musig2 to be the stable version

sputn1ck:
This PR changes the stable protocol version to be the Musig2 protocol.

The experimental version is set to the stable version in order for the flag to still work if a user has it set.


GeorgeTsagk (approved):
:ok:

ellemouton:
just for my understanding - so this wont be opt out at all? ie, the flag will do nothing right?

bhandras:
> just for my understanding - so this wont be opt out at all? ie, the flag will do nothing right?

Yes there's no opt out with this change, but we keep the `--experimental` flag. It'll however from now on just default to the stable protocol version and essentially do nothing.

bhandras (approved):
LGTM, thanks @sputn1ck 💯

GeorgeTsagk (approved):

PR #580, opened 2023-05-02
liquidity: reduce miner fee multiplier for fee percentage

GeorgeTsagk:
## Description

When using the fee percentage option for autoloop we scale the calculated miner fee by some multiplier as a pre-caution against fee spikes. This multiplier was 100X and is being changed to 10X with this PR. This is done in order to allow for more small swaps to be executed and not cut by the fee check.

### DIFF
--- a/liquidity/liquidity_test.go
+++ b/liquidity/liquidity_test.go
@@ -1555,7 +1555,7 @@ func TestFeePercentage(t *testing.T) {
 			quote: &loop.LoopOutQuote{
 				SwapFee:      60,
 				PrepayAmount: 30,
-				MinerFee:     1,
+				MinerFee:     10,
 			},
 			suggestions: &Suggestions{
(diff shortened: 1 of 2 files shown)


liquidity/liquidity_test.go
```diff
@@ -1555,7 +1555,7 @@ func TestFeePercentage(t *testing.T) {
 			quote: &loop.LoopOutQuote{
 				SwapFee:      60,
 				PrepayAmount: 30,
-				MinerFee:     1,
+				MinerFee:     10,
```
GeorgeTsagk, line 1558: Since we changed 100X to 10X this now needs to be bumped for the test to return the expected error

GeorgeTsagk:
Closing this as this fix should not be done in such a direct way, as we still need to consider the full (100x) amount for fee bumping.

PR #585, opened 2023-05-17
Migrate from boltdb to sql

sputn1ck:
This PR adds the migration from boltdb to a local sqlite db or a postgres db. This will make it easier for future improvements such as taproots assets and other protocol changes.

Currently the migration is only done once from boltdb to the specified sql store. We should add a migrate command to loop which will allow the migration from sqlite to postgres and back.

#### TODO
- [x] cleanup commits
- [x] better comments
- [x] write migration from boltdb to sqlite
- [x] use sqlite as base db
- [x] update release notes

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -15,7 +15,7 @@ This file tracks release notes for the loop client.
 ## Next release
 
 #### New Features
-
+* Deprecated boltdb, we now support both sqlite and postgres. On first startup, loop will automatically migrate your database to sqlite. If you want to use postgres, you can set the `--databasebackend` flag to `postgres` and set the `--postgres.host`, `--postgres.port`, `--postgres.user`, `--postgres.password` and `--postgres.dbname` flags to connect to your postgres instance. Your boltdb file will be saved as a backup in the loop directory. NOTE: we're currently only supporting migrating once from boltdb to a sql backend. A manual migration between postgres and sqlite will be supported in the future.
 #### Breaking Changes
 
 #### Bug Fixes


loopd/migration.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	err = sqliteDb.Close()
+	if err != nil {
+		return err
+	}
+
+	// If the migration was successfull we'll rename the bolt db to
+	// loop.db.bk.
+	err = os.Rename(
```
GeorgeTsagk, line 61: is keeping the backup what usually happens with migrations?

what if db was super big and keeping a duplicate causes problems?
(e.g a user with many swaps like muun)
is the following a possible scenario:
- muun upgrades & triggers migration
- migration fails due to running out of space
- now loop falls back to using boltdb (?)

loopd/migration.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	// Create a new migrator manager.
+	migrator := loopdb.NewMigratorManager(boltdb, sqliteDb)
+
+	// Run the migration.
+	err = migrator.RunMigrations(ctx)
+	if err != nil {
+		return err
```
GeorgeTsagk, line 45: perhaps clean-up the not-complete sqlite db?

GeorgeTsagk (commented):
cACK, not super expert in db migrations

will do another review run soon so I can also provide a tACK

loopdb/sqlite/queries/swaps.sql
```diff
@@ -0,0 +1,166 @@
+
+
+-- name: GetLoopOutSwaps :many
+SELECT 
+    s.*,
+    COALESCE(GROUP_CONCAT(swap_updates.update_state),'') AS update_states,
```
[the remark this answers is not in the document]
Roasbeef: Commented offline, but IMO this seems like a bit of an anti-pattern vs just getting the swaps, then doing a join in another query for all the updates for a given swap. 

I think this query can also be restructred to instead use a CTE to get all the swaps, then in another query, join against each of those and return the swap updates along with that swap (tho you'd duplicate some of the rows).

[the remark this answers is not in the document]
Roasbeef: We'd also want to add pagination here from the get go as well since there's a gRPC level limit hard coded in rn.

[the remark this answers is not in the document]
Roasbeef: Also re the comma pattern: wouldn't this require additional parsing for the response to get it in a useable state vs returning all the updates for a swap in a distinct query? 

Also IIUC, most of the them we only either care about the base swap itself, or the very last update as well.

sqlc.yaml
```diff
@@ -0,0 +1,10 @@
+version: "2"
+sql:
+  - engine: "sqlite"
```
Roasbeef, line 3: See the `tapd` config, we can get by with just the `postgres` engine. Then we also don't need to segment things for sqlite vs not, can just live under a `sqlc` folder.

loopdb/sqlite/queries/liquidity_params.sql
```diff
@@ -0,0 +1,6 @@
+-- name: PutLiquidityParams :exec
+INSERT OR REPLACE INTO liquidity_params (id, params)
```
Roasbeef, line 2: Should be an upsert for portability.

loopdb/migrate.go
```diff
@@ -0,0 +1,404 @@
...
+func (m *MigratorManager) migrateLoopOuts(ctx context.Context) error {
+	// Fetch all loop outs from the fromStore.
+	loopOuts, err := m.fromStore.FetchLoopOutSwaps(ctx)
+	if err != nil {
+		return err
+	}
+
+	// For each loop out, create a new loop out in the toStore.
+	for _, loopOut := range loopOuts {
+		err := m.toStore.CreateLoopOut(ctx, loopOut.Hash, loopOut.Contract)
```
Roasbeef, line 101: In order to speed this up, we want to be able to batch all the insertion here. Otherwise, you're doing a single db transaction for each swap. Instead a series of swaps can be created, within a single DB transaction.

Roasbeef (dismissed):

loopdb/sqlite_store.go
```diff
@@ -0,0 +1,739 @@
...
+// convertOutgoingChanSet converts a comma separated string of channel IDs into
+// a ChannelSet.
+func convertOutgoingChanSet(outgoingChanSet string) (ChannelSet, error) {
+	// Split the string into a slice of strings
+	chanStrings := strings.Split(outgoingChanSet, ",")
+	channels := make([]uint64, len(chanStrings))
+
+	// Iterate over the chanStrings slice and convert each string to ChannelID
+	for i, chanString := range chanStrings {
+		chanID, err := strconv.Atoi(chanString)
```
bhandras, line 671: I think we want to use `strconv.ParseInt(chanString, 10, 64)` since `Atoi` returns just an `int`.

bhandras (commented):
Awesome work! First pass looks really good already! Very close 🚀

client.go
```diff
@@ -192,13 +192,13 @@ func NewClient(dbDir string, cfg *ClientConfig) (*Client, func(), error) {
 }
 
 // FetchSwaps returns all loop in and out swaps currently in the database.
-func (s *Client) FetchSwaps() ([]*SwapInfo, error) {
-	loopOutSwaps, err := s.Store.FetchLoopOutSwaps()
+func (s *Client) FetchSwaps(ctx context.Context) ([]*SwapInfo, error) {
```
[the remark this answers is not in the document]
bhandras: What I mean is that we could perhaps wrap the `ctx` to a timeout context (per call to the store), to avoid hanging if the DB is for any reason unresponsive.

loopdb/sqlite_test.go
```diff
@@ -0,0 +1,380 @@
...
+				if field.CanSet() {
+					field.Set(reflect.ValueOf(time.Now()))
+				}
+			}
+			if field.Type() == reflect.TypeOf(route.Vertex{}) {
+				if field.CanSet() {
+					field.Set(reflect.ValueOf(route.Vertex{1, 2, 3}))
+				}
+			}
+
```
bhandras, line 375: nit: extra nl

loopd/migration.go
bhandras: I agree with @sputn1ck, it's important to keep the old DB around. It's also likely not large enough to cause any issues regarding storage space and if it is, then the migration above will already fail since this is just a rename.

loopd/utils.go
```diff
@@ -2,34 +2,69 @@ package loopd
...
+		MaxLsatFee:          btcutil.Amount(cfg.MaxLSATFee),
+		LoopOutMaxParts:     cfg.LoopOutMaxParts,
+		TotalPaymentTimeout: cfg.TotalPaymentTimeout,
+		MaxPaymentRetries:   cfg.MaxPaymentRetries,
 	}
 
-	swapClient, cleanUp, err := loop.NewClient(config.DataDir, clientConfig)
+	// Now that we know where the database will live, we'll go ahead and
+	// open up the default implementation of it.
+	var db loopdb.SwapStore
```
bhandras, line 37: nit: these `var` definitions could be merged for readability.

loopdb/macaroons.go
```diff
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"io"
 
+	"github.com/jackc/pgx/v4"
```
bhandras, line 9: nit: commit message: encrpytion => encryption

loopdb/postgres.go
```diff
@@ -65,7 +65,8 @@ type PostgresStore struct {
 
 // NewPostgresStore creates a new store that is backed by a Postgres database
 // backend.
-func NewPostgresStore(cfg *PostgresConfig, network *chaincfg.Params) (*PostgresStore, error) {
+func NewPostgresStore(cfg *PostgresConfig, network *chaincfg.Params,
+	opts ...BaseDBOptionsOptionFunc) (*PostgresStore, error) {
```
bhandras, line 69: nit: add an extra nl after the arguments

GeorgeTsagk (approved):
lgtm + tACK

Pending: decide on migrating macaroons db or not

loopdb/sqlc/migrations/000001_swap.up.sql
```diff
@@ -0,0 +1,166 @@
...
+	-- allow the server to delay the publication in exchange for possibly
+	-- lower fees.
+	publication_deadline TIMESTAMP NOT NULL
+);
+
+
+-- htlc_keys stores public and private keys used when construcing swap HTLCs.
+CREATE TABLE htlc_keys (
+    -- swap_hash points to the parent swap hash.
+	swap_hash BLOB PRIMARY KEY REFERENCES swaps(swap_hash),
```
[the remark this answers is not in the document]
bhandras: Maybe your editor uses tabs and spaces mixed?

loopdb/sql_store.go
```diff
@@ -0,0 +1,722 @@
...
+
+	err := s.ExecTx(ctx, NewSqlReadOpts(), func(*sqlc.Queries) error {
+		swaps, err := s.Queries.GetLoopOutSwaps(ctx)
+		if err != nil {
+			return err
+		}
+
+		loopOuts = make([]*LoopOut, 0, len(swaps))
+
+		for _, swap := range swaps {
```
bhandras, line 33: nit: not causing any issues right now, but would be nicer to avoid this pattern and use an indexed for loop instead, since we're appending a converted loopvar to the `loopOuts` slice, if we ever end up adding a byte slice to the contract then it may cause problems. Similar comment for `FetchLoopIns`.

loopdb/sql_store.go
```diff
@@ -0,0 +1,722 @@
...
+		}
+
+		return nil
+	})
+}
+
+// loopToInsertArgs converts a SwapContract struct to the arguments needed to
+// insert it into the database.
+func loopToInsertArgs(hash lntypes.Hash,
+	swap SwapContract) sqlc.InsertSwapParams {
```
bhandras, line 392: nit: could receive a pointer instead.

loopdb/sql_store.go
```diff
@@ -0,0 +1,722 @@
...
+		loopInInsertParams.LastHop = loopIn.LastHop[:]
+	}
+
+	return loopInInsertParams
+}
+
+// swapToHtlcKeysInsertArgs extracts the htlc keys from a SwapContract struct
+// and converts them to the arguments needed to insert them into the database.
+func swapToHtlcKeysInsertArgs(hash lntypes.Hash,
+	swap SwapContract) sqlc.InsertHtlcKeysParams {
```
bhandras, line 447: nit: could receive a pointer here too.

loopdb/sql_store.go
```diff
@@ -0,0 +1,722 @@
...
+			swap.HtlcKeys.ClientScriptKeyLocator.Index,
+		),
+	}
+}
+
+// convertLoopOutRow converts a database row containing a loop out swap to a
+// LoopOut struct.
+func (s *BaseDB) convertLoopOutRow(row sqlc.GetLoopOutSwapRow,
+	updates []sqlc.SwapUpdate) (
+	*LoopOut, error) {
```
bhandras, line 467: nit: no need to bring return values to the next line.

loopdb/sql_store.go
```diff
@@ -0,0 +1,722 @@
...
+					Server:   btcutil.Amount(updates[i].ServerCost),
+					Onchain:  btcutil.Amount(updates[i].OnchainCost),
+					Offchain: btcutil.Amount(updates[i].OffchainCost),
+				},
+			},
+			Time: updates[i].UpdateTimestamp.UTC(),
+		}
+
+		if updates[i].HtlcTxhash != "" {
+			chainHash, err := chainhash.NewHashFromStr(updates[i].HtlcTxhash)
```
bhandras, line 633: 🎉 glad we store as string.

loopdb/sql_test.go
```diff
@@ -0,0 +1,389 @@
+package loopdb
```
bhandras, line 1: nit: commit message missing `:`

loopd/migration.go
```diff
@@ -0,0 +1,93 @@
...
+	"path/filepath"
+
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightninglabs/loop/loopdb"
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// migrateBoltdb migrates the boltdb to sqlite.
+func migrateBoltdb(ctx context.Context, cfg *Config) error {
+
```
bhandras, line 16: nit: extra nl

loopd/migration.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// needSqliteMigration checks if the boltdb exists at it's default location
+// and returns true if it does.
+func needSqliteMigration(cfg *Config) bool {
```
bhandras, line 80: nit: maybe rename to `needSqlMigration`.

loopd/migration.go
```diff
@@ -0,0 +1,93 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightninglabs/loop/loopdb"
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// migrateBoltdb migrates the boltdb to sqlite.
+func migrateBoltdb(ctx context.Context, cfg *Config) error {
```
bhandras, line 15: Perhaps as a separate PR we can also add the option to migrate sqlite to postgres. Not needed now since there're no users yet on sqlite.

bhandras (approved):
tACK, LGTM! Excellent work @sputn1ck 🎉 🎉 🎉 

Please don't forget to update the release notes as well.

PR #590, opened 2023-05-29
liquidity: get autoloop flag directly from params

GeorgeTsagk:
## Description

Previously we would exclusively pass the autoloop boolean to multiple functions while they had directly access to the manager's parameters. With this commit we remove this explicit flag from the various function interfaces and retrieve the value directly from the parameters.

> NOTE: no changes in tests are required. Actually, we need to see all tests passing to verify that this diff didn't introduce any behavioral changes.


bhandras (approved):
LGTM 🎉

hieblmi (approved):
LGTM!

PR #591, opened 2023-05-29
Autoloop easy tag

GeorgeTsagk:
### Description

Based on https://github.com/lightninglabs/loop/pull/590

This PR adds a new label for easy autoloops. This helps us identify which loops were dispatched from easy-autoloop through the `user_agent` field

### DIFF
--- a/liquidity/liquidity.go
+++ b/liquidity/liquidity.go
@@ -1052,7 +1052,7 @@ func (m *Manager) checkExistingAutoLoops(ctx context.Context,
 	var summary existingAutoLoopSummary
 
 	for _, out := range loopOuts {
-		if out.Contract.Label != labels.AutoloopLabel(swap.TypeOut) {
+		if !isAutoloopLabel(out.Contract.Label) {
 			continue
 		}
 
@@ -1082,7 +1082,7 @@ func (m *Manager) checkExistingAutoLoops(ctx context.Context,
 	}
 
 	for _, in := range loopIns {
-		if in.Contract.Label != labels.AutoloopLabel(swap.TypeIn) {
+		if !isAutoloopLabel(in.Contract.Label) {
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


hieblmi:
Could you briefly explain in the PR description what this new label is used for so new folks get an idea?

liquidity/liquidity.go
```diff
@@ -1052,7 +1052,8 @@ func (m *Manager) checkExistingAutoLoops(ctx context.Context,
 	var summary existingAutoLoopSummary
 
 	for _, out := range loopOuts {
-		if out.Contract.Label != labels.AutoloopLabel(swap.TypeOut) {
+		if out.Contract.Label != labels.AutoloopLabel(swap.TypeOut) &&
+			out.Contract.Label != labels.EasyAutoloopLabel(swap.TypeOut) {
```
hieblmi, line 1056: nit: line length

maybe do something like 
```
label := out.Contract.Label
outLabel := labels.AutoloopLabel(swap.TypeOut)
easyOutLabel := labels.EasyAutoloopLabel(swap.TypeOut)
if label != outLabel && label != easyOutLabel {
	continue
}
```

liquidity/liquidity.go
```diff
@@ -1082,7 +1083,8 @@ func (m *Manager) checkExistingAutoLoops(ctx context.Context,
 	}
 
 	for _, in := range loopIns {
-		if in.Contract.Label != labels.AutoloopLabel(swap.TypeIn) {
+		if in.Contract.Label != labels.AutoloopLabel(swap.TypeIn) &&
+			in.Contract.Label != labels.EasyAutoloopLabel(swap.TypeIn) {
```
hieblmi, line 1087: nit: line length, same as above

hieblmi (commented):
Could you briefly explain in the PR description what this new label is used for so new folks get an idea?

liquidity/liquidity.go
GeorgeTsagk: I created a helper, as we might get even more labels in the future, so we can isolate all checks in there

hieblmi (approved):

PR #594, opened 2023-06-02
Autoloop-in flake fix

GeorgeTsagk:
After the FeePPM miner fee fix there has probably been a race condition in this test. The test would expect only one of the two quotes to work, as one had a marginally high miner fee. After the FeePPM fix that miner fee was considered valid and we dispatched two swaps instead of one, with the test setup examining the first one that reached the go channel.


bhandras (approved):
LGTM 🎉

PR #595, opened 2023-06-14
Easy autoloop feeppm parameter

GeorgeTsagk:
## Description

Previously we would always use a hard-coded 20_000ppm swap fee for automatically dispatched swaps of easy autoloop. With this PR we will respect the `feepercent` parameter, if set.

### DIFF
--- a/docs/autoloop.md
+++ b/docs/autoloop.md
@@ -25,6 +25,10 @@ channel balance of 1M sats. Keep in mind that on first time use this will use
 the default budget parameters. If you wish to configure a custom budget you can
 find more info in the [Budget](#budget) section.
 
+In addition, if you want to control the maximum amount of fees you're willing to
+spend per swap, as a percentage of the total swap amount, you can use the
+`feepercent` parameter as explained in the [Fees](#fees) section.
+
 ## Liquidity Rules
 
 At present, autoloop can be configured to either acquire incoming liquidity 
@@ -89,10 +93,11 @@ to a percentage of the swap amount using the fee percentage parameter:
 loop setparams --feepercent={percentage of swap amount}
 ```
 
-If you would like finer grained control over swap fees, there are multiple fee
-related settings which can be used to tune the autolooper to your preference.
-The sections that follow explain these settings in detail. Note that these fees 
-are expressed on a per-swap basis, rather than as an overall budget. 
+If you are not using easyautoloop and you would like finer grained control over
+swap fees, there are multiple fee related settings which can be used to tune the
+autolooper to your preference. The sections that follow explain these settings
+in detail. Note that these fees are expressed on a per-swap basis, rather than
+as an overall budget. 
 
 ### On-Chain Fees
 When performing a successful loop out swap, the loop client needs to sweep the 
--- a/liquidity/autoloop_test.go
+++ b/liquidity/autoloop_test.go
@@ -1419,6 +1419,33 @@ func TestEasyAutoloop(t *testing.T) {
 
 	c.easyautoloop(step, true)
 	c.stop()
+
+	// Restore the local balance to a higher value that will trigger a swap.
+	easyChannel2.LocalBalance = btcutil.Amount(95000)
+	channels = []lndclient.ChannelInfo{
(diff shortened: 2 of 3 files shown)


liquidity/liquidity.go
```diff
@@ -600,11 +600,21 @@ func (m *Manager) dispatchBestEasyAutoloopSwap(ctx context.Context) error {
...
-	easyParams.FeeLimit = &FeePortion{
-		PartsPerMillion: defaultFeePPM,
+
+	switch easyParams.FeeLimit.(type) {
+	case *FeePortion:
+		if easyParams.FeeLimit.(*FeePortion).PartsPerMillion == 0 {
+			easyParams.FeeLimit = &FeePortion{
+				PartsPerMillion: defaultFeePPM,
+			}
+		}
```
hieblmi, line 613: nit: newline after case

liquidity/autoloop_test.go
```diff
@@ -1419,6 +1419,33 @@ func TestEasyAutoloop(t *testing.T) {
 
 	c.easyautoloop(step, true)
 	c.stop()
+
+	// Restore the local balance to a higher value that requires for a swap
```
hieblmi, line 1423: ...that is required for a swap...?

liquidity/autoloop_test.go
```diff
@@ -1419,6 +1419,33 @@ func TestEasyAutoloop(t *testing.T) {
...
+	// should not be sufficient for the swap to be dispatched.
+	step = &easyAutoloopStep{
+		minAmt: 1,
+		maxAmt: 50000,
+		// Since we have the exact same balance as the first step, we
+		// can reuse the quoteOut1 for the expected loop out quote.
+		quotesOut: quotesOut1,
+	}
+
+	c.easyautoloop(step, false)
```
hieblmi, line 1447: In the test above it says 'no action' equals noop to `true`. Shouldn't this be true here too?

GeorgeTsagk: Right, I might need to polish the doc a bit. Noop means that easyautoloop decided to exit early because no autoloop was needed.

In this case, an loop was going to be dispatched, but fees were not sufficient.

Thanks for pointing out :star2:

liquidity/liquidity.go
```diff
@@ -600,11 +600,21 @@ func (m *Manager) dispatchBestEasyAutoloopSwap(ctx context.Context) error {
...
 
-	// Override our current parameters in order to use the const percent
-	// limit of easy-autoloop.
+	// If no fee is set, override our current parameters in order to use the
+	// default percent limit of easy-autoloop.
 	easyParams := m.params
-	easyParams.FeeLimit = &FeePortion{
-		PartsPerMillion: defaultFeePPM,
+
+	switch easyParams.FeeLimit.(type) {
```
bhandras, line 607: nit: you can avoid the second type cast below by changing this type switch a bit as explained here: https://go.dev/tour/methods/16

docs/autoloop.md
```diff
@@ -25,6 +25,10 @@ channel balance of 1M sats. Keep in mind that on first time use this will use
 the default budget parameters. If you wish to configure a custom budget you can
 find more info in the [Budget](#budget) section.
 
+In addition, if you want to control the maximum amount of fees you want to spend
```
bhandras, line 28: nit: "the maximum amount of fees you want to spend" => "the maximum amount of fees you're willing to spend".

docs/autoloop.md
```diff
@@ -89,7 +93,7 @@ to a percentage of the swap amount using the fee percentage parameter:
 loop setparams --feepercent={percentage of swap amount}
 ```
 
-If you would like finer grained control over swap fees, there are multiple fee
+If you are not using easyautoloop and you would like finer grained control over swap fees, there are multiple fee
```
bhandras, line 96: nit: I'd reformat to 80 cols to align with the existing style.

bhandras (approved):
LGTM 🎉

hieblmi (approved):
Thanks for the explanation. LGTM

PR #596, opened 2023-06-14
liquidity: add false-positive checks on autoloop tests

GeorgeTsagk:
## Description

Previously we would only check the mocked lnd channels for dispatched swaps against our **expected** swaps in the tests. If our expected swaps were `nil` we would never proceed with checking whether the channels had values ready to read or not.

This PR adds extra checks on the `autoloop` and `easyautoloop` step functions to also check for false-positive dispatched swaps.

### DIFF
--- a/liquidity/autoloop_testcontext_test.go
+++ b/liquidity/autoloop_testcontext_test.go
@@ -344,6 +344,24 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
 		return c.manager.numActiveStickyLoops() == 0
 	}, defaultEventuallyTimeout, defaultEventuallyInterval, "failed to"+
 		" wait for sticky loop counter")
+
+	// Since we're checking if any false-positive swaps were dispatched we
+	// need to give some time to autoloop to possibly dispatch them.
+	select {
+	case <-c.outRequest:
+		c.t.Fatal("expected no more loopout requests")
+
+	case <-c.inRequest:
+		c.t.Fatal("expected no more loopin requests")
+
+	case <-c.quoteRequestIn:
+		c.t.Fatal("expected no more loopout quote requests")
+
+	case <-c.quoteRequest:
+		c.t.Fatal("expected no more loopin quote requests")
+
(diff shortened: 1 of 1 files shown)


liquidity/autoloop_testcontext_test.go
```diff
@@ -344,6 +344,21 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
 		return c.manager.numActiveStickyLoops() == 0
 	}, defaultEventuallyTimeout, defaultEventuallyInterval, "failed to"+
 		" wait for sticky loop counter")
+
+	// Since we're checking if any false-positive swaps were dispatched we
+	// need to give some time to autoloop to possibly dispatch them.
+	time.Sleep(500 * time.Millisecond)
```
bhandras, line 350: nit: perhaps it's better to do a `<-time.Afer(500 * time.Millisecond)` inside the `select`.

liquidity/autoloop_testcontext_test.go
```diff
@@ -344,6 +344,21 @@ func (c *autoloopTestCtx) autoloop(step *autoloopStep) {
...
+	select {
+	case <-c.outRequest:
+		c.t.Fatal("expected no more requests")
```
bhandras, line 353: nit: could use separate explanations given these are separate requests. Also add a nl between cases.

liquidity/autoloop_testcontext_test.go
```diff
@@ -383,6 +398,17 @@ func (c *autoloopTestCtx) easyautoloop(step *easyAutoloopStep, noop bool) {
 			actual.OutgoingChanSet,
 		)
 	}
+
+	// Since we're checking if any false-positive swaps were dispatched we
+	// need to give some time to autoloop to possibly dispatch them.
+	time.Sleep(500 * time.Millisecond)
```
bhandras, line 404: nit: similarly could be added to the `select`.

bhandras (approved):
LGTM 💯

PR #598, opened 2023-06-16
loopd: make term and quote calls readable

hieblmi:
Removing loop permissions from the quote and term calls.

### DIFF
--- a/loopd/perms/perms.go
+++ b/loopd/perms/perms.go
@@ -34,30 +34,18 @@ var RequiredPermissions = map[string][]bakery.Op{
 	"/looprpc.SwapClient/LoopOutTerms": {{
 		Entity: "terms",
 		Action: "read",
-	}, {
-		Entity: "loop",
-		Action: "out",
 	}},
 	"/looprpc.SwapClient/LoopOutQuote": {{
 		Entity: "swap",
 		Action: "read",
-	}, {
-		Entity: "loop",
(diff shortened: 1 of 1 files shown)


loopd/perms/perms.go
```diff
@@ -34,30 +34,18 @@ var RequiredPermissions = map[string][]bakery.Op{
...
-	}, {
-		Entity: "loop",
-		Action: "in",
 	}},
 	"/looprpc.SwapClient/GetLoopInQuote": {{
 		Entity: "swap",
 		Action: "read",
-	}, {
-		Entity: "loop",
-		Action: "in",
```
GeorgeTsagk, line 60: I wonder if there was a strong rationale on placing these here?

GeorgeTsagk (approved):
LGTM 🕸️

loopd/perms/perms.go
hieblmi: See my comment here, https://github.com/lightninglabs/terminal-web/issues/662#issuecomment-1594284365. We have to find out :-)

alexbosworth (approved):

hieblmi:
Instead of removing the `loop:in`/`loop:out` perms we preserve them and add these to the client side macaroon permissions. `loop:in`/`loop:out` stay in place here to be able to scope for only one of the directions in the client macaroon.

PR #601, opened 2023-06-22
multi: expose L402 id in grpc interface

sputn1ck:
This commit adds an additional field to the tokens in order to fetch the relevant l402 ids for the users tokens.

We should also consider starting with the naming conversion.


looprpc/client.proto
```diff
@@ -743,6 +743,11 @@ message LsatToken {
     file name of the token where it's stored on the file system.
     */
     string storage_name = 8;
+
+    /*
+    The l402 Id of the token.
```
bhandras, line 748: nit: Id => ID.

bhandras (approved):
LGTM 🌮

GeorgeTsagk (approved):

PR #613, opened 2023-07-26
Remove elapsed reason for autoloop budget

GeorgeTsagk:
## Description

After [introducing recurring budget](https://github.com/lightninglabs/loop/pull/551) the error "elapsed budget" is no longer applicable, as it refers to the old model of an once-off budget with an expiry date.

### DIFF
--- a/docs/autoloop.md
+++ b/docs/autoloop.md
@@ -279,8 +279,6 @@ following reasons will be displayed:
 * Budget not started: if the start date for your budget is in the future,
   no swaps will be executed until the start date is reached. See [budget](#budget) to
   update.
-* Budget elapsed: if the autolooper has elapsed the budget assigned to it for 
-  fees, this reason will be returned. See [budget](#budget) to update.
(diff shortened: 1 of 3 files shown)


docs/autoloop.md
```diff
@@ -290,8 +288,7 @@ following reasons will be displayed:
   [in flight limit](#in-flight-limit) to update.
 * Budget insufficient: if there is not enough remaining budget for a swap, 
   including the amount currently reserved for in flight swaps, an insufficient
```
hieblmi, line 290: nit wording: an insufficient reason => a reason indicating insufficient budget

hieblmi (approved):
LGTM 🚀

bhandras (approved):
Looks good, but this is a breaking change and should be synced with lightning terminal too.

levmi:
ack - will chat with terminal team to see path forward on that front, thanks for the fix!

GeorgeTsagk:
Closing, as seems to no longer be needed

PR #614, opened 2023-07-27
Allow loopd autogenerated TLS cert validity override with a new flag

gcaracuel:
Add a new flag `--tlsvalidity` that allows override default ~14months loopd auto generated certificate validity period.

Use case example:

When using lndconnect string to reach loopd some user will like to have a valid string for longer of 14 months. For those that has a system to rotate this config it might be useful to change this validity period.


gcaracuel:
Guys a compilation of loopserver for arm64 platform will be really welcome :) it has been painful to test this thing on M1/2 chipsets.

loopd/config.go
```diff
@@ -99,9 +99,8 @@ var (
 	)
 
 	// DefaultAutogenValidity is the default validity of a self-signed
-	// certificate. The value corresponds to 14 months
-	// (14 months * 30 days * 24 hours).
-	DefaultAutogenValidity = 14 * 30 * 24 * time.Hour
+	// certificate in number of days. The value corresponds to 420 days (~14 months)
+	DefaultAutogenValidity = 420
```
GeorgeTsagk, line 103: Why not align this with 1 year i.e `365` ?

loopd/config.go
```diff
@@ -152,6 +151,7 @@ type Config struct {
 	TLSExtraDomains    []string `long:"tlsextradomain" description:"Adds an extra domain to the generated certificate."`
 	TLSAutoRefresh     bool     `long:"tlsautorefresh" description:"Re-generate TLS certificate and key if the IPs or domains are changed."`
 	TLSDisableAutofill bool     `long:"tlsdisableautofill" description:"Do not include the interface IPs or the system hostname in TLS certificate, use first --tlsextradomain as Common Name instead, if set."`
+	TLSValidity        int      `long:"tlsvalidity" description:"Loop's TLS certificate validity period in days. Defaults to 430 (roughly 14 months)"`
```
GeorgeTsagk, line 154:
```suggestion
	TLSValidity        int      `long:"tlsvalidity" description:"Loop's TLS certificate validity period in days. Defaults to 420 (roughly 14 months)"`
```

GeorgeTsagk (approved):
lgtm :white_check_mark:

loopd/config.go
bhandras, line 154: nit: Alternatively this could be a `time.Duration` variable so could be configured as:
```tlsvalidity=365d```

**Review by bhandras — APPROVED** [review]

File: loopd/config.go (hunk 1 above)
Comment (gcaracuel) [reply loopd/config.go:103]: I am using current default here. Happy to change that to 365

File: loopd/config.go (hunk 2 above)
Comment (gcaracuel) [reply loopd/config.go]: So? moving to `time.Duration` and `365d` as other comments suggested?

**Comment by gcaracuel** [comment]
Refactored PR with your suggestion, it looks way better now, thanks!

File: loopd/config.go [hunk 3]
```diff
@@ -348,7 +349,12 @@ func Validate(cfg *Config) error {
 
 	// At least one retry.
 	if cfg.MaxPaymentRetries < 1 {
-		return fmt.Errorf("max payment retries must be positive")
+		return fmt.Errorf("max payment retries must be at least 1")
+	}
+
+	// TLS Valisity period to be at least 24 hours
```
GeorgeTsagk, line 355:
```suggestion
	// TLS Validity period to be at least 24 hours
```

PR #617, opened 2023-08-07
loopdb: fix faulty timestamps on startup

sputn1ck:
This PR addresses https://github.com/lightninglabs/lightning-terminal/issues/615 .

For now I only fix the faulty timestamps on startup. We need to discuss whether loop should check if the timestamp is in milliseconds (and use seconds instead) or if we want to change how lightning-terminal constructs the loop out swap.

### DIFF
--- a/loopdb/sql_test.go
+++ b/loopdb/sql_test.go
@@ -346,13 +346,11 @@ func TestIssue615(t *testing.T) {
 					Index:  2,
 				},
 			},
-			MaxMinerFee: 10,
-			MaxSwapFee:  20,
-
+			MaxMinerFee:      10,
+			MaxSwapFee:       20,
 			InitiationHeight: 99,
-
-			InitiationTime:  time.Now(),
-			ProtocolVersion: ProtocolVersionMuSig2,
+			InitiationTime:   time.Now(),
+			ProtocolVersion:  ProtocolVersionMuSig2,
 		},
 		MaxPrepayRoutingFee:     40,
 		PrepayInvoice:           "prepayinvoice",
--- a/loopdb/sqlite.go
+++ b/loopdb/sqlite.go
@@ -316,6 +316,11 @@ func (r *SqliteTxOptions) ReadOnly() bool {
 func parseSqliteTimeStamp(dateTimeStr string) (time.Time, error) {
 	// Split the date and time parts.
 	parts := strings.Fields(strings.TrimSpace(dateTimeStr))
+	if len(parts) <= 2 {
+		return time.Time{}, fmt.Errorf("invalid timestamp format: %v",
+			dateTimeStr)
+	}
+
 	datePart, timePart := parts[0], parts[1]
 
 	return parseTimeParts(datePart, timePart)
@@ -328,6 +333,11 @@ func parseSqliteTimeStamp(dateTimeStr string) (time.Time, error) {
 func parsePostgresTimeStamp(dateTimeStr string) (time.Time, error) {
 	// Split the date and time parts.
 	parts := strings.Split(dateTimeStr, "T")
+	if len(parts) != 2 {
+		return time.Time{}, fmt.Errorf("invalid timestamp format: %v",
+			dateTimeStr)
+	}
+
 	datePart, timePart := parts[0], strings.TrimSuffix(parts[1], "Z")
 
 	return parseTimeParts(datePart, timePart)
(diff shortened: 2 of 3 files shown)


loopdb/sqlite.go
```diff
@@ -191,6 +206,82 @@ func (db *BaseDB) ExecTx(ctx context.Context, txOptions TxOptions,
...
+		faultyTime := parseTimeFunc(swap.PublicationDeadline)
+
+		// Skip if the time is not faulty.
+		if !isMilisecondsTime(faultyTime.Unix()) {
+			continue
+		}
+
+		// Update the faulty time to a valid time.
+		secs := faultyTime.Unix() / 1000
+		nsecs := (faultyTime.Unix() % 1000) * 1e6
```
bhandras, line 259: Since we currently ignore anything less than a second in the parse functions we'll always end up with zero nanosecs.

loopdb/sqlite.go
```diff
@@ -219,3 +310,70 @@ func NewSqlReadOpts() *SqliteTxOptions {
...
+// parseSqliteTimeStamp parses a timestamp string in the format of
+// "YYYY-MM-DDTHH:MM:SSZ" and returns a time.Time value.
+// NOTE: we can't use time.Parse() because it doesn't support having years
+// with more than 4 digits.
+func parsePostgresTimeStamp(dateTimeStr string) time.Time {
+	// Split the date and time parts.
+	parts := strings.Split(dateTimeStr, "T")
+	datePart, timePart := parts[0], strings.TrimSuffix(parts[1], "Z")
+
+	// Parse the date.
```
bhandras, line 350: The rest of this function is the same as for the sqlite case. Perhaps makes sense to extract to a separate function and also add the error handling in so any malformed stings wont bite us.

loopdb/sql_test.go
```diff
@@ -318,6 +318,76 @@ func TestSqliteTypeConversion(t *testing.T) {
...
+// Reference: https://github.com/lightninglabs/lightning-terminal/issues/615
+func TestIssue615(t *testing.T) {
+	ctxb := context.Background()
+
+	// Create a new sqlite store for testing.
+	sqlDB := NewTestDB(t)
+
+	// Create a faulty loopout swap.
+	destAddr := test.GetDestAddr(t, 0)
+	faultyTime := parseSqliteTimeStamp("55563-06-27 02:09:24 +0000 UTC")
```
bhandras, line 331: Did you mean 5563 for the year?

bhandras (commented):
Thanks for the quick fix! 🥇

loopdb/sql_test.go
bhandras: Ah indeed, missed that part on the screenshot!

bhandras (approved):
LGTM, great work on this fix! 💯

loopdb/sql_test.go
```diff
@@ -318,6 +318,78 @@ func TestSqliteTypeConversion(t *testing.T) {
...
+				ClientScriptKeyLocator: keychain.KeyLocator{
+					Family: 1,
+					Index:  2,
+				},
+			},
+			MaxMinerFee: 10,
+			MaxSwapFee:  20,
+
+			InitiationHeight: 99,
+
```
GeorgeTsagk, line 353: nit: extra nl here and above

GeorgeTsagk (approved):

loopdb/sqlite.go
```diff
@@ -219,3 +308,85 @@ func NewSqlReadOpts() *SqliteTxOptions {
...
+	datePart, timePart := parts[0], strings.TrimSuffix(parts[1], "Z")
+
+	return parseTimeParts(datePart, timePart)
+}
+
+// parseTimeParts takes a datePart string in the format of "YYYY-MM-DD" and
+// a timePart string in the format of "HH:MM:SS" and returns a time.Time value.
+func parseTimeParts(datePart, timePart string) (time.Time, error) {
+	// Parse the date.
+	dateParts := strings.Split(datePart, "-")
```
hieblmi, line 340: Check length of `dateParts` to avoid index out of range`.

loopdb/sqlite.go
```diff
@@ -219,3 +308,85 @@ func NewSqlReadOpts() *SqliteTxOptions {
...
+		return time.Time{}, err
+	}
+
+	day, err := strconv.Atoi(dateParts[2])
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	// Parse the time.
+	timeParts := strings.Split(timePart, ":")
```
hieblmi, line 358: Check length of `timeParts` to avoid `index out of range`.

PR #619, opened 2023-08-08
loopdb: fix timestamp parsing

sputn1ck:
This PR changes how the timestamp parsing is managed, by allowing timestamps to be displayed with nanoseconds.
Additionally we now try both timestamp parsing functions on any given string.

### DIFF
--- a/loopdb/sqlite.go
+++ b/loopdb/sqlite.go
@@ -122,7 +122,7 @@ func NewSqliteStore(cfg *SqliteConfig, network *chaincfg.Params) (*SqliteSwapSto
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
-	err = baseDB.FixFaultyTimestamps(ctx, parseSqliteTimeStamp)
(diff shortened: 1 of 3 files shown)


loopdb/sqlite.go
```diff
@@ -309,14 +308,30 @@ func (r *SqliteTxOptions) ReadOnly() bool {
...
 
+// parseTimeStamp tries to parse a timestamp string with both the
+// parseSqliteTimeStamp and parsePostgresTimeStamp functions.
+// If both fail, it returns an error.
+func parseTimeStamp(dateTimeStr string) (time.Time, error) {
+	t, err := parseSqliteTimeStamp(dateTimeStr)
+	if err != nil {
+		t, err = parsePostgresTimeStamp(dateTimeStr)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("invalid timestamp format: %v",
```
bhandras, line 319: nit: i think we can jus return the error here since it's already the same string.

bhandras (approved):
LGTM 🎉

guggero (approved):
tACK, node starts up correctly again.

PR #622, opened 2023-08-10
trivial: clean up typos and comments

hieblmi:
Simple typo fixes and code comments.


loopin.go
```diff
@@ -459,10 +459,9 @@ func (s *loopInSwap) sendUpdate(ctx context.Context) error {
 	info := s.swapInfo()
 	s.log.Infof("Loop in swap state: %v", info.State)
 
+	info.HtlcAddressP2WSH = s.htlcP2WSH.Address
```
bhandras, line 462: nit: can't do this as `s.htlcP2WSH` and `s.htlcP2TR` is not always valid, they could be nil depending on the contract type.

bhandras (approved):
LGTM, with one nit 🎉

GeorgeTsagk (approved):
Lgtm :star:

PR #623, opened 2023-08-11
Correct loop setrule command and styling in Autoloop guide

Liongrass:
The main purpose of this pull request is to fix the `loop setrule --type={in|out}` command. H/t to @hieblmi for finding the error!

The pull request also standardizes some language and capitalization, including
- Capitalizing Loop In and Loop Out
- Capitalizing Autoloop
- Referring to Autoloop as Autoloop, rather then the autolooper


hieblmi (approved):
Thanks for fixing the docs!

PR #625, opened 2023-08-17
Async loopin sweep

hieblmi:
Prior to this PR new auto-loopins were only dispatched once the on-chain htlc was either swept by the server or the time-out path was claimed.

This PR extracts the on-chain notification part into its own goroutine, so that once the swap invoice is paid by the server the swap executor is able to dispatch the next loop-in while the sweep processing is handled in the background.

I tested this PR in regtest with the following auto-loopout configuration:

Set parameters
```
loop --network regtest setparams --autoloop=true --autobudget 10000000 --maxamt 1000000 --feepercent 20 --autoinflight 10
```

Set rule
```
loop --network regtest setrule --type in --outgoing_threshold 99 node_key_to_loop_in_to
```
Result:
A new swap is dispatched although the previous swap is still in a pending state
```
LOOP_IN SUCCESS 0.003 BTC - P2TR: bcrt1p7awp9kaujtdr5h2pzlcpzyuypjsnh35trv5ma49mznfkl6yntntsh739lh (cost: server 130, onchain 7700, offchain 0)
LOOP_IN INVOICE_SETTLED 0.003 BTC - P2TR: bcrt1para8f4fy3r9qtrvvpmrf0z3uzwh776zv7577fsx24k8mvtxfagrqhp6w87 (cost: server -299870, onchain 7700, offchain 0)
LOOP_IN HTLC_PUBLISHED 0.003 BTC - P2TR: bcrt1pzapdkea2gamkf5f7p52lwhx0p55ry57g59qnyugxda6rn7m6295q6zpd9g
```

Accounting considerations @levmi:

To deal with pending swap fees the autoloop engine temporarily assigns pending swaps the maximum allowed fees so to not overshoot the set budget. Once the swap finalizes the actual fees(likely lower) will be set on the swap and the auto-loop dispatcher will reevaluate the creation of a new swap if it was previously blocked by an elapsed budget.
As an example, we could create many swaps in settled state that temporarily chew up our budget - but once the swaps finalize and capital frees new swaps might be dispatched.


loopin_test.go
```diff
@@ -315,6 +320,9 @@ func testLoopInTimeout(t *testing.T, externalValue int64) {
 	require.NoError(t, err)
 	cost.Onchain += fee
 
+	// TODO(hieblmi) remove Sleep and fix flake
```
hieblmi, line 323: I am not clear on why this Sleep here fixes the test flake in Github CI. Locally it does not appear.

alexbosworth (approved):

loopin.go
```diff
@@ -835,44 +831,113 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
+			)
+			if err != nil {
+				s.log.Infof("register spend ntfn: %v", err)
+				return
+			}
+
+			select {
+			// The htlc spend is confirmed. Inspect the spending tx
+			// to determine the final swap state.
+			case spendDetails := <-spendChan:
```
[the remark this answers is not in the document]
hieblmi: I see. I added a case for when the context is cancelled and the Done() channel is closed, so we unblock.

loopin.go
```diff
@@ -835,44 +831,117 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
+	if sweepFee > 0 {
+		// We published the timeout transaction.
+		sweepFeeWg.Done()
+	}
+
+	// We monitor the htlc spend by the server or the timeout sweep in a
+	// separate go routine since we don't want to block on them in order to
+	// continue with successive loop ins once the invoice is settled.
+	executorWg.Add(1)
+	go func(s *loopInSwap) {
```
bhandras, line 848: nit: we don't need to pass in `s` as it's already in scope.

loopin.go
```diff
@@ -835,44 +831,117 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
+		defer executorWg.Done()
+		spent := false
+		notifier := s.lnd.ChainNotifier
+		for !spent {
```
bhandras, line 852: nit: I'd consider extracting the whole goroutine to a method to make it more readable.

loopin.go
```diff
@@ -835,44 +831,117 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
 			s.height = notification.(int32)
 
-			sweepFee, err = publishTxOnTimeout()
-			if err != nil {
-				return err
+			// If we haven't published the timeout transaction
+			// earlier we check if we can publish it. If we publish
+			// it here we mark the sweep fee as set so the htlc
+			// spend-processing goroutine can continue.
+			if sweepFee == 0 {
```
bhandras, line 928: Note that this changes the flow quite a bit. Previously we'd publush the timeout on each block. This is useful too, since the tx may get evicted from the mempool if there's a surge of incoming transactions on the network.

loopin.go
```diff
@@ -835,44 +831,117 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
+				// transaction we wait until the sweep fee was
+				// set by the htlc publisher.
+				isSuccessWitness := s.htlc.IsSuccessWitness(
+					htlcInput.Witness,
+				)
+				if !isSuccessWitness {
+					sweepFeeWg.Wait()
+				}
+
+				err := s.processHtlcSpend(
```
bhandras, line 885: Perhaps worth doing some deeper refactoring here since `processHtlcSpend` again checks if it's a success witness.

loopin.go
hieblmi: That's a good catch, we wouldn't publish again after the first time. Will change that.

loopin.go
hieblmi: Yes, I will redo this in an extra method.

loopin.go
```diff
@@ -835,44 +831,117 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
+					ctx, isSuccessWitness, htlcValue,
+					sweepFee,
```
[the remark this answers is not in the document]
hieblmi: Thank you for the valuable comments. I've now extracted the timeout tx publishing together with the processing of the spend into a separate method which is called as a goroutine. The sweep fee is not set only within this extracted method so we are sure to set the right fee of the confirmed tx for accounting purposes.

loopin.go
```diff
@@ -487,7 +496,7 @@ func (s *loopInSwap) sendUpdate(ctx context.Context) error {
 // execute starts/resumes the swap. It is a thin wrapper around executeSwap to
 // conveniently handle the error case.
 func (s *loopInSwap) execute(mainCtx context.Context,
-	cfg *executeConfig, height int32) error {
+	executorWg *sync.WaitGroup, cfg *executeConfig, height int32) error {
```
bhandras, line 499: We could completely rely on the exectur's wait group so the `loopInSwap.wg` could be removed altogether. And then maybe we could just rename `exectutorWg` to `wg` for simplicity.

loopin.go
```diff
@@ -794,85 +804,49 @@ func getTxFee(tx *wire.MsgTx, fee chainfee.SatPerKVByte) btcutil.Amount {
...
+	}()
 
-			htlcSpend = true
+	isFinalized := func() bool {
+		progress.Lock()
+		defer progress.Unlock()
+		return progress.invoiceFinalized
 
+	}
+	for !isFinalized() {
```
bhandras, line 848: Another way to do this loop could be to move it completely to the async goroutine and just have a single `invoiceFinalized` channel that we wait on here. That way the `htlcKeyRevealed` wouldn't be passed around and we can simplify the code by removing the `swapProgress` struct altogether.

hieblmi: That sounds much simpler! I should have done this from the beginning :)

loopin.go
```diff
@@ -601,17 +602,42 @@ func (s *loopInSwap) executeSwap(globalCtx context.Context) error {
...
-		return err
-	}
+	// receive the preimage and sweep the htlc. We won't wait for the htlc
+	// sweep to occur but instead consider the swap successful if the
+	// invoice is settled. Therefor we monitor the invoice settlement and
+	// htlc process in a separate goroutine. If the invoice is settled or
+	// cancelled we notify the main goroutine on the invFinalizedChan to
+	// continue processing the next swap. The separate goroutine will
+	// continue to watch out for a htlc spend while simultaneously watching
+	// the htlcs expiry height. When the htlc expires, we will publish a
```
hieblmi, line 612: typo

loopin.go
```diff
@@ -601,17 +602,42 @@ func (s *loopInSwap) executeSwap(globalCtx context.Context) error {
...
-	// receive the preimage and sweep the htlc. We are waiting for this to
-	// happen and simultaneously watch the htlc expiry height. When the htlc
-	// expires, we will publish a timeout tx to reclaim the funds.
-	err = s.waitForSwapComplete(globalCtx, htlcOutpoint, htlcValue)
-	if err != nil {
-		return err
-	}
+	// receive the preimage and sweep the htlc. We won't wait for the htlc
+	// sweep to occur but instead consider the swap successful if the
+	// invoice is settled. Therefor we monitor the invoice settlement and
```
hieblmi, line 607: or cancelled

loopin.go
```diff
@@ -829,20 +859,21 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
 
 	// Check timeout at current height. After a restart we may want to
 	// publish the tx immediately.
-	var sweepFee btcutil.Amount
-	sweepFee, err = publishTxOnTimeout()
+	sweepFee, err := publishTxOnTimeout()
 	if err != nil {
-		return err
+		return fmt.Errorf("publishing timout tx error: %v", err)
```
hieblmi, line 864: typo

loopin_test.go
```diff
@@ -63,9 +63,11 @@ func testLoopInSuccess(t *testing.T) {
 
 	ctx.store.assertLoopInStored()
 
-	errChan := make(chan error)
+	errChan := make(chan error, 1)
```
hieblmi, line 66: revert

loopin_test.go
```diff
@@ -193,7 +195,7 @@ func TestLoopInTimeout(t *testing.T) {
 }
 
 func testLoopInTimeout(t *testing.T, externalValue int64) {
-	defer test.Guard(t)()
+	//defer test.Guard(t)()
```
hieblmi, line 198: revert

hieblmi:
@bhandras Thanks for the suggestions. Notice that I removed the previously introduced method `waitForSuccessOrTimeout` and now run `waitForSwapComplete` in its own go routine from which is signaled with two channels if the invoice is settled/cancelled or if an error occurred during invoice processing.
Also note that only invoice related errors are returned from `waitForSwapComplete` since the `errChan` might be closed once the spend processing related errors might trigger.

loopin.go
```diff
@@ -487,9 +485,9 @@ func (s *loopInSwap) sendUpdate(ctx context.Context) error {
 // execute starts/resumes the swap. It is a thin wrapper around executeSwap to
 // conveniently handle the error case.
 func (s *loopInSwap) execute(mainCtx context.Context,
-	cfg *executeConfig, height int32) error {
+	wg *sync.WaitGroup, cfg *executeConfig, height int32) error {
 
-	defer s.wg.Wait()
+	defer wg.Wait()
```
bhandras, line 490: Since the executor will wait for any goroutines started in the swap we don't need this wait.

loopin.go
```diff
@@ -500,9 +498,9 @@ func (s *loopInSwap) execute(mainCtx context.Context,
 	subCtx, cancel := context.WithCancel(mainCtx)
 	defer cancel()
 
-	s.wg.Add(1)
+	wg.Add(1)
 	go func() {
-		defer s.wg.Done()
+		defer wg.Done()
```
bhandras, line 503: So below we use the `subCtx` which will be canceled as soon as the invoice is finalized. I think instead we'd want to use a different context. The `mainCtx` that we pass in to this function lives the whole time the executor is alive so perhaps we need to make a cancellable context per swap that actually cancels when the swap execution is really finished.

loopin.go
```diff
@@ -516,11 +514,12 @@ func (s *loopInSwap) execute(mainCtx context.Context,
...
 	// error occurs.
-	err = s.executeSwap(mainCtx)
+	err = s.executeSwap(mainCtx, wg)
 
 	// Sanity check. If there is no error, the swap must be in a final
-	// state.
-	if err == nil && s.state.Type() == loopdb.StateTypePending {
+	// state. A settled invoice for a loop-in is considered a final state.
+	if err == nil && s.state.Type() == loopdb.StateTypePending &&
+		s.state != loopdb.StateInvoiceSettled {
```
bhandras, line 522: nit: extra nl below.

loopin.go
```diff
@@ -923,10 +960,23 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
 				invoiceFinalized = true
 				htlcKeyRevealed = s.tryPushHtlcKey(ctx)
 
+				// Notify the main goroutine that the swap
+				// is finalized from the client's perspective. A
+				// new swap can be started while this goroutine
+				// is waiting for processing the on-chain spend.
+				invFinalizedChan <- invoiceFinalized
```
bhandras, line 967: There is a rare case this could deadlock. If this goroutine is already in this `case` but the parent context is canceled then we may not ever try to read this channel. I recommend putting this in a select like this:

```
select {
       case invFinalizedChan <- invoiceFinalized:
       case <-ctx.Done():
}
```

liquidity/liquidity.go
```diff
@@ -1202,17 +1202,22 @@ func (m *Manager) currentSwapTraffic(loopOut []*loopdb.LoopOut,
...
+			state  = in.State().State
+		)
 		switch {
-		// Include any pending swaps in our ongoing set of swaps.
-		case in.State().State.Type() == loopdb.StateTypePending:
+		// Include any pending swaps in our ongoing set of swaps. We do
+		// not consider InvoiceSettled a pending state in the context of
+		// loop-ins.
+		case state.Type() == loopdb.StateTypePending &&
+			state != loopdb.StateInvoiceSettled:
```
bhandras, line 1214: nit: add extra nl.

version.go
```diff
@@ -27,7 +27,7 @@ const (
 	// Note: please update release_notes.md when you change these values.
 	appMajor uint = 0
 	appMinor uint = 26
-	appPatch uint = 2
+	appPatch uint = 3
```
bhandras, line 30: No need for bumping here, we usually bump when we tag the next release.

bhandras (commented):
Very close! Just a few nits.

loopin.go
hieblmi: I guess you are referring to this part? https://github.com/lightninglabs/loop/blob/57c9537d49d7ecdae5ade5a76cdb103ede9437d4/executor.go#L219-L222

loopin.go
hieblmi: Added this. I thought now where we manually cancel the parent context once the swap is completed we don't need this. But the parent context could also be canceled in the `subscribeAndLogUpdates` method.

loopin.go
hieblmi: When we pass `subCtx` to `executeSwap` and cancel the context when the swap ends it works for all cases except for when we resume swaps: https://github.com/lightninglabs/loop/blob/master/client.go#L373
Here we pass the executor context, I haven't figured out yet why that deadlocks the swap execution.

hieblmi:
This issue will be addressed by a simpler fix in https://github.com/lightninglabs/loop/pull/636

Issue #626, opened 2023-08-17
Loop DB migration cases


kornpow:
We ran into some issues with migrating to sqlite, because upon reboot we would have the `loop.db` restored, which was previously renamed `loop.db.bk` by the migrator function.

# Idea for a fix
Would it be possible to add additional checks to the migration script which prevents attempting migration of `loop.db` if there already exists a migrated (non-empty) `loop_sqlite.db` file?

# What would've worked better if done differently (for us) 
If the migrator copied the db to loop.db.bk and then migrated the db in-place to sqlite for a final destination of `loop.db`.
(Or something like that)

PR #627, opened 2023-08-17
loopd: always skip migration if `loop_sqlite.db` file already exists

bhandras:
This commit ensures that we don't even check for the existence of the loop.db file if `loop_sqlite.db` already present. This is to ensure that on systems that may by any trigger restore deleted files we don't try to re-run migration which would otherwise make `loopd` unable to start up properly.

Fixes https://github.com/lightninglabs/loop/issues/626

#### Pull Request Checklist

### DIFF
--- a/loopd/migration.go
+++ b/loopd/migration.go
@@ -82,6 +82,17 @@ func needSqlMigration(cfg *Config) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


GeorgeTsagk (approved):
Lgtm :ok:

Issue #628, opened 2023-08-21
After upgrading to v0.10.5, I can't start litd with error "unable to start loop in integrated mode: unable to open database: invalid time format"


MakerRain:
**Expected behavior**
Before upgrade, I encounter "Unable to start loop in integrated mode: sql: Scan error on column index 20, name 'publication_deadline' ". As stated in lightninglabs/lightning-terminal#615 .

That issue solved after upgrade to v0.10.5. But after reboot system, litd cannot startup.

**Actual behavior**
Loop failed with error:
2023-08-21 13:59:46.903 [ERR] LITD: Error starting Lightning Terminal: could not start subservers: unable to start loop in integrated mode: unable to open database: invalid time format: 16:07:49+08:00

**To reproduce**
Every time I start litd, this error occurred. 

**System information**
raspiberry pi 4, Linux version 5.10.103-v7l+.

I use Lightning Terminal followed by raspibolt and together with Ride the Lightning.

MakerRain:
@hieblmi , thanks for your help. I created swaps through Lightning Terminal on local web. The format such as "16:07:49+08:00" lies in my log file. Detail logs below:

8月 21 12:26:19 raspberrypi systemd[1]: Started Lightning Terminal Daemon.
8月 21 12:26:19 raspberrypi litd[14758]: 2023-08-21 12:26:19.962 [INF] LITD: LiT is running in mainnet, the taproot assets subserver do not support the `mainnet` network yet, disabling taproot assets subserver
8月 21 12:26:19 raspberrypi litd[14758]: 2023-08-21 12:26:19.964 [INF] LITD: LiT version: 0.10.5-alpha commit=v0.10.5-alpha
8月 21 12:26:19 raspberrypi litd[14758]: 2023-08-21 12:26:19.983 [INF] LITD: Listening for http_tls on: 127.0.0.1:8443
8月 21 12:26:19 raspberrypi litd[14758]: 2023-08-21 12:26:19.985 [INF] LITD: Listening for http on: 127.0.0.1:8442
8月 21 12:26:20 raspberrypi litd[14758]: 2023-08-21 12:26:20.007 [INF] SESS: Checking for schema update: latest_version=0, db_version=0
8月 21 12:26:20 raspberrypi litd[14758]: 2023-08-21 12:26:20.008 [INF] LITD: Dialing lnd gRPC server at 127.0.0.1:10009
[...]

PR #630, opened 2023-08-22
sqldb: fix timeparsing edge cases

sputn1ck:
This PR adds additional fixes for more timestamp formats.

I was not able to reproduce https://github.com/lightninglabs/loop/issues/628, but I added improved logging, so should the error persist we would be more able to fully grasp the failed string.

### DIFF
--- a/loopdb/sql_test.go
+++ b/loopdb/sql_test.go
@@ -328,8 +328,10 @@ func TestIssue615(t *testing.T) {
 
 	// Create a faulty loopout swap.
 	destAddr := test.GetDestAddr(t, 0)
-	faultyTime, err := parseSqliteTimeStamp("55563-06-27 02:09:24 +0000 UTC")
-	require.NoError(t, err)
+	// Corresponds to 55563-06-27 02:09:24 +0000 UTC.
+	faultyTime := time.Unix(1691247002964, 0)
+
+	t.Log(faultyTime.Unix())
 
 	unrestrictedSwap := LoopOutContract{
 		SwapContract: SwapContract{
@@ -362,7 +364,7 @@ func TestIssue615(t *testing.T) {
 		SwapPublicationDeadline: faultyTime,
 	}
 
-	err = sqlDB.CreateLoopOut(ctxb, testPreimage.Hash(), &unrestrictedSwap)
+	err := sqlDB.CreateLoopOut(ctxb, testPreimage.Hash(), &unrestrictedSwap)
 	require.NoError(t, err)
 
 	// This should fail because of the faulty timestamp.
@@ -441,7 +443,7 @@ func TestTimeConversions(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		time, err := parseTimeStamp(test.timeString)
+		time, err := fixTimeStamp(test.timeString)
 		require.NoError(t, err)
 		require.Equal(t, test.expectedTime, time)
 	}
--- a/loopdb/sqlite.go
+++ b/loopdb/sqlite.go
@@ -248,19 +248,25 @@ func (b *BaseDB) FixFaultyTimestamps(ctx context.Context) error {
 	defer tx.Rollback() //nolint: errcheck
 
 	for _, swap := range loopOutSwaps {
-		faultyTime, err := parseTimeStamp(swap.PublicationDeadline)
+
+		// Get the year of the timestamp.
+		year, err := getTimeStampYear(swap.PublicationDeadline)
 		if err != nil {
 			return err
 		}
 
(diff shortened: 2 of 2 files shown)


loopdb/sqlite.go
```diff
@@ -377,12 +378,25 @@ func parseTimeParts(datePart, timePart string) (time.Time, error) {
 		return time.Time{}, err
 	}
 
-	day, err := strconv.Atoi(dateParts[2])
+	dateIndex := 2
```
hieblmi, line 381: I find it hard to parse what the code here is doing. Maybe we can add the template date string that is handled here to clarify?

hieblmi: Also, maybe in the two returns above maybe we can add `unable to atoi month` and `unable to atoi year` to the error string.

hieblmi (commented):
I am thinking that rather than adjusting the date parsing manually for every new issue we could maybe use a battle tested date parsing module here?

hieblmi (approved):
lgtm ✅

loopdb/sqlite.go
```diff
@@ -369,20 +370,27 @@ func parseTimeParts(datePart, timePart string) (time.Time, error) {
...
 	day, err := strconv.Atoi(dateParts[2])
 	if err != nil {
-		return time.Time{}, err
+		return time.Time{}, fmt.Errorf("unable to parse day: %v", err)
 	}
 
 	// Parse the time.
+
+	// Remove the timezone part if it exists. E.g. 2023-08-04 16:07:49+0800 CST
+	if strings.Contains(timePart, "+") {
```
bhandras, line 389: I'm afraid there's some more things to consider. I was trying to look up where these things come from and it seems to me that this is governed by the ISO 8601 standard (https://en.wikipedia.org/wiki/ISO_8601). According to that specification the offset may also be negative.

GeorgeTsagk: +1

GeorgeTsagk (approved):
Lgtm, pending the timezone nit

loopdb/sqlite.go
```diff
@@ -335,9 +336,13 @@ func parseSqliteTimeStamp(dateTimeStr string) (time.Time, error) {
...
 
-	datePart, timePart := parts[0], parts[1]
+	dateTimeStr, err := replaceFaultyYear(parts[0], dateTimeStr)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("%v: unable to replace faulty "+
+			"year: %v", dateTimeStr, err)
+	}
 
-	return parseTimeParts(datePart, timePart)
+	return time.Parse("2006-01-02 15:04:05.99999 -0700 MST", dateTimeStr)
```
bhandras, line 345: Instead of this, I'm wondering if we should try to instead parse the time based on the supported RFC's as documented here: https://pkg.go.dev/time#pkg-constants

loopdb/sqlite.go
```diff
@@ -311,117 +312,87 @@ func (r *SqliteTxOptions) ReadOnly() bool {
...
+		)
 	}
 
-	datePart, timePart := parts[0], strings.TrimSuffix(parts[1], "Z")
+	var layouts []string
 
-	return parseTimeParts(datePart, timePart)
-}
+	// Add the custom go sqlite layout.
+	layouts = append(layouts, "2006-01-02 15:04:05.99999 -0700 MST")
```
bhandras, line 341: nit: maybe we could add this layout to the `DefaultLayouts` function too.

loopdb/sqlite.go
```diff
@@ -311,117 +312,87 @@ func (r *SqliteTxOptions) ReadOnly() bool {
...
+func DefaultAllLayouts() []string {
+	return []string{
+		time.RFC3339Nano,
+		time.RFC3339,
+		time.RFC1123Z,
+		time.RFC1123,
+		time.RFC850,
+		time.RFC822Z,
+		time.RFC822,
+		time.Layout,
```
bhandras, line 378: So I wonder if we need all these layouts?

loopdb/sql_test.go
```diff
@@ -516,3 +519,68 @@ func randomStruct(v interface{}) error {
 
 	return nil
 }
+
+// parseTimeParts takes a datePart string in the format of "YYYY-MM-DD" and
+// a timePart string in the format of "HH:MM:SS" and returns a time.Time value.
+func parseTimeParts(dateTimeStr string) (time.Time, error) {
```
bhandras, line 525: Do we need this function?

bhandras (approved):
Fix LGTM, just a few nits! 🎉

PR #641, opened 2023-09-28
multi: bump lnd version to v0.17.0-beta

guggero:
Prepares everything to update to `lnd v0.17.0-beta` once the final release is out (we'll want to wait with merging this until then).

#### Pull Request Checklist


loopdb/sqlite.go
```diff
@@ -211,14 +211,19 @@ func (db *BaseDB) ExecTx(ctx context.Context, txOptions TxOptions,
...
 func (b *BaseDB) FixFaultyTimestamps(ctx context.Context) error {
-
 	// Manually fetch all the loop out swaps.
 	rows, err := b.DB.QueryContext(
 		ctx, "SELECT swap_hash, publication_deadline FROM loopout_swaps",
 	)
 	if err != nil {
 		return err
 	}
+	if rows.Err() != nil {
```
bhandras, line 221: Nice catch! I believe instead what we'd want is first go through the rows (line 237), then after that `for` loop we'd:
```
	if err := rows.Close(); err != nil {
		return nil, err
	}
	if err := rows.Err(); err != nil {
		return nil, err
	}
```

bhandras (approved):
Nice fixes, thanks! 🎉

loopdb/sqlite.go
guggero: Ah, good point. Though the linter really wants me to call `Close` with a `defer`... So how does that latest version look to you?

bhandras (approved):

loopdb/sqlite.go
bhandras: Looks good!

ellemouton (commented):
lgtm once updated to rc6 ⚡

guggero:
Bumped to final version `lnd v0.17.0-beta`, will merge after green CI.

PR #642, opened 2023-09-28
[1/?] Static Loop-In Address - Create

hieblmi:
This draft PR implements static address generation for loop-ins.
It includes new command line options for `loop`, namely `~$ loop in static new` a.k.a `~$ loop in s n`.

Missing features to consider this PR complete are:
- [x] unit tests / mocks


cmd/loop/staticaddr.go
```diff
@@ -0,0 +1,75 @@
...
+
+	resp, err := client.NewAddress(
+		ctxb, &looprpc.NewAddressRequest{},
+	)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Received a new static loop-in address from the server: "+
+		"%s.\nEach individual deposit to this static address "+
```
hieblmi, line 53: nit: remove period after address.

staticaddr/actions.go
```diff
@@ -0,0 +1,86 @@
...
+		return a.HandleError(err)
+	}
+
+	a.addrParams.ClientPubkey = clientKey.PubKey
+	a.addrParams.KeyLocator = &clientKey.KeyLocator
+
+	return OnInitAddressSuccess
+}
+
+// CreateAddressAction is the action that is executed when the
```
hieblmi, line 40: nit: early line break

cmd/loop/staticaddr.go
```diff
@@ -0,0 +1,76 @@
...
+		// spendOutputsCommand,
+	},
+}
+
+var newStaticAddressCommand = cli.Command{
+	Name:      "new",
+	ShortName: "n",
+	Usage:     "Create a new static loop in address.",
+	Description: `
+	Requests a new static loop in address from the server. Musig 2:2...
```
hieblmi, line 28: nit: better description

staticaddr/address_interface.go
```diff
@@ -0,0 +1,78 @@
+package staticaddr
```
hieblmi, line 1: rename file to interface.go

staticaddr/script/script.go
```diff
@@ -0,0 +1,115 @@
...
+
+	taprootKey, _, _, err := TaprootKey(expiry, serverKey, clientKey)
+	if err != nil {
+		return nil, err
+	}
+	return payToWitnessTaprootScript(taprootKey)
+}
+
+// TaprootKey returns the aggregated MuSig2 combined internal key and the
+// tweaked Taproot key of a reusable address output, as well as the expiry
```
bhandras, line 53: nit: comment is wrong as the returned key is already tweaked. Added the same comment to the reservation's PR, imo it's better to just return the musig2 AggregateKey struct.

bhandras (commented):
Great work! Did just a very light pass to see where the PR is.

loopdb/sqlc/migrations/000003_static_address.up.sql
```diff
@@ -0,0 +1,62 @@
...
+    -- server_pubkey is the server side public taproot key that is used to
+    -- construct the 2-of-2 MuSig2 taproot output that represents the static
+    -- address.
+    server_pubkey BYTEA NOT NULL,
+
+    -- expiry denotes the CSV delay at which funds at a specific static address
+    -- can be swept back to the client.
+    expiry INT NOT NULL,
+
+    -- server_key_family is the key family of the server public key from the
```
hieblmi, line 24: client_key_family

staticaddr/script/script.go
hieblmi: I see, the `aggregateKey` is already tweaked.

hieblmi:
Accidentally force-pushed without commits. Reopening...

staticaddr/manager.go
```diff
@@ -0,0 +1,127 @@
...
+// Run runs the address manager.
+func (m *AddressManager) Run(ctx context.Context) error {
+	log.Debugf("Starting address manager...")
+
+	runCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	m.Lock()
+	m.runCtx = runCtx
+	m.Unlock()
```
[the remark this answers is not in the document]
hieblmi: oh so that the `manager.Run()` caller continues once the init channel is closed?

cmd/loop/staticaddr.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+var staticAddressCommands = cli.Command{
+	Name:      "static",
+	ShortName: "s",
+	Usage:     "manage static loop-in addresses",
+	Category:  "StaticAddress",
+	Subcommands: []cli.Command{
+		newStaticAddressCommand,
+		// listOutputsCommand,
```
hieblmi, line 18: nit: remove

staticaddr/manager.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightninglabs/loop"
+	"github.com/lightninglabs/loop/staticaddr/script"
+	staticaddressrpc "github.com/lightninglabs/loop/swapserverrpc"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+type ManagerConfig struct {
+	AddressClient staticaddressrpc.StaticAddressServerClient
```
hieblmi, line 18: godoc

staticaddr/manager.go
```diff
@@ -0,0 +1,152 @@
...
+	<-runCtx.Done()
+
+	return nil
+}
+
+// NewAddress starts a new address creation flow.
+func (m *AddressManager) NewAddress(ctx context.Context) (*AddressParameters,
+	error) {
+
+	localPubKey := m.cfg.Lnd.NodePubkey
```
hieblmi, line 69: comment about probing

loopdb/sqlite.go
```diff
@@ -293,6 +293,10 @@ func (b *BaseDB) FixFaultyTimestamps(ctx context.Context) error {
 	return tx.Commit()
 }
 
+func (db *BaseDB) GetNetwork() *chaincfg.Params {
```
bhandras, line 296: nit: missing godoc.

staticaddr/sql_store.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/jackc/pgx/v4"
+	"github.com/lightninglabs/loop/loopdb"
+	"github.com/lightninglabs/loop/loopdb/sqlc"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+// SqlStore is the backing store for static addresses.
+type SqlStore struct {
+	baseDB  *loopdb.BaseDB
+	nowTime func() time.Time
```
bhandras, line 19: Or alternatively we could use LND's `clock.Clock` (which also has a nice test version `clock.TestClock`).

staticaddr/sql_store.go
```diff
@@ -0,0 +1,194 @@
...
+		ClientKeyIndex:  int32(addrParams.KeyLocator.Index),
+		Pkscript:        addrParams.PkScript,
+		AddressHrp:      addrParams.Address.Hrp(),
+		AddressWitnessVersion: int16(
+			addrParams.Address.WitnessVersion(),
+		),
+		AddressWitnessProgram: addrParams.Address.WitnessProgram(),
+	}
+
+	return s.ExecTx(ctx, func(q *sqlc.Queries) error {
```
bhandras, line 92: nit: don't need the `ExecTx` since it's just one call.

staticaddr/sql_store.go
```diff
@@ -0,0 +1,194 @@
...
+// GetAllStaticAddresses returns all address known to the server.
+func (s *SqlStore) GetAllStaticAddresses(ctx context.Context) (
+	[]*AddressParameters, error) {
+
+	staticAddresses, err := s.baseDB.Queries.AllStaticAddresses(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []*AddressParameters
```
bhandras, line 136: nit: could be presized.

staticaddr/script/script.go
```diff
@@ -0,0 +1,113 @@
+package script
```
bhandras, line 1: nit: ideally we have some basic tests for these scripts. There are examples of how to do that in `swap/htlc_test.go`.

staticaddr/manager.go
```diff
@@ -0,0 +1,152 @@
...
+// NewAddressManager creates a new address manager.
+func NewAddressManager(cfg *ManagerConfig) *AddressManager {
+	return &AddressManager{
+		cfg:      cfg,
+		initChan: make(chan struct{}),
+	}
+}
+
+// Run runs the address manager.
+func (m *AddressManager) Run(ctx context.Context) error {
```
bhandras, line 47: Will the `AddressManager` have a runloop? If not, then a simple `Init` call would suffice (and also no need for storing the `runCtx`).

swapserverrpc/server.proto
```diff
@@ -617,3 +617,28 @@ message ServerPushKeyReq {
...
 }
+
+service StaticAddressServer {
+    rpc ServerNewAddress (ServerNewAddressRequest)
+        returns (ServerNewAddressResponse);
+}
+
+message ServerNewAddressRequest {
+    // The client key for the MuSig2 static address output.
+    bytes client_key = 1;
```
bhandras, line 628: Ideally we have a protocol version here just in case the API changes.

bhandras (commented):
Great work! Left some minor comments. Perhaps most important is to have some test coverage for the static addr script.

staticaddr/manager.go
hieblmi: Yes, the manager will have a run loop to handle different actions of a static address like timeout, confirmations, coop-spends and loopins.

staticaddr/script/script.go
hieblmi: Done, wrote test cases in `script_test.go` for success/error of coop and timeout spend.

staticaddr/protocol_version.go
```diff
@@ -0,0 +1,29 @@
+package staticaddr
+
+type ProtocolVersion uint32
```
bhandras, line 3: Why not just use the the `loopdb` provided protocol version instead?

staticaddr/sql_store.go
```diff
@@ -0,0 +1,195 @@
...
+			log.Errorf("unable to rollback db tx: %v", err)
+		}
+	}()
+
+	if err := txBody(s.baseDB.Queries.WithTx(tx)); err != nil {
+		return err
+	}
+
+	// Commit transaction.
+	if err = tx.Commit(); err != nil {
```
bhandras, line 66: nit: can just return the error here.

staticaddr/script/script_test.go
```diff
@@ -0,0 +1,349 @@
+package script
```
bhandras, line 1: Very nice!!! 🥇

swapserverrpc/server.proto
```diff
@@ -617,3 +617,42 @@ message ServerPushKeyReq {
...
+        returns (ServerNewAddressResponse);
+}
+
+/**
+This enum defines the protocol versions that clients may adhere to. Note that
+this is not a flagged enum. If a particular protocol version adds a feature,
+then in general all the preceding features are also supported. Exception to this
+is when features get deprecated.
+*/
+enum AddressProtocolVersion {
```
bhandras, line 632: So instead of this we could just use the existing protocol version.

staticaddr/manager.go
bhandras: nit: missing godocs.

staticaddr/manager.go
bhandras: If that's coming in follow up PR's then perhaps worth thinking about whether we need the embedded context there.

staticaddr/manager.go
```diff
@@ -0,0 +1,167 @@
...
+
+// NewAddress starts a new address creation flow.
+func (m *AddressManager) NewAddress(ctx context.Context) (*AddressParameters,
+	error) {
+
+	// Before we request a new address we ask the loop server to probe us.
+	// This ensures connectivity and obtains a L402 ID if we haven't gotten
+	// one earlier.
+	localPubKey := m.cfg.Lnd.NodePubkey
+	err := m.cfg.LoopClient.Server.Probe(ctx, 1, localPubKey, nil, nil)
```
bhandras, line 75: Do we need the probe? Since we're later also calling the server so L402 will be obtained there.

bhandras (commented):
Very nice! Looks pretty close imo.

fsm/fsm.go
```diff
@@ -81,6 +81,11 @@ type Observer interface {
 	Notify(Notification)
 }
 
+// StateMachineOpts represents options that can be passed to the state machine.
```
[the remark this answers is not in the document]
hieblmi: nope!

staticaddr/manager.go
hieblmi: The L402 is used server-side to evaluate if the client already has a static address or if a new one has to be created.

loopdb/sqlite.go
```diff
@@ -194,7 +194,7 @@ func (db *BaseDB) ExecTx(ctx context.Context, txOptions TxOptions,
 
 	// Rollback is safe to call even if the tx is already closed, so if
 	// the tx commits successfully, this is a no-op.
-	defer tx.Rollback() //nolint: errcheck
+	defer tx.Rollback() // nolint: errcheck
```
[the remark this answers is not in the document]
hieblmi: my ide settings do this automatically, struggling to get this space in

staticaddr/manager.go
hieblmi: I've removed the embedded context for now.

staticaddr/manager.go
hieblmi: Using @sputn1ck's fetch-method once available.

swapserverrpc/server.proto
hieblmi: I thought we keep the address versions separated from the normal loop features but I guess new address features can just be part of the existing protocol versions.

staticaddr/manager.go
hieblmi: Done!

swapserverrpc/server.proto
hieblmi: Done, added a new version `ProtocolVersionStaticAddress`.

staticaddr/protocol_version.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// AddressProtocolVersion represents the protocol version (declared on rpc
+// level) that the client declared to us.
+type AddressProtocolVersion uint32
+
+const (
+	// ProtocolVersionLegacy indicates that the client is a legacy version
+	// that did not report its protocol version.
+	ProtocolVersionLegacy AddressProtocolVersion = 0
```
bhandras, line 16: This (and unrecorded) can't happen as any client we release that supports static address will report their static addr protocol version. So I'd suggest to just add the first version with an appropriate name, such as STATIC_ADDR_V1 or anything you feel is better (value can be 0 or 1 too).

staticaddr/protocol_version.go
```diff
@@ -0,0 +1,63 @@
...
+
+// CurrentRPCProtocolVersion returns the RPC protocol version selected to be
+// used for new swaps.
+func CurrentRPCProtocolVersion() looprpc.StaticAddressProtocolVersion {
+	return currentRPCProtocolVersion
+}
+
+// ResetCurrentProtocolVersion resets the current protocol version to the stable
+// protocol. Note: used in integration tests only!
+func ResetCurrentProtocolVersion() {
```
bhandras, line 42: Don't think we even need this, this is coming from the loop protocol version code to aid testing of experimental features.

staticaddr/sql_store.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/lightninglabs/loop/loopdb"
+	"github.com/lightninglabs/loop/loopdb/sqlc"
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+// SqlStore is the backing store for static addresses.
+type SqlStore struct {
+	baseDB  *loopdb.BaseDB
+	nowTime clock.Clock
```
bhandras, line 19: nit: instead of `nowTime` could be named as `clock`.

staticaddr/sql_store.go
```diff
@@ -0,0 +1,181 @@
...
+
+// ExecTx is a wrapper for txBody to abstract the creation and commit of a db
+// transaction. The db transaction is embedded in a `*sqlc.Queries` that txBody
+// needs to use when executing each one of the queries that need to be applied
+// atomically.
+func (s *SqlStore) ExecTx(ctx context.Context,
+	txBody func(queries *sqlc.Queries) error) error {
+
+	// Create the db transaction.
+	writeOpts := &loopdb.SqliteTxOptions{}
```
bhandras, line 39: The tx options should be an argument so we can bundle both reads and writes.

staticaddr/sql_store.go
```diff
@@ -0,0 +1,181 @@
...
+		err := tx.Rollback()
+		switch {
+		// If the tx was already closed (it was successfully executed)
+		// we do not need to log that error.
+		case errors.Is(err, pgx.ErrTxClosed):
+			return
+
+		// If this is an unexpected error, log it.
+		case err != nil:
+			log.Errorf("unable to rollback db tx: %v", err)
```
bhandras, line 57: 👍 nice!

staticaddr/script/script.go
```diff
@@ -0,0 +1,200 @@
...
+func parsePubKey(muSig2Version input.MuSig2Version, key [33]byte) (
+	*btcec.PublicKey, error) {
+
+	// Make sure that we have the correct public keys depending on the
+	// MuSig2 version.
+	switch muSig2Version {
+	case input.MuSig2Version100RC2:
+		return btcec.ParsePubKey(key[:])
+
+	case input.MuSig2Version040:
```
bhandras, line 126: Don't think we should support v0.4.

staticaddr/manager.go
```diff
@@ -0,0 +1,167 @@
...
+	// LoopClient provides loop rpc functionality.
+	SwapClient *loop.Client
+
+	// Store is the database store that is used to store static address
+	// related records.
+	Store AddressStore
+
+	// Lnd is the lnd client that is used to derive new keys from lnd's
+	// wallet.
+	Lnd lndclient.LndServices
```
bhandras, line 34: Would be enough to just store the walletkit until we need more interfaces.

staticaddr/manager.go
```diff
@@ -0,0 +1,167 @@
...
+	return nil
+}
+
+// NewAddress starts a new address creation flow.
+func (m *AddressManager) NewAddress(ctx context.Context) (*AddressParameters,
+	error) {
+
+	// We are fetching a new L402 token from the server. There is one static
+	// address per L402 token allowed.
+	err := m.cfg.SwapClient.Server.FetchL402(ctx)
```
bhandras, line 74: Cool!

bhandras (commented):
Looks very-very good! 🎉

staticaddr/manager.go
hieblmi: 🧊

staticaddr/manager.go
hieblmi: We also need the chain params in this PR, and in the follow-up PRs the Chainnotifier and others.

What's the drawback in importing all of the `LndServices`? I like it cause it keeps the manager config more compact.

staticaddr/manager.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+
+	return addrParams, nil
+}
+
+// WaitInitComplete waits until the address manager has completed its setup.
+func (m *AddressManager) WaitInitComplete() {
+	defer log.Debugf("Address manager initiation complete.")
+	<-m.initChan
```
hieblmi, line 164: pass in context and wait on `Done()`.

staticaddr/interface.go
```diff
@@ -0,0 +1,84 @@
...
+	// ClientPubkey is the client's pubkey for the static address. It is
+	// used for the 2-of-2 funding output as well as for the client's
+	// timeout path.
+	ClientPubkey *btcec.PublicKey
+
+	// ClientPubkey is the client's pubkey for the static address. It is
+	// used for the 2-of-2 funding output.
+	ServerPubkey *btcec.PublicKey
+
+	// Expiry is the CSV timout value at which the client can claim the
```
bhandras, line 69: I think you meant CLTV (so absolute time) right?

staticaddr/manager.go
```diff
@@ -0,0 +1,165 @@
...
+	// lnd's wallet.
+	WalletKit lndclient.WalletKitClient
+
+	// ChainParams is the chain configuration(mainnet, testnet...) this
+	// manager uses.
+	ChainParams *chaincfg.Params
+}
+
+// AddressManager manages the address state machines.
+type AddressManager struct {
```
bhandras, line 43: nit: either we should call this `Manager` or the config `AddressManagerConfig` for consistency.

bhandras (commented):
Looks good, just a few nits!

staticaddr/interface.go
hieblmi: CSV here is correct since the static address doesn't change, hence the expiry can't have absolute time since that would change the script hash every time.

cmd/loop/staticaddr.go
```diff
@@ -0,0 +1,77 @@
...
+	resp, err := client.NewAddress(
+		ctxb, &looprpc.NewAddressRequest{},
+	)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Received a new static loop-in address from the server: "+
+		"%s\nEach individual deposit to this static address "+
+		"expires after %d blocks.\n", resp.Address, resp.Expiry)
```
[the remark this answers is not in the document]
hieblmi: The expiry here is CSV so only relative blocks are possible to display. Once a deposit happens we can determine the absolute expiry. Will keep track of it on the deposit in a future PR.

staticaddr/manager.go
```diff
@@ -0,0 +1,165 @@
...
+	// AddressClient is the client that communicates with the loop server
+	// to manage static addresses.
+	AddressClient staticaddressrpc.StaticAddressServerClient
+
+	// LoopClient provides loop rpc functionality.
+	SwapClient *loop.Client
+
+	// Store is the database store that is used to store static address
+	// related records.
+	Store AddressStore
```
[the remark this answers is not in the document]
hieblmi: We'll init these from the daemon, so that's where they are accessed.

cmd/loop/staticaddr.go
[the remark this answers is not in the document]
bhandras: Thanks for the clarification, sgtm.

staticaddr/manager.go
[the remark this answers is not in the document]
bhandras: imho, it's better to pass dependencies to the constructor as that will create a compile time check (that the constructor is called correctly) vs just filling members which we'd need runtime checks for.

bhandras (approved):
LGTM (just needs clean rebase to a fresh master) 🎉 🎉

loopdb/sqlc/migrations/000007_static_address.up.sql
```diff
@@ -0,0 +1,49 @@
+-- static_address stores the static loop-in addresses that clients
+-- cooperatively created with the server.
+CREATE TABLE IF NOT EXISTS static_addresses (
+    -- id is the auto-incrementing primary key for a static address.
+    id INTEGER PRIMARY KEY,
+
+    -- address_id is a unique id to identify a specif address.
+    address_id BYTEA NOT NULL UNIQUE,
```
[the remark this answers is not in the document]
hieblmi: If multiple addresses per L402 become a feature shouldn't we already have a unique ID in place? Or should we add it once we go that route?

swapserverrpc/server.proto
```diff
@@ -652,4 +652,32 @@ message FetchL402Response {
...
+    bytes client_key = 2;
+}
+
+message ServerNewAddressResponse {
+    ServerAddressParameters params = 1;
+}
+
+message ServerAddressParameters {
+    // A unique identifier for static address parameters assigned by the server.
+    bytes id = 1;
```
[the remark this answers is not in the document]
hieblmi: I thought this might be handy once we enable multple static addresses per client?

looprpc/staticaddressclient.pb.json.go
```diff
@@ -0,0 +1,48 @@
+// Code generated by falafel 0.9.1. DO NOT EDIT.
```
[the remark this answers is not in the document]
hieblmi: For me it gets generated.

loopdb/sqlc/migrations/000007_static_address.up.sql
```diff
@@ -0,0 +1,49 @@
...
+
+    -- client_key_index is the key index of the client public key from the
+    -- client's lnd wallet.
+    client_key_index INT NOT NULL,
+
+    -- pkscript is the witness program that represents the static address.
+    pkscript BYTEA NOT NULL,
+
+    -- address_hrp is the hrp part of the taproot address.
+    address_hrp TEXT NOT NULL,
```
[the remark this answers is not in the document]
hieblmi: Yes we could create the address from keys+expiry, we are storing it out of convenience to not have to regenerate it when restoring from db after a restart.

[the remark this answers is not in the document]
hieblmi: Good point. I removed the address from the schema and added a method to generate the address from the keys+expiry when needed.

swapserverrpc/server.proto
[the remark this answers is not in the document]
hieblmi: Removed this ID

staticaddr/manager.go
```diff
@@ -0,0 +1,165 @@
...
+
+	<-ctx.Done()
+
+	return nil
+}
+
+// NewAddress starts a new address creation flow.
+func (m *Manager) NewAddress(ctx context.Context) (*AddressParameters,
+	error) {
+
```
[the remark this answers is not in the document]
hieblmi: done

loopdb/sqlc/migrations/000007_static_address.up.sql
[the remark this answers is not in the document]
hieblmi: done

PR #644, opened 2023-10-03
liquidity: dest address support for easy autloop

bhandras:
This PR adds support for destination address when using "easy autoloop".

#### Pull Request Checklist


liquidity/liquidity.go
```diff
@@ -653,6 +653,11 @@ func (m *Manager) dispatchBestEasyAutoloopSwap(ctx context.Context) error {
 		return fmt.Errorf("unexpected swap suggestion type: %T", t)
 	}
 
+	// Check if the parameter for custom address is defined for loop outs.
+	if m.params.DestAddr != nil {
```
hieblmi, line 657: should we point out to the user in the parameter descriptions that this leads to address use?

GeorgeTsagk: I'm not sure if this piece of code should be part of the loopout builder `buildSwap` method, since there we also handle the lnd account for xpub addresses

GeorgeTsagk (approved):
Lgtm :pretzel:

liquidity/liquidity.go
hieblmi: +1 I think it makes sense to handle address setup in one place. And we don't have to check there if both `accounts` and `destAddr` are set since this is already done in the `cli` and in a `validate` method in parameters.

bhandras: done

hieblmi (approved):

PR #649, opened 2023-10-12
db: refactor db opening

hieblmi:
In preparation of https://github.com/lightninglabs/loop/pull/642, this PR extracts the db opening into its own util function and exposes the `BaseDB` struct in the daemon.

### DIFF
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -392,8 +392,22 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 		log.Infof("Successfully migrated boltdb")
 	}
 
+	// Now that we know where the database will live, we'll go ahead and
+	// open up the default implementation of it.
+	chainParams, err := lndclient.Network(d.cfg.Network).ChainParams()
+	if err != nil {
+		return err
+	}
+
+	swapDb, _, err := openDatabase(d.cfg, chainParams)
+	if err != nil {
+		return err
+	}
+
 	// Create an instance of the loop client library.
-	swapclient, clientCleanup, err := getClient(d.cfg, &d.lnd.LndServices)
+	swapClient, clientCleanup, err := getClient(
+		d.cfg, swapDb, &d.lnd.LndServices,
+	)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


loopd/daemon.go
```diff
@@ -388,11 +388,22 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
...
+		return err
+	}
+
+	swapDb, _, err := openDatabase(d.cfg, network)
 	if err != nil {
 		return err
 	}
+
+	// Create an instance of the loop client library.
+	swapClient, clientCleanup, err := getClient(
```
bhandras, line 404: nit: need to check err

loopd/utils.go
```diff
@@ -31,26 +31,40 @@ func getClient(cfg *Config, lnd *lndclient.LndServices) (*loop.Client,
...
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return swapClient, cleanUp, nil
+}
+
+func openDatabase(cfg *Config, network *chaincfg.Params) (
+	loopdb.SwapStore, *loopdb.BaseDB, error) {
```
bhandras, line 45: nit: linter also complains here about BaseDB not being used anywhere.

loopd/utils.go
```diff
@@ -31,26 +31,40 @@ func getClient(cfg *Config, lnd *lndclient.LndServices) (*loop.Client,
...
+		cfg.DataDir, swapDb, clientConfig,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return swapClient, cleanUp, nil
+}
+
+func openDatabase(cfg *Config, network *chaincfg.Params) (
```
bhandras, line 44: nit: instead of `network` I'd suggest calling this `chainParams` or `params` (we use this naming elsewhere in the codebase).

bhandras (approved):
LGTM, pending nits.

loopd/utils.go
hieblmi: This will be used by the PR that's building on this, so for now I've added `//nolint:unparam` which I will remove in my next commits.

loopd/daemon.go
```diff
@@ -392,8 +392,22 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 		log.Infof("Successfully migrated boltdb")
 	}
 
+	// Now that we know where the database will live, we'll go ahead and
+	// open up the default implementation of it.
+	network, err := lndclient.Network(d.cfg.Network).ChainParams()
```
hieblmi, line 397: change `network` to `chainParams`.

GeorgeTsagk (approved):
LGTM :heavy_check_mark:

PR #659, opened 2023-11-10
Fix pub deadline timestamp

bitbandi:
This pr fix the publish_deadline timestamp from swap invoice datetime. (normally +30 minutes, if i check good)


bhandras (approved):
LGTM 🎉

GeorgeTsagk (commented):
Should we close this in favor of #682 ?

PR #661, opened 2023-11-13
Abandon API for pending Loop-in swaps

hieblmi:
This PR introduces a `AbandonSwap` Api that currently supports only loop-in swaps. Any pending swap's status can be overridden with a new database state called `StateFailAbandoned`. This command relies on the users judgement to only abandon swaps that don't lead to loss of funds, e.g. if a swap is abandoned that had it's time out path not swept yet.


loopd/swapclient_server.go
```diff
@@ -508,6 +512,57 @@ func (s *swapClientServer) SwapInfo(_ context.Context,
...
+		return nil, fmt.Errorf("please read the AbandonSwap Api " +
+			"documentation")
+	}
+
+	swapHash, err := lntypes.MakeHash(req.Id)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing swap hash: %v", err)
+	}
+
+	// Just return the server's in-memory cache here too as we also want to
```
hieblmi, line 530: nit: remove comment

looprpc/client.proto
```diff
@@ -43,6 +43,11 @@ service SwapClient {
     */
     rpc SwapInfo (SwapInfoRequest) returns (SwapStatus);
 
+    /* loop: `abandonswap`
+    AbandonSwap allows the client to abandon a swap.
+   */
```
bhandras, line 48: nit: formatting.

looprpc/client.proto
```diff
@@ -1180,3 +1191,21 @@ message SuggestSwapsResponse {
...
+    /*
+    The swap identifier which currently is the hash that locks the HTLCs. When
+    using REST, this field must be encoded as URL safe base64.
+    */
+    bytes id = 1;
+
+    /*
+    A flag that tries to ensure that the client understands that they are
+    risking of loosing funds by abandoning a swap. This could happen if an
+    abandoned would wait on a timeout sweep by the client.
```
bhandras, line 1205: nit: abandoned swap

looprpc/client.proto
```diff
@@ -43,6 +43,11 @@ service SwapClient {
     */
     rpc SwapInfo (SwapInfoRequest) returns (SwapStatus);
 
+    /* loop: `abandonswap`
```
bhandras, line 46: Commenting here because github hides it. The server.proto changes are missing from the next commit.

client.go
```diff
@@ -316,11 +320,15 @@ func (s *Client) Run(ctx context.Context, statusChan chan<- SwapInfo) error {
...
+	// determine if a particular swap should be cancelled.
+	s.abandonChan = abandonChan
+
 	// Main event loop.
-	err = s.executor.run(mainCtx, statusChan)
+	err = s.executor.run(mainCtx, statusChan, abandonChan)
 
 	// Consider canceled as happy flow.
-	if err == context.Canceled {
+	if errors.Is(err, context.Canceled) {
```
bhandras, line 331: 👍

loopd/swapclient_server.go
```diff
@@ -508,6 +512,57 @@ func (s *swapClientServer) SwapInfo(_ context.Context,
 	return s.marshallSwap(&swp)
 }
 
+// AbandonSwap requests the server to abandon a swap with the given hash.
+func (s *swapClientServer) AbandonSwap(_ context.Context,
+	req *clientrpc.AbandonSwapRequest) (*clientrpc.AbandonSwapResponse,
+	error) {
+
+	if !req.IKnowWhatIAmDoing {
+		return nil, fmt.Errorf("please read the AbandonSwap Api " +
```
bhandras, line 521: nit: s/Api/API

loopd/swapclient_server.go
```diff
@@ -508,6 +512,57 @@ func (s *swapClientServer) SwapInfo(_ context.Context,
...
+	}
+
+	swapHash, err := lntypes.MakeHash(req.Id)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing swap hash: %v", err)
+	}
+
+	// Just return the server's in-memory cache here too as we also want to
+	// return temporary failures to the client.
+	swap, ok := s.swaps[swapHash]
```
bhandras, line 532: This must be guarded by the `swapsLock`.

bhandras (commented):
Did a quick pass, looking good so far! I think we'll need some test coverage on the client side too. Essentially we want to make sure that calling the API in an on-going swap will indeed abandon it.

looprpc/client.proto
hieblmi: 😬 how did this happen

client.go
```diff
@@ -753,3 +761,15 @@ func (s *Client) Probe(ctx context.Context, req *ProbeRequest) error {
...
 }
+
+// AbandonSwap sends the hash of the swap to abandon to all active swaps. Only
+// the addressed swap will abandon itself.
+func (s *Client) AbandonSwap(req *AbandonSwapRequest) error {
+	if req == nil {
+		return errors.New("no request provided")
+	}
+
+	s.abandonChan <- req.SwapHash
```
GeorgeTsagk, line 772: I think you need to change the core structure here. A single channel won't notify all swaps that are waiting to read from it.

Here's an example:
We have 3 loop-ins active, `in1` `in2` and `in3`
All of them are waiting to read from `s.abandonChan` in case the user requests for one of them to be abandoned

User wants to abandon `in2` so they call `AbandonSwap` method and that method writes to `s.abandonChan` as shown in the line above.

A **random** swap will read that value **once**, and then the channel is going to be clear. Let's say that in the above example `in1` is reading the value, it checks that `in1` != `in2` and continues with a noop. The intended swap never received the value over the channel because it was consumed only once by `in1`.

Here's 2 ways you can proceed with this:
1. Enhance the "abandonChan" into a structure that is able to hold multiple channels for each subscriber (i.e you could have a `map[lntypes.Hash]chan Bool` which means one dedicated channel for each swap). When user tries to `AbandonSwap` we will first find the correct channel by accessing the map, and then try to write to that channel.

2. The more hacky approach is for each loop-in to re-write the value of `lntypes.Hash` into the channel when they check it and realise it doesn't match theirs. In that case the value would be written only once on the channel from this speciifc line above, but then it would ping-pong between all active loop-ins until it would **eventually** reach the intended target. (**hacky and dirty**, but only 1 line of code)

hieblmi: Thank you for the thorough review and the nice catch. I prefer option 1.) to fix this.

loopd/swapclient_server.go
```diff
@@ -508,6 +512,59 @@ func (s *swapClientServer) SwapInfo(_ context.Context,
 	return s.marshallSwap(&swp)
 }
 
+// AbandonSwap requests the server to abandon a swap with the given hash.
+func (s *swapClientServer) AbandonSwap(_ context.Context,
```
GeorgeTsagk, line 516: Here we are not actually making use of the request context, but the current implementation of `AbandonSwap` is writing to a channel that is not buffered. Why not pass down the request context to the methods?

client.go
```diff
@@ -753,3 +765,15 @@ func (s *Client) Probe(ctx context.Context, req *ProbeRequest) error {
...
 }
+
+// AbandonSwap sends the hash of the swap to abandon to all active swaps. Only
+// the addressed swap will abandon itself.
+func (s *Client) AbandonSwap(req *AbandonSwapRequest) error {
+	if req == nil {
+		return errors.New("no request provided")
+	}
+
+	s.abandonChans[req.SwapHash] <- struct{}{}
```
GeorgeTsagk, line 776: According to my previous comment we could make use of `ctx` here and do smth like this 
```suggestion
	select {
	case s.abandonChans[req.SwapHash] <- struct{}{}:
	case ctx.Done():
	    return ctx.Err()
	}
```
in order to avoid blocking indefinitely.

This also raises a new topic: How are channels erased from memory?
And I think an important parameter here is whether the channel is buffered or not (i.e whether you block while writing to it or not)

1. Buffered channel: That means you immediately return from the above line thus there is also no need for using `ctx`
    Note: How/When could we delete entries from the map in this case?
2. Unbuffered channel: That means we block in the above line and when the execution resumes either
    i) ctx is done
    ii) swap actually read the signal
    and since the channel is only used once we could safely then delete the channel from the map

Seems to me like sticking with 2 (unbuffered channels) is more simple, so I would suggest adding this line on top of the function
```golang
// When this function returns there is no use of the swap's abandon chan.
defer delete(s.abandonChans, req.SwapHash)
```
given that the upper block of code with the `select` statement and `ctx` is used too

GeorgeTsagk: Or if there is another place in the code where `client` receives signals from swaps when they reach a final state we could delete the map entry there, as it is no longer of use

GeorgeTsagk (commented):
Looks good! just 1 last comment about deleting abandon chans

client.go
hieblmi: Great points @GeorgeTsagk, thank you!

I think the `defer delete` however should only be applied if the channel was successfully abandoned, which could still fail even if it was sent on abandon channel.

Let me check out the places where we delete from the map.

Thanks again.

hieblmi: The deletion introduces another problem, namely concurrent access to the channel map. I solved this by introducing a Mutx to the `executor` that is locked when a loop-in is created and locked again when a swap finishes.

Would be interested in alternative solutions.

client.go
```diff
@@ -753,3 +767,22 @@ func (s *Client) Probe(ctx context.Context, req *ProbeRequest) error {
 		req.RouteHints,
 	)
 }
+
+// AbandonSwap sends the hash of the swap to abandon to all active swaps. Only
```
hieblmi, line 771: nit: fix description

looprpc/client.proto
```diff
@@ -1180,3 +1191,64 @@ message SuggestSwapsResponse {
...
+    The result of the abandon operation.
+    */
+    AbandonResult result = 1;
+}
+
+enum AbandonResult {
+    ABANDON_IN_PROGRESS = 0;
+
+    /*
+    ABANDON_FAILURE indicates that the swap was not abandoned due to an unknown
```
bhandras, line 1221: nit: doc seems to be out of sync.

looprpc/client.proto
```diff
@@ -1180,3 +1191,64 @@ message SuggestSwapsResponse {
...
+    reason.
+     */
+    FAILURE_REASON_MISSING_FLAG = 1;
+
+    /*
+    ABANDON_FAILURE indicates that the swap was not abandoned because it is
```
bhandras, line 1227: nit: seems out of sync.

loopdb/swapstate.go
```diff
@@ -98,6 +102,18 @@ func (s SwapState) Type() SwapStateType {
 	return StateTypeFail
 }
 
+// IsPending returns true if the swap is in a pending state.
+func (s SwapState) IsPending() bool {
```
bhandras, line 106: Nice 🎉

client.go
```diff
@@ -65,8 +66,9 @@ var (
 // Client performs the client side part of swaps. This interface exists to be
 // able to implement a stub.
 type Client struct {
-	started uint32 // To be used atomically.
-	errChan chan error
+	started      uint32 // To be used atomically.
+	errChan      chan error
+	abandonChans map[lntypes.Hash]chan struct{}
```
bhandras, line 71: I think this justifies a comment since it's not trivial to understand what we store in the map and why.

executor.go
```diff
@@ -167,6 +171,18 @@ func (s *executor) run(mainCtx context.Context,
...
+				// If a loop-in ended we have to remove its
+				// abandon channel from our abandonChans map
+				// since the swap finalized.
+				if swap, ok := newSwap.(*loopInSwap); ok {
+					s.Lock()
+					delete(
+						abandonChans,
+						swap.hash,
+					)
+					s.Unlock()
```
GeorgeTsagk, line 183: :+1:

client.go
```diff
@@ -753,3 +763,22 @@ func (s *Client) Probe(ctx context.Context, req *ProbeRequest) error {
...
+// the addressed swap will abandon itself.
+func (s *Client) AbandonSwap(ctx context.Context,
+	req *AbandonSwapRequest) error {
+
+	if req == nil {
+		return errors.New("no request provided")
+	}
+
+	select {
+	case s.abandonChans[req.SwapHash] <- struct{}{}:
```
bhandras, line 781: Access here should also be guarded by the mutex.

GeorgeTsagk (approved):
Great work, LGTM :pinched_fingers:

client.go
bhandras: There's also an issue where if the abandon chan is not read immediately then this call might block and hold the mutex. A buffered channel can help here with the addition of using a non-blocking write:
```

s.executor.Lock()
defer s.executor.Unlock()

select {
    case s.abandonChans[req.SwapHash] <- struct{}{}:
    case <-ctx.Done():
                   return ctx.Err()
    default:
         // This is to avoid writing to a full channel.
    }
```

loopin.go
```diff
@@ -1062,6 +1118,22 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
...
 }
 
+// setStateAbandoned stores the abandoned state in the database and announces
+// it.
+func (s *loopInSwap) setStateAbandoned(ctx context.Context) error {
+	if s.state.IsPending() {
+		s.log.Infof("Abandoning swap %v", s.hash)
+		s.setState(loopdb.StateFailAbandoned)
+
+		err := s.persistAndAnnounceState(ctx)
```
bhandras, line 1132: nit: can just return here.

loopin.go
```diff
@@ -840,6 +876,26 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
 	htlcKeyRevealed := false
 	for !htlcSpend || !invoiceFinalized {
 		select {
+		// If the client requested the swap to be abandoned, we override
+		// the status in the database.
+		case <-s.abandonChan:
+			err = s.setStateAbandoned(ctx)
```
bhandras, line 886: Cannot really happen here but just for future proofing, if the state isn't pending (or considered abandonable) then we'd still cancel the invoice etc here.

loopin.go
```diff
@@ -1062,6 +1118,22 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
 	return fee, nil
 }
 
+// setStateAbandoned stores the abandoned state in the database and announces
+// it.
+func (s *loopInSwap) setStateAbandoned(ctx context.Context) error {
+	if s.state.IsPending() {
```
bhandras, line 1128: Does it make sense to abandon the swap is the htlc is already published? Maybe we should instead create a function whether the swap is "abandonable".

client.go
```diff
@@ -380,6 +380,10 @@ func (s *Client) resumeSwaps(ctx context.Context,
 			continue
 		}
 
+		// Store the swap's abandon channel so that the client can
+		// abandon the swap by providing the swap hash.
+		s.abandonChans[swap.hash] = swap.abandonChan
```
bhandras, line 385: Locking is also needed here as this function runs in its own goroutine and the executor might already be active.

cmd/loop/swaps.go
```diff
@@ -90,3 +90,71 @@ func swapInfo(ctx *cli.Context) error {
...
+}
+
+func abandonSwap(ctx *cli.Context) error {
+	args := ctx.Args()
+
+	var id string
+	switch {
+	case ctx.IsSet("id"):
+		id = ctx.String("id")
+	case ctx.NArg() > 0:
```
bhandras, line 122: nit: add nl between cases.

cmd/loop/swaps.go
```diff
@@ -90,3 +90,71 @@ func swapInfo(ctx *cli.Context) error {
 	printRespJSON(resp)
 	return nil
 }
+
+var abandonSwapCommand = cli.Command{
+	Name:  "abandonswap",
+	Usage: "abandon a swap with a given swap hash",
+	Description: "This command overrides the database and abandons a " +
+		"swap with a given swap hash.\n\n" +
+		"!!! This command might potentially lead to loss of funds if" +
```
bhandras, line 99: nit: add space at the end.

cmd/loop/swaps.go
```diff
@@ -90,3 +90,71 @@ func swapInfo(ctx *cli.Context) error {
...
+		"it is applied to swaps that are still waiting for pending " +
+		"user funds. Before executing this command make sure that " +
+		"no funds are locked by the swap",
```
bhandras, line 102: nit: missing period.

cmd/loop/swaps.go
```diff
@@ -90,3 +90,71 @@ func swapInfo(ctx *cli.Context) error {
...
+	ArgsUsage: "ID",
+	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name: "i_know_what_i_am_doing",
+			Usage: "Specify this flag if you made sure that you " +
+				"read and understood the following " +
+				"consequence of applying this command",
```
bhandras, line 109: nit: missing period.

bhandras (commented):
Looks really good now! Just a question on my side whether we should allow abandoning a loopin swap if the client already published the htlc (because they won't sweep it back then and it'll just hang in limbo) and some smallish comments.

client.go
hieblmi: Thanks for the explanation. I've also changed the channel in the resume-swaps case to a buffered one.

loopin.go
hieblmi: Nice catch!

loopin.go
hieblmi: For our resolution of an issue the swap is caught in a pending state although the swap itself hasn't been published, so there it makes sense to abandon a swap in "published" state.

loopin.go
hieblmi: see https://github.com/lightninglabs/loop/pull/661#discussion_r1403395076

loopin.go
bhandras: Still applies.

bhandras (commented):
LGTM, pending decision about the status enum vs just returning an error.

loopin.go
hieblmi: weird, I though I fixed this

loopin.go
```diff
@@ -999,6 +981,23 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
 	return nil
 }
 
+// abandonSwapCancelInvoice abandons the swap and cancels the swap invoice.
```
bhandras, line 984: nit: either move in previous commit or don't move here to make the diff smaller and consistent with the commit intent.

bhandras (approved):
LGTM, pending the last few comments 🤓 Great work! 🎉

loopin.go
```diff
@@ -840,6 +890,18 @@ func (s *loopInSwap) waitForSwapComplete(ctx context.Context,
...
 		select {
+		// If the client requested the swap to be abandoned, we override
+		// the status in the database.
+		case <-s.abandonChan:
+			err = s.abandonSwapCancelInvoice(ctx)
+			if err != nil {
+				return err
+			}
+
+			return fmt.Errorf("swap hash "+
```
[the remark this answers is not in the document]
hieblmi: We rely on the error here, it might be that `abandonSwapCancelInvoice()` is not returning an error. Also the function might fail due db issues which we want to catch here.

loopin.go
```diff
@@ -1062,6 +1124,18 @@ func (s *loopInSwap) publishTimeoutTx(ctx context.Context,
 	return fee, nil
 }
 
+// setStateAbandoned stores the abandoned state and announces it.
+func (s *loopInSwap) setStateAbandoned(ctx context.Context) error {
```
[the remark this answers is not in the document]
hieblmi: Ah right, `setStateAbandoned` could just handle state and invoice cancellation.

PR #665, opened 2023-11-23
Finalize swap if wallet holds insufficient confirmed funds

hieblmi:
The loop daemon knows a swap state `StateFailTemporary` which indicates a failure that might need manual intervention such as a restart to resolve.

This PR excludes swap failures due to an insufficient confirmed wallet balance, since this might be resolved once the next block is mined.

This resolves an issue with auto-loop in if two swaps are performed without the first one being confirmed on-chain.


bhandras (approved):
LGTM, pending base PR 🎉

GeorgeTsagk (approved):
Looks good! :raised_hands:

PR #667, opened 2023-11-30
fsm: expand fsm

sputn1ck:
This PR adds:
- a default observer to the FSM
- more info to the action entry and exit funcs
- an optional initial wait time for the WaitForState function

### DIFF
--- a/fsm/observer.go
+++ b/fsm/observer.go
@@ -46,9 +46,60 @@ func (c *CachedObserver) GetCachedNotifications() []Notification {
 	return c.cachedNotifications.Get()
 }
 
+// WaitForStateOption is an option that can be passed to the WaitForState
+// function.
+type WaitForStateOption interface {
+	apply(*fsmOptions)
+}
+
+// fsmOptions is a struct that holds all options that can be passed to the
+// WaitForState function.
+type fsmOptions struct {
+	initialWait time.Duration
+}
+
+// InitialWaitOption is an option that can be passed to the WaitForState
+// function to wait for a given duration before checking the state.
+type InitialWaitOption struct {
+	initialWait time.Duration
+}
+
+// WithWaitForStateOption creates a new InitialWaitOption.
+func WithWaitForStateOption(initialWait time.Duration) WaitForStateOption {
+	return &InitialWaitOption{
+		initialWait,
+	}
+}
+
+// apply implements the WaitForStateOption interface.
+func (w *InitialWaitOption) apply(o *fsmOptions) {
+	o.initialWait = w.initialWait
+}
+
 // WaitForState waits for the state machine to reach the given state.
+// If the optional initialWait parameter is set, the function will wait for
+// the given duration before checking the state. This is useful if the
+// function is called immediately after sending an event to the state machine
+// and the state machine needs some time to process the event.
 func (s *CachedObserver) WaitForState(ctx context.Context,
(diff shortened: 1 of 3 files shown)


fsm/observer.go
```diff
@@ -46,9 +46,43 @@ func (c *CachedObserver) GetCachedNotifications() []Notification {
 	return c.cachedNotifications.Get()
 }
 
+// WaitForStateOption is an option for the WaitForState function.
+type WaitForStateOption struct {
```
bhandras, line 50: Almost, but what you want is:

```
type FSMOption interface {
    apply(*fsmOptions) 
}

type fsmOptions struct {
   initialWait time.Duration
}

type WaitForStateOption struct {
   initialWait time.Duration
}

func WithWaitForStateOption(initialWait time.Duration) FSMOption {
   return &WaitForStateOption{
         initialWait,
   }
}

func (w *WaitForStateOption) apply(o *fsmOptions) {
   o.initialWait = w.initialWait
}

```

And then below in the constructor you can pass in `opts ...FSMOption` and just call `apply()` on them.
This pattern is the same e.g. grpc uses with `DialOption` and we use the same pattern in LND a lot.

bhandras (approved):
LGTM, pending the option fix. 🌮

fsm/observer.go
```diff
@@ -46,9 +46,43 @@ func (c *CachedObserver) GetCachedNotifications() []Notification {
...
+
+	var initialWait time.Duration
+	if len(opts) > 0 {
+		for _, opt := range opts {
+			initialWait = opt.initialWait
+		}
+	}
+
+	// Wait for the initial wait duration if set.
+	if initialWait > 0 {
```
hieblmi, line 78: Not sure this check is needed.

fsm/observer.go
```diff
@@ -46,9 +46,43 @@ func (c *CachedObserver) GetCachedNotifications() []Notification {
...
+// the given duration before checking the state. This is useful if the
+// function is called immediately after sending an event to the state machine
+// and the state machine needs some time to process the event.
 func (s *CachedObserver) WaitForState(ctx context.Context,
-	timeout time.Duration, state StateType) error {
+	timeout time.Duration, state StateType,
+	opts ...WaitForStateOption) error {
+
+	var initialWait time.Duration
+	if len(opts) > 0 {
```
hieblmi, line 71: Can omit this check.

fsm/observer.go
hieblmi: Also nit on fix comment: https://github.com/lightninglabs/loop/pull/667/files#diff-28f5dffa528d8282177bcb103631cef19c6b2ae6d20abca5f559be88209469d6R129

fsm/observer.go
```diff
@@ -46,9 +46,60 @@ func (c *CachedObserver) GetCachedNotifications() []Notification {
...
+// and the state machine needs some time to process the event.
 func (s *CachedObserver) WaitForState(ctx context.Context,
-	timeout time.Duration, state StateType) error {
+	timeout time.Duration, state StateType,
+	opts ...InitialWaitOption) error {
+
+	var options fsmOptions
+
+	for _, opt := range opts {
+		opt.apply(&options)
```
bhandras, line 91: Cool! Looks good now!

bhandras (approved):
🚀 🚀

PR #672, opened 2023-12-11
swapserverrpc: Add Fetchl402 call

sputn1ck:
This commit adds a new FetchL402 call which allows a client to retrieve a l402 token from the loop server.

### DIFF
--- a/server_mock_test.go
+++ b/server_mock_test.go
@@ -273,3 +273,7 @@ func (s *serverMock) PushKey(_ context.Context, _ loopdb.ProtocolVersion,
 
 	return nil
 }
+
+func (s *serverMock) FetchL402(_ context.Context) error {
+	return nil
+}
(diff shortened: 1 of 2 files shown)


swapserverrpc/server.proto
```diff
@@ -47,6 +47,10 @@ service SwapServer {
     rpc MuSig2SignSweep (MuSig2SignSweepReq) returns (MuSig2SignSweepRes);
 
     rpc PushKey (ServerPushKeyReq) returns (ServerPushKeyRes);
+
+    // FetchL402 is a simple non-l402-allowlisted request that is required
+    // in order to force the creation of an l402.
+    rpc FetchL402 (FetchL402Request) returns (FetchL402Response);
```
bhandras, line 53: Don't we need an empty implementation for this in `swapserver.go`?

bhandras: Ah never mind, I see the other PR!

bhandras (approved):
Thanks 🙏

hieblmi (approved):
Tvm @sputn1ck.

PR #676, opened 2023-12-20
loopin: sweep incorrect amount on timeout

hieblmi:
`loopd` currently does not sweep externally provided loop-in funds if they weren't matching the swap contract amount.
Instead it will mark the swap with `StateFailIncorrectHtlcAmt` and finalize it, leaving the user with manual intervention via `chantools recoverloopin` to reclaim funds.

This PR adds functionality to publish a timeout transaction for the mismatched swap amount once the htlc is timed out.


loopin.go
```diff
@@ -625,7 +625,39 @@ func (s *loopInSwap) executeSwap(globalCtx context.Context) error {
 	// amount. Otherwise, fail the swap immediately.
 	if htlcValue != s.LoopInContract.AmountRequested {
 		s.setState(loopdb.StateFailIncorrectHtlcAmt)
-		return s.persistAndAnnounceState(globalCtx)
+		var publishedSweep bool
+		for {
```
hieblmi, line 629: should we move this into a separate method?

loopdb/swapstate.go
```diff
@@ -110,7 +115,7 @@ func (s SwapState) Type() SwapStateType {
 func (s SwapState) IsPending() bool {
 	return s == StateInitiated || s == StateHtlcPublished ||
 		s == StatePreimageRevealed || s == StateFailTemporary ||
-		s == StateInvoiceSettled
+		s == StateInvoiceSettled || s == StateFailIncorrectHtlcAmt
```
hieblmi, line 118: This state is declared 'pending' so that after a restart we pick up the swap again and check if it timed out.

loopin.go
```diff
@@ -624,8 +624,35 @@ func (s *loopInSwap) executeSwap(globalCtx context.Context) error {
...
+					err := s.publishIncorrectAmountSweep(
+						globalCtx, htlcOutpoint,
+						htlcValue,
+					)
+					if err != nil {
+						return err
+					}
+
+					publishedSweep = true
+					break
```
bhandras, line 645: Would it make sense to republish each block?

bhandras (commented):
Looks good, just one q.

looprpc/client.proto
```diff
@@ -541,6 +541,12 @@ enum FailureReason {
     wasn't published due to insufficient confirmed balance.
     */
     FAILURE_REASON_INSUFFICIENT_CONFIRMED_BALANCE = 8;
+
+    /*
+    FAILURE_REASON_INCORRECT_HTLC_AMT_SWEPT indicates that a swap
+    wasn't published due to insufficient confirmed balance.
+    */
+    FAILURE_REASON_INCORRECT_HTLC_AMT_SWEPT = 9;
```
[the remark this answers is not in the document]
GeorgeTsagk: ok after reading a bit more I think I got my answer:

Before sweep is confirmed we have the plain "incorrect htlc amt" state, then we mark it as swept by using the above state

GeorgeTsagk (approved):
Looks good!

loopin.go
bhandras: I think it may be cleaner if we unify this sweeping with the normal timeout sweep. We'd then republish on each block too.

hieblmi: Sounds good, will take a look.

loopdb/swapstate.go
hieblmi: from a stale commit

bhandras (approved):
LGTM, pending @sputn1ck's comment 🎉

PR #677, opened 2023-12-20
[3/?] StaticAddr: Deposit tracking and timeout sweep

hieblmi:
It introduces tracking of deposits to a static address and monitoring of the CSV expiry. Instead of using `RegisterConfirmationsNtfn` to detect spends to a pkScript, we poll the lnd account of the imported static address taproot script and notify the main event loop of the manager about new deposits to a static address.

~Todo:~

~- state machine recovery~
~- listening for timeout tx confirmations to be sure that the client claimed funds.~


staticaddr/actions.go
```diff
@@ -0,0 +1,109 @@
...
+
+	f.Debugf("publishing timeout sweep with txid: %v", msgTx.TxHash())
+
+	return OnSwept
+}
+
+// FinalizeDeposit is the first action that is executed when the instant
+// out FSM is started. It will send the instant out request to the server.
+func (f *FSM) FinalizeDeposit(eventCtx fsm.EventContext) fsm.EventType {
+
```
hieblmi, line 100: nit: extra line

staticaddr/actions.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightninglabs/loop/fsm"
+	"github.com/lightninglabs/loop/staticaddr/script"
+)
+
+const (
+	defaultConfTarget = 3
+)
+
+// SweepExpiredDepositAction ...
```
hieblmi, line 17: nit: godoc

staticaddr/manager.go
```diff
@@ -35,40 +54,123 @@ type ManagerConfig struct {
...
+		case <-m.runCtx.Done():
+			return nil
+		}
+	}
 }
 
-// NewAddress starts a new address creation flow.
+// NewAddress starts a new address creation flow with the server and starts a
+// deposit listener if the address creation succeeded.
 func (m *AddressManager) NewAddress(ctx context.Context) (*AddressParameters,
```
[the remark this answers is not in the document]
hieblmi: The server ensures that there is only one static address per L402. If a second address is requested the server won't hand out another server pubkey.

loopdb/sqlc/migrations/000003_static_address.up.sql
```diff
@@ -46,4 +46,22 @@ CREATE TABLE IF NOT EXISTS static_addresses (
...
     -- use.
     protocol_version INTEGER NOT NULL
-);
\ No newline at end of file
+);
+
+-- deposits stores historic and unspent static address outputs.
+CREATE TABLE IF NOT EXISTS deposits (
+    -- id is the auto-incrementing primary key for a static address.
+    id INTEGER PRIMARY KEY,
```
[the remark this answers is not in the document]
hieblmi: Since there is only one entry in the `static_address` table, do we need that link from the `deposits` table?

staticaddr/manager.go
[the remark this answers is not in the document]
hieblmi: As @bhandras mentioned it is a good idea to serialize calls to creating a new address with a channel to the manager's run loop.

staticaddr/actions.go
```diff
@@ -0,0 +1,109 @@
...
+func (f *FSM) SweepExpiredDepositAction(
+	eventCtx fsm.EventContext) fsm.EventType {
+
+	depoCtx, ok := eventCtx.(*DepositContext)
+	if !ok {
+		return f.HandleError(fsm.ErrInvalidContextType)
+	}
+	f.Debugf("SweepExpiredDepositAction: %v", depoCtx)
+
+	msgTx := wire.NewMsgTx(2)
```
hieblmi, line 27: extract tx creation into separate method.

loopdb/sqlc/migrations/000003_static_address.down.sql
```diff
@@ -1 +1,2 @@
-DROP TABLE IF EXISTS static_addresses;
\ No newline at end of file
+DROP TABLE IF EXISTS static_addresses;
+DROP TABLE IF EXISTS deposits;
```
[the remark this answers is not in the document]
hieblmi: done

loopdb/sqlc/queries/static_addresses.sql
```diff
@@ -34,4 +34,29 @@ INSERT INTO static_addresses (
...
+) VALUES (
+             $1,
+             $2,
+             $3,
+             $4,
+             $5,
+             $6
+         );
+
+-- name: DeleteDeposit :exec
```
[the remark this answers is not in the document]
hieblmi: removed it.

staticaddr/manager.go
```diff
@@ -221,3 +414,189 @@ func (m *AddressManager) ListUnspentRaw(ctx context.Context, minConfs,
...
+		}
+
+		select {
+		case <-ctx.Done():
+			return err
+
+		default:
+			// Send the new deposit to the
+			// manager event loop.
+			m.depositChan <- deposit
```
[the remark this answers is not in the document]
hieblmi: Creating the fsm here makes it simpler. Thanks

staticaddr/manager.go
```diff
@@ -172,9 +259,100 @@ func (m *Manager) NewAddress(ctx context.Context) (*AddressParameters,
...
+	}
+
+	// Recover deposits.
+	deposits, err := m.cfg.Store.AllDeposits(ctx)
+	if err != nil {
+		return err
+	}
+
+	for i := range deposits {
+		m.deposits[deposits[i].OutPoint] = &deposits[i]
```
hieblmi, line 347: instantiate fsms for unspent deposits

hieblmi: Also recovered the timeout script on the deposit.

staticaddr/actions.go
```diff
@@ -0,0 +1,109 @@
...
+	f.Debugf("publishing timeout sweep with txid: %v", msgTx.TxHash())
+
+	return OnSwept
+}
+
+// FinalizeDeposit is the first action that is executed when the instant
+// out FSM is started. It will send the instant out request to the server.
+func (f *FSM) FinalizeDeposit(eventCtx fsm.EventContext) fsm.EventType {
+
+	depoCtx, ok := eventCtx.(*DepositContext)
```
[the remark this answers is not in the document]
hieblmi: removed

staticaddr/manager.go
```diff
@@ -37,37 +59,121 @@ type ManagerConfig struct {
...
-	return nil
+	for {
+		select {
+		case height := <-newBlockChan:
+			m.Lock()
+			m.currentHeight = uint32(height)
+			m.Unlock()
+
+			// Inform all active deposits about a new block arrival.
+			for _, fsm := range m.activeDeposits {
```
hieblmi, line 156: lock `activeDeposits`

staticaddr/deposit/fsm.go
```diff
@@ -0,0 +1,281 @@
...
+	OnStart           = fsm.EventType("OnStart")
+	OnExpiry          = fsm.EventType("OnExpiry")
+	OnExpiryPublished = fsm.EventType("OnExpiryPublished")
+	OnExpirySwept     = fsm.EventType("OnExpirySwept")
+	OnRecover         = fsm.EventType("OnRecover")
+)
+
+// FSM is the state machine that handles the instant out.
+type FSM struct {
+	*fsm.StateMachine
```
hieblmi, line 44: add protocol version

staticaddr/deposit/fsm.go
```diff
@@ -0,0 +1,281 @@
...
+	depoFsm := &FSM{
+		cfg:                  cfg,
+		deposit:              deposit,
+		params:               params,
+		address:              address,
+		ctx:                  ctx,
+		blockNtfnChan:        make(chan uint32),
+		finalizedDepositChan: finalizedDepositChan,
+	}
+
```
hieblmi, line 87: add protocol version switch for `DepositStatesV1()` etc.

staticaddr/deposit/fsm.go
```diff
@@ -0,0 +1,297 @@
...
+			depositStates, DefaultObserverSize,
+		)
+	}
+
+	depoFsm.ActionEntryFunc = depoFsm.updateDeposit
+
+	go func() {
+		for {
+			select {
+			case currentHeight := <-depoFsm.blockNtfnChan:
```
bhandras, line 120: Seems like the `blockNtfnChan` is unset?

staticaddr/deposit/fsm.go
```diff
@@ -0,0 +1,297 @@
...
+	ctx context.Context
+
+	blockNtfnChan chan uint32
+
+	finalizedDepositChan chan wire.OutPoint
+}
+
+// NewFSM creates a new state machine that can action on all static address
+// feature requests.
+func NewFSM(ctx context.Context, deposit *Deposit, cfg *ManagerConfig,
```
bhandras, line 71: So IIUC the plan is to have an FSM per deposit, meaning we'd sweep them back individually if they expire? Perhaps worth thinking about efficiency here, so we could possibly have a single FSM and sweep back all expired deposits? wdyt? cc: @sputn1ck

staticaddr/deposit/manager.go
```diff
@@ -0,0 +1,421 @@
...
+		select {
+		case height := <-newBlockChan:
+			m.Lock()
+			m.currentHeight = uint32(height)
+			m.Unlock()
+
+			// Inform all active deposits about a new block arrival.
+			for _, fsm := range m.activeDeposits {
+				select {
+				case fsm.blockNtfnChan <- uint32(height):
```
bhandras, line 145: Ah nvm, I see the chan is fed here.

staticaddr/deposit/manager.go
```diff
@@ -0,0 +1,421 @@
...
+	log.Infof("Recovering static address parameters and deposits...")
+
+	// Recover deposits.
+	deposits, err := m.cfg.Store.AllDeposits(ctx)
+	if err != nil {
+		return err
+	}
+
+	for i, d := range deposits {
+
```
bhandras, line 178: nit: extra nl

looprpc/client.proto
```diff
@@ -127,6 +127,16 @@ service SwapClient {
     */
     rpc InstantOutQuote (InstantOutQuoteRequest)
         returns (InstantOutQuoteResponse);
+
+    /* loop: `static newaddress`
+    NewAddress requests a new static address for loop-ins from the server.
+    */
+    rpc NewAddress (NewAddressRequest) returns (NewAddressResponse);
```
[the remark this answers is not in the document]
hieblmi: So these two rpcs are refactored to be in the `swapClient` rather than in the separate `StaticAddressClient` where they resided before. Their purpose is as their comments suggest.

Overall static address and deposit info will be introduced in https://github.com/lightninglabs/loop/pull/721.

staticaddr/deposit/sql_store.go
```diff
@@ -0,0 +1,256 @@
...
+
+		clock: clock.NewDefaultClock(),
+	}
+}
+
+// ExecTx is a wrapper for txBody to abstract the creation and commit of a db
+// transaction. The db transaction is embedded in a `*sqlc.Queries` that txBody
+// needs to use when executing each one of the queries that need to be applied
+// atomically.
+func (s *SqlStore) ExecTx(ctx context.Context, txOptions loopdb.TxOptions,
```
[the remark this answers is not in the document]
hieblmi: I just see it already is, so I will reuse that one.

staticaddr/deposit/sql_store.go
```diff
@@ -0,0 +1,256 @@
...
+		func(q *sqlc.Queries) error {
+			var err error
+
+			deposits, err := q.AllDeposits(ctx)
+			if err != nil {
+				return err
+			}
+
+			for _, deposit := range deposits {
+				updates, err := q.GetDepositUpdates(
```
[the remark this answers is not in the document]
hieblmi: done

staticaddr/deposit/actions.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+const (
+	defaultConfTarget = 3
+)
+
+// PublishExpiredDepositAction creates and publishes the timeout transaction
+// that spends the deposit from the static address timeout leaf to the
+// predefined timeout sweep pkscript.
+func (f *FSM) PublishExpiredDepositAction(_ fsm.EventContext) fsm.EventType {
```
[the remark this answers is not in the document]
hieblmi: yes that makes a lot more sense.

staticaddr/deposit/fsm.go
hieblmi: Yes your understanding is correct. Since each deposit has the same relative timelock but most likely a different effective timelock we'd somehow have to batch all the expired deposits. So the user would have to wait an additional amount of blocks until our threshold for sweeping timouts is reached and a bunch of deposits is batched together.

looprpc/client.proto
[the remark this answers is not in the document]
hieblmi: see my previous comment. I mentioned previously offline that this is part of a small refactoring.

hieblmi:
> So far this looks good to me! Gonna try to do a test run in a bit

Thank you, for convenience, here are some steps I take to test this.
 - reduce the CSV value on the server to be able to timeout after 100 blocks or so.
 - create a static address on the client side `loop --network=regtest s n`
 - send coins to it `reg_bob sendcoins --amt 9111123 --addr static_addr_here --min_confs 0`
 - mine 3 blocks, otherwise the coins are not showing up in `loop --network=regtest s list`
 - check the deposit with `loop --network=regtest s list`
 - mine enough blocks to timeout the deposit...

looprpc/client.proto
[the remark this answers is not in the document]
hieblmi: Yes you are right, previously in the static address server, the context was clear and the names made sense in that context. Now I'd call it `NewStaticAddress` and `ListUnspentDeposits`.

staticaddr/deposit/sql_store.go
```diff
@@ -0,0 +1,208 @@
...
+	}, nil
+}
+
+// Close closes the database connection.
+func (s *SqlStore) Close() {
+	s.baseDB.DB.Close()
+}
+
+// marshalSqlNullInt64 converts an int64 to a sql.NullInt64.
+func marshalSqlNullInt64(i int64) sql.NullInt64 {
```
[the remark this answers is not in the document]
hieblmi: I removed this function as we only call it once.

staticaddr/deposit/actions.go
```diff
@@ -0,0 +1,140 @@
...
+	msgTx.TxIn[0].Witness, err = address.GenTimeoutWitness(sig)
+	if err != nil {
+		return f.HandleError(err)
+	}
+
+	err = f.cfg.WalletKit.PublishTransaction(
+		f.ctx, msgTx, f.deposit.OutPoint.Hash.String()+"-close-sweep",
+	)
+	if err != nil {
+		return f.HandleError(err)
```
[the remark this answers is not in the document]
hieblmi: added logging

staticaddr/deposit/deposit.go
```diff
@@ -0,0 +1,66 @@
...
+	TimeOutSweepPkScript []byte
+}
+
+// IsPending returns true if the deposit is pending.
+func (d *Deposit) IsPending() bool {
+	return !d.IsFinal()
+}
+
+// IsFinal returns true if the deposit is final.
+func (d *Deposit) IsFinal() bool {
```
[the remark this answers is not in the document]
hieblmi: I'll rename this to `IsInFinalState` to make it clearer. This method is used when we recover deposits from the DB. Deposits that are in final state machine states(Expired, Withdrawn, LoopedIn etc.) are not eligible to be recovered, so this method helps to identify those.

staticaddr/deposit/actions.go
```diff
@@ -0,0 +1,140 @@
...
+		return f.HandleError(fmt.Errorf("timeout sweep fee "+
+			"estimation failed: %v", err))
+	}
+
+	weight := script.ExpirySpendWeight()
+
+	fee := feeRateEstimator.FeeForWeight(weight)
+
+	// We cap the fee at 20% of the deposit value.
+	if fee > f.deposit.Value/5 {
```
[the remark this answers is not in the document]
hieblmi: This would be a conf flag that could be used in a bunch of other places, for example in reservations or loop-in expiries. I will open an issue to address this in a separate PR.

staticaddr/deposit/deposit.go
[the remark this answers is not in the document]
bhandras: Could a deposit in `Expired` state still be expired but not swept back? I guess it's possible that we publish the sweep but it's evicted from the mempool. I suppose we could consider `Expired` non-final iiuc and add a real final state which means that a deposit is either spent or swept back?

staticaddr/deposit/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+func (f *FSM) handleBlockNotification(currentHeight uint32) error {
+	params, err := f.cfg.AddressManager.GetStaticAddressParameters(f.ctx)
+	if err != nil {
+		return err
+	}
+
+	// If the deposit is expired but not yet sufficiently confirmed, we
+	// republish the expiry sweep transaction.
+	if f.deposit.isExpired(currentHeight, params.Expiry) {
+		if f.deposit.State == WaitForExpirySweep {
```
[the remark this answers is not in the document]
hieblmi: >  we could just send the onexpiry event and in the fsm call the action itself again when the event is sent.

Wouldn't the new merged state action also wait for the confirmation of the sweep and hence block processing of new events until the state action returns a.k.a the expiry got confirmed? So we couldn't call `buildAndPublishTimeoutSweep` again since we wouldn't receive the event till the sweep is confirmed.

What do you think about extracting the sweep tx once we are in the final state `Expired` and then update the deposit with it, rather than needing to keep track of replacement txns while the sweep is unconfirmed?

staticaddr/deposit/actions.go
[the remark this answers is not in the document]
hieblmi: Added an issue to track this https://github.com/lightninglabs/loop/issues/729

staticaddr/deposit/deposit.go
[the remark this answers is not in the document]
hieblmi: > I guess it's possible that we publish the sweep but it's evicted from the mempool. 

The only way we end up in state `Expired` is if the sweep tx _PkScirpt_ got 3 confirmations by then `WaitForExpirySweepAction` returned `OnExpirySwept`. Until that is not the case we keep re-publishing the sweep tx. So the scenario in your first question is not possible imo.

For the second question, a state without state transitions other than self is a final state, so it is obvious from there. Do you still think we should add an extra `Final` state? From the last update in the DB it wouldn't be obvious what caused the final transition...

staticaddr/deposit/sql_store.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/loop/fsm"
+	"github.com/lightninglabs/loop/loopdb"
+	"github.com/lightninglabs/loop/loopdb/sqlc"
+	"github.com/lightningnetwork/lnd/clock"
+)
+
+// SqlStore is the backing store for static addresses.
```
hieblmi, line 16: fix comment

staticaddr/deposit/sql_store.go
```diff
@@ -0,0 +1,213 @@
...
+// to the underlying driver which can be postgres or sqlite.
+func NewSqlStore(db *loopdb.BaseDB) *SqlStore {
+	return &SqlStore{
+		baseDB: db,
+
+		clock: clock.NewDefaultClock(),
+	}
+}
+
+// CreateDeposit creates a static address record in the database.
```
hieblmi, line 33: fix comment

looprpc/client.proto
```diff
@@ -127,6 +127,18 @@ service SwapClient {
...
+    /* loop: `static newstaticaddress`
+    NewStaticAddress requests a new static address for loop-ins from the server.
+    */
+    rpc NewStaticAddress (NewStaticAddressRequest)
+        returns (NewStaticAddressResponse);
+
+    /* loop: `static listunspentdeposits`
+    ListUnspentDeposits returns a list of utxos deposited at a static address.
+    */
+    rpc ListUnspentDeposits (ListUnspentDepositsRequest)
```
bhandras, line 140: Nice, I think the new naming is much cleaner!

loopdb/sqlc/queries/static_address_deposits.sql
```diff
@@ -0,0 +1,66 @@
...
+             $1,
+             $2,
+             $3,
+             $4,
+             $5,
+             $6,
+             $7
+         );
+
+-- name: UpdateDeposit :exec
```
bhandras, line 20: It's simpler to use just one `UpdateDeposit` function but perhaps worth considering a more granular approach where depending on the state change we only update the confirmation height or the expiry txid. WDYT? Obviously this would also mean a slight change in how the FSM works and I could also argue its a bit overly cautious.

staticaddr/address/log.go
```diff
@@ -0,0 +1,29 @@
+package address
```
bhandras, line 1: Another approach to subpackage logging could be to just have one logger in the `staticaddr` package and then each subpackage just defines a global variable (`log`) which we assign with a modified `UseLogger` function.

staticaddr/deposit/actions.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return f.HandleError(err)
+	}
+
+	txLabel := fmt.Sprintf("timeout sweep for deposit %v",
+		f.deposit.OutPoint)
+
+	err = f.cfg.WalletKit.PublishTransaction(f.ctx, msgTx, txLabel)
+	if err != nil {
+		if !strings.Contains(err.Error(), "output already spent") {
```
bhandras, line 96: 👍

staticaddr/deposit/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+		deposit:              deposit,
+		params:               params,
+		address:              address,
+		ctx:                  ctx,
+		blockNtfnChan:        make(chan uint32),
+		finalizedDepositChan: finalizedDepositChan,
+	}
+
+	depositStates := depoFsm.DepositStatesV0()
+	switch params.ProtocolVersion {
```
bhandras, line 97: nit: this check could go just below where we call `GetStaticAddressParameters()`.

staticaddr/deposit/manager.go
```diff
@@ -0,0 +1,420 @@
...
+	MaxConfs = 0
+)
+
+// ManagerConfig holds the configuration for the address manager.
+type ManagerConfig struct {
+	// AddressClient is the client that communicates with the loop server
+	// to manage static addresses.
+	AddressClient staticaddressrpc.StaticAddressServerClient
+
+	AddressManager AddressManager
```
bhandras, line 40: nit: missing godoc

staticaddr/deposit/fsm.go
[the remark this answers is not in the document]
bhandras: As a side note this is also not safe as we're running on the block notification goroutine.

bhandras (commented):
Looking really good now! Would be nice to add some test coverage too, but can come in a follow up PR as well given the side branch.

staticaddr/address/log.go
hieblmi: That reminds me that I didn't want to clutter the logs too much with subsystem logs. I think we can just summarize all deposit/withdraw/loop-in operations inside the `SADDR` logger. I think the log becomes easier to read then.

hieblmi: I placed a single logger into the `staticaddr` package. For the loop-in I might add a separate logger, but creation/deposits/withdrawals I think can be under the same sublogger.

staticaddr/deposit/fsm.go
hieblmi: I think it has to stay here since if there'd be a `V1` state machine we need the `depoFsm` struct to retrieve those states.

staticaddr/deposit/fsm.go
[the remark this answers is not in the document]
hieblmi: Thanks yes, I've added a mutex to the deposit and locked all operations on the state.

hieblmi:
Thanks for the review @bhandras, I plan to add some unit tests to increase test coverage.

staticaddr/address/manager.go
```diff
@@ -46,30 +51,20 @@ type ManagerConfig struct {
...
-	initChan chan struct{}
-
 	sync.Mutex
 }
 
-// NewAddressManager creates a new address manager.
-func NewAddressManager(cfg *ManagerConfig) *Manager {
+// NewManager creates a new address manager.
+func NewManager(cfg *ManagerConfig) *Manager {
+	log = staticaddr.GetLogger()
```
bhandras, line 59: nit: alternatively, you could do this in the package's `init()` function too, so it's executed only once at the static initialization stage.

staticaddr/deposit/deposit.go
```diff
@@ -0,0 +1,107 @@
...
+	defer d.Unlock()
+
+	return d.State
+}
+
+func (d *Deposit) setState(state fsm.StateType) {
+	d.Lock()
+	defer d.Unlock()
+
+	d.State = state
```
bhandras, line 92: nit: given we have setter and getter maybe we don't even need to export `State`.

staticaddr/address/manager_test.go
```diff
@@ -0,0 +1,147 @@
...
+	in *swapserverrpc.ServerNewAddressRequest, opts ...grpc.CallOption) (
+	*swapserverrpc.ServerNewAddressResponse, error) {
+
+	args := m.Called(ctx, in, opts)
+
+	return args.Get(0).(*swapserverrpc.ServerNewAddressResponse),
+		args.Error(1)
+}
+
+func TestManager(t *testing.T) {
```
bhandras, line 45: nit: missing godoc

bhandras (approved):
Awesome work on the deposit FSM 🚀  LGTM ⚡

staticaddr/address/manager.go
hieblmi: good idea, obvious in hindsight.

Issue #680, opened 2023-12-24
Container 3bf773dd5d54debe3086f30feb2061db89ccd6ef6a03d24dd96910fed13628b8 is not running


1nrho12:
Hello, I am trying to run ./regtest.sh start to start on regtest mode, but every time I get:

`./regtest.sh start
[+] Running 5/5
 ✔ Container regtest-bitcoind-1    Started                                                                                                                                                             0.4s 
 ✔ Container regtest-lndclient-1   Started                                                                                                                                                            10.9s 
 ✔ Container regtest-lndserver-1   Started                                                                                                                                                            10.6s 
 ✔ Container regtest-loopserver-1  Started                                                                                                                                                             0.7s 
 ✔ Container regtest-loopclient-1  Started                                                                                                                                                             0.5s 
Waiting for nodes to start
Error response from daemon: Container 3bf773dd5d54debe3086f30feb2061db89ccd6ef6a03d24dd96910fed13628b8 is not running`

> OS: macOS

> lnd --version
lnd version 0.17.3-beta commit=v0.17.3-beta

> loopd --version
loopd version 0.12.0-beta commit=

Could someone, please, help?

1nrho12:
Additional info:

lnd.conf:

`[Application Options]
listen=0.0.0.0:9734
rpclisten=localhost:10009
restlisten=0.0.0.0:8180

[Bitcoin]
bitcoin.active=1
bitcoin.regtest=1
bitcoin.node=bitcoind

[Bitcoind]
bitcoind.rpchost=localhost
bitcoind.rpcuser=bitcoinrpc
bitcoind.rpcpass=password
bitcoind.zmqpubrawblock=tcp://127.0.0.1:29000
bitcoind.zmqpubrawtx=tcp://127.0.0.1:28333
`

loopd.conf:
`[Bitcoin]
bitcoin.active=1
bitcoin.regtest=1
zmqpubrawblock=tcp://127.0.0.1:29000
zmqpubrawtx=tcp://127.0.0.1:28333

[Routing]
routing.assumechanvalid=1

[Loop]
loop.outgoingchanmonitor=1

[Lnd]
lnd.adminmacaroonpath=/Users/macos/Library/Application Support/Lnd/data/chain/bitcoin/regtest/admin.macaroon
`

`docker logs 3bf773dd5d54
[...]

PR #683, opened 2024-01-05
docker-regtest: fix loop-server config

sputn1ck:
This PR fixes the `docker-compose.yml` file for recent loop-server changes.

addresses #680


bhandras (approved):
Thanks! :)

PR #692, opened 2024-01-31
StaticAddr: Dependencies

hieblmi:
This PR introduces protocol versioning and a address key family for static addresses.


staticaddr/protocol_version.go
```diff
@@ -0,0 +1,48 @@
...
+type AddressProtocolVersion uint32
+
+const (
+	// ProtocolVersion_V0 indicates that the client is a legacy version
+	// that did not report its protocol version.
+	ProtocolVersion_V0 AddressProtocolVersion = 0
+
+	// stableRPCProtocolVersion defines the current stable RPC protocol
+	// version.
+	stableRPCProtocolVersion = looprpc.StaticAddressProtocolVersion_V0
```
GeorgeTsagk, line 18: undefined references here and below, missing `go.mod` entry?

hieblmi: Hmm ya, they will be used in a loop follow-up PR, but used meanwhile in a different project.

GeorgeTsagk (approved):

PR #698, opened 2024-02-08
instantout: Check reservation expiry and improve FSM UX

sputn1ck:
*only last 2 commits are new*

This PR adds an additional client side check to make sure the reservation expires later than the requested instantout.

I'm also expanding the fsm observer to allow for an early return, should it see an `fsm.OnError` event. This allows us to abort early and give users a quicker feedback,

### DIFF
--- a/instantout/actions.go
+++ b/instantout/actions.go
@@ -21,7 +21,7 @@ import (
 	"github.com/lightningnetwork/lnd/lntypes"
 )
 
-var (
+const (
 	// Define route independent max routing fees. We have currently no way
 	// to get a reliable estimate of the routing fees. Best we can do is
 	// the minimum routing fees, which is not very indicative.
@@ -46,6 +46,10 @@ var (
 	// defaultPollPaymentTime is the default time to poll the server for the
 	// payment status.
 	defaultPollPaymentTime = time.Second * 15
+
+	// htlcExpiryDelta is the delta in blocks we require between the htlc
+	// expiry and reservation expiry.
+	htlcExpiryDelta = int32(40)
 )
 
 // InitInstantOutCtx contains the context for the InitInstantOutAction.
@@ -96,6 +100,15 @@ func (f *FSM) InitInstantOutAction(eventCtx fsm.EventContext) fsm.EventType {
 		reservationAmt += uint64(res.Value)
 		reservationIds = append(reservationIds, resId[:])
 		reservations = append(reservations, res)
+
+		// Check that the reservation expiry is larger than the cltv
+		// expiry of the swap, with an additional delta to allow for
+		// preimage reveal.
+		if int32(res.Expiry) < initCtx.cltvExpiry+htlcExpiryDelta {
+			return f.HandleError(fmt.Errorf("reservation %x has "+
+				"expiry %v which is less than the swap expiry %v",
+				resId, res.Expiry, initCtx.cltvExpiry))
+		}
 	}
 
 	// Create the preimage for the swap.
--- a/instantout/manager.go
+++ b/instantout/manager.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/loop/fsm"
 	"github.com/lightninglabs/loop/instantout/reservation"
 	"github.com/lightninglabs/loop/swapserverrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
@@ -169,15 +170,10 @@ func (m *Manager) NewInstantOut(ctx context.Context,
 	// waiting for sweepless sweep to be confirmed.
 	err = instantOut.DefaultObserver.WaitForState(
 		ctx, defaultStateWaitTime, WaitForSweeplessSweepConfirmed,
+		fsm.WithAbortEarlyOnErrorOption(),
 	)
 	if err != nil {
-		if instantOut.LastActionError != nil {
-			return instantOut, fmt.Errorf(
-				"error waiting for sweepless sweep "+
-					"confirmed: %w", instantOut.LastActionError,
-			)
-		}
-		return instantOut, nil
+		return nil, err
 	}
 
 	return instantOut, nil


fsm/fsm.go
```diff
@@ -320,3 +320,16 @@ func NewErrWaitingForStateTimeout(expected,
 		expected, actual,
 	))
 }
+
+type ErrWaitingForStateEarlyAbortError error
```
bhandras, line 324: nit: godoc

bhandras: Also consider the usual pattern where we create a new error global with `errors.New` and then when you add context you just `fmt.Errorf("blah %w blah ...", myError, ...)`

fsm/observer.go
```diff
@@ -130,7 +158,12 @@ func (s *CachedObserver) WaitForState(ctx context.Context,
 		return NewErrWaitingForStateTimeout(
 			state, s.lastNotification.NextState,
 		)
-	case <-ch:
+	case errorOccured := <-ch:
```
bhandras, line 161: nit: add nl

fsm/observer.go
```diff
@@ -76,6 +77,24 @@ func (w *InitialWaitOption) apply(o *fsmOptions) {
...
+	abortEarlyOnError bool
+}
+
+// apply implements the WaitForStateOption interface.
+func (a *AbortEarlyOnErrorOption) apply(o *fsmOptions) {
+	o.abortEarlyOnError = a.abortEarlyOnError
+}
+
+// WithAbortEarlyOnErrorOption creates a new AbortEarlyOnErrorOption.
+func WithEarlyAbort() WaitForStateOption {
```
bhandras, line 92: nit: godoc and function name is not in sync

fsm/observer.go
```diff
@@ -130,7 +158,12 @@ func (s *CachedObserver) WaitForState(ctx context.Context,
 		return NewErrWaitingForStateTimeout(
 			state, s.lastNotification.NextState,
 		)
-	case <-ch:
+	case errorOccured := <-ch:
+		if errorOccured {
+			return NewErrWaitingForStateEarlyAbortError(
```
bhandras, line 163: Could we just return the last error that made the fsm abort?

instantout/actions.go
```diff
@@ -96,6 +96,14 @@ func (f *FSM) InitInstantOutAction(eventCtx fsm.EventContext) fsm.EventType {
 		reservationAmt += uint64(res.Value)
 		reservationIds = append(reservationIds, resId[:])
 		reservations = append(reservations, res)
+
+		// Check that the reservation expiry is larger than the cltv
+		// expiry of the swap.
+		if int32(res.Expiry) <= initCtx.cltvExpiry {
```
bhandras, line 102: So I still think we should apply the delta here no?

instantout/actions.go
```diff
@@ -96,6 +100,15 @@ func (f *FSM) InitInstantOutAction(eventCtx fsm.EventContext) fsm.EventType {
 		reservationAmt += uint64(res.Value)
 		reservationIds = append(reservationIds, resId[:])
 		reservations = append(reservations, res)
+
+		// Check that the reservation expiry is larger than the cltv
+		// expiry of the swap, with an additional delta to allow for
+		// preimage reveal.
+		if int32(res.Expiry) < initCtx.cltvExpiry+htlcExpiryDelta {
```
bhandras, line 107: Looks good!

instantout/actions.go
```diff
@@ -96,6 +100,15 @@ func (f *FSM) InitInstantOutAction(eventCtx fsm.EventContext) fsm.EventType {
...
+			return f.HandleError(fmt.Errorf("reservation %x has "+
+				"expiry %v which is less than the swap expiry %v",
```
bhandras, line 109: nit: maybe mention the delta here so it's not confusing if someone gets this error.

bhandras (approved):
LGTM 🎉

fsm/observer.go
```diff
@@ -115,10 +135,18 @@ func (s *CachedObserver) WaitForState(ctx context.Context,
 		for {
 			// Check if the last state is the desired state
 			if s.lastNotification.NextState == state {
-				ch <- struct{}{}
+				ch <- nil
```
hieblmi, line 138: do we have to check `ctx.Done()` here?

hieblmi, line 138: do we have to check `timeoutCtx.Done()` here?

fsm/observer.go
```diff
@@ -115,10 +135,18 @@ func (s *CachedObserver) WaitForState(ctx context.Context,
...
 				return
 			}
 
+			// Check if an error occurred
+			if s.lastNotification.Event == OnError {
+				if options.abortEarlyOnError {
+					ch <- s.lastNotification.LastActionError
```
hieblmi, line 145: do we have to check `timeoutCtx.Done()` here?

fsm/observer.go
```diff
@@ -115,10 +135,18 @@ func (s *CachedObserver) WaitForState(ctx context.Context,
...
 			// Check if the last state is the desired state
 			if s.lastNotification.NextState == state {
-				ch <- struct{}{}
+				ch <- nil
 				return
 			}
 
+			// Check if an error occurred
+			if s.lastNotification.Event == OnError {
+				if options.abortEarlyOnError {
```
hieblmi, line 144: I think we should also return if an error occured and no option is set?

instantout/actions.go
```diff
@@ -46,6 +46,10 @@ var (
 	// defaultPollPaymentTime is the default time to poll the server for the
 	// payment status.
 	defaultPollPaymentTime = time.Second * 15
+
+	// htlcExpiryDelta is the delta in blocks we require between the htlc
+	// expiry and reservation expiry.
+	htlcExpiryDelta = int32(40)
```
hieblmi, line 52: shouldn't these `var`s all be `const`?

hieblmi (commented):
Left a question regarding the timeout context in `WaitForState`.

fsm/observer.go
hieblmi: Okay that makes sense

hieblmi (approved):
LGTM 💾

PR #699, opened 2024-02-09
reservation: fix exiting on fetchL402

sputn1ck:
This PR changes the way the reservation manager works, by not exiting when the `fetchL402` call fails. This ensures that a client can still run loop, even if they don't have outbound capacity or the loop server is offline.


bhandras (approved):
LGTM 💯

PR #703, opened 2024-02-14
version: bump version to v0.27.1-beta

alexbosworth:


bhandras (approved):

hieblmi:
Some release notes:

`This release adds automatic sweeping of incorrectly deposited amounts to external loop in addresses.
Previously, a mismatch in the contract amount and the actually deposited amount required external tools
to recover the client funds. With this release the client automatically sweeps the funds back to the
wallet upon contract expiry.`

PR #709, opened 2024-03-01
instantout: add custom address to send funds to

sputn1ck:
This PR adds the ability to set a custom address that instantout will sweep to.


instantout/actions.go
```diff
@@ -59,6 +59,7 @@ type InitInstantOutCtx struct {
 	initationHeight int32
 	outgoingChanSet loopdb.ChannelSet
 	protocolVersion ProtocolVersion
+	sweepAddress    string
```
bhandras, line 62: nit: alternatively could use `*btcutil.Address` pointer so deconding can happen on the CLI.

bhandras (approved):
LGTM 🚀

instantout/actions.go
```diff
@@ -165,11 +166,21 @@ func (f *FSM) InitInstantOutAction(eventCtx fsm.EventContext) fsm.EventType {
 	}
 
 	// Create the address that we'll send the funds to.
-	sweepAddress, err := f.cfg.Wallet.NextAddr(
-		f.ctx, "", walletrpc.AddressType_TAPROOT_PUBKEY, false,
-	)
-	if err != nil {
-		return f.HandleError(err)
+	var sweepAddress btcutil.Address
+	if initCtx.sweepAddress != "" {
```
hieblmi, line 170: I think this should be `==`

hieblmi (commented):
I think there's a small error in how we check the provided sweep address.

instantout/actions.go
```diff
@@ -165,11 +167,17 @@ func (f *FSM) InitInstantOutAction(eventCtx fsm.EventContext) fsm.EventType {
 	}
 
 	// Create the address that we'll send the funds to.
-	sweepAddress, err := f.cfg.Wallet.NextAddr(
-		f.ctx, "", walletrpc.AddressType_TAPROOT_PUBKEY, false,
-	)
-	if err != nil {
-		return f.HandleError(err)
+	var sweepAddress btcutil.Address
```
hieblmi, line 170: Ultra-mirco nit, could we simplify to

```suggestion
        var sweepAddress = initCtx.sweepAddress
	if sweepAddress == nil {
		sweepAddress, err = f.cfg.Wallet.NextAddr(
			f.ctx, lnwallet.DefaultAccountName,
			walletrpc.AddressType_TAPROOT_PUBKEY, false,
		)
		if err != nil {
			return f.HandleError(err)
		}
	}
```

hieblmi (approved):
great feature, LGTM

PR #719, opened 2024-03-15
[4/?] StaticAddr: Instant deposit withdrawals

hieblmi:
This PR enables instant withdrawals of funds at static address deposits, allowing the client to omit the CSV timeout period of their Musig2 outpoints and cooperatively spend selected deposit outpoints cooperatively with the server.

Withdrawing can be done for selected or all eligible deposits. Examples:

Selected deposit:
```
loop --network=regtest s withdraw --utxo 0b879cb14939f98ad359063d60e648b3db72a1dc32bf9a1fbafa1a0d87ec1db:0
```

All deposits:
```
loop --network=regtest s withdraw --all
```


swapserverrpc/withdraw.proto
```diff
@@ -0,0 +1,57 @@
...
+
+service WithdrawalServer {
+    // WithdrawDeposits allows to cooperatively sweep deposits that haven't
+    // timed out yet to the client's wallet. The server will generate the
+    // partial sigs for the client's selected deposits.
+    rpc WithdrawDeposits (ServerWithdrawRequest)
+        returns (ServerWithdrawResponse);
+}
+
+message ServerWithdrawRequest {
```
bhandras, line 19: My main q here (commented on the server side too) whether it'd be better to just send the psbt of the sweep and the nonces.

hieblmi: As discussed we don't draw an advantage here by using PSBT since the withdrawal tx won't dynamically add any inputs during signing.
The current implementation is also aligned with the reservations flow, so it might become easier to review.

staticaddr/deposit/fsm.go
```diff
@@ -250,6 +274,16 @@ func (f *FSM) updateDeposit(notification fsm.Notification) {
 	}
 }
 
+/*func isRecoverable(state fsm.StateType) bool {
```
hieblmi, line 277: remove

cmd/loop/staticaddr.go
```diff
@@ -105,3 +111,113 @@ func listUnspent(ctx *cli.Context) error {
...
+	Usage:     "Withdraw from static address deposits.",
+	Description: `
+	Withdraws from all or selected static address deposits by sweeping them 
+	back to our lnd wallet.
+	`,
+	Flags: []cli.Flag{
+		cli.StringSliceFlag{
+			Name: "utxo",
+			Usage: "specify utxos as outpoints(tx:idx) which will" +
+				"be closed.",
```
hieblmi, line 127: ...be withdrawn

loopd/daemon.go
```diff
@@ -561,6 +563,18 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 			Signer:         d.lnd.Signer,
 		}
 		depositManager = deposit.NewManager(depoCfg)
+
+		// Static address deposit withdrawal manager setup.
+		closeCfg := &withdraw.ManagerConfig{
```
hieblmi, line 568: `withdrawalCfg`

swapserverrpc/server.proto
```diff
@@ -677,4 +686,37 @@ message ServerAddressParameters {
 
     // The CSV expiry for the MuSig2 static address output.
     uint32 expiry = 2;
+}
+message ServerWithdrawRequest {
+    // The deposit outpoints the client wishes to close.
```
hieblmi, line 691: ...to withdraw

swapserverrpc/server.proto
```diff
@@ -677,4 +686,37 @@ message ServerAddressParameters {
...
+    repeated ServerOutPoint outpoints = 1;
+
+    // The nonces that the client used to generate the coop close tx sigs.
```
hieblmi, line 694: ...withdrawal tx...

cmd/loop/staticaddr.go
```diff
@@ -105,3 +111,113 @@ func listUnspent(ctx *cli.Context) error {
...
+	}
+
+	client, cleanup, err := getClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	var (
+		req            = &looprpc.WithdrawDepositsRequest{}
```
hieblmi, line 149: remove `req`

loopdb/sqlc/migrations/000008_static_address_deposits.up.sql
```diff
@@ -24,7 +24,11 @@ CREATE TABLE IF NOT EXISTS deposits (
     timeout_sweep_pk_script BYTEA NOT NULL,
 
     -- expiry_sweep_txid is the transaction id of the expiry sweep.
-    expiry_sweep_txid BLOB
+    expiry_sweep_txid BLOB,
+
+    -- withdrawal_sweep_pk_script is the address that will be used to sweep the
+    -- deposit cooperatively with the server before it has expired.
```
starius, line 30: > before it has expired

Is it possible to withdraw after the deposit has expired?

hieblmi: The client currently prevents selecting deposits for withdrawal that are in the expired state.

swapserverrpc/server.proto
```diff
@@ -677,4 +686,37 @@ message ServerAddressParameters {
...
+    repeated bytes client_nonces = 2;
+
+    // The address that the client wants to sweep the static address deposits
+    // to.
+    string client_sweep_addr = 3;
+
+    // The fee rate in sat/kw that the client wants to use for the sweep.
+    uint64 musig_tx_fee_rate = 4;
```
starius, line 702: I suggest to rename to `musig2_tx_fee_rate` not to imply that MuSig1 protocol is used. Or just `tx_fee_rate`.

There is another instance of a field with name `musig_tx_fee_rate` in `instantout.proto`. I suggest to rename here as well.

swapserverrpc/server.proto
```diff
@@ -677,4 +686,37 @@ message ServerAddressParameters {
...
+    // The nonces that the server used to generate the sweepless sweep sigs.
+    repeated bytes server_nonces = 2;
+}
+
+message ServerOutPoint {
+    // Raw bytes representing the transaction id.
+    bytes txid_bytes = 1;
+
+    // Hex-encoded string representing the transaction id.
+    string txid_str = 2;
```
starius, line 718: These two fields seem to store the same information (txid) in different formats. Can we keep only one of them or use proto3's [oneof operator](https://protobuf.dev/programming-guides/proto3/#oneof)?

staticaddr/address/manager.go
```diff
@@ -60,6 +56,7 @@ type Manager struct {
 
 // NewManager creates a new address manager.
 func NewManager(cfg *ManagerConfig) *Manager {
+	log = staticaddr.GetLogger()
```
starius, line 59: Why is this change needed? Isn't `sync.Once` needed here to prevent `log` from being simultaneously overwritten and used for logging?

staticaddr/deposit/actions.go
```diff
@@ -146,3 +146,19 @@ func (f *FSM) SweptExpiredDepositAction(_ fsm.EventContext) fsm.EventType {
...
+// the manager that the deposit has been swept and the FSM can be removed. It
+// also ends the state machine main loop by cancelling its context.
+func (f *FSM) WithdrawnDepositAction(_ fsm.EventContext) fsm.EventType {
+	select {
+	case <-f.ctx.Done():
+		return fsm.OnError
+
+	default:
+		f.finalizedDepositChan <- f.deposit.OutPoint
+		f.ctx.Done()
```
starius, line 160: `f.ctx.Done()` does not close the context. The description of the function says "ends the state machine main loop by cancelling its context", but this call just returns the context, it is a no-op.

loopdb/sqlc/migrations/000008_static_address_deposits.up.sql
starius: > But that is useless to the client.

Wouldn't it save some sats for the client? IIUC, this is a way to spend the UTXO using MuSig2 instead of using tapscript (i.e. cheaper and more private).

staticaddr/address/manager.go
starius: There is another instance in `deposit` package.

staticaddr/deposit/manager.go
```diff
@@ -430,3 +428,85 @@ func (m *Manager) finalizeDeposit(outpoint wire.OutPoint) {
...
+	defer m.Unlock()
+
+	var deposits []*Deposit
+	for _, fsm := range m.activeDeposits {
+		if fsm.deposit.GetState() != stateFilter {
+			continue
+		}
+		deposits = append(deposits, fsm.deposit)
+	}
+
```
starius, line 446: This function returns deposits in arbitrary order, getting them by traversing a map (`m.activeDeposits`). We can sort them by ID to make the function deterministic.

staticaddr/deposit/manager.go
```diff
@@ -430,3 +428,85 @@ func (m *Manager) finalizeDeposit(outpoint wire.OutPoint) {
...
+// are active and in the specified state.
+func (m *Manager) AllOutpointsActiveDeposits(outpoints []wire.OutPoint,
+	stateFilter fsm.StateType) (
+
+	[]*Deposit, bool) {
+
+	m.Lock()
+	defer m.Unlock()
+
+	var deposits []*Deposit
```
starius, line 465: I suggest this:
```go
deposits := make([]*Deposit, 0, len(outpoints))
```
to preallocate slice and to make sure it is never `nil`.

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,726 @@
...
+	withdrawalAddress btcutil.Address) error {
+
+	withdrawalPkScript, err := txscript.PayToAddrScript(withdrawalAddress)
+	if err != nil {
+		return err
+	}
+
+	confChan, errChan, err := m.cfg.ChainNotifier.RegisterConfirmationsNtfn(
+		m.runCtx, nil, withdrawalPkScript, MinConfs,
+		int32(m.initiationHeight),
```
starius, line 381: Access to initiationHeight without a lock.

I suggest to use [atomic.Uint32](https://pkg.go.dev/sync/atomic@go1.19.10#Uint32) for initiationHeight and currentHeight.

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,726 @@
...
+		if !strings.Contains(err.Error(), "output already spent") {
+			log.Errorf("%v: %v", txLabel, err)
+			return err
+		}
+	} else {
+		log.Debugf("published deposit withdrawal with txid: %v",
+			finalizedWithdrawalTx.TxHash())
+	}
+
+	return err
```
starius, line 368:
```go
	if err != nil {
		if !strings.Contains(err.Error(), "output already spent") {
			log.Errorf("%v: %v", txLabel, err)
		}

		return err
	}
	
	log.Debugf("published deposit withdrawal with txid: %v",
		finalizedWithdrawalTx.TxHash())

	return nil
```

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,726 @@
...
+
+	// Add the deposit inputs to the transaction in the order the server
+	// signed them.
+	for _, outpoint := range outpoints {
+		msgTx.AddTxIn(&wire.TxIn{
+			PreviousOutPoint: outpoint,
+		})
+	}
+
+	// Estimate the fee
```
starius, line 540: Add '".".

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,726 @@
...
+		}
+		prevOuts[outpoint] = txOut
+	}
+
+	return prevOuts
+}
+
+func (m *Manager) republishWithdrawals() error {
+	m.Lock()
+	defer m.Unlock()
```
starius, line 707: m.publishWithdrawal call also happens under the lock. It is a long running call, so I suggest to avoid calling it under the lock.

loopd/swapclient_server.go
```diff
@@ -1336,6 +1339,67 @@ func (s *swapClientServer) ListUnspentDeposits(ctx context.Context,
...
 
+// WithdrawDeposits tries to obtain a partial signature from the server to spend
+// the selected deposits to the client's wallet.
+func (s *swapClientServer) WithdrawDeposits(ctx context.Context,
+	req *clientrpc.WithdrawDepositsRequest) (
+	*clientrpc.WithdrawDepositsResponse, error) {
+
+	var (
+		isAllSelected  = req.All
+		isUtxoSelected = req.Outpoints != nil
```
starius, line 1350:
```go
isUtxoSelected = len(req.Outpoints) > 0
```

An empty slice can be not equal to `nil`.

cmd/loop/staticaddr.go
```diff
@@ -105,3 +111,113 @@ func listUnspent(ctx *cli.Context) error {
...
+		return err
+	}
+
+	printRespJSON(resp)
+
+	return nil
+}
+
+func utxosToOutpoints(utxos []string) ([]*looprpc.OutPoint, error) {
+	var outpoints []*looprpc.OutPoint
```
starius, line 188:
```go
outpoints := make([]*looprpc.OutPoint, 0, len(utxos))
```

cmd/loop/staticaddr.go
```diff
@@ -105,3 +111,113 @@ func listUnspent(ctx *cli.Context) error {
...
+	}
+	txid := parts[0]
+	if hex.DecodedLen(len(txid)) != chainhash.HashSize {
+		return nil, fmt.Errorf("invalid hex-encoded txid %v", txid)
+	}
+	outputIndex, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return nil, fmt.Errorf("invalid output index: %v", err)
+	}
+	return &looprpc.OutPoint{
```
starius, line 219: Add an empty line before `return`.

swapserverrpc/server.proto
hieblmi: Thanks for the suggestion, I used the `oneof` operator.

staticaddr/address/manager.go
hieblmi: `address`, `deposit` and `withdraw` all share the same instance of the logger. I think `GetLogger` returns a single instance of the logger from the `init()` call.

How else would you access the common log instance from `staticaddr/log.go`? Pass it into each of the managers?

staticaddr/deposit/actions.go
hieblmi: Good catch, removed it. After this call the FSM is in a final state.

staticaddr/withdraw/manager.go
hieblmi: nice catch

staticaddr/deposit/sql_store.go
```diff
@@ -39,6 +39,9 @@ func (s *SqlStore) CreateDeposit(ctx context.Context, deposit *Deposit) error {
 		Amount:               int64(deposit.Value),
 		ConfirmationHeight:   deposit.ConfirmationHeight,
 		TimeoutSweepPkScript: deposit.TimeOutSweepPkScript,
+		WithdrawalSweepAddress: sql.NullString{
+			String: deposit.WithdrawalSweepAddress,
```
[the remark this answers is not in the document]
hieblmi: It looks like we have to tell the db here if the string is valid or not. If it is valid(meanning != "") but we set Valid: false then its written empty to the db it seems.

swapserverrpc/server.proto
```diff
@@ -677,4 +686,34 @@ message ServerAddressParameters {
 
     // The CSV expiry for the MuSig2 static address output.
     uint32 expiry = 2;
+}
+message ServerWithdrawRequest {
```
bhandras, line 690: nit: add nl

staticaddr/address/manager_test.go
```diff
@@ -32,6 +32,15 @@ type mockStaticAddressClient struct {
 	mock.Mock
 }
 
+func (m *mockStaticAddressClient) ServerWithdrawDeposits(ctx context.Context,
+	in *swapserverrpc.ServerWithdrawRequest,
+	opts ...grpc.CallOption) (*swapserverrpc.ServerWithdrawResponse,
+	error) {
+
+	// TODO implement me
+	panic("implement me")
```
bhandras, line 41: nit: missing impl

staticaddr/deposit/manager.go
```diff
@@ -420,3 +422,88 @@ func (m *Manager) finalizeDeposit(outpoint wire.OutPoint) {
...
+		m.Unlock()
+		if !ok {
+			return fmt.Errorf("deposit not found")
+		}
+
+		err := sm.SendEvent(event, nil)
+		if err != nil {
+			return err
+		}
+		err = sm.DefaultObserver.WaitForState(
```
bhandras, line 496: Do we expect that the one minute will be enough? Perhaps could be a timeout param?

staticaddr/deposit/manager.go
```diff
@@ -420,3 +422,88 @@ func (m *Manager) finalizeDeposit(outpoint wire.OutPoint) {
...
+		)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (m *Manager) UpdateDeposit(d *Deposit) error {
```
bhandras, line 507: nit: missing godoc

staticaddr/deposit/manager.go
```diff
@@ -420,3 +422,88 @@ func (m *Manager) finalizeDeposit(outpoint wire.OutPoint) {
...
+	defer m.Unlock()
+
+	deposits := make([]*Deposit, 0, len(outpoints))
+	for _, o := range outpoints {
+		if _, ok := m.activeDeposits[o]; !ok {
+			return nil, false
+		}
+
+		deposit := m.deposits[o]
+		if deposit.GetState() != stateFilter {
```
bhandras, line 469: Such check also assumes that the state machine does not advance in the background iiuc. I guess it's OK in specific cases where we expect that the deposits are in the correct state.

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,723 @@
...
+	MinConfs int32 = 3
+
+	// Is the default confirmation target for the fee estimation of the
+	// withdrawal transaction.
+	defaultConfTarget int32 = 3
+)
+
+// ManagerConfig holds the configuration for the address manager.
+type ManagerConfig struct {
+	StaticAddressServerClient staticaddressrpc.StaticAddressServerClient
```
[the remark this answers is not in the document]
bhandras: +1

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,723 @@
...
+		}
+	}
+}
+
+func (m *Manager) recover() error {
+	// To recover withdrawals we skim through all active deposits and check
+	// if they have a withdrawal address set. For the ones that do we
+	// cluster those with equal withdrawal addresses and kick-off
+	// their withdrawal. Each cluster represents a separate withdrawal
+	// intend by the user.
```
bhandras, line 143: nit: intent

staticaddr/withdraw/manager.go
```diff
@@ -0,0 +1,723 @@
...
+	txLabel := fmt.Sprintf("deposit-withdrawal-%v",
+		finalizedWithdrawalTx.TxHash())
+
+	// Publish the withdrawal sweep transaction.
+	err = m.cfg.WalletKit.PublishTransaction(
+		m.runCtx, finalizedWithdrawalTx, txLabel,
+	)
+
+	if err != nil {
+		if !strings.Contains(err.Error(), "output already spent") {
```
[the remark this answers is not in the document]
bhandras: +1

staticaddr/withdraw/manager.go
bhandras: On the other hand I think it is nice that all state is locked together. I think it's probably better to keep so and not introduce atomics for these specific use cases.

bhandras (approved):
Really nice work @hieblmi 🎉 LGTM, pending a few nits and comments ✅

staticaddr/address/manager.go
hieblmi: Thank you, I've applied this change.

staticaddr/deposit/fsm.go
```diff
@@ -41,6 +45,8 @@ var (
 // Events.
 var (
 	OnStart           = fsm.EventType("OnStart")
+	OnWithdraw        = fsm.EventType("OnWithdraw")
+	OnWithdrawn       = fsm.EventType("OnWithdrawn")
```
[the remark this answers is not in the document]
hieblmi: Technically not published yet, I renamed it to `OnWithdrawInitiated`.

bhandras (approved):
LGTM 🌊

Issue #722, opened 2024-03-19
Not all payments were made from the selected channels when executing loop out


uheqiang:
Not all payments were made from the selected channel when executing loop out. This makes it inconvenient for us users to control the selection of channels and track the income and expenditure of channel assets. I hope you can evaluate this issue. Thank you.


**【1】LN has two channels:**  
```
"chan_id": "2839500873376923649",
"chan_id": "2839628416710541312"
```

**【2】Loop out params:** 
```
{
    "max_swap_fee": 2500,
    "outgoing_chan_set": [
        "2839628416710541312"
    ],
    "max_prepay_routing_fee": 1000,
    "max_swap_routing_fee": 1000,
    "max_prepay_amt": 35000,
    "swap_publication_deadline": 1710845991,
    "amt": 250000,
    "dest": "tb1qkljku7hjfnec37lrth38x44mkd0umdx4hl33pd",
    "max_miner_fee": 1000
}
```

**【3】Loop result**
```
{
    "code": "200",
    "success": true,
    "msg": "ok",
    "data": {
        "amt": "250000",
        "id": "2fd6feebe4196544a73eab4661b339bdf8ad1e05f89ca66268e780d1c253e6e4",
        "id_bytes": "L9b+6+QZZUSnPqtGYbM5vfitHgX4nKZiaOeA0cJT5uQ=",
        "type": "LOOP_OUT",
        "state": "SUCCESS",
        "failure_reason": "FAILURE_REASON_NONE",
        "initiation_time": "1710845996063412194",
        "last_update_time": "1710847271336933349",
        "htlc_address": "tb1pryjjz8t7nlplj3qjmp0gzd7wm06ghn3pehhdkh523unjvfhpgeuq5ar9c0",
        "htlc_address_p2wsh": "tb1pryjjz8t7nlplj3qjmp0gzd7wm06ghn3pehhdkh523unjvfhpgeuq5ar9c0",
        "htlc_address_p2tr": "",
        "cost_server": "654",
        "cost_onchain": "121",
        "cost_offchain": "252",
        "last_hop": "",
        "outgoing_chan_set": [
            "2839628416710541312"
        ],
        "label": ""
    }
}
```

**【3】Loop logs**
```
2024-03-19 10:59:55.175 [INF] LOOPD: Loop out request received                                                                                                  
[...]

PR #727, opened 2024-04-15
Send loop out prepay over selected outgoing chan set

hieblmi:
Fixes https://github.com/lightninglabs/loop/issues/722.


bhandras (approved):
Nice catch! LGTM 🥇

hieblmi:
Tested this on regtest successfully.

PR #730, opened 2024-04-17
multi: replace occurrences of "LSAT" to "L402"

starius:
# Description

Aperture was updated to include https://github.com/lightninglabs/aperture/pull/135 . Occurrences of "LSAT" in the loop's code were replaced with "L402".

This PR depends on https://github.com/lightninglabs/aperture/pull/135 . Commit "update aperture to include lsat to l402 renaming" should be updated after https://github.com/lightninglabs/aperture/pull/135 is merged.

Some unrelated changes:

  * looprpc,swapserverrpc: update Go image to 1.19.10 (needed for the following item)
  * go: bump protobuf to v1.33.0-hex-display (needed to update to the recent aperture version)
  * multi: apply make fmt

#### Pull Request Checklist

This is a breaking change!

In loopd.conf file `maxlsatcost` and `maxlsatfee` were renamed to `maxl402cost` and `maxl402fee` accordingly. If they have been changed locally, the file has to be updated for loopd to recognize the options.

The path in looprpc "/v1/lsat/tokens" was renamed to "/v1/l402/tokens" and the corresponding method was renamed from `GetLsatTokens` to `GetL402Tokens`. Update `loop` and `loopd` simultaneously otherwise this RPC won't work.


hieblmi (approved):
Thank you very much for the PR, tACK from my side. I tested your changes on my static address branch which fetches an L402 from a local instance. Don't see any issues. The only remaining `lsat` references in code are in release notes, so LGTM

hieblmi:
On another note, we should check if that these changes don't break the server.

looprpc/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16.3-buster
+FROM golang:1.19.10-buster
```
guggero, line 1: Here's an example for the changes required to the `protobuf-compiler` and `clang-format` versions if you wanted to go to a `-bookworm` (Go 1.2x) version: https://github.com/lightninglabs/taproot-assets/blob/main/taprpc/Dockerfile#L1
It's not too involved, so I would suggest doing it.

go.mod
```diff
@@ -204,4 +204,7 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 
 replace github.com/lightninglabs/loop/swapserverrpc => ./swapserverrpc
 
+// Temporary added until https://github.com/lightninglabs/aperture/pull/135 is merged.
+replace github.com/lightninglabs/aperture => github.com/starius/aperture v0.3.2-rc1-beta
```
guggero, line 208: Reminder to not merge this PR until the dependent PR was merged.

loopd/config.go
```diff
@@ -160,8 +160,8 @@ type Config struct {
 	MaxLogFileSize int    `long:"maxlogfilesize" description:"Maximum logfile size in MB."`
 
 	DebugLevel  string `long:"debuglevel" description:"Logging level for all subsystems {trace, debug, info, warn, error, critical} -- You may also specify <subsystem>=<level>,<subsystem2>=<level>,... to set the log level for individual subsystems -- Use show to list available subsystems"`
-	MaxLSATCost uint32 `long:"maxlsatcost" description:"Maximum cost in satoshis that loopd is going to pay for an LSAT token automatically. Does not include routing fees."`
-	MaxLSATFee  uint32 `long:"maxlsatfee" description:"Maximum routing fee in satoshis that we are willing to pay while paying for an LSAT token."`
+	MaxL402Cost uint32 `long:"maxl402cost" description:"Maximum cost in satoshis that loopd is going to pay for an L402 token automatically. Does not include routing fees."`
```
guggero, line 163: This might cause users that have the old flag set to not be able to start their daemon after upgrading. Should we keep the old flags but set them to `hidden:true` and log a deprecation warning if they are set? Then remove the actual flags in a later release?

looprpc/client.yaml
```diff
@@ -26,8 +26,8 @@ http:
       get: "/v1/loop/in/probe/{amt}"
     - selector: looprpc.SwapClient.GetInfo
       get: "/v1/loop/info"
-    - selector: looprpc.SwapClient.GetLsatTokens
-      get: "/v1/lsat/tokens"
+    - selector: looprpc.SwapClient.GetL402Tokens
+      get: "/v1/l402/tokens"
```
guggero, line 30: Should we keep the old mapping (can have `additional_bindings` to map multiple URIs to the same method, see https://github.com/lightninglabs/taproot-assets/blob/main/taprpc/taprootassets.yaml#L66 for example) to not break any client applications?

release_notes.md
```diff
@@ -18,6 +18,14 @@ This file tracks release notes for the loop client.
 
 #### Breaking Changes
 
+In loopd.conf file `maxlsatcost` and `maxlsatfee` were renamed to `maxl402cost`
+and `maxl402fee` accordingly. If they have been changed locally, the file has
+to be updated for loopd to recognize the options.
+
+The path in looprpc "/v1/lsat/tokens" was renamed to "/v1/l402/tokens" and
```
guggero, line 25: Hmm, okay, so we warn about breaking changes here. So I guess it's up to the team to decide whether that's enough or we want to keep backward compatibility as suggested in the comments above.

guggero (commented):
Code looks good to me. See inline discussion about potential backward compatibility fixes.

Going to approve once dependent PR is merged to avoid merging prematurely.

looprpc/Dockerfile
starius: Fixed

go.mod
starius: Fixed

loopd/config.go
starius: The flags were re-added in commit "loopd: recorgnize maxlsatcost and maxlsatfee flags".
I ran loopd locally to verify that if both old and new variants of an option are used simultaneously, it fails to start; and if only the old variant is provided, it starts with a deprecation warning.

looprpc/client.yaml
starius: I added `additional_bindings`.

But it does not provide backward compatibility for GRPC. When I run new loopd with old loop, I get the following error, trying to to run `loop listauth`:
```
[loop] rpc error: code = Unimplemented desc = unknown method GetLsatTokens for service looprpc.SwapClient
```

release_notes.md
starius: I updated the release notes. Now the only broken scenario that I found is `loop listauth` if the `loop` binary is from a previous release, see above.

swapserverrpc/Dockerfile
```diff
@@ -1,21 +1,25 @@
-FROM golang:1.16.3-buster
+FROM golang:1.21.9-bookworm
```
guggero, line 1: Reminder to push a new tag for the `swapserverrpc` module after merging this PR.

looprpc/client.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
+// 	protoc-gen-go v1.33.0
```
guggero, line 3: micro-nit (non-blocking): This should probably go into the previous commit.

loopd/perms/perms.go
```diff
@@ -69,7 +69,7 @@ var RequiredPermissions = map[string][]bakery.Op{
 		Entity: "loop",
 		Action: "in",
 	}},
-	"/looprpc.SwapClient/GetLsatTokens": {{
+	"/looprpc.SwapClient/GetL402Tokens": {{
```
guggero, line 72: Hmm, I didn't really think about this in my last review. But I think this might break some client applications (like Terminal Web, RTL or others that have Loop integrated) that rely on the old RPC name. To avoid headaches, I recommend to not rename this. Maybe we can add a new RPC and declare this one as deprecated, then remove it in a couple of releases.

loopd/utils.go
```diff
@@ -34,6 +52,17 @@ func getClient(cfg *Config, swapDb loopdb.SwapStore,
 		MaxPaymentRetries:   cfg.MaxPaymentRetries,
 	}
 
+	if cfg.MaxL402Cost == defaultCost && cfg.MaxLSATCost != 0 {
```
guggero, line 55: :+1:

looprpc/client.yaml
guggero: Yeah, this might be a problem for some web UIs that use the RPC method directly, like Terminal Web (see my comment above).

guggero (commented):
Code looks great now, thanks for the updates!
A remaining question around backward compatibility of the RPC interface. Normally we don't break backward compatibility for at least two major releases (at least that's the idea in `lnd` but in practice we end up rarely removing/breaking anything at all)...

looprpc/client.pb.go
starius: This is actually updated in this commit.

loopd/perms/perms.go
starius: I added GetLsatTokens back as a wrapper. Updated release notes file.

Tested with old `loop` and new `loopd`:

```
$ loop listauth
[]
```

Loopd log:

```
2024-04-25 13:52:27.367 [WRN] LOOPD: Received deprecated call GetLsatTokens. Please update the client software. Calling GetL402Tokens now.
```

looprpc/client.proto
```diff
@@ -80,6 +80,15 @@ service SwapClient {
     */
     rpc GetL402Tokens (TokensRequest) returns (TokensResponse);
 
+    /* loop: `listauth`
```
guggero, line 83: I think we can remove the `loop: listauth` part, as that's for the API docs generator only, which should use the new RPC.

guggero (approved):
Nice, LGTM :tada:

looprpc/client.proto
starius: Fixed

PR #736, opened 2024-04-26
daemon: fix wrapped errors and typos

hieblmi:
Some minor chore.

### DIFF
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -51,7 +51,7 @@ type ListenerCfg struct {
 	// on the passed TLS configuration.
 	restListener func(*tls.Config) (net.Listener, error)
 
-	// getLnd returns a grpc connection to an lnd instance.
+	// getLnd returns a grpc connection to a lnd instance.
 	getLnd func(lndclient.Network, *lndConfig) (*lndclient.GrpcLndServices,
 		error)
 }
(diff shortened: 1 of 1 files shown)


loopd/daemon.go
```diff
@@ -343,7 +343,7 @@ func (d *Daemon) startWebServers() error {
 
 		log.Infof("RPC server listening on %s", d.grpcListener.Addr())
 		err = d.grpcServer.Serve(d.grpcListener)
-		if err != nil && err != grpc.ErrServerStopped {
+		if err != nil && !errors.Is(grpc.ErrServerStopped, err) {
```
bhandras, line 346: nit: I think you meant `!errors.Is(err, grpc.ErrServerStopped)`.

bhandras (approved):

loopd/daemon.go
```diff
@@ -324,7 +324,7 @@ func (d *Daemon) startWebServers() error {
 			err := d.restServer.Serve(d.restListener)
 			// ErrServerClosed is always returned when the proxy is
 			// shut down, so don't log it.
-			if err != nil && err != http.ErrServerClosed {
+			if err != nil && !errors.Is(http.ErrServerClosed, err) {
```
starius, line 327: Isn't the order of arguments reversed?

hieblmi: fixed

PR #741, opened 2024-05-07
fsm: add WaitForStateAsync to the cached observer

bhandras:
By adding WaitForStateAsync to the observer we can always observe state changes in an atomic way without relying on the observer's internal cache.

#### Pull Request Checklist

### DIFF
--- a/fsm/observer.go
+++ b/fsm/observer.go
@@ -100,12 +100,11 @@ func WithAbortEarlyOnErrorOption() WaitForStateOption {
 // the given duration before checking the state. This is useful if the
 // function is called immediately after sending an event to the state machine
 // and the state machine needs some time to process the event.
-func (s *CachedObserver) WaitForState(ctx context.Context,
+func (c *CachedObserver) WaitForState(ctx context.Context,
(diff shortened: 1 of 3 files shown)


fsm/observer.go
```diff
@@ -120,61 +119,80 @@ func (s *CachedObserver) WaitForState(ctx context.Context,
...
+		return err
+	}
+}
+
+// WaitForStateAsync waits asynchronously until the passed context is canceled
+// or the expected state is reached. The function returns a channel that will
+// receive an error if the expected state is reached or an error occurred. If
+// the context is canceled before the expected state is reached, the channel
+// will receive an ErrWaitingForStateTimeout error.
+func (s *CachedObserver) WaitForStateAsync(ctx context.Context, state StateType,
```
hieblmi, line 143: nit: rename receiver `c`.

bhandras: done

hieblmi (approved):
tACK, tested these changes to the fsm with the static address withdrawal flow without issues.
Thanks for the explanation of the deadlock scenario.

PR #743, opened 2024-05-10
loopout: configurable payment timeout for off-chain payments

bhandras:
With this client side RPC option the client will be able to set a payment timeout for individual off-chain payments. This applies both to the prepayment and the swap payment. With the swap payment, we may try to pay the swap invoice multiple times, but each attempt will use this configurable timeout.

#### Pull Request Checklist


loopdb/sql_store.go
```diff
@@ -443,6 +443,7 @@ func loopOutToInsertArgs(hash lntypes.Hash,
 		PrepayInvoice:       loopOut.PrepayInvoice,
 		MaxPrepayRoutingFee: int64(loopOut.MaxPrepayRoutingFee),
 		PublicationDeadline: loopOut.SwapPublicationDeadline.UTC(),
+		PaymentTimeout:      int32(loopOut.PaymentTimeout.Seconds()),
```
hieblmi, line 446: what happens for older entries? I guess the timeout is `0` then?

loopout.go
```diff
@@ -647,8 +650,8 @@ func (s *loopOutSwap) payInvoice(ctx context.Context, invoice string,
 		var result paymentResult
 
 		status, err := s.payInvoiceAsync(
```
hieblmi, line 652: If we follow `s.payInvoiceAsync` and the `paymentTimeout` it ends up in `awaitSendPayment`. The context here is passed to `s.lnd.Router.SendPayment(ctx, *req)`.
Question: Does it make sense to cancel this context once the `paymentTimeout` is reached, given that we think about making this context cancelable to properly shutdown `lnd`'s payment loop? The related PR is here: https://github.com/lightningnetwork/lnd/pull/8734

hieblmi (commented):
Very nice feature! I have a question regarding the cancellation of the payment context in addition to the `paymentTimeout`. 
Since we are thinking about properly exiting `lnd`'s payment loop in this PR https://github.com/lightningnetwork/lnd/pull/8734 it might make sense to also make the context of `awaitSendPayment` bound to the timeout.

loopout.go
bhandras: Yeah I think this could be a good idea, but given that not all users switch to the latest LND release, we need to still rely on the request's timeout. For now I'll keep it as is and we can add it in a separate PR once we bump the minimum LND version.

loopd/swapclient_server.go
```diff
@@ -101,6 +101,13 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 
 	log.Infof("Loop out request received")
 
+	paymentTimeout := time.Duration(in.PaymentTimeout) * time.Second
+	// Make sure we don't exceed the total allowed payment timeout.
+	if paymentTimeout > s.config.TotalPaymentTimeout {
+		return nil, fmt.Errorf("payment timeout exceeds maximum "+
+			"allowed timeout of %v", s.config.TotalPaymentTimeout)
```
starius, line 108: I propose to include `paymentTimeout` into the error message:
```
return nil, fmt.Errorf("payment timeout %v exceeds maximum " ...
```

loopd/swapclient_server.go
```diff
@@ -101,6 +101,13 @@ func (s *swapClientServer) LoopOut(ctx context.Context,
 
 	log.Infof("Loop out request received")
 
+	paymentTimeout := time.Duration(in.PaymentTimeout) * time.Second
+	// Make sure we don't exceed the total allowed payment timeout.
+	if paymentTimeout > s.config.TotalPaymentTimeout {
```
starius, line 106: I propose to add a comment that `in.PaymentTimeout` is uint32, so it is >= 0 (no need to check that separately).

loopout.go
```diff
@@ -705,8 +709,27 @@ func (s *loopOutSwap) payInvoiceAsync(ctx context.Context,
...
+				time.Duration(maxRetries)
+		}
+
+		// If the payment timeout is too long, we need to adjust the
+		// number of retries to ensure we don't exceed the total
+		// payment timeout.
+		if paymentTimeout*time.Duration(maxRetries) >
+			totalPaymentTimeout {
+
+			maxRetries = int(totalPaymentTimeout / paymentTimeout)
```
starius, line 726: I propose to add an info or warning here.

starius, line 726: If totalPaymentTimeout is not divisible by paymentTimeout, the remainder of totalPaymentTimeout will be removed. Does it make sense for the last attempt to use a shorter timeout instead?

cmd/loop/loopout.go
```diff
@@ -235,6 +243,15 @@ func loopOut(ctx *cli.Context) error {
 		}
 	}
 
+	paymentTimeout := 0
+	if ctx.IsSet("payment_timeout") {
+		paymentTimeout = int(ctx.Duration("payment_timeout").Seconds())
+		if paymentTimeout < 0 {
```
starius, line 249: If user passes `--payment_timeout=0s`, it should also fail, I think. So this check should be `if paymentTimeout <= 0`

cmd/loop/loopout.go
```diff
@@ -235,6 +243,15 @@ func loopOut(ctx *cli.Context) error {
 		}
 	}
 
+	paymentTimeout := 0
+	if ctx.IsSet("payment_timeout") {
+		paymentTimeout = int(ctx.Duration("payment_timeout").Seconds())
```
starius, line 248: I propose to add a check that the value is divisible by a second. The current implementation treats `--payment_timeout=99.999s` as `--payment_timeout=99s` silently. I think it should crash instead, complaining that sub-second accuracy is not supported.

starius, line 248: I propose to check that `paymentTimeout < math.MaxUint32` just in case, to make sure that it does not overflow in the conversion to `uint32` below. It should be always satisfied, because time.Duration does not support such large values.

starius (approved):
LGTM 🎉

Left few small comments.

hieblmi (approved):
LGTM 💾

PR #754, opened 2024-05-24
sweepbatcher: do not fail on restoring empty batches

bhandras:
Previously stroing an empty batch would make the batcher fail to start as spinning up a restored batch assumes that there's a primary sweep added already. As there's no point in spinning up such batch we can just skip over it.
Furthermore we'll ensure that we won't try to ever publish an empty batch to avoid setting the fee rate too early.

#### Pull Request Checklist


sweepbatcher/sweep_batcher.go
```diff
@@ -407,23 +416,16 @@ func (b *Batcher) spinUpBatch(ctx context.Context) (*batch, error) {
...
-	}
-
 	dbSweeps, err := b.store.FetchBatchSweeps(ctx, batch.id)
 	if err != nil {
 		return err
 	}
 
 	if len(dbSweeps) == 0 {
-		return fmt.Errorf("batch %d has no sweeps", batch.id)
+		log.Infof("skipping restored batch %d as it has no sweeps",
```
GeorgeTsagk, line 425: maybe also delete from storage

sweepbatcher/sweep_batcher.go
```diff
@@ -321,7 +330,7 @@ func (b *Batcher) handleSweep(ctx context.Context, sweep *sweep,
 	// If one of the batches accepts the sweep, we provide it to that batch.
 	for _, batch := range b.batches {
 		accepted, err := batch.addSweep(ctx, sweep)
-		if err != nil && err != ErrBatchShuttingDown {
+		if err != nil {
```
GeorgeTsagk, line 333: we don't want to crash here, as it's possible that the batch confirmed by the time the sweep tries to enter it, returning an `ErrBatchShuttingDown`

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1029,3 +1030,38 @@ func TestGetFeePortionForSweep(t *testing.T) {
...
+		defer wg.Done()
+		err = batcher.Run(ctx)
+	}()
+
+	// Wait for the batcher to be initialized.
+	<-batcher.initDone
+
+	// Now make it quit by canceling the context.
+	cancel()
+	wg.Wait()
```
GeorgeTsagk, line 1064: could go a bit further

add a sweep and verify that we now have 1 batch in memory, and 2 in storage

sweepbatcher/sweep_batcher.go
GeorgeTsagk: it's not critical, but no need to cause an all-around system restart 

could perhaps add some simple coverage for this, not sure how involved the unit test would be though

sweepbatcher/sweep_batcher_test.go
bhandras: done

sweepbatcher/sweep_batch.go
```diff
@@ -493,7 +493,7 @@ func (b *batch) Run(ctx context.Context) error {
 			b.currentHeight = height
 
 		case <-timerChan:
-			if b.state == Open {
+			if b.state == Open && len(b.sweeps) > 0 {
```
hieblmi, line 496: maybe we can move the length check inside `publish` so that other calls to it also check for the length?

sweepbatcher/sweep_batcher.go
```diff
@@ -458,6 +465,11 @@ func (b *Batcher) spinUpBatchFromDB(ctx context.Context, batch *batch) error {
 		log:              batchPrefixLogger(fmt.Sprintf("%d", batch.id)),
 	}
 
+	cfg := batchConfig{
+		maxTimeoutDistance: batch.cfg.maxTimeoutDistance,
+		batchConfTarget:    defaultBatchConfTarget,
```
starius, line 470: Why setting `defaultBatchConfTarget` is needed here? If it is used, it means, that `confTarget`s of sweeps were not taken into account.

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1029,3 +1030,83 @@ func TestGetFeePortionForSweep(t *testing.T) {
...
+
+	batcher := NewBatcher(lnd.WalletKit, lnd.ChainNotifier, lnd.Signer,
+		testMuSig2SignSweep, nil, lnd.ChainParams, batcherStore, store)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+		err = batcher.Run(ctx)
```
starius, line 1056: Let's name it to `runErr` just in case another `err` is created between this line and the check of the error in the end.

sweepbatcher/sweep_batcher.go
starius: The source of confTarget is `fetchSweep` method. I think, after resuming from DB we should make sure that confTarget is filled from `fetchSweep` before it is used to calculate fee rate (i.e. before `publish`). Then we can remove `const defaultBatchConfTarget`.

GeorgeTsagk: We update this when the first sweep comes in, see [here](https://github.com/lightninglabs/loop/blob/master/sweepbatcher/sweep_batch.go#L408)

sweepbatcher/sweep_batcher.go
starius: In the previous loop (trying to addSweep (update) to the batch to which it already belongs), the error from `batch.addSweep` is checked with regular `err != nil` check. Should the checks of errors from `batch.addSweep` in the two loops be the same?

Another question about that loop. Can we return earlier there, if the sweep was added successfuly?

```diff
diff --git a/sweepbatcher/sweep_batcher.go b/sweepbatcher/sweep_batcher.go
index 0251101..6019f6c 100644
--- a/sweepbatcher/sweep_batcher.go
+++ b/sweepbatcher/sweep_batcher.go
@@ -388,6 +388,9 @@ func (b *Batcher) handleSweep(ctx context.Context, sweep *sweep,
                                        "accepted by batch %d", sweep.swapHash[:6],
                                        batch.id)
                        }
+
+                       // The sweep was updated in the batch, our job is done.
+                       return nil
                }
        }
 ```

File: sweepbatcher/sweep_batch.go (hunk 4 above)
Comment (bhandras) [reply sweepbatcher/sweep_batch.go]: Good idea, done.

File: sweepbatcher/sweep_batcher.go (hunk 5 above)
Comment (bhandras) [reply sweepbatcher/sweep_batcher.go:473]: Note that this is just a reorder, no added code.

File: sweepbatcher/sweep_batcher.go (hunk 2 above)
Comment (bhandras) [reply sweepbatcher/sweep_batcher.go:334]: Note that there's a check below for `accepted` and if that's true we return early.

File: sweepbatcher/sweep_batcher_test.go (hunk 6 above)
Comment (bhandras) [reply sweepbatcher/sweep_batcher_test.go]: done

File: sweepbatcher/sweep_batcher.go (hunk 5 above)
Comment (GeorgeTsagk) [reply sweepbatcher/sweep_batcher.go:473]: yeah, ideally let's not introduce unwanted changes

File: sweepbatcher/sweep_batcher.go (hunk 1 above)
Comment (bhandras) [reply sweepbatcher/sweep_batcher.go:426]: done

File: sweepbatcher/sweep_batcher_test.go (hunk 3 above)
Comment (bhandras) [reply sweepbatcher/sweep_batcher_test.go:1111]: I've changed the test a bit since the dormant batch is now dropped from the db.

File: sweepbatcher/store.go [hunk 7]
```diff
@@ -108,6 +111,12 @@ func (s *SQLStore) InsertSweepBatch(ctx context.Context, batch *dbBatch) (int32,
 	return s.baseDb.InsertBatch(ctx, batchToInsertArgs(*batch))
 }
 
+// DropBatch drops a batch from the database. Note that we only use this call
+// for batches that have no sweeps and so we'd not be able to resume.
+func (s *SQLStore) DropBatch(ctx context.Context, id int32) error {
```
starius, line 116: I propose to add a check that the number of sweep in the batch is 0. If the batch is not empty, return an error.

sweepbatcher/sweep_batcher.go
starius: Should this check also be `if err != nil && !errors.Is(err, ErrBatchShuttingDown) {`?

starius (approved):
LGTM 🏆

sweepbatcher/store.go
bhandras: Good idea, added!

sweepbatcher/sweep_batcher.go
bhandras: Added the error check to both code paths.

PR #755, opened 2024-05-24
sweepbatcher: small refactorings

starius:
These changes are mostly cosmetic and they don't change the behaviour.

#### Pull Request Checklist


sweepbatcher/sweep_batcher.go
```diff
@@ -324,6 +331,9 @@ func (b *Batcher) handleSweep(ctx context.Context, sweep *sweep,
 					"accepted by batch %d", sweep.swapHash[:6],
 					batch.id)
 			}
+
+			// The sweep was updated in the batch, our job is done.
+			return nil
```
hieblmi, line 336: Isn't this a bug that should be addressed in a separate commit?
Is it right that without this change, we might add the sweep to another batch in the loop below even though it already existed in another batch?

starius: I think it is possible. I'll try to create a unit test to trigger the bug condition. I think it worth doing it in a separate PR.

hieblmi: Yeah make sense. IIUC it might be fine if a sweep exists in multiple batches as a sweep spend would notify all batches, cc @GeorgeTsagk

sweepbatcher/sweep_batcher.go
```diff
@@ -82,6 +82,14 @@ type BatcherStore interface {
 	TotalSweptAmount(ctx context.Context, id int32) (btcutil.Amount, error)
 }
 
+// LoopOutFetcher is used to load LoopOut swaps from the database.
+// It is implemented by loopdb.SwapStore.
+type LoopOutFetcher interface {
```
hieblmi, line 87: 👍

hieblmi (approved):

bhandras (approved):
LGTM, nice 🎉

GeorgeTsagk (approved):
:broom: :ant:

starius:
Rebased

PR #759, opened 2024-05-29
sweepbatcher: exit early in handleSweep

starius:
Avoid adding a sweep to two batches:
 - existing in the first `for` loop,
 - any other batch in the second `for`

If the sweep was successfully updated in the batch, no need to try to add it to all other batches. So `return` in the first loop in case of acceptance.

Added a test reproducing adding a sweep to both batches without this change.

Also fixed some long lines of code (>80 chars) in a separate commit.

This PR was moved from https://github.com/lightninglabs/loop/pull/755#discussion_r1614040088

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/log.go
+++ b/sweepbatcher/log.go
@@ -17,7 +17,8 @@ func init() {
 	UseLogger(build.NewSubLogger("SWEEP", nil))
 }
 
-// batchPrefixLogger returns a logger that prefixes all log messages with the ID.
+// batchPrefixLogger returns a logger that prefixes all log messages with
+// the ID.
 func batchPrefixLogger(batchID string) btclog.Logger {
 	return build.NewPrefixLog(fmt.Sprintf("[Batch %s]", batchID), log)
 }
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -592,9 +592,11 @@ func (b *batch) publishBatch(ctx context.Context) (btcutil.Amount, error) {
 	batchTx.LockTime = uint32(b.currentHeight)
 
 	var (
-		batchAmt     btcutil.Amount
-		prevOuts     = make([]*wire.TxOut, 0, len(b.sweeps))
-		signDescs    = make([]*lndclient.SignDescriptor, 0, len(b.sweeps))
+		batchAmt  btcutil.Amount
+		prevOuts  = make([]*wire.TxOut, 0, len(b.sweeps))
+		signDescs = make(
+			[]*lndclient.SignDescriptor, 0, len(b.sweeps),
+		)
 		sweeps       = make([]sweep, 0, len(b.sweeps))
 		fee          btcutil.Amount
 		inputCounter int
--- a/sweepbatcher/sweep_batcher_test.go
+++ b/sweepbatcher/sweep_batcher_test.go
@@ -182,7 +182,8 @@ func TestSweepBatcherBatchCreation(t *testing.T) {
 	<-lnd.RegisterSpendChannel
 
 	require.Eventually(t, func() bool {
-		// Verify that each batch has the correct number of sweeps in it.
+		// Verify that each batch has the correct number of sweeps
+		// in it.
 		for _, batch := range batcher.batches {
 			switch batch.primarySweepID {
 			case sweepReq1.SwapHash:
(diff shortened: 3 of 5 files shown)


sweepbatcher/sweep_batcher_test.go
```diff
@@ -1111,3 +1111,191 @@ func TestRestoringEmptyBatch(t *testing.T) {
...
+		runErr = batcher.Run(ctx)
+	}()
+
+	// Wait for the batcher to be initialized.
+	<-batcher.initDone
+
+	const shortCltv = 111
+	const longCltv = 111 + defaultMaxTimeoutDistance + 6
+
+	// Create two sweep request with CltvExpiry distant from each other
```
hieblmi, line 1181: nit: requests

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1111,3 +1111,191 @@ func TestRestoringEmptyBatch(t *testing.T) {
...
+		}
+
+		// Make sure the second batch has the second sweep.
+		sweep2, has := batches[1].sweeps[sweepReq2.SwapHash]
+		if !has {
+			return false
+		}
+
+		// Make sure the second sweep's timeout has been updated.
+		if sweep2.timeout != shortCltv {
```
hieblmi, line 1288: Is it correct that without this change, `sweep2` would've existed in two batches because we'd try and add it to all existing batches _and_ because the `shortCltv` on `sweep2` now fits into the first batch?
@bhandras(and @GeorgeTsagk if bandwidth), did we want that behavior? E.g. Would the sweep be removed from the second batch if the first batch(with duplicated `sweep2`) would get confirmed? 

In other words, is the original behavior of adding the sweep to multiple batches actually intended?

sweepbatcher/log.go
```diff
@@ -17,7 +17,8 @@ func init() {
 	UseLogger(build.NewSubLogger("SWEEP", nil))
 }
 
-// batchPrefixLogger returns a logger that prefixes all log messages with the ID.
+// batchPrefixLogger returns a logger that prefixes all log messages with
```
hieblmi, line 20: ultra nit: do you have the line break set to 80 chars? The `the` of `the ID` technically fits into the first line. Also seeing early line breaks elsewhere.

sweepbatcher/sweep_batch.go
```diff
@@ -591,11 +591,11 @@ func (b *batch) publishBatch(ctx context.Context) (btcutil.Amount, error) {
 	batchTx := wire.NewMsgTx(2)
 	batchTx.LockTime = uint32(b.currentHeight)
 
+	prevOuts := make([]*wire.TxOut, 0, len(b.sweeps))
```
hieblmi, line 594: could just do
```
var (
		batchAmt     btcutil.Amount
		fee          btcutil.Amount
		inputCounter int
		addrOverride bool
		prevOuts     = make([]*wire.TxOut, 0, len(b.sweeps))
		sweeps       = make([]sweep, 0, len(b.sweeps))
		signDescs    = make(
			[]*lndclient.SignDescriptor, 0, len(b.sweeps),
		)
	)
```

hieblmi (commented):
Very nice test 👌 That got me thinking if the original behavior is actually intended so would be good to double check that.

sweepbatcher/log.go
starius: The length of "// batchPrefixLogger returns a logger that prefixes all log messages with the ID." is 81 character. It has a period in the end.

I have the following settings in vim to highlight with red last few characters up until 80th. And if something does not fit into "red zone", it means the line is tool long.

```
match ErrorMsg /\%<81v.\%>77v/
```

sweepbatcher/sweep_batcher_test.go
starius: Fixed.

sweepbatcher/sweep_batch.go

sweepbatcher/sweep_batcher_test.go
starius: Also, one possible solution is to only keep it in the first batch and remove from the second batch. If a sweep is kept in two batches, then they will interfere with each other RBF. RBF of one batch may fail, because it double spends the same UTXO and does not take fees of another batch into account. It may end up in paying higher fee-rate.

hieblmi: RBF conflict is a great point, so yes we'd want the sweep to only exist in one batch.

sweepbatcher/log.go
hieblmi: Ah ok, so this
```
// batchPrefixLogger returns a logger that prefixes all log messages with the 
// ID.
```
would be valid according to your settings?

hieblmi (approved):

PR #760, opened 2024-05-29
sweepbatcher: load from DB preserves confTarget

starius:
After loading from DB the value of `batchConfTarget` used to be set to default (`defaultBatchConfTarget` = 12) which
could in theory affect the fee rate if `updateRbfRate()` and `publish()` were not called before the batch was saved. It was unlikely, but possible in theory. Also it was error-prone to have wrong value in that field, in case it was used for something in the future.

Constant `defaultBatchConfTarget` was removed, it is never used before is overwritten in any non-buggy scenario. Added a check that `batchConfTarget != 0` (if 0, `updateRbfRate` ~fails~ prints a warning).

Added a test checking that batch confTarget survives DB store-load cycle.
To do it, `StoreMock` was updated to attach `LoopOut` in `FetchBatchSweeps` method (otherwise `convertSweep` crashes). Now `sweepbatcher.NewStoreMock` has an argument (loopdb).

Re-added commit [fix too long lines](https://github.com/lightninglabs/loop/commit/8d6e86ba983daac7bb784620f20675d9d405c5cc) from https://github.com/lightninglabs/loop/pull/759 since it can take longer to review.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -316,7 +316,22 @@ func NewBatch(cfg batchConfig, bk batchKit) *batch {
 }
 
 // NewBatchFromDB creates a new batch that already existed in storage.
-func NewBatchFromDB(cfg batchConfig, bk batchKit) *batch {
+func NewBatchFromDB(cfg batchConfig, bk batchKit) (*batch, error) {
+	// Make sure the batch is not empty.
+	if len(bk.sweeps) == 0 {
+		// This should never happen, as this precondition is already
+		// ensured in spinUpBatchFromDB.
+		return nil, fmt.Errorf("empty batch is not allowed")
+	}
+
+	// Assign batchConfTarget to primary sweep's confTarget.
+	for _, sweep := range bk.sweeps {
+		if sweep.swapHash == bk.primaryID {
+			cfg.batchConfTarget = sweep.confTarget
+			break
+		}
+	}
+
 	return &batch{
 		id:               bk.id,
 		state:            bk.state,
@@ -343,7 +358,7 @@ func NewBatchFromDB(cfg batchConfig, bk batchKit) *batch {
 		store:            bk.store,
 		log:              bk.log,
 		cfg:              &cfg,
-	}
+	}, nil
 }
 
 // addSweep tries to add a sweep to the batch. If this is the first sweep being
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -35,10 +35,6 @@ const (
 	// fee rate is increased when an rbf is attempted.
 	defaultFeeRateStep = chainfee.SatPerKWeight(100)
 
-	// defaultBatchConfTarget is the default confirmation target of the
-	// batch transaction.
-	defaultBatchConfTarget = 12
-
 	// batchConfHeight is the default confirmation height of the batch
 	// transaction.
 	batchConfHeight = 3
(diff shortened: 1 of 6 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -316,7 +316,20 @@ func NewBatch(cfg batchConfig, bk batchKit) *batch {
 }
 
 // NewBatchFromDB creates a new batch that already existed in storage.
-func NewBatchFromDB(cfg batchConfig, bk batchKit) *batch {
+func NewBatchFromDB(cfg batchConfig, bk batchKit) (*batch, error) {
+	// Make sure the batch is not empty. We have checked this in
+	// spinUpBatchFromDB before calling NewBatchFromDB.
+	if len(bk.sweeps) == 0 {
```
bhandras, line 318: nit: I'd add a comment emphasizing that this should never happen as this precondition is already ensured in `spinUpBatchFromDB`.

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1115,3 +1115,119 @@ func TestRestoringEmptyBatch(t *testing.T) {
...
+			len(batcher.batches[0].sweeps) == 1 &&
+			batcher.batches[0].cfg.batchConfTarget == 123
+	}, test.Timeout, eventuallyCheckFrequency)
+
+	// Make sure we have stored the batch.
+	batches, err := batcherStore.FetchUnconfirmedSweepBatches(ctx)
+	require.NoError(t, err)
+	require.Len(t, batches, 1)
+
+	// Now make it quit by canceling the context.
```
bhandras, line 1194: nit: make it quit => make the batcher quit

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1115,3 +1115,119 @@ func TestRestoringEmptyBatch(t *testing.T) {
...
+	cancel()
+	wg.Wait()
+
+	// Make sure the batcher existed without an error.
```
bhandras, line 1198: nit: existed => exited

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1115,3 +1115,119 @@ func TestRestoringEmptyBatch(t *testing.T) {
...
+	require.Equal(t, 123, int(batcher.batches[0].cfg.batchConfTarget))
+
+	// Expect registration for spend notification.
+	<-lnd.RegisterSpendChannel
+
+	// Now make it quit by canceling the context.
+	cancel()
+	wg.Wait()
+
+	// Make sure the batcher existed without an error.
```
bhandras, line 1231: nit: exited

bhandras (approved):
LGTM, thanks for the fix 🥇

sweepbatcher/sweep_batch.go
```diff
@@ -316,7 +312,20 @@ func NewBatch(cfg batchConfig, bk batchKit) *batch {
...
+		return nil, fmt.Errorf("empty batch is not allowed")
+	}
+
+	// Assign batchConfTarget to primary sweep's confTarget.
+	for _, sweep := range bk.sweeps {
+		if sweep.swapHash == bk.primaryID {
+			cfg.batchConfTarget = sweep.confTarget
```
hieblmi, line 325: nit: I think we can break here.

hieblmi (approved):
Thanks for the fix, the commit message makes this review very easy.

sweepbatcher/sweep_batch.go
starius: Fixed

sweepbatcher/sweep_batcher_test.go
starius: Fixed.

sweepbatcher/sweep_batcher_test.go
starius: Fixed.

sweepbatcher/sweep_batcher_test.go
starius: Also changes this comment in previous tests.

sweepbatcher/sweep_batcher_test.go
starius: Fixed

sweepbatcher/sweep_batch.go
starius: Fixed.

PR #778, opened 2024-06-17
sweepbatcher: unit tests with SQL loopdb as well

starius:
This is needed to cover the code of SQLStore with tests.

To achieve compatibility with loopdb (SQLite), the following changes were done:

 - DestAddr is filled to avoid crash in SQL layer
 - Preimage is filled to avoid uniqueness checks by the DB
 - the code working with batch IDs was changed to work correctly when batch_id starts with 1 instead of 0
 - SQL swap store has to have a swap with swap_hash from sweep to satisfy foreign key constraint of the DB

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batcher_test.go
+++ b/sweepbatcher/sweep_batcher_test.go
@@ -3,6 +3,7 @@ package sweepbatcher
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 	"testing"
 	"time"
@@ -30,6 +31,17 @@ const (
 	ntfnBufferSize = 1024
 )
 
+// destAddr is a dummy p2wkh address to use as the destination address for
+// the swaps.
+var destAddr = func() btcutil.Address {
+	p2wkhAddr := "bcrt1qq68r6ff4k4pjx39efs44gcyccf7unqnu5qtjjz"
+	addr, err := btcutil.DecodeAddress(p2wkhAddr, nil)
+	if err != nil {
+		panic(err)
+	}
+	return addr
+}()
+
 func testMuSig2SignSweep(ctx context.Context,
 	protocolVersion loopdb.ProtocolVersion, swapHash lntypes.Hash,
 	paymentAddr [32]byte, nonce []byte, sweepTxPsbt []byte,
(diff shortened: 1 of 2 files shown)


starius:
The test will be fixed after https://github.com/lightninglabs/loop/pull/780 is merged.
It fixes the case of using read only transaction instead of write transaction, which is caught by postgres test.
For some reason sqlite works even without this fix. Maybe it ignore the read only flag on the transaction?

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1538,3 +1725,135 @@ func TestSweepFetcher(t *testing.T) {
...
+	// AssertLoopOutStored asserts that a swap is stored.
+	AssertLoopOutStored()
+}
+
+type loopdbStore struct {
+	loopdb.SwapStore
+}
+
+// AssertLoopOutStored asserts that a sweep is stored.
+func (s *loopdbStore) AssertLoopOutStored() {
```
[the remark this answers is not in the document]
starius: Fixed

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1164,23 +1246,67 @@ func (s *loopStoreMock) putLoopOutSwap(hash lntypes.Hash, out *loopdb.LoopOut) {
...
+		return
+	}
+
+	if _, ok := s.backend.(*loopdb.StoreMock); ok {
+		// Do not create a fake loop in loopdb.StoreMock, because it
+		// blocks on notification channels and this is not needed.
+		return
+	}
+
+	// Put a swap with the same ID to backend store to satisty SQL foreign
```
hieblmi, line 1264: typo: satisfy

sweepbatcher/sweep_batcher_test.go
```diff
@@ -94,6 +119,7 @@ func TestSweepBatcherBatchCreation(t *testing.T) {
 			AmountRequested: 111,
 		},
 
+		DestAddr:    destAddr,
```
hieblmi, line 122: Is `DestAddr` always stored in the DB or only if there's one provided by the client?

sweepbatcher/sweep_batcher_test.go
```diff
@@ -1300,24 +1429,29 @@ func TestHandleSweepTwice(t *testing.T) {
 			return false
 		}
 
+		// Find the batch with largest ID. It must be the second batch.
+		var secondBatch *batch
+		for _, batch := range batches {
+			if secondBatch == nil || batch.id > secondBatch.id {
```
hieblmi, line 1435: interesting way to find it. Little hard to read imo, could we
```
secondBatch := batches[0]
if batches[1].id > batches[0].id {
        secondBatch = batches[1]
}
```

hieblmi (approved):
LGTM 💾

sweepbatcher/sweep_batcher_test.go
starius: If I don't set it in tests, `CreateLoopOut` fails in SQL version of the store, inside function [`loopOutToInsertArgs`](https://github.com/lightninglabs/loop/blob/v0.28.5-beta/loopdb/sql_store.go#L491), which attempts to convert the address to string:

```go
DestAddress:         loopOut.DestAddr.String(),
```

This code assumes that it is always filled. I guess, if the client doesn't fill the field manually when creating a swap, it is filled by the `loopd` software before the swap is written to DB.

sweepbatcher/sweep_batcher_test.go
starius: Variable batches is a map, not a slice, and we don't know the keys (in practice 0,1 or 1,2) so we have to visit all the items and find the one with maximum id.

sweepbatcher/sweep_batcher_test.go
starius: Fixed, thanks!

PR #779, opened 2024-06-17
sweepbatcher: fix race condition when stopping

starius:
The race was detected in CI and locally when running with `-race`. It happened between the following calls:

```
WARNING: DATA RACE
Write at 0x00c0003e6638 by goroutine 1374:
  runtime.racewrite()
      <autogenerated>:1 +0x1e
  github.com/lightninglabs/loop/sweepbatcher.(*batch).Wait()
      sweepbatcher/sweep_batch.go:463 +0x6e
  github.com/lightninglabs/loop/sweepbatcher.(*Batcher).Run.func1()
      sweepbatcher/sweep_batcher.go:272 +0x10e

Previous read at 0x00c0003e6638 by goroutine 1388:
  runtime.raceread()
      <autogenerated>:1 +0x1e
  github.com/lightninglabs/loop/sweepbatcher.(*batch).monitorConfirmations()
      sweepbatcher/sweep_batch.go:1144 +0x285
  github.com/lightninglabs/loop/sweepbatcher.(*batch).handleSpend()
      sweepbatcher/sweep_batch.go:1309 +0x10e4
  github.com/lightninglabs/loop/sweepbatcher.(*batch).Run()
      sweepbatcher/sweep_batch.go:526 +0xb04
  github.com/lightninglabs/loop/sweepbatcher.(*Batcher).spinUpBatch.func1()
      sweepbatcher/sweep_batcher.go:455 +0xbd
```

The race was caused because `wg.Add(1)` and `wg.Wait()` were running from different goroutines (one goroutine was running `batch.Run()` and another - `batcher.Run()`).

To avoid this scenario, `wg.Wait()` call was moved into `batch.Run()` call, so it waits itself for its children goroutines, after which the channel b.finished is closed, and it serves a signal for external waiters (the batcher, calling `batch.Wait()`).

Also the channel batch.stopped was renamed to batch.stopping to better reflect its nature.

Also added TestSweepBatcherCloseDuringAdding to make sure adding a sweep during shutting down does not cause a crash. The test did not catch the originalrace condition.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -193,8 +193,12 @@ type batch struct {
 	// main event loop.
 	callLeave chan struct{}
 
-	// stopped signals that the batch has stopped.
-	stopped chan struct{}
+	// stopping signals that the batch is stopping.
+	stopping chan struct{}
+
+	// finished signals that the batch has stopped and all child goroutines
+	// have finished.
+	finished chan struct{}
 
 	// quit is owned by the parent batcher and signals that the batch must
 	// stop.
@@ -273,7 +277,10 @@ func (b *batch) scheduleNextCall() (func(), error) {
 	case <-b.quit:
 		return func() {}, ErrBatcherShuttingDown
 
-	case <-b.stopped:
+	case <-b.stopping:
+		return func() {}, ErrBatchShuttingDown
+
+	case <-b.finished:
 		return func() {}, ErrBatchShuttingDown
 	}
 
@@ -297,7 +304,8 @@ func NewBatch(cfg batchConfig, bk batchKit) *batch {
 		errChan:          make(chan error, 1),
 		callEnter:        make(chan struct{}),
 		callLeave:        make(chan struct{}),
-		stopped:          make(chan struct{}),
+		stopping:         make(chan struct{}),
+		finished:         make(chan struct{}),
 		quit:             bk.quit,
 		batchTxid:        bk.batchTxid,
 		wallet:           bk.wallet,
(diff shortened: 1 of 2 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -460,16 +466,20 @@ func (b *batch) sweepExists(hash lntypes.Hash) bool {
 // Wait waits for the batch to gracefully stop.
 func (b *batch) Wait() {
 	b.log.Infof("Stopping")
-	b.wg.Wait()
```
bhandras, line 463: I doubt this has caused a race, wait groups are supposed to be used from different goroutines.See also implementation: https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/sync/waitgroup.go;l=43

starius: If Add(1) and Wait() are called in parallel, the outcome for Wait() depends on the order of execution, which is a race condition: it can be either a no-op (if performed before Add), or a wait (if performed after). In practice, this means, that some goroutines of the batch could run away and outlive `batch.Wait()`.

See https://pkg.go.dev/sync#WaitGroup.Add

> If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned.

In the [implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/sync/waitgroup.go;l=56-59) of WaitGroup.Add(1) there is the following part for race detection:

```go
	if race.Enabled && delta > 0 && v == int32(delta) {
		// The first increment must be synchronized with Wait.
		// Need to model this as a read, because there can be
		// several concurrent wg.counter transitions from 0.
		race.Read(unsafe.Pointer(&wg.sema))
	}
```

I think, this `race.Read` caused the crash above.

bhandras: Indeed we can't guarantee the order, so moving the `Wait()` into `Run`'s `defer` can be a good solution, however I think you'll also need to change `scheduleNextCall()` in order to correctly avoid adding a sweep while the batcher is shutting down.

starius: I added
```go
case <-b.finished:
  return func() {}, ErrBatchShuttingDown
```
to `scheduleNextCall()`, however it is not necessary in the current implementation, because channel `finished` is closed after channel `stopping` and there is already a case for `stopping`.

Also added TestSweepBatcherCloseDuringAdding to make sure adding a sweep during shutting down does not cause a crash. The test did not catch the originalrace condition.

bhandras (approved):
LGTM 🥇

PR #780, opened 2024-06-17
sweepbatcher: use write tx in DropBatch

starius:
This fixes a crash on postgres, can be observed in this PR adding sweepbatcher tests with SQL backends: https://github.com/lightninglabs/loop/pull/778

For some reason sqlite works even without this fix. Maybe it ignore the read only flag on the transaction?

#### Pull Request Checklist

### DIFF
--- a/loopdb/sqlite.go
+++ b/loopdb/sqlite.go
@@ -323,11 +323,19 @@ type SqliteTxOptions struct {
 	readOnly bool
 }
 
-// NewSqlReadOpts returns a new KeyStoreTxOptions instance triggers a read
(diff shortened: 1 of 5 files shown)


loopdb/sqlite.go
```diff
@@ -331,6 +331,14 @@ func NewSqlReadOpts() *SqliteTxOptions {
 	}
 }
 
+// NewSqlWriteOpts returns a new KeyStoreTxOptions instance triggers a write
```
hieblmi, line 334: nit: that triggers a write...

hieblmi (approved):
Thanks for the fix!

PR #783, opened 2024-06-25
sweepbatcher: add option WithNoBumping

starius:
It disables fee bumping. It is useful when fee rate is set externally.

Add test for fee bumping, testing both modes.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batcher_test.go
+++ b/sweepbatcher/sweep_batcher_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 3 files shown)


sweepbatcher/sweep_batcher_test.go
```diff
@@ -245,6 +251,123 @@ func testSweepBatcherBatchCreation(t *testing.T, store testStore,
...
+
+	// Deliver sweep request to batcher.
+	require.NoError(t, batcher.AddSweep(&sweepReq1))
+
+	// Since a batch was created we check that it registered for its primary
+	// sweep's spend.
+	<-lnd.RegisterSpendChannel
+
+	// Wait for sweep tx to appear in the wallet.
+	require.Eventually(t, func() bool {
```
bhandras, line 341: nit: there's no need for the eventually check here and below, since the tx can be read from `lnd.TxPublishChannel` and the guard will fail if the test timeouts.

bhandras (approved):
LGTM 🎉

sweepbatcher/sweep_batcher_test.go
starius: Thanks! Fixed.

sweepbatcher/sweep_batcher_test.go
```diff
@@ -245,6 +251,113 @@ func testSweepBatcherBatchCreation(t *testing.T, store testStore,
 	require.True(t, batcherStore.AssertSweepStored(sweepReq3.SwapHash))
 }
 
+// testFeeBumping tests that sweep is RBFed with slightly higher fee rate
```
hieblmi, line 254: nit: early line break

hieblmi (approved):

sweepbatcher/sweep_batcher_test.go
starius: Thanks! Fixed.

PR #785, opened 2024-06-26
sweepbatcher: add field SweepInfo.MinFeeRate 

starius:
MinFeeRate is minimum fee rate that must be used by a batch of the sweep. If it is specified, confTarget is ignored.

This is useful for external source of fees.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -97,6 +97,10 @@ type sweep struct {
 	// notifier is a collection of channels used to communicate the status
 	// of the sweep back to the swap that requested it.
 	notifier *SpendNotifier
+
+	// minFeeRate is minimum fee rate that must be used by a batch of
+	// the sweep. If it is specified, confTarget is ignored.
+	minFeeRate chainfee.SatPerKWeight
 }
 
 // batchState is the state of the batch.
@@ -399,9 +403,10 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
 		b.sweeps[sweep.swapHash] = *sweep
 
 		// If this is the primary sweep, we also need to update the
-		// batch's confirmation target.
+		// batch's confirmation target and fee rate.
 		if b.primarySweepID == sweep.swapHash {
 			b.cfg.batchConfTarget = sweep.confTarget
+			b.rbfCache.FeeRate = sweep.minFeeRate
 		}
 
 		return true, nil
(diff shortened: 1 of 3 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -456,6 +462,11 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
 	b.log.Infof("adding sweep %x", sweep.swapHash[:6])
 	b.sweeps[sweep.swapHash] = *sweep
 
+	// Update FeeRate.
+	if sweep.minFeeRate > b.rbfCache.FeeRate {
```
bhandras, line 466: nit: I'd change the order for readability (so: `if b.rbfCache.FeeRate < sweep.minFeeRate`).

bhandras (approved):
LGTM 🎉

sweepbatcher/sweep_batch.go
starius: Thanks! Fixed this and another instance in `sweep_batcher.go`.

sweepbatcher/sweep_batch.go
```diff
@@ -402,6 +406,7 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
 		// batch's confirmation target.
 		if b.primarySweepID == sweep.swapHash {
 			b.cfg.batchConfTarget = sweep.confTarget
+			b.rbfCache.FeeRate = sweep.minFeeRate
```
hieblmi, line 409: nit: update comment as well.

sweepbatcher/sweep_batch.go
```diff
@@ -456,6 +462,11 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
 	b.log.Infof("adding sweep %x", sweep.swapHash[:6])
 	b.sweeps[sweep.swapHash] = *sweep
 
+	// Update FeeRate.
```
hieblmi, line 465: nit: a more descriptive comment would be nice, mentioning that the sweeps min fee rate is the basis for fee bumps or something like that.

hieblmi (approved):
The changes look good, just more detailed comments would be great.

By referring to external fee sources, were you referring to `type Oracle struct`?
Would it make sense to include a fee estimator interface in the batch and batcher instead of a field `minFeeRate`?

sweepbatcher/sweep_batch.go
starius: Fixed

sweepbatcher/sweep_batch.go

starius:
> By referring to external fee sources, were you referring to `type Oracle struct`? Would it make sense to include a fee estimator interface in the batch and batcher instead of a field `minFeeRate`?

Discussed offline. We need a separate feerate for each sweep, so we can't have an interface assigning feerate to a batch. Keeping the current approach.

PR #789, opened 2024-07-11
sweepbatcher: fix spawning condition "unattached" spend notifiers

bhandras:
Previously, when handling a sweep we assumed that if a sweep status was completed, the parent batch was also finished. However, since the batch confirmation status depends on three on-chain confirmations, it is possible that a spend notifier was started for a sweep of an active batch. The notifier would fetch the parent batch from the database, but because we incorrectly assumed that  the parent was confirmed (when it was not), the DB call would fail with a 'no rows returned' error. This failure would cause the sweep to fail and the sweep batcher to stop, resulting in a permanent failure state.

#### Pull Request Checklist

### DIFF
--- a/loopdb/sqlc/queries/batch.sql
+++ b/loopdb/sqlc/queries/batch.sql
@@ -73,11 +73,7 @@ FROM
 JOIN
         sweeps ON sweep_batches.id = sweeps.batch_id
 WHERE
-        sweeps.swap_hash = $1
-AND
-        sweeps.completed = TRUE
-AND
-        sweep_batches.confirmed = TRUE;
+        sweeps.swap_hash = $1;
 
 -- name: GetBatchSweptAmount :one
 SELECT
--- a/sweepbatcher/store.go
+++ b/sweepbatcher/store.go
@@ -196,10 +196,6 @@ func (s *SQLStore) GetParentBatch(ctx context.Context, swapHash lntypes.Hash) (
 		return nil, err
 	}
 
-	if err != nil {
-		return nil, err
-	}
-
 	return convertBatchRow(batch), nil
 }
 
--- a/sweepbatcher/sweep_batcher.go
+++ b/sweepbatcher/sweep_batcher.go
@@ -418,7 +418,24 @@ func (b *Batcher) handleSweep(ctx context.Context, sweep *sweep,
 	// can't attach its notifier to the batch as that is no longer running.
 	// Instead we directly detect and return the spend here.
 	if completed && *notifier != (SpendNotifier{}) {
-		return b.monitorSpendAndNotify(ctx, sweep, notifier)
+		// Verify that the parent batch is confirmed. Note that a batch
+		// is only considered confirmed after it has received three
+		// on-chain confirmations to prevent issues caused by reorgs.
+		parentBatch, err := b.store.GetParentBatch(ctx, sweep.swapHash)
+		if err != nil {
+			log.Errorf("unable to get parent batch for sweep %x: "+
+				"%v", sweep.swapHash[:6], err)
+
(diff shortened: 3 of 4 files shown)


sweepbatcher/store.go
```diff
@@ -196,10 +196,6 @@ func (s *SQLStore) GetParentBatch(ctx context.Context, swapHash lntypes.Hash) (
 		return nil, err
 	}
 
-	if err != nil {
```
hieblmi, line 199: 👍

sweepbatcher/sweep_batcher.go
```diff
@@ -418,7 +418,22 @@ func (b *Batcher) handleSweep(ctx context.Context, sweep *sweep,
 	// can't attach its notifier to the batch as that is no longer running.
 	// Instead we directly detect and return the spend here.
 	if completed && *notifier != (SpendNotifier{}) {
-		return b.monitorSpendAndNotify(ctx, sweep, notifier)
+		// Verify that the parent batch is confirmed. Note that a batch
+		// is only considered confirmed after it has received three
+		// on-chain confirmations to prevent issues caused by reorgs.
+		parentBatch, err := b.store.GetParentBatch(ctx, sweep.swapHash)
```
hieblmi, line 424: we could pass the `parentBatch` into `monitorSpendAndNotify`, since the method is called in only one place. Then  remove the parent batch call in that method.

hieblmi (approved):
LGTM 💾

sweepbatcher/sweep_batcher.go
```diff
@@ -418,7 +418,22 @@ func (b *Batcher) handleSweep(ctx context.Context, sweep *sweep,
...
+			log.Errorf("unable to get parent batch for sweep %x: "+
+				"%v", sweep.swapHash[:6], err)
+
+			return err
+		}
+
+		// The parent batch is indeed confirmed, meaning it is complete
+		// and we won't be able to attach this sweep to it.
+		if parentBatch.State == batchConfirmed {
+			return b.monitorSpendAndNotify(ctx, sweep, notifier)
```
GeorgeTsagk, line 435: 👍 Checked that it's safe to proceed with re-adding the sweep to a closed batch, as we update the batch sweep before checking if batch is closed

loopdb/sqlc/queries/batch.sql
```diff
@@ -73,11 +73,7 @@ FROM
 JOIN
         sweeps ON sweep_batches.id = sweeps.batch_id
 WHERE
-        sweeps.swap_hash = $1
-AND
-        sweeps.completed = TRUE
-AND
-        sweep_batches.confirmed = TRUE;
+        sweeps.swap_hash = $1;
```
GeorgeTsagk, line 76: This looks more simple, not every parent has to be a completed one

GeorgeTsagk (approved):
> Previously, when handling a sweep we assumed that if a sweep status was completed, the parent batch was also finished.

That's probably the root cause for the bug, the swap is satisfied with 1 conf, while the sweep batcher is the one that takes it all the way to 3 (with re-orgs in mind)

LGTM!

loopdb/sqlc/queries/batch.sql
bhandras: The reason of that extra condition was that it was assumed (wrongly) that the parent is confirmed by the time the sweep is confirmed. Given that we move the assumption check before we monitor it is not needed in the DB.

sweepbatcher/sweep_batcher.go
bhandras: Good idea, passed in the parent batch ID since that's what we need.

PR #793, opened 2024-07-16
instantout: assign chaincfg in sql store

hieblmi:
Minor fix to instantiate the instantout sql store with network params.


bhandras (approved):
LGTM, nice catch!

PR #796, opened 2024-07-22
StaticAddr: Split withdrawal creation and publishing

hieblmi:
In this PR we separate the withdrawal tx creation
and publication in order to avoid the coop signing
process on each re-publishing.

P.S. Once approved/merged I will squash this commit with the one that introduced the withdrawal manager on the staging branch.

### DIFF
--- a/cmd/loop/quote.go
+++ b/cmd/loop/quote.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/lightninglabs/loop"
 	"github.com/lightninglabs/loop/looprpc"
 	"github.com/lightningnetwork/lnd/routing/route"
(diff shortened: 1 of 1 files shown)


bhandras (approved):
LGTM 🎉

staticaddr/withdraw/manager.go
```diff
@@ -402,9 +434,12 @@ func (m *Manager) handleWithdrawal(deposits []*deposit.Deposit,
 					err)
 			}
 
-			// Remove the withdrawal from the active withdrawals.
+			// Remove the withdrawal from the active withdrawals and
+			// remove its finalized to stop republishing it on block
```
hieblmi, line 438: nit: finalized withdrawal transaction

staticaddr/withdraw/manager.go
```diff
@@ -710,17 +745,13 @@ func (m *Manager) toPrevOuts(deposits []*deposit.Deposit,
...
-			withdrawalAddress, m.cfg.ChainParams,
-		)
-		if err != nil {
-			log.Errorf("Error decoding withdrawal address: %v", err)
-
-			return err
+	for _, finalizedTx := range m.finalizedWithdrawalTxns {
+		if finalizedTx == nil {
+			log.Warnf("Finalized withdrawal tx is nil")
+			continue
```
starius, line 751: `m.finalizedWithdrawalTxns` is not persisted, so tx is not re-published if the daemon restarts. IIUC, before this commit it would recreate a transaction and re-publish it.

hieblmi: `m.finalizedWithdrawalTxns` is populated by `createFinalizedWithdrawalTx` which is also called in `recover`, so it will still be republished here.

PR #803, opened 2024-08-05
cmd/loop: fix path options and loopDir interaction

starius:
According to the comment in the code, loopdir should affect tls cert path and macaroon path only if custom values were not set for them. However the actual code ignored custom tls cert path and macaroon path if loopdir and/or network was changed. For instance, the following call:

```
loop --network regtest --tlscertpath tls.cert --macaroonpath loop.macaroon
```

resulted in using `~/.loop/regtest/tls.cert` and `~/.loop/regtest/loop.macaroon` instead of the files provided.

This commit fixes the code to match the description in the comment. Also the comment was updated to mention custom network setting in addition to custom loopdir.

#### Pull Request Checklist


bhandras (approved):
LGTM, nice catch 🚀

hieblmi (approved):
LGTM!

PR #791, opened 2024-07-12
sweepbatcher: add mixed batches as an option

starius:
Option `WithMixedBatch` instructs `sweepbatcher` to create mixed batches with regard to cooperativeness. Such a batch can include both sweeps signed both cooperatively and non-cooperatively. If cooperative signing fails for a sweep, transaction is updated to sign that sweep non-cooperatively and another round of cooperative signing runs on the remaining sweeps. The remaining sweeps are signed in non-cooperative (more expensive) way. If the whole procedure fails for whatever reason, the batch is signed non-cooperatively (the fallback).

Updated greedy batch selection algorithm to take into account whether mixed batches are enabled.

#### Pull Request Checklist
- [ ] Update `release_notes.md` if your PR contains major features, breaking changes or bugfixes

PR #804, opened 2024-08-05
swap: fix leaf estimation in AddSuccessToEstimator

starius:
According to `AddTapscriptInput`'s godoc, the first argument (`leafWitnessSize`) must include not the whole size of witness, but only the size of data consumed by the revealed script. Previously the value was too high, because it also included the following extra elements: number_of_witness_elements (1 byte), witness script and its size, control block and its size.

Tests for greedy_batch_selection were affected by this change.

Factored out from https://github.com/lightninglabs/loop/pull/791

#### Pull Request Checklist


bhandras (approved):
LGTM, nice catch! :tada:

starius:
Rebased

PR #810, opened 2024-08-13
Add swaps 2.0

sputn1ck:
This PR adds the new hyperloop functionality to loop. Hyperloops are a new MuSig2 based swap type, where users of a batch share a single MuSig2 output. On cooperation this output gets swept directly to the users wallet addresses, without the need for HTLCs and sweeps.

This PR implements the cooperative path for a hyperloop. Future PRs will include the other non-cooperative paths.


hieblmi:
!lightninglabs-deploy mute

PR #813, opened 2024-08-26
sweepbatcher: set max batch size to 1000 sweeps

starius:
This is needed to avoid non-standard batch transactions (larger than 400k wu). A non-cooperative input is 393 wu, so 1000 inputs are still under 400k wu.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -45,6 +45,11 @@ const (
 	// maxFeeToSwapAmtRatio is the maximum fee to swap amount ratio that
 	// we allow for a batch transaction.
 	maxFeeToSwapAmtRatio = 0.2
+
+	// MaxSweepsPerBatch is the maximum number of sweeps in a single batch.
+	// It is needed to prevent sweep tx from becoming non-standard. Max
+	// standard transaction is 400k wu, a non-cooperative input is 393 wu.
+	MaxSweepsPerBatch = 1000
 )
 
 var (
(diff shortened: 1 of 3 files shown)


sweepbatcher/sweep_batcher_test.go
```diff
@@ -1176,6 +1176,156 @@ func testDelays(t *testing.T, store testStore, batcherStore testBatcherStore) {
...
+	var runErr error
+	go func() {
+		defer wg.Done()
+		runErr = batcher.Run(ctx)
+	}()
+
+	// Wait for the batcher to be initialized.
+	<-batcher.initDone
+
+	const swapsNum = 2500
```
bhandras, line 1227: To reduce test runtime, would just 1001 work? Also the test seems to be logging excessively, so perhaps we could reduce that too. wdyt?

starius: Great idea! I changed the number of swaps to 1001. Disabling logging has negligible effect.

bhandras: Disabling logging is helpful when running `make unit` locally, as it's easier to follow test logs.

bhandras (approved):

sweepbatcher/sweep_batcher_test.go
starius: Disabled logging in the test.

sweepbatcher/sweep_batch.go
```diff
@@ -462,6 +467,13 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
 		return true, nil
 	}
 
+	// Enforce MaxSweepsPerBatch. If there are already too many sweeps in
+	// the batch, do not add another sweep to prevent the tx from becoming
+	// non-standard.
+	if len(b.sweeps) >= MaxSweepsPerBatch {
+		return false, nil
```
hieblmi, line 474: does it make sense to log such an occurence?

hieblmi (approved):
LGTM 💾

sweepbatcher/sweep_batch.go
starius: I added Info log message in cases when addSweep returns accept=false and err=nil in a separate commit "sweepbatcher: log the reason of acceptance failure".

PR #818, opened 2024-09-02
sweepbatcher: log confirmed batch tx id

hieblmi:
Log confirmed batch tx ids.


bhandras (approved):

sweepbatcher/sweep_batch.go
```diff
@@ -1997,7 +1997,7 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
 // handleConf handles a confirmation notification. This is the final step of the
 // batch. Here we signal to the batcher that this batch was completed.
 func (b *batch) handleConf(ctx context.Context) error {
-	b.log.Infof("confirmed")
+	b.log.Infof("confirmed in txid %s", b.batchTxid.String())
```
starius, line 2000: I propose to remove `.String()`. If for some reason batchTxid is nil, `b.batchTxid.String()` panics. If we pass `b.batchTxid`, it just prints as `<nil>`.

starius (approved):

PR #823, opened 2024-09-05
sweepbatcher: add option WithWaitForAddSweep

starius:
`WithWaitForAddSweep` instructs sweepbatcher to wait for all existing sweeps to be `AddSweep`ed in Run before actual starting. This is needed in setups where `AddSweep` is called after setting up dependencies of `FetchSweep`, so `FetchSweep` would fail if called before `AddSweep`.

#### Pull Request Checklist


starius:
Replaced with another approach.

PR #826, opened 2024-09-10
Generalized notif stream

sputn1ck:
This PR adds a generalized notif stream that replaces specific streams such as the reservations listener.

### DIFF
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -21,6 +21,7 @@ import (
 	"github.com/lightninglabs/loop/loopd/perms"
 	"github.com/lightninglabs/loop/loopdb"
 	loop_looprpc "github.com/lightninglabs/loop/looprpc"
+	"github.com/lightninglabs/loop/notifications"
 	loop_swaprpc "github.com/lightninglabs/loop/swapserverrpc"
 	"github.com/lightninglabs/loop/sweepbatcher"
 	"github.com/lightningnetwork/lnd/clock"
@@ -501,21 +502,41 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 		}
 	}
 
+	// Start the notification manager.
+	notificationCfg := &notifications.Config{
+		Client:    loop_swaprpc.NewSwapServerClient(swapClient.Conn),
+		FetchL402: swapClient.Server.FetchL402,
+	}
+	notificationManager := notifications.NewManager(notificationCfg)
+
+	d.wg.Add(1)
+	go func() {
+		defer d.wg.Done()
+
+		log.Info("Starting notification manager")
+		err := notificationManager.Run(d.mainCtx)
+		if err != nil {
+			d.internalErrChan <- err
+			log.Errorf("Notification manager stopped: %v", err)
+		}
+	}()
+
 	var (
 		reservationManager *reservation.Manager
 		instantOutManager  *instantout.Manager
 	)
+
 	// Create the reservation and instantout managers.
 	if d.cfg.EnableExperimental {
 		reservationStore := reservation.NewSQLStore(
 			loopdb.NewTypedStore[reservation.Querier](baseDb),
 		)
 		reservationConfig := &reservation.Config{
-			Store:             reservationStore,
-			Wallet:            d.lnd.WalletKit,
-			ChainNotifier:     d.lnd.ChainNotifier,
-			ReservationClient: reservationClient,
-			FetchL402:         swapClient.Server.FetchL402,
+			Store:               reservationStore,
+			Wallet:              d.lnd.WalletKit,
+			ChainNotifier:       d.lnd.ChainNotifier,
+			ReservationClient:   reservationClient,
+			NotificationManager: notificationManager,
 		}
 
 		reservationManager = reservation.NewManager(
--- a/loopd/log.go
+++ b/loopd/log.go
@@ -10,6 +10,7 @@ import (
 	"github.com/lightninglabs/loop/instantout/reservation"
 	"github.com/lightninglabs/loop/liquidity"
 	"github.com/lightninglabs/loop/loopdb"
+	"github.com/lightninglabs/loop/notifications"
 	"github.com/lightninglabs/loop/sweepbatcher"
 	"github.com/lightningnetwork/lnd"
 	"github.com/lightningnetwork/lnd/build"
@@ -47,6 +48,9 @@ func SetupLoggers(root *build.RotatingLogWriter, intercept signal.Interceptor) {
 	)
 	lnd.AddSubLogger(
 		root, instantout.Subsystem, intercept, instantout.UseLogger,
+	)
+	lnd.AddSubLogger(
+		root, notifications.Subsystem, intercept, notifications.UseLogger,
 	)
 }
 


notifications/log.go
```diff
@@ -0,0 +1,26 @@
+package notifications
+
+import (
+	"github.com/btcsuite/btclog"
+	"github.com/lightningnetwork/lnd/build"
+)
+
+// Subsystem defines the sub system name of this package.
+const Subsystem = "NTF"
```
starius, line 9: I propose to make it 4-letter or 5-letter not to be confused with log levels (ERR/WRN/DBG). E.g. NOTIF or NTFC.

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+// NewManager creates a new notification manager.
+func NewManager(cfg *Config) *Manager {
+	return &Manager{
+		cfg:         cfg,
+		subscribers: make(map[NotificationType]map[string]func(*swapserverrpc.SubscribeNotificationsResponse)), //nolint:lll
+	}
+}
+
+// SubscribeReservations subscribes to the reservation notifications.
+func (n *Manager) SubscribeReservations(
```
starius, line 50: I suggest to change the interface to accept context and reservationResChan.

Then we don't need a callback, ID, and Unsubscribe method.

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+// RegisterNotifications registers a new server notification stream.
+func (n *Manager) RegisterNotifications() error {
```
starius, line 107: This method can be private.

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, subscribers := range n.subscribers {
+		delete(subscribers, id)
+	}
+}
+
+// Run starts the notification manager.
+func (n *Manager) Run(ctx context.Context, readyChan chan struct{},
+) error {
+
+	n.runCtx = ctx
```
starius, line 86: It is better not to store ctx in a field.

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+	// fetched a l402 token and have subscribed to the notifications at least
+	// once.
+	close(readyChan)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		}
+	}
```
starius, line 103: Can we just run the event loop here instead of creating a goroutine?

This will also fix the lint warning:
```
S1000: should use for range instead of for { select {} } (gosimple)
```

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+
+// Unsubscribe removes a subscription based on its ID.
+func (n *Manager) Unsubscribe(id string) {
+	for _, subscribers := range n.subscribers {
+		delete(subscribers, id)
+	}
+}
+
+// Run starts the notification manager.
+func (n *Manager) Run(ctx context.Context, readyChan chan struct{},
```
starius, line 83: `readyChan` can be declared as a write-only chan (`readyChan chan<- struct{}`).

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+			return
+
+		case <-timer.C:
+			err := m.cfg.FetchL402(ctx)
+			if err != nil {
+				log.Warnf("Error fetching L402: %v", err)
+				timer.Reset(time.Second * 10)
+				continue
+			}
+			m.hasL402 = true
```
starius, line 187: Add an empty line before `return`.

notifications/manager.go
```diff
@@ -0,0 +1,214 @@
...
+// handleNotification handles an incoming notification from the server,
+// forwarding it to the appropriate subscribers.
+func (n *Manager) handleNotification(notification *swapserverrpc.
+	SubscribeNotificationsResponse) {
+
+	switch notification.Notification.(type) {
+	case *swapserverrpc.SubscribeNotificationsResponse_ReservationNotification:
+		for _, fn := range n.subscribers[NotificationTypeReservation] {
+			fn(notification)
+		}
```
starius, line 202: I propose to add a `default` case and log an error there (unknown notification type).

This will also fix the lint warning:
```
singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
```

loopd/daemon.go
```diff
@@ -498,21 +499,48 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
...
+		}
+	}()
+
+	select {
+	case <-notificationInitChan:
+		break
+	case <-time.After(15 * time.Second):
+		return fmt.Errorf("notification manager not ready")
+	case <-d.mainCtx.Done():
+		return fmt.Errorf("notification manager not ready: %v", d.mainCtx.Err())
```
starius, line 524: Is the error message correct? If this happen, the root cause is `mainCtx` cancellation, not notification manager readiness.

starius (approved):
LGTM 🎉

Great work! I like the `oneof` approach in schema.
Left few suggestions.

notifications/manager.go
```diff
@@ -0,0 +1,205 @@
...
+	// caller.
+	notifChan := make(
+		chan *swapserverrpc.ServerReservationNotification, //nolint:lll
+	)
+	sub := subscriber{
+		subCtx:   ctx,
+		recvChan: notifChan,
+	}
+
+	m.subscribers[NotificationTypeReservation] = append(
```
starius, line 72: I think this is a race, because `m.subscribers` is accessed here and also in `handleNotification`.

To avoid this, we can either protect it with a mutex or (more common in Loop codebase) pass new subscriptions to the event loop via a channel.

notifications/manager.go
```diff
@@ -0,0 +1,205 @@
...
+		// We'll forward the reservation notification to all subscribers.
+		// Cleaning up any subscribers that have been canceled.
+		newSubs := []subscriber{}
+		for _, sub := range n.subscribers[NotificationTypeReservation] {
+			recvChan := sub.recvChan.(chan *swapserverrpc.
+				ServerReservationNotification)
+
+			select {
+			case <-sub.subCtx.Done():
+				continue
```
starius, line 193: I propose to close recvChan here.

Note that it will be closed not right after ctx expires, but after first notification of this type is received. This is fine, because receiving side (the caller) is likely to wait on the same ctx expiry anyway in `select` statement.

notifications/manager.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	callCtx, cancel := context.WithCancel(mainCtx)
+
+	// We'll now subscribe to the swap server notifications.
+	notifStream, err := n.cfg.Client.SubscribeNotifications(
+		callCtx, &swapserverrpc.SubscribeNotificationsRequest{},
+	)
+	if err != nil {
+		cancel()
```
starius, line 106: Add an empty line

notifications/manager.go
starius: I propose to re-organize the code to have only one event loop and avoid new goroutines launch.

In case a reconnect is needed, the code in that loop can do the actions needed to reconnect and reassign some variables and then proceed with receiving notifications in new iteration.

I think it will simplify the code and make it more reader friendly. It would be easier to follow the flow of code.

starius (commented):
Few suggestions

notifications/manager.go
starius: I agree, it makes sense not to add a new line here.

On the other hand, manually called `cancel` (not in `defer`) looks suspicious. I think, it won't be needed if the code is reorganized into a single event loop (and goroutine).

notifications/manager.go
```diff
@@ -0,0 +1,205 @@
...
+// handleNotification handles an incoming notification from the server,
+// forwarding it to the appropriate subscribers.
+func (n *Manager) handleNotification(notification *swapserverrpc.
+	SubscribeNotificationsResponse) {
+
+	switch notification.Notification.(type) {
+	case *swapserverrpc.SubscribeNotificationsResponse_ReservationNotification:
+		// We'll forward the reservation notification to all subscribers.
+		// Cleaning up any subscribers that have been canceled.
+		newSubs := []subscriber{}
```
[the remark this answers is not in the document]
starius: We can allocate for `len(n.subscribers[NotificationTypeReservation])` items (using it as slice capacity).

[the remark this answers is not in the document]
starius: I mean this:

```go
newSubs := make([]subscriber, 0, len(n.subscribers[NotificationTypeReservation]))
```

Then `append` won't reallocate, otherwise it works the same.

In most cases there were no cancellations, so the space is fully utilized.

notifications/manager.go
```diff
@@ -0,0 +1,193 @@
...
+		reservationNtfn := notification.GetReservationNotification()
+		for _, sub := range n.subscribers[NotificationTypeReservation] {
+			recvChan := sub.recvChan.(chan *swapserverrpc.
+				ServerReservationNotification)
+
+			select {
+			case <-sub.subCtx.Done():
+				close(recvChan)
+				continue
```
starius, line 179: Should it be `break` here? Otherwise we can close recvChan multiple times (panic).

loopd/daemon.go
```diff
@@ -498,21 +499,53 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
...
+		err := notificationManager.Run(d.mainCtx, initChan)
+		if err != nil {
+			d.internalErrChan <- err
+			log.Errorf("Notification manager stopped: %v", err)
+		}
+	}()
+
+	select {
+	case <-time.After(10 * time.Second):
+		return fmt.Errorf("notification manager not ready")
```
starius, line 524: Is 10 seconds sufficient? If L402 fails and retries, `fetchL402` already waits for 10 seconds, so it will crash here by the time of the second attempt. Also in Tor 10 seconds is not enough.

notifications/manager.go
```diff
@@ -0,0 +1,193 @@
...
+				log.Debugf("Received notification: %v", notification)
+				n.handleNotification(notification)
+				continue
+			}
+			log.Errorf("Error receiving "+
+				"notification: %v", err)
+
+			cancel()
+			// Wait for a bit before reconnecting.
+			<-time.After(time.Second * 10)
```
starius, line 131: I propose this:
```go
select {
case <-time.After(time.Second * 10):
case <-ctx.Done():
}
```
not to keep this goroutine running for 10 seconds after main context is cancelled.

notifications/manager.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Start the notification runloop.
+	for {
+		// Return if the context has been canceled.
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+
```
starius, line 103: Nit: I propose to move this code into a new method. Then we can `defer cancel()` in that method.

notifications/manager.go
```diff
@@ -0,0 +1,193 @@
...
+	m.subscribers[NotificationTypeReservation] = append(
+		m.subscribers[NotificationTypeReservation],
+		sub,
+	)
+	m.Unlock()
+
+	return notifChan
+}
+
+// Run starts the notification manager.
```
starius, line 84: I propose to add to godoc, that it closes readyChan as soon as it makes first successful connection to the stream. Also worth noting, that the function is long-running (not launching a goroutine and returning soon).

starius (approved):
LGTM 🌴
Final micro-nits

notifications/manager.go
starius: You are right! I confused sub.subCtx with ctx :)

hieblmi (approved):
Great work @sputn1ck, indeed the `oneof` approach is nicely extensible. I have only one suggestion to make the `Subscribe` method more generic for other users of the notification manager.

notifications/manager.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+}
+
+type subscriber struct {
+	subCtx   context.Context
+	recvChan interface{}
+}
+
+// SubscribeReservations subscribes to the reservation notifications.
+func (m *Manager) SubscribeReservations(ctx context.Context,
```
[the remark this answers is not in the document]
hieblmi: As discussed, lets keep it simple with a subscriber per notification type.

bhandras (approved):
LGTM 🥇

notifications/manager.go
```diff
@@ -0,0 +1,210 @@
...
+) <-chan *swapserverrpc.ServerReservationNotification {
+
+	notifChan := make(chan *swapserverrpc.ServerReservationNotification, 1)
+	sub := subscriber{
+		subCtx:   ctx,
+		recvChan: notifChan,
+	}
+
+	m.Lock()
+	m.subscribers[NotificationTypeReservation] = append(
```
bhandras, line 79: nit:  consider also extracting this as a method so we have `addSubscriber` and `removeSubscriber` for when we add more subscription types.

notifications/manager_test.go
```diff
@@ -0,0 +1,173 @@
...
+		err := mgr.Run(ctx)
+		require.NoError(t, err)
+	}()
+
+	// Wait a bit to ensure manager is running and has subscribed
+	require.Eventually(t, func() bool {
+		mgr.Lock()
+		defer mgr.Unlock()
+		return len(mgr.subscribers[NotificationTypeReservation]) > 0
+	}, time.Second*20, 10*time.Millisecond)
```
bhandras, line 128: nit: can just be 5 sec

bhandras (approved):

PR #829, opened 2024-09-20
loopd: allow setting RPC batch size when running cost migration

bhandras:
Previously we defaulted to fetching 1000 payments at a time when running the cost migration however due to slowness on LND side this may prevent the migration from completing due to RPC timeout kicking in. This commit adds a new config value to give more control to the user by setting the batch size as needed.

#### Pull Request Checklist


djkazic:
LGTM, got my loopd back online with this and batch size = 10 (`postgres_kvdb` backend on LND)

hieblmi (approved):
LGTM!

PR #830, opened 2024-10-07
swapserverrpc: refactor reservation notifications

sputn1ck:
This PR refactors the proto definitions of the reservation notifications to use a more generic
mechanism that can be used for other types of notifications as well.

The old stream is kept and deprecated to not make old clients crash.


bhandras (approved):
LGTM 🎉

swapserverrpc/server.proto
```diff
@@ -654,6 +658,17 @@ message FetchL402Request {
 // confirm the lnd L402.
 message FetchL402Response {
 }
+// SubscribeNotificationsRequest is a request to subscribe to notifications.
```
hieblmi, line 661: nit: nl above

hieblmi (approved):
LGTM.

PR #832, opened 2024-10-14
loopout: trace logs for hasBandwidth mpp check

hieblmi:
Adds trace logs for the `hasBandwidth` function that checks for mpp feasibility within `validateLoopOutRequest`.


bhandras (approved):
LGTM 🎉 Alternatively we could just log the local balances on startup and recreate the problematic case in a unit test.

hieblmi:
Added more logging.

PR #844, opened 2024-11-04
static addr loopin: add support for sweepless sweep batching

sputn1ck:
This PR adds support for serverside sweepless sweep batching, by listening for notifications from the server and responding
with the signature.

Only the 8 last commits are for review


staticaddr/loopin/interface.go
```diff
@@ -64,6 +64,10 @@ type StaticAddressLoopInStore interface {
 
 	// IsStored checks if the loop-in is already stored in the database.
 	IsStored(ctx context.Context, swapHash lntypes.Hash) (bool, error)
+
+	// FetchLoopInByHash returns the loop-in swap with the given hash.
+	FetchLoopInByHash(ctx context.Context, swapHash lntypes.Hash) (
+		*StaticAddressLoopIn, error)
```
starius, line 71: This commit does not build:
```
# github.com/lightninglabs/loop/staticaddr/loopin
staticaddr/loopin/actions.go:791:23: undefined: looprpc.FetchSweeplessSweepTxRequest
```

I think it should include SQL changes.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+			}
+
 		case <-ctx.Done():
 			return ctx.Err()
 		}
 	}
 }
 
+// notifyNotFinished notifies the server that a swap is not finished by
+// sending an empty signature map.
```
starius, line 223: "by sending an empty signature map"
This is a part of API.

I propose to add this to description of PushStaticAddressSweeplessSigsRequest.signing_info.

Also, we can make this more explicit by making a oneof: a map or an error (string). If the swap is not finished or any other error happened during signing, we would send PushStaticAddressSweeplessSigsRequest with an error message instead of a map.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+	sweepPacket, err := psbt.NewFromRawBytes(reader, false)
+	if err != nil {
+		return err
+	}
+
+	sweepTx := sweepPacket.UnsignedTx
+
+	// If the loop-in is not in the Succeeded state we return an
+	// error.
+	if loopIn.state != Succeeded {
```
starius, line 279: ~Should it also accept FetchSignPushSweeplessSweepTx and SucceededSweeplessSigFailed states?~

They are removed in a subsequent commit. Maybe it worth adding them here and removing in commit "staticaddr/loopin: update fsm for server ntfn sigs".

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+			}
+		}
+
+		if !foundDepositInTx {
+			return fmt.Errorf("deposit outpoint not part of " +
+				"sweep tx")
+		}
+	}
+
+	prevoutMap := make(map[wire.OutPoint]*wire.TxOut)
```
starius, line 326: I propose to preallocate the map, using len(req.PrevoutInfo) as its capacity.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+	for reqOutpoint := range req.DepositToNonces {
+		hasDeposit := false
+		for _, depositOutpoint := range loopIn.DepositOutpoints {
+			if depositOutpoint == reqOutpoint {
+				hasDeposit = true
+				break
+			}
+		}
+		if !hasDeposit {
+			return fmt.Errorf("deposit outpoint not part of " +
```
starius, line 306: I propose to include the outpoint of extra deposit to the error message.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+		for i, txIn := range sweepTx.TxIn {
+			if txIn.PreviousOutPoint.String() == reqOutpoint {
+				depositToIdxMap[reqOutpoint] = i
+				foundDepositInTx = true
+				break
+			}
+		}
+
+		if !foundDepositInTx {
+			return fmt.Errorf("deposit outpoint not part of " +
```
starius, line 321: I propose to include the outpoint of missing deposit to the error message.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+
+		for i, txIn := range sweepTx.TxIn {
+			if txIn.PreviousOutPoint.String() == reqOutpoint {
+				depositToIdxMap[reqOutpoint] = i
+				foundDepositInTx = true
+				break
+			}
+		}
+
+		if !foundDepositInTx {
```
starius, line 320: Do we need this check? IIUC it prevents sweeping a subset of deposits. What if we need to do this? E.g. because the number of inputs in a batch transaction is too high and we can accept some items to an existing batch, but not all. I think, current sweepbatcher logic can do this.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+		if err != nil {
+			return err
+		}
+
+		prevoutMap[wire.OutPoint{
+			Hash:  *txid,
+			Index: prevout.OutputIndex,
+		}] = &wire.TxOut{
+			Value:    int64(req.PrevoutInfo[i].Value),
+			PkScript: req.PrevoutInfo[i].PkScript,
```
starius, line 342: You can use `prevout` instead of `req.PrevoutInfo[i]` in these two lines.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+			return fmt.Errorf("deposit outpoint not part of " +
+				"sweep tx")
+		}
+	}
+
+	prevoutMap := make(map[wire.OutPoint]*wire.TxOut)
+
+	// Set all the prevouts in the prevout map.
+	for i := range req.PrevoutInfo {
+		prevout := req.PrevoutInfo[i]
```
starius, line 330:
```go
for i, prevout := range req.PrevoutInfo {
```

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+		prevoutMap,
+	)
+
+	sigHashes := txscript.NewTxSigHashes(
+		sweepPacket.UnsignedTx, prevOutputFetcher,
+	)
+
+	// We'll now sign for every deposit that is part of the loop-in.
+	responseMap := make(map[string]*looprpc.ClientSweeplessSigningInfo)
+	for depositOutpoint, nonce := range req.DepositToNonces {
```
starius, line 356: I propose to factor out the body of this `for` to a function.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+	// prevout info.
+	if len(sweepTx.TxIn) != len(req.PrevoutInfo) {
+		return fmt.Errorf("expected %v inputs, got %v",
+			len(req.PrevoutInfo), len(sweepTx.TxIn))
+	}
+
+	// Check if all the deposits requested are part of the loop-in and
+	// find them in the requested sweep.
+	depositToIdxMap := make(map[string]int)
+	for reqOutpoint := range req.DepositToNonces {
```
starius, line 297: I propose to factor out the body of this `for` to a function.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+		if !hasDeposit {
+			return fmt.Errorf("deposit outpoint not part of " +
+				"loop-in")
+		}
+
+		foundDepositInTx := false
+
+		for i, txIn := range sweepTx.TxIn {
+			if txIn.PreviousOutPoint.String() == reqOutpoint {
+				depositToIdxMap[reqOutpoint] = i
```
starius, line 314: Let's also check that `req.DepositToNonces` doesn't include duplicate deposits. We can add a check that `depositToIdxMap[reqOutpoint]` does not pre-exist here.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+			return err
+		}
+
+		var (
+			serverNonce [musig2.PubNonceSize]byte
+			sigHash     [32]byte
+		)
+
+		copy(serverNonce[:], nonce)
+		musig2Session, err := loopIn.createMusig2Session(
```
starius, line 371: I propose to cleanup the musig2 session after we finished using it.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
...
+			Sig:   sig,
+		}
+	}
+
+	txHash := sweepTx.TxHash()
+
+	_, err = m.cfg.Server.PushStaticAddressSweeplessSigs(
+		ctx, &looprpc.PushStaticAddressSweeplessSigsRequest{
+			SwapHash:    loopIn.SwapHash[:],
+			Txid:        txHash[:],
```
starius, line 413: IIUC, server knows txhash already, because it depends on unsigned parts only. Can we remove the field?

starius (approved):
LGTM! It's very cool!

Added several comments.

staticaddr/loopin/manager.go
```diff
@@ -189,12 +206,217 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
 				return ctx.Err()
 			}
 
+		case sweepReq := <-sweepReqs:
+			err = m.handleLoopInSweepReq(ctx, sweepReq)
+			if err != nil {
+				log.Errorf("Error handling loop-in sweep request: %v",
```
hieblmi, line 218: nit: lll

staticaddr/loopin/manager.go
hieblmi, line 284: nit: should do `if !loopIn.IsInState(Succeeded)`

staticaddr/loopin/manager.go
hieblmi: +1

staticaddr/loopin/manager.go
hieblmi: Isn't the pkscript the same for all deposits. If so couldn't we just call
```
loopIn.toPrevOuts(loopIn.Deposits, req.PrevoutInfo[0].PkScript)
```

hieblmi (commented):
I left some nits and potential simplifications. Will test it now.

hieblmi (approved):
tAck thanks for the batching preparations!

We should fix the mentioned clean-up of musig sessions, they get auto-cleaned if the signing runs smoothly, and then cleaning fails with the mentioned error. We should only cancel sessions if the signing doesn't take place.

We should also fix the minor issues from the previous review.

notifications/manager.go
```diff
@@ -189,10 +218,24 @@ func (m *Manager) handleNotification(notification *swapserverrpc.
 
 			recvChan <- reservationNtfn
 		}
+	case *swapserverrpc.SubscribeNotificationsResponse_StaticLoopInSweep: // nolint: lll
+		// We'll forward the static loop in sweep request to all
+		// subscribers.
+		staticLoopInSweepRequestNtfn := ntfn.
```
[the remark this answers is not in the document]
hieblmi: still an issue

swapserverrpc/staticaddr.proto
```diff
@@ -184,43 +179,26 @@ message ClientHtlcSigningInfo {
...
     // sigs for.
     bytes swap_hash = 1;
 
-    // A map of fee rate to the nonces.
-    map<uint64, ClientSweeplessSigningInfo> signing_info = 2;
+    // The txid of the sweep tx that the client wants to push the sweepless
+    // sigs for.
+    bytes txid = 2;
+
+    // A map of deposits in formad txid:idx to the nonces.
```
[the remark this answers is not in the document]
hieblmi: still an issue

staticaddr/loopin/interface.go
```diff
@@ -64,6 +64,10 @@ type StaticAddressLoopInStore interface {
 
 	// IsStored checks if the loop-in is already stored in the database.
 	IsStored(ctx context.Context, swapHash lntypes.Hash) (bool, error)
+
+	// FetchLoopInByHash returns the loop-in swap with the given hash.
+	FetchLoopInByHash(ctx context.Context, swapHash lntypes.Hash) (
```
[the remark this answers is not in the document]
hieblmi: still an issue

PR #846, opened 2024-11-05
staticaddr: Minor staging fixes

hieblmi:
This PR fixes some minor issues in the staging branch of static addr.
- [x]  default 6 deposit confs
- [x]  extract static addr service in its own proto file


staticaddr/loopin/fsm.go
```diff
@@ -355,7 +355,7 @@ func (f *FSM) Warnf(format string, args ...interface{}) {
 // Errorf logs an error message with the loop-in swap hash.
 func (f *FSM) Errorf(format string, args ...interface{}) {
 	if f.loopIn == nil {
-		log.Errorf(format, args...)
+		log.Infof(format, args...)
```
bhandras, line 358: Is there a strong reason to change to `Infof`?

hieblmi: in the case of error I guess better to keep it. Will change.

hieblmi:
Added another commit to increase the confirmation target from 3 to 6.

PR #852, opened 2024-11-18
loopcli/staticaddr: explicit loop in command

sputn1ck:
This PR changes static addr loop in to use a explicit command.

Only _last_ commit is new

### DIFF
--- a/cmd/loop/staticaddr.go
+++ b/cmd/loop/staticaddr.go
@@ -29,38 +29,8 @@ var staticAddressCommands = cli.Command{
 		listStaticAddressSwapsCommand,
 		withdrawalCommand,
 		summaryCommand,
+		staticAddressLoopInCommand,
 	},
-	Description: `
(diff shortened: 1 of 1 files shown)


hieblmi (approved):
✅ tAck, thanks for the change!

cmd/loop/staticaddr.go
```diff
@@ -427,9 +397,45 @@ func NewProtoOutPoint(op string) (*looprpc.OutPoint, error) {
 	}, nil
 }
 
+var staticAddressLoopInCommand = cli.Command{
+	Name:  "in",
+	Usage: "Loop in funds from static address deposits.",
+	Description: `
+	Requests a loop-in swap based on static address deposits. After the
+	creation of a static address funds can be send to it. Once the funds are
```
bhandras, line 405: nit: funds can be sent to it.

bhandras (approved):

PR #853, opened 2024-11-19
StaticAddr: merge staging to master

hieblmi:
This PR merges the static address loop-in feature to the main repository branch.


cmd/loop/staticaddr.go
```diff
@@ -0,0 +1,635 @@
+package main
```
hieblmi, line 1: add build flag

bhandras (approved):

PR #860, opened 2024-12-02
StaticAddr: Withdraw arbitrary amounts

hieblmi:
This PR adds an amount field to the withdrawal rpc.

It allows to withdraw a fraction of the sum of selected deposits to withdraw. The change is sent back to the static address.
If the change is below the dust limit it goes towards miner fees, but warns the user beforehand.

```
loop static withdraw --utxo XXX ---utxo YYY --dest_addr ZZZ --amount 1500000
```


swapserverrpc/staticaddr.proto
```diff
@@ -74,10 +74,16 @@ message ServerWithdrawRequest {
...
-    // The fee rate in sat/kw that the client wants to use for the sweep.
-    uint64 tx_fee_rate = 4;
+    // The amount that the client wants to withdraw, in satoshis. This amount
+    // can be a fraction of the total deposit amount.
+    int64 withdraw_amount = 4;
+
+    // The change amount will be sent back to the static address. The client
+    // ensures that the change amount covers the transaction fee, and that the
+    // change after fees is at least the dust limit.
+    int64 change_amount = 5;
```
starius, line 86: I propose to replace `client_withdrawal_addr`, `withdraw_amount` and `change_amount` with a psbt, so the UTXO could be used for whatever purpose, e.g. batch sending, coinjoin/payjoin, channel opening etc.

staticaddr/withdraw/manager.go
```diff
@@ -380,32 +383,49 @@ func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
...
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("couldn't get confirmation height for "+
+			"deposit, %w", err)
 	}
 
-	addressParams, err := m.cfg.AddressManager.GetStaticAddressParameters(
-		ctx,
+	// Send change back to the static address.
+	changeAddress, err := m.cfg.AddressManager.GetTaprootAddress(
```
starius, line 395: Why `m.cfg.AddressManager.GetStaticAddress(ctx)` can't be used here?

hieblmi: Good catch, I changed it.

staticaddr/withdraw/manager.go
```diff
@@ -628,33 +656,101 @@ func (m *Manager) createWithdrawalTx(outpoints []wire.OutPoint,
...
+			// If the change is dust, we give it to the miners.
+			hasChange = false
+			withdrawalAmount = selectedWithdrawalAmount
+
+		default:
+			// If the fees eat into our withdrawal amount, we fail
+			// the withdrawal.
+			return nil, 0, 0, fmt.Errorf("the change doesn't " +
+				"cover for fees. Consider lowering the fee " +
+				"rate or increase the withdrawal amount")
```
starius, line 707: s/increase/decrease/ IIUC

staticaddr/withdraw/manager.go
```diff
@@ -628,33 +656,101 @@ func (m *Manager) createWithdrawalTx(outpoints []wire.OutPoint,
...
+			changeAmount = change - feeWithChange
+			withdrawalAmount = selectedWithdrawalAmount
+
+		case change-feeWithChange >= 0:
+			// If the change is dust, we give it to the miners.
+			hasChange = false
+			withdrawalAmount = selectedWithdrawalAmount
+
+		default:
+			// If the fees eat into our withdrawal amount, we fail
```
starius, line 703: Actually, the feeWithChange fees eat into our withdrawal amount, but this doesn't mean that feeWithoutChange do, since it is lower than feeWithChange. This means, that some amounts fail, even though it is possible to withdraw without change. And we fallback to withdrawing without change in general case when it is not enough funds for the change. So I think for consistency we should do the same here. Can we replace `case change-feeWithChange >= 0:` with `case change-feeWithoutChange >= 0:` above?

starius (approved):
LGTM 🎉
I left few comments.
Also I notices that there are no tests in `staticaddr/withdraw` package. Does it make sense to add some to cover all withdrawing cases (full amount, with change, gifting dust change to miners)?

bhandras:
Needs a rebase.

bhandras:
Needs a rebase 🙏

staticaddr/withdraw/manager.go
```diff
@@ -261,8 +263,8 @@ func (m *Manager) WaitInitComplete() {
 
 // WithdrawDeposits starts a deposits withdrawal flow.
 func (m *Manager) WithdrawDeposits(ctx context.Context,
-	outpoints []wire.OutPoint, destAddr string, satPerVbyte int64) (string,
-	string, error) {
+	outpoints []wire.OutPoint, destAddr string, satPerVbyte int64,
+	amount int64) (string, string, error) {
```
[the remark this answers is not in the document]
hieblmi: good point.

staticaddr/withdraw/manager.go
hieblmi: you are correct, so that the remaining amount can cover for fees.

staticaddr/withdraw/manager.go
hieblmi: This is a good catch 🙏 You are right in that the withdrawal shouldn't fail until the `feeWithoutChange` eats into our withdrawal amount.

Changed it.

hieblmi:
Rebased and addressed Boris' comments, thanks for the improvement @starius.

staticaddr/withdraw/manager.go
```diff
@@ -649,25 +655,101 @@ func (m *Manager) createWithdrawalTx(outpoints []wire.OutPoint,
...
 
-	msgTx.AddTxOut(sweepOutput)
+	// Create the withdrawal output.
+	msgTx.AddTxOut(&wire.TxOut{
+		Value:    int64(withdrawalAmount),
+		PkScript: withdrawScript,
+	})
+
+	if hasChange {
+		changeScript, err := txscript.PayToAddrScript(changeAddress)
```
bhandras, line 736: I think we should just get the change address here (so no need to pass it around and it is not generated if unused).

bhandras (approved):
LGTM 🎉

staticaddr/withdraw/manager.go
hieblmi: I moved all change address data into this if-branch.

staticaddr/withdraw/manager.go
```diff
@@ -649,25 +640,117 @@ func (m *Manager) createWithdrawalTx(outpoints []wire.OutPoint,
...
 
-	msgTx.AddTxOut(sweepOutput)
+	if withdrawalAmount < dustLimit {
+		return nil, 0, 0, fmt.Errorf("withdrawal amount is below " +
+			"dust limit")
+	}
+
+	if changeAmount < 0 {
+		return nil, 0, 0, fmt.Errorf("change amount is negative")
+	}
```
starius, line 707: Optional. Can we add a check that `changeAmount` is lower than each input UTXO? If it is larger, this means we could save on fees by not including that UTXO into the transaction. Can we show a warning to the user asking to remove some UTXOs or to confirm that they want to proceed.

starius (approved):
LGTM!

I added one proposal. It is not blocking.

staticaddr/withdraw/manager.go
hieblmi: That's a neat tweak, I will add that.

PR #864, opened 2024-12-09
StaticAddr: fix deposit detection

hieblmi:
The height hint for deposit confirmations was previously set to the initiation height of the deposit manager, which does not work in case the manager is restarted before a deposit was confirmed.

### DIFF
--- a/staticaddr/address/interface.go
+++ b/staticaddr/address/interface.go
@@ -54,4 +54,7 @@ type Parameters struct {
 
 	// ProtocolVersion is the protocol version of the static address.
 	ProtocolVersion version.AddressProtocolVersion
+
+	// InitiationHeight is the height at which the address was initiated.
+	InitiationHeight int32
 }
--- a/staticaddr/address/manager.go
+++ b/staticaddr/address/manager.go
@@ -40,6 +40,10 @@ type ManagerConfig struct {
 	// ChainParams is the chain configuration(mainnet, testnet...) this
 	// manager uses.
 	ChainParams *chaincfg.Params
+
+	// ChainNotifier is the chain notifier that is used to listen for new
+	// blocks.
+	ChainNotifier lndclient.ChainNotifierClient
 }
 
 // Manager manages the address state machines.
@@ -47,6 +51,8 @@ type Manager struct {
 	cfg *ManagerConfig
 
 	sync.Mutex
+
+	currentHeight int32
 }
 
 // NewManager creates a new address manager.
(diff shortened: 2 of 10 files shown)


staticaddr/address/manager.go
```diff
@@ -58,8 +64,26 @@ func NewManager(cfg *ManagerConfig) *Manager {
 
 // Run runs the address manager.
 func (m *Manager) Run(ctx context.Context) error {
-	<-ctx.Done()
-	return nil
+	newBlockChan, newBlockErrChan, err :=
+		m.cfg.ChainNotifier.RegisterBlockEpochNtfn(ctx)
```
bhandras, line 68: I think this crashes in the unit tests.

staticaddr/address/interface.go
```diff
@@ -18,7 +18,8 @@ var (
 type Store interface {
 	// CreateStaticAddress inserts a new static address with its parameters
 	// into the store.
-	CreateStaticAddress(ctx context.Context, addrParams *Parameters) error
+	CreateStaticAddress(ctx context.Context, addrParams *Parameters,
+		currentHeight int32) error
```
starius, line 22: Why a separate argument `currentHeight` is needed? `addrParams` has new field `InitiationHeight`.

hieblmi: removed `currentHeight` and use the `InitiationHeight`.

staticaddr/address/manager.go
```diff
@@ -147,6 +171,7 @@ func (m *Manager) NewAddress(ctx context.Context) (*btcutil.AddressTaproot,
 		ProtocolVersion: version.AddressProtocolVersion(
 			protocolVersion,
 		),
+		InitiationHeight: m.currentHeight,
```
starius, line 174: Can `currentHeight` be uninitialized (0) at this point? IMHO we should either return an error here if currentHeight is 0 (if we are sure that this won't happen) or wait for its initialization (e.g. using a cond var).

bhandras (approved):
LGTM, nice catch! 🎉

staticaddr/address/manager_test.go
```diff
@@ -97,7 +97,7 @@ func TestManager(t *testing.T) {
 	// Start the manager.
 	go func() {
 		err := testContext.manager.Run(ctxb)
-		require.NoError(t, err)
+		require.ErrorContains(t, err, "context canceled")
```
starius, line 100: nit: `require.ErrorIs(t, err, context.Canceled)`

starius (approved):
LGTM! 🏆

I propose to add a test reproducing the bug being fixed.

staticaddr/address/manager_test.go
hieblmi: fixed.

PR #865, opened 2024-12-11
StaticAddr: warn user when creating a static address

hieblmi:
Hint to the user that they need the L402 file to loop-in or withdraw funds sent to a static address.


bhandras (approved):
LGTM 🌮

PR #868, opened 2024-12-13
StaticAddr: fix rc review comments

hieblmi:
Gather PR to fix upcoming rc testing issues.

- checks dust limits before withdrawing deposits
- removes the `--experimental` flag for static addresses
- fixes text formatting

### DIFF
--- a/staticaddr/withdraw/manager.go
+++ b/staticaddr/withdraw/manager.go
@@ -380,7 +380,39 @@ func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
 		).FeePerKWeight()
 	}
 
+	// We'll now check the selected fee rate leaves a withdrawal output that
+	// is above the dust limit. If not we cancel the withdrawal instead of
+	// requesting a signature from the server.
+	addressParams, err := m.cfg.AddressManager.GetStaticAddressParameters(
+		ctx,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get confirmation height for "+
+			"deposit, %w", err)
+	}
+
+	// Calculate the fee value in satoshis.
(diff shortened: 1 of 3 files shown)


staticaddr/withdraw/manager.go
```diff
@@ -380,7 +380,39 @@ func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
...
+		totalValue  = withdrawalValue(prevOuts)
+		outputValue = int64(totalValue) - int64(feeValue)
+		// P2TRSize calculates a dust limit based on a 40 byte maximum
+		// size witness output.
+		dustLimit = lnwallet.DustLimitForSize(input.P2TRSize)
+	)
+
+	if outputValue < int64(dustLimit) {
+		return nil, fmt.Errorf("withdrawal output value %v sats "+
+			"below dust limit %v sats", outputValue, dustLimit)
```
starius, line 413: %v prints amouts in BTC with " BTC" suffix. So this error message would look like "below dust limit 0.00000294 BTC sats".
I propose to use "%d sats" in both cases.

staticaddr/withdraw/manager.go
```diff
@@ -380,7 +380,39 @@ func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
 		).FeePerKWeight()
 	}
 
+	// We'll now check the selected fee rate leaves a withdrawal output that
```
starius, line 383: I propose to cover the logic and various sub-cases in this method with unit tests. Maybe factor out a pure function to make it unit tests friendly.

starius (approved):
LGTM! 🏒
Added a couple of comments.

PR #881, opened 2025-01-31
autoloop: ignore asset channels

sputn1ck:
This PR adds a check for autoloop to ignore the balance in channels that have custom data (such as asset channels)
As asset channels contain a sat balance, we don't want them to be considered for default btc autoloops.

### DIFF
--- a/liquidity/autoloop_test.go
+++ b/liquidity/autoloop_test.go
@@ -1287,9 +1287,20 @@ func TestEasyAutoloop(t *testing.T) {
 		Capacity:      100000,
 	}
 
+	// The custom channel should be ignored.
+	easyChannelCustom := lndclient.ChannelInfo{
+		Active:            true,
+		ChannelID:         chanID1.ToUint64(),
+		PubKeyBytes:       peer1,
+		LocalBalance:      50000,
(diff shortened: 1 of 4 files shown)


liquidity/liquidity.go
```diff
@@ -556,6 +556,12 @@ func (m *Manager) dispatchBestEasyAutoloopSwap(ctx context.Context) error {
 
 	localTotal := btcutil.Amount(0)
 	for _, channel := range channels {
+		// If the channel has custom channel data, the channel is a
```
bhandras, line 559: nit: I think this code block is worth extracting to a helper function so we spare the copy-paste of the comment.

liquidity/autoloop_test.go
```diff
@@ -1287,9 +1287,19 @@ func TestEasyAutoloop(t *testing.T) {
 		Capacity:      100000,
 	}
 
+	easyChannelCustom := lndclient.ChannelInfo{
```
hieblmi, line 1290: maybe leave a comment that this channel should be ignored in the autoloop step.

hieblmi (approved):
LGTM!

PR #882, opened 2025-01-31
sweepbatcher: always enabled mixed batches

starius:
Mixed batches approach covers the case where all the inputs are cooperative (function `publishBatchCoop`) and it is better than fully non-cooperative case (function `publishBatch`), but it can also create such transaction if needed, i.e. if all the sweeps are non-cooperative. So we can remove functions `publishBatch`, `publishBatchCoop`, option `WithMixedBatch` and associated code in greedy batch selection algorithm.

This simplifies the code ahead of the upcoming changes.

#### Pull Request Checklist


starius:
> I missed why there were two different batches live in the first place. Isn't the mixed batcher just a more generalized coop-batcher?

@hieblmi yes, the mixed batcher is a generalized coop-batcher and it is also a generalized noncoop-batcher. It can produce both these variants when possible/needed.

bhandras (approved):
LGTM 🍨

PR #884, opened 2025-02-06
notifications: do not hog the CPU if there's no L402 token yet

bhandras:
#### Pull Request Checklist

### DIFF
--- a/notifications/manager.go
+++ b/notifications/manager.go
@@ -149,6 +149,7 @@ func (m *Manager) Run(ctx context.Context) error {
 		waitTime time.Duration
 		backoff  time.Duration
 		attempts int
+		timer    = time.NewTimer(0)
 	)
 
 	// Start the notification runloop.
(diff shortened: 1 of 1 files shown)


notifications/manager.go
```diff
@@ -149,14 +149,17 @@ func (m *Manager) Run(ctx context.Context) error {
...
+		timer    = time.NewTimer(0)
 	)
 
 	// Start the notification runloop.
 	for {
 		// Increase the wait time for the next iteration.
 		backoff = waitTime + time.Duration(attempts)*time.Second
 		waitTime = 0
-		timer := time.NewTimer(backoff)
+
```
hieblmi, line 160: I don't think we need the `waitTime = 0` above here.

hieblmi (approved):
LGTM pending one nit.

notifications/manager.go
bhandras: We want to reset the `waitTime` here which is really used in line 224 if we were able to connect the stream but upon first read it disconnected.

PR #887, opened 2025-02-20
staticaddr: fractional swap amount

hieblmi:
_the first commits are from the previous PR_

This PR adds an `amount` flag to static loop-in swaps that allows the caller to only swap a fraction of the value of the selected deposits. If only `amount` is selected, the client selects deposits automatically to cover for the swaps amount.

To ensure backwards UX compatibility, omitting the new flag (`--amount=0`) selects the total deposit amount for swapping.

This PR also adds a method `SelectDeposits` to the loop-in manager that is used to coin-select deposits in case the user
only provides a static swap amount.

TODO
- [x] Sign sweepless sweep and htlc tx with change output if needed
- [x] integration test


hieblmi:
Rebased

hieblmi:
Rebased

hieblmi:
Rebased

hieblmi:
Rebased.

hieblmi:
Rebased

loopdb/sqlc/migrations/000015_static_selected_amount.down.sql
```diff
@@ -0,0 +1,3 @@
+-- selected_amount is a fractinal amount amongst selected deposits of a static
```
bhandras: nit: fractinal => fractional

loopdb/sqlc/migrations/000015_static_selected_amount.up.sql
```diff
@@ -0,0 +1,4 @@
+-- selected_amount is a fractional amount amongst selected deposits of a static
+-- address loop-in swap. A selected amount of 0 indicates that the total amount
+-- of deposits is selected for the swap.
+ALTER TABLE static_address_swaps ADD selected_amount BIGINT NOT NULL DEFAULT 0;
```
bhandras, line 4: Perhaps we could create a SQL migration that adds the actual sum of deposits instead of zero. I think it should be doable.

staticaddr/deposit/manager.go
```diff
@@ -578,3 +578,28 @@ func (m *Manager) toActiveDeposits(outpoints *[]wire.OutPoint) ([]*FSM,
 
 	return fsms, deposits
 }
+
+// DepositsForOutpoints returns all deposits that are behind the given
+// outpoints.
+func (m *Manager) DepositsForOutpoints(ctx context.Context,
+	outpoints []string) ([]*Deposit, error) {
+
+	deposits := make([]*Deposit, 0, len(outpoints))
```
bhandras, line 581: Should we sanity check that the input contains no duplicates?

loopd/swapclient_server.go
```diff
@@ -913,20 +920,45 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
+		// input fee.
+		for _, deposit := range depositList.FilteredDeposits {
+			totalDepositAmount += btcutil.Amount(
+				deposit.Value,
+			)
+		}
+
+		// If the selected amount would leave a dust change output or
+		// exceeds the total deposits value, we return an error.
+		dustLimit := lnwallet.DustLimitForSize(input.P2TRSize)
```
bhandras: nit: I think the amount check we could potentially move to the manager.

cmd/loop/staticaddr.go
```diff
@@ -591,36 +607,61 @@ func staticAddressLoopIn(ctx *cli.Context) error {
...
 
-	var sum int64
-	depositMap := make(map[string]*looprpc.Deposit)
+	// Select the deposits that are needed to cover the swap amount without
+	// leaving a dust change.
+	var selectedDeposits []string
+	var selectedAmount int64
 	for _, deposit := range deposits {
-		depositMap[deposit.Outpoint] = deposit
+		if selectedAmount >= amount+int64(dustLimit) {
```
bhandras: I think this is wrong, because if the user wants to loop in 1 btc and they have two 1 btc deposits then we'll select both.

cmd/loop/staticaddr.go
```diff
@@ -591,36 +607,61 @@ func staticAddressLoopIn(ctx *cli.Context) error {
...
-			return true
-		}
-		found[outpoint] = struct{}{}
-	}
+// selectDeposits sorts the deposits by amount in descending order, then by
+// blocks-until-expiry in ascending order. It then selects the deposits that
+// are needed to cover the amount requested without leaving a dust change. It
+// returns an error if the sum of deposits minus dust is less than the requested
+// amount.
+func selectDeposits(deposits []*looprpc.Deposit, amount int64) ([]string,
```
bhandras: Could you please add a unit test for this function?

bhandras (commented):
Just a few comments, otherwise look pretty good!

loopdb/sqlc/migrations/000015_static_selected_amount.down.sql
hieblmi: fixed

loopdb/sqlc/migrations/000015_static_selected_amount.up.sql
hieblmi: I've added a migration as you suggested and also added a unit test for it.

staticaddr/deposit/manager.go
hieblmi: done

loopd/swapclient_server.go
hieblmi: Done.

cmd/loop/staticaddr.go
hieblmi: Great catch, I've reworked the function and added a unit test.

cmd/loop/staticaddr.go
hieblmi: done

staticaddr/loopin/manager.go
```diff
@@ -562,8 +606,33 @@ func (m *Manager) initiateLoopIn(ctx context.Context,
 	}
 	totalDepositAmount := tmp.TotalDepositAmount()
 
+	// If the selected amount would leave a dust change output or exceeds
+	// the total deposits value, we return an error.
+	dustLimit := lnwallet.DustLimitForSize(input.P2TRSize)
```
bhandras: nit: why not use SwapAmountFromSelectedAmount?

cmd/loop/staticaddr.go
```diff
@@ -624,6 +640,47 @@ func staticAddressLoopIn(ctx *cli.Context) error {
 	return nil
 }
 
+// selectDeposits sorts the deposits by amount in descending order, then by
+// blocks-until-expiry in ascending order. It then selects the deposits that
+// are needed to cover the amount requested without leaving a dust change. It
+// returns an error if the sum of deposits minus dust is less than the requested
+// amount.
+func selectDeposits(deposits []*looprpc.Deposit, targetAmount int64) ([]string,
```
bhandras: (see my next comment too)

Ideally this should be solved with a knapsack algorithm, but since that's a bit difficult, maybe we can use a variance of this heuristic to first try to find a covering deposit?

staticaddr/loopin/selected_amount_migration.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	// Now we'll calculate the cost for each swap and finally update the
+	// costs in the database.
+	// TODO(hieblmi): normalize swap hash and deposit ids.
+	updateAmounts := make(map[lntypes.Hash]btcutil.Amount)
+	for _, swap := range swaps {
+		for _, outpoint := range swap.DepositOutpoints {
+			op := strings.Split(outpoint, ":")
```
bhandras: nit: use `wire.NewOutPointFromString`.

staticaddr/loopin/selected_amount_migration.go
```diff
@@ -0,0 +1,99 @@
...
+	// amount migration.
+	selectedAmountMigrationID = "selected_amount"
+)
+
+// MigrateSelectedSwapAmount will update the selected swap amount of past swaps
+// with the sum of the values of the deposits they swapped.
+func MigrateSelectedSwapAmount(ctx context.Context, db loopdb.SwapStore,
+	depositStore *deposit.SqlStore, swapStore *SqlStore) error {
+
+	migrationDone, err := db.HasMigration(ctx, selectedAmountMigrationID)
```
bhandras, line 25: Imo the whole migration should be done in a single transaction.

staticaddr/loopin/manager.go
hieblmi: Sorry, what do you mean by SwapAmountFromSelectedAmount? I can't follow.

cmd/loop/staticaddr.go
hieblmi: The current implementation sorts values descending and then fills the sack until the target is met. This might be a good enough approximation of a knapsack solver, see here: https://en.wikipedia.org/wiki/Knapsack_problem#Greedy_approximation_algorithm

staticaddr/loopin/selected_amount_migration.go
hieblmi: fixed

staticaddr/loopin/selected_amount_migration.go
hieblmi: The actual migration is done in a single batch in `BatchUpdateSelectedSwapAmounts`. Is that what you mean?

hieblmi:
Rebased

hieblmi:
Rebased master

hieblmi:
Rebased

loopd/swapclient_server.go
```diff
@@ -922,20 +962,34 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
-		// The requested amount should be 0 here if the request
-		// contained deposit outpoints.
-		if amount != 0 && len(depositList.FilteredDeposits) > 0 {
-			return nil, fmt.Errorf("amount should be 0 for " +
-				"deposit quotes")
+		if len(req.DepositOutpoints) !=
+			len(depositList.FilteredDeposits) {
+
+			return nil, fmt.Errorf("expected %d deposits, got %d",
+				numDeposits, len(depositList.FilteredDeposits))
```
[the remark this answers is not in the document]
hieblmi: fixed

loopd/swapclient_server.go
```diff
@@ -2284,10 +2345,17 @@ func validateConfTarget(target, defaultTarget int32) (int32, error) {
 // validateLoopInRequest fails if the mutually exclusive conf target and
 // external parameters are both set.
 func validateLoopInRequest(htlcConfTarget int32, external bool,
-	numDeposits uint32, amount int64) (int32, error) {
+	numDeposits uint32, amount int64, autoSelectDeposits bool) (int32,
+	error) {
```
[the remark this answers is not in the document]
hieblmi: added

hieblmi:
Address gemini comments and rebased.

starius:
> first two commits are from the previous PR

Which PR includes commit "sqlc: add amount column to static swaps"?

looprpc/client.proto
```diff
@@ -833,6 +837,14 @@ message QuoteRequest {
...
     */
     AssetLoopOutRequest asset_info = 9;
+
+    /*
+    This field indicates whether the swap amount should be covered by our lnd
+    wallet or by static address deposits. If set to true, the swap amount
+    will be selected from the static address deposits. This field cannot be used
+    in conjunction with deposit_outpoints.
+    */
+    bool select_deposits = 10;
```
starius: This logic is hard to understand.
IIUC, we have 3 cases how this RPC is used with loop-ins:

 - legacy loop-in (deposit_outpoints=[], select_deposits=false)
 - static loop-in, deposits autoselected (deposit_outpoints=[], select_deposits=true)
 - static loop-in, deposits selected manually (deposit_outpoints is provided, select_deposits=false)

Also in both static loop-in cases depending on `amt` value it can result in partial static loop-in (have change).

Can we provide `oneof` with these 3 cases of loop-in to make it more explicit:
```protobuf
oneof loopin {
  LegacyIn legacy = 10;
  StaticInAutoselect static_autoselect = 11;
  StaticInManualSelection static_manual_selection = 12;
}
```

staticaddr/loopin/loopin.go
```diff
@@ -308,12 +325,20 @@ func (l *StaticAddressLoopIn) createHtlcTx(chainParams *chaincfg.Params,
...
 	// Create the sweep output
 	sweepOutput := &wire.TxOut{
-		Value:    int64(l.TotalDepositAmount()) - int64(fee),
+		Value:    int64(swapAmt - fee),
 		PkScript: pkscript,
 	}
 
 	msgTx.AddTxOut(sweepOutput)
 
+	// We expect change to be sent back to our static address output script.
```
starius, line 336: > We expect change to be sent back to our static address output script.

I guess in the multi address world this will be fresh static address, right?

I propose to prepare a bit: add a field holding change pkscript (and for now just assign it to `l.AddressParams.PkScript`). We should use it everywhere and pass instead of `hasChange` to `htlcWeight`. Then we don't have to assume there that is is P2TR and we can use method [weightEstimator.AddOutput](https://pkg.go.dev/github.com/lightningnetwork/lnd/input#TxWeightEstimator.AddOutput).

staticaddr/loopin/manager.go
```diff
@@ -305,6 +317,34 @@ func (m *Manager) handleLoopInSweepReq(ctx context.Context,
...
+	changeAmt := totalDepositAmount - loopIn.SelectedAmount
+	if changeAmt > 0 && changeAmt < totalDepositAmount {
+		var foundChange bool
+		changePkScript := loopIn.AddressParams.PkScript
+
+		for _, out := range sweepTx.TxOut {
+			if out.Value == int64(changeAmt) &&
+				bytes.Equal(out.PkScript, changePkScript) {
+
+				foundChange = true
```
starius: What is two swaps have equal change outputs? IIRC, we need to check all swaps taking part in the transaction to make sure that each of them has a separate change output.

staticaddr/loopin/manager.go
```diff
@@ -531,29 +571,83 @@ func (m *Manager) DeliverLoopInRequest(ctx context.Context,
...
+		}
 	}
 
-	// Calculate the total deposit amount.
+	// Calculate the total deposit amount and check if the selected amount
+	// would leave a dust output.
 	tmp := &StaticAddressLoopIn{
-		Deposits: deposits,
+		Deposits: selectedDeposits,
+	}
```
starius: I propose to make a function:
```go
func sumOfDeposits(deposts []Deposit) btcutil.Amount
```

We could factor out `tmp` there.

staticaddr/loopin/manager.go
```diff
@@ -710,6 +805,85 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
+
+	return nil, fmt.Errorf("not enough deposits to cover "+
+		"requested amount or prevent dust change, have %d but need %d",
+		selectedAmount, targetAmount)
+}
+
+// SwapAmountFromSelectedAmount calculates the swap amount based on the
+// selected amount and the total deposit amount. It checks if the selected
+// amount leaves a dust change output or exceeds the total deposits value.
+func SwapAmountFromSelectedAmount(totalDepositAmount btcutil.Amount,
```
starius: `s/SwapAmountFromSelectedAmount/DeduceSwapAmount/`

It deduces not only from selected amount, so the current name is misleading.

staticaddr/loopin/manager.go
```diff
@@ -710,6 +805,85 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
+	}
+
+	return nil, fmt.Errorf("not enough deposits to cover "+
+		"requested amount or prevent dust change, have %d but need %d",
+		selectedAmount, targetAmount)
+}
+
+// SwapAmountFromSelectedAmount calculates the swap amount based on the
+// selected amount and the total deposit amount. It checks if the selected
+// amount leaves a dust change output or exceeds the total deposits value.
```
starius: Note that selectedAmount=0 means "swap everything".

What does selectedAmount=totalDepositAmount mean? Is it allowed? Please also note it here.

staticaddr/loopin/manager.go
```diff
@@ -710,6 +805,85 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
+func SwapAmountFromSelectedAmount(totalDepositAmount btcutil.Amount,
+	selectedAmount btcutil.Amount) (btcutil.Amount, error) {
+
+	// If the selected amount would leave a dust change output or exceeds
+	// the total deposits value, we return an error.
+	swapAmount := selectedAmount
+	dustLimit := lnwallet.DustLimitForSize(input.P2TRSize)
```
starius: We can factor out `dustLimit` to an external var to cache the value.

staticaddr/loopin/manager.go
```diff
@@ -710,6 +805,85 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
+	remainingAmount := totalDepositAmount - selectedAmount
+	switch {
+	case remainingAmount < 0:
+		return 0, fmt.Errorf("selected amount %v exceeds total "+
+			"deposit value %v", selectedAmount, totalDepositAmount)
+
+	case remainingAmount > 0 && remainingAmount < dustLimit:
+		return 0, fmt.Errorf("selected amount %v leaves dust change "+
+			"%v", selectedAmount, remainingAmount)
+
```
starius, line 934: Other edge cases:
 - selectedAmount < 0
 - 0 < selectedAmount < dustLimit
 - totalDepositAmount < dustLimit

I propose to add a unit test for this function to cover all the cases.

hieblmi:
It is included in this PR. I changed the PR description.

staticaddr/loopin/manager.go
```diff
@@ -619,8 +713,9 @@ func (m *Manager) initiateLoopIn(ctx context.Context,
 	}
 
 	swap := &StaticAddressLoopIn{
-		DepositOutpoints:      req.DepositOutpoints,
-		Deposits:              deposits,
+		SelectedAmount:        req.SelectedAmount,
+		DepositOutpoints:      selectedOutpoints,
```
starius, line 766: BTW, can we replace field `DepositOutpoints` with getter method `DepositOutpoints()`?

staticaddr/loopin/manager_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/lightninglabs/loop/staticaddr/deposit"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/stretchr/testify/require"
+)
+
+type testCase struct {
```
starius, line 17: I propose to add `name string` field and use it in `t.Run` to facilitate finding a failing test case.

staticaddr/loopin/manager_test.go
```diff
@@ -0,0 +1,126 @@
...
+		Value:              1_000_000,
+		ConfirmationHeight: 1000,
+	}, &deposit.Deposit{
+		Value:              2_000_000,
+		ConfirmationHeight: 2000,
+	}, &deposit.Deposit{
+		Value:              3_000_000,
+		ConfirmationHeight: 3000,
+	},
+		&deposit.Deposit{
```
starius: inconsistency in formatting.

I propose to change it to separate assignments:
```go
var (
  d1 = ...
  d2 = ...
  ...
)
```

loopd/swapclient_server.go
```diff
@@ -893,22 +893,62 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 	infof("Loop in quote request received")
 
 	var (
-		numDeposits = uint32(len(req.DepositOutpoints))
-		err         error
+		selectedAmount     = btcutil.Amount(req.Amt)
+		totalDepositAmount btcutil.Amount
+		autoSelectDeposits = req.SelectDeposits
```
starius: "autoselect" is a better name for this feature than just "select". Since "select" may also mean "manual select". Let's rename it everywhere to autoselect?

loopd/swapclient_server.go
```diff
@@ -2284,10 +2343,21 @@ func validateConfTarget(target, defaultTarget int32) (int32, error) {
 // validateLoopInRequest fails if the mutually exclusive conf target and
 // external parameters are both set.
 func validateLoopInRequest(htlcConfTarget int32, external bool,
-	numDeposits uint32, amount int64) (int32, error) {
+	numDeposits uint32, amount int64, autoSelectDeposits bool) (int32,
```
starius: s/amount int64/amount btcutil.Amount/

Please update godoc and add what this functino returns.

staticaddr/loopin/manager.go
```diff
@@ -710,6 +805,85 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
+			return iExp < jExp
+		}
+		return deposits[i].Value > deposits[j].Value
+	})
+
+	// Select the deposits that are needed to cover the swap amount without
+	// leaving a dust change.
+	var selectedDeposits []*deposit.Deposit
+	var selectedAmount btcutil.Amount
+	dustLimit := lnwallet.DustLimitForSize(input.P2TRSize)
```
starius: Can factor out `dustLimit`.

cmd/loop/staticaddr.go
```diff
@@ -478,6 +478,13 @@ var staticAddressLoopInCommand = cli.Command{
 				"The client can retry the swap with adjusted " +
 				"parameters after the payment timed out.",
 		},
+		cli.IntFlag{
```
starius: Uint64Flag?
To make sure it is non-negative.

starius (approved):
**LGTM!** Great work! 💾
I left several comments.

looprpc/client.proto
hieblmi: ~Great suggestion! I've added an enum `FundingOption` that covers these cases. It is evaluated in `GetLoopInQuote`.~

I've renamed `select_deposits` to `auto_select_deposits` and added clarifying comments on the usage of it. If we introduce a enum type instead of the bool we will have to evaluate enums against paramters set in `deposit_outpoints` and `amt`, which I find confusing. So I'd rely on the clarifying comments, unless there is a strong opinion to change it.

staticaddr/loopin/loopin.go
hieblmi: > I guess in the multi address world this will be fresh static address, right?
> 
It will be, yes. I am working on this on an offline PR.

> I propose to prepare a bit: add a field holding change pkscript (and for now just assign it to `l.AddressParams.PkScript`). We should use it everywhere and pass instead of `hasChange` to `htlcWeight`. Then we don't have to assume there that is is P2TR and we can use method [weightEstimator.AddOutput](https://pkg.go.dev/github.com/lightningnetwork/lnd/input#TxWeightEstimator.AddOutput).

I will leave this for the future PR, just so that we don't overload this already full PR.

staticaddr/loopin/manager.go
hieblmi: I will address this in a follow-up PR.

staticaddr/loopin/manager.go
hieblmi: done

staticaddr/loopin/manager.go
hieblmi: Certainly, I added a TODO to in `StaticAddressLoopIn`.

staticaddr/loopin/manager.go
hieblmi: done

staticaddr/loopin/manager.go
hieblmi: done

staticaddr/loopin/manager.go
hieblmi: done

staticaddr/loopin/manager.go
hieblmi: makes sense, changed to `DeduceSwapAmount`.

staticaddr/loopin/manager.go
hieblmi: Thanks for listing those, I added all checks and added unit tests for those.

staticaddr/loopin/manager_test.go
hieblmi: added

staticaddr/loopin/manager_test.go
hieblmi: Fixed the formatting

loopd/swapclient_server.go
hieblmi: done

loopd/swapclient_server.go
hieblmi: done

staticaddr/loopin/selected_amount_migration.go
```diff
@@ -0,0 +1,84 @@
...
+				return fmt.Errorf("deposit for outpoint %s "+
+					"not found", outpoint)
+			}
+
+			// Set the selected amount to the value of the deposit.
+			updateAmounts[swap.SwapHash] += deposit.Value
+		}
+	}
+
+	log.Infof("Updating selected swap amounts for %d loop out swaps",
```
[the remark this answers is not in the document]
hieblmi: fixed

cmd/loop/staticaddr.go
hieblmi: done

staticaddr/loopin/selected_amount_migration.go
bhandras: So it's possible that we crash before we call `db.SetMigration()`, but now double checked and iiuc it's harmess as we'd just set the selected_amount again to sum of used deposits, so should be good 👍

bhandras (approved):
Awesome work! LGTM 🎉

cmd/loop/staticaddr.go
starius: I have the code for coin selection based on knapsack algorithm. It finds a combination of coins such minimum sum which is higher or equal to the target amount. The goal is to avoid change creation. But sum can be slightly higher.

If we use such an algorithm here, what do we do if it finds a combination which is slightly larger than needed? It would make sense to increase the swap amount, but this may be undesirable by the user in edge cases (e.g. not enough inbound liquidity).

@hieblmi and me discussed this offline. We can postpone implementing knapsack algorithm for a future PR.

What we can do do address the UX issue is to provide a separate command for selecting coins:
```
$ loop static selectdeposits 500000
Found a combination of deposits of total amount 500100 sats:
...
To loop-in them, run the following command: ...
```

litbot-9000:

staticaddr/loopin/manager.go
hieblmi: I've added method `checkChange` that ensures that for all inputs that associate a swap there is the expected change output per swap. Since there is only one static address at the moment we can sum up all change of associated swaps and check if there is a consolidated change output for us.

sweepbatcher/sweep_batcher.go
```diff
@@ -1354,9 +1354,8 @@ func (b *Batcher) monitorConfAndNotify(ctx context.Context, sweep *sweep,
 
 	batchTxid := spendTx.TxHash()
 
-	if len(spendTx.TxOut) != 1 {
-		return fmt.Errorf("unexpected number of outputs in batch: %d, "+
-			"want %d", len(spendTx.TxOut), 1)
+	if len(spendTx.TxOut) < 1 {
+		return fmt.Errorf("expected at least one output in batch")
```
starius, line 1358: :+1:

staticaddr/loopin/manager.go
```diff
@@ -465,6 +446,73 @@ func (m *Manager) handleLoopInSweepReq(ctx context.Context,
...
+		}
+	}
+
+	if expectedChange == 0 {
+		return nil
+	}
+
+	for _, out := range sweepTx.TxOut {
+		if out.Value == int64(expectedChange) &&
+			bytes.Equal(out.PkScript, changeAddr.PkScript) {
```
starius, line 505: This will be updated in multi-address world, right? There might be multiple change outputs.

We can write the code already with multi-address in mind. We just need to collect change addresses and change amounts from all the `loopIn` variables in the previous for loop and verify that each of them is present. In single address world this would work identically to the current implementation.

starius (approved):
**LGTM!** 💾
I checked two last commits.
I propose to adjust the code so it works both now and in multi-address world.

staticaddr/loopin/manager.go
hieblmi: Yes I had that in mind and agree, however, this requires some refactoring around loading the address parameters per swap which is introduced in the multi-address PRs. So I'd like to defer that.

PR #888, opened 2025-02-21
sweepbatcher: add mode with presigned transactions and CPFP

starius:
In this mode sweepbatcher uses transactions provided by the CPFP helper, which may be pre-signed and not replace-by-fee (RBF) compatible. In such cases, CPFP may be necessary.

A single Batcher instance can handle both CPFP and regular batches. Currently CPFP and non-CPFP sweeps never appear in the same batch.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/cpfp.go
+++ b/sweepbatcher/cpfp.go
@@ -1,0 +1,651 @@
+package sweepbatcher
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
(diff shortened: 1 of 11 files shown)


levmi:
Can we close this PR since we have chosen a different path forward IIUC?

starius:
Replaced with https://github.com/lightninglabs/loop/pull/891

PR #890, opened 2025-02-26
multi: fix race conditions and run unit tests with -race

starius:
#### Pull Request Checklist

### DIFF
--- a/staticaddr/address/manager.go
+++ b/staticaddr/address/manager.go
@@ -57,10 +57,13 @@ type Manager struct {
 }
 
 // NewManager creates a new address manager.
-func NewManager(cfg *ManagerConfig) *Manager {
-	return &Manager{
+func NewManager(cfg *ManagerConfig, currentHeight int32) *Manager {
+	m := &Manager{
 		cfg: cfg,
 	}
+	m.currentHeight.Store(currentHeight)
+
+	return m
 }
 
 // Run runs the address manager.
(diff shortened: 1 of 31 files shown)


staticaddr/address/manager.go
```diff
@@ -74,7 +75,7 @@ func (m *Manager) Run(ctx context.Context) error {
 	for {
 		select {
 		case currentHeight := <-newBlockChan:
-			m.currentHeight = currentHeight
+			m.currentHeight.Store(currentHeight)
```
bhandras, line 78: I think we should also initialize the current height by querying best height at the start on `Run`.

staticaddr/address/manager.go
```diff
@@ -111,7 +112,7 @@ func (m *Manager) NewAddress(ctx context.Context) (*btcutil.AddressTaproot,
 	m.Unlock()
 
 	// Ensure that we have that we have a sane current block height.
-	if m.currentHeight == 0 {
+	if m.currentHeight.Load() == 0 {
```
bhandras, line 115: This check could be removed if current height is correctly initialized.

bhandras (approved):
LGTM 🎉

staticaddr/address/manager.go
starius: `Run` already does this as soon as possible. Actually this is the only thing Run does - keeping `currentHeight` up-to-date.

staticaddr/address/manager.go
starius: `Run` runs in background, so the caller doesn't know when `Run` has initialized `currentHeight`.
Additionally, it is better to double check even if it was initialized before, to detect a wrong order of methods called.

staticaddr/address/manager.go
```diff
@@ -52,14 +53,17 @@ type Manager struct {
...
 	sync.Mutex
 
-	currentHeight int32
+	currentHeight atomic.Int32
 }
 
 // NewManager creates a new address manager.
-func NewManager(cfg *ManagerConfig) *Manager {
-	return &Manager{
+func NewManager(cfg *ManagerConfig, currentHeight int32) *Manager {
```
bhandras, line 60: Nice, thanks!

bhandras (approved):

starius:
Rebased

hieblmi (approved):
LGTM!

PR #891, opened 2025-02-27
sweepbatcher: add mode with presigned transactions

starius:
In this mode sweepbatcher uses transactions provided by the Presigned helper.
Transactions are signed upon adding an input to a batch.

A single Batcher instance can handle both presigned and regular batches.
Currently presigned and non-presigned sweeps never appear in the same batch.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -1,0 +1,630 @@
+package sweepbatcher
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ensurePresigned checks that there is a presigned transaction spending the
+// inputs of this group only. If allowNonEmptyBatch is false, the batch must be
+// empty.
+func (b *batch) ensurePresigned(ctx context.Context, newSweeps []*sweep,
+	allowNonEmptyBatch bool) error {
+
+	if b.cfg.presignedHelper == nil {
+		return fmt.Errorf("presignedHelper is not installed")
+	}
+	if len(b.sweeps) != 0 && !allowNonEmptyBatch {
+		return fmt.Errorf("ensurePresigned should be done when " +
+			"adding to an empty batch")
+	}
+
+	return ensurePresigned(
+		ctx, newSweeps, b.cfg.presignedHelper, b.cfg.chainParams,
+	)
+}
+
+// presignedTxChecker has methods to check if the inputs are presigned.
+type presignedTxChecker interface {
+	destPkScripter
+
+	// SignTx signs an unsigned transaction or returns a pre-signed tx.
+	// It is only called with loadOnly=true by ensurePresigned.
+	SignTx(ctx context.Context, primarySweepID wire.OutPoint,
+		tx *wire.MsgTx, inputAmt btcutil.Amount,
+		minRelayFee, feeRate chainfee.SatPerKWeight,
+		loadOnly bool) (*wire.MsgTx, error)
+}
+
+// ensurePresigned checks that there is a presigned transaction spending the
+// inputs of this group only.
+func ensurePresigned(ctx context.Context, newSweeps []*sweep,
+	presignedTxChecker presignedTxChecker,
+	chainParams *chaincfg.Params) error {
+
+	sweeps := make([]sweep, len(newSweeps))
+	for i, s := range newSweeps {
+		sweeps[i] = sweep{
+			outpoint:  s.outpoint,
+			value:     s.value,
+			presigned: s.presigned,
+		}
+	}
+
+	// The sweeps are ordered inside the group, the first one is the primary
+	// outpoint in the batch.
+	primarySweepID := sweeps[0].outpoint
+
+	// Cache the destination address.
+	destAddr, err := getPresignedSweepsDestAddr(
+		ctx, presignedTxChecker, primarySweepID, chainParams,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to find destination address: %w", err)
+	}
+
+	// Set LockTime to 0. It is not critical.
+	const currentHeight = 0
+
+	// Check if we can sign with minimum fee rate.
+	const feeRate = chainfee.FeePerKwFloor
+
+	tx, _, _, _, err := constructUnsignedTx(
+		sweeps, destAddr, currentHeight, feeRate,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to construct unsigned tx "+
+			"for feeRate %v: %w", feeRate, err)
+	}
+
+	// Check of a presigned transaction exists.
+	var batchAmt btcutil.Amount
+	for _, sweep := range newSweeps {
+		batchAmt += sweep.value
+	}
+	const loadOnly = true
+	signedTx, err := presignedTxChecker.SignTx(
+		ctx, primarySweepID, tx, batchAmt, feeRate, feeRate, loadOnly,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to find a presigned transaction "+
+			"for feeRate %v, txid of the template is %v, inputs: %d, "+
+			"outputs: %d: %w", feeRate, tx.TxHash(),
+			len(tx.TxIn), len(tx.TxOut), err)
+	}
+
+	// Check the SignTx worked correctly.
+	err = CheckSignedTx(tx, signedTx, batchAmt, feeRate)
+	if err != nil {
+		return fmt.Errorf("signed tx doesn't correspond the "+
+			"unsigned tx: %w", err)
+	}
+
+	return nil
+}
+
+// getOrderedSweeps returns the sweeps of the batch in the order they were
+// added. The method must be called from the event loop of the batch.
+func (b *batch) getOrderedSweeps(ctx context.Context) ([]sweep, error) {
+	// We use the DB just to know the order. Sweeps are copied from RAM.
+	utxo2sweep := make(map[wire.OutPoint]sweep, len(b.sweeps))
+	for _, s := range b.sweeps {
+		utxo2sweep[s.outpoint] = s
+	}
+
+	dbSweeps, err := b.store.FetchBatchSweeps(ctx, b.id)
+	if err != nil {
+		return nil, fmt.Errorf("FetchBatchSweeps(%d) failed: %w", b.id,
+			err)
+	}
+	if len(dbSweeps) != len(utxo2sweep) {
+		return nil, fmt.Errorf("FetchBatchSweeps(%d) returned %d "+
+			"sweeps, len(b.sweeps) is %d", b.id, len(dbSweeps),
+			len(utxo2sweep))
+	}
+
+	orderedSweeps := make([]sweep, len(dbSweeps))
+	for i, dbSweep := range dbSweeps {
+		// Sanity check: make sure dbSweep.ID grows.
+		if i > 0 && dbSweep.ID <= dbSweeps[i-1].ID {
+			return nil, fmt.Errorf("sweep ID does not grow: %d->%d",
+				dbSweeps[i-1].ID, dbSweep.ID)
+		}
+
+		s, has := utxo2sweep[dbSweep.Outpoint]
+		if !has {
+			return nil, fmt.Errorf("FetchBatchSweeps(%d) returned "+
+				"unknown sweep %v", b.id, dbSweep.Outpoint)
+		}
+		orderedSweeps[i] = s
+	}
+
+	return orderedSweeps, nil
+}
+
+// getSweepsGroups returns groups in which sweeps were added to the batch.
+// All the sweeps are sorted by addition order and grouped by swap.
+// The method must be called from the event loop of the batch.
+func (b *batch) getSweepsGroups(ctx context.Context) ([][]sweep, error) {
+	orderedSweeps, err := b.getOrderedSweeps(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getOrderedSweeps(%d) failed: %w", b.id,
+			err)
+	}
+
+	groups := [][]sweep{}
+	for _, s := range orderedSweeps {
+		index := len(groups) - 1
+
+		// Start new group if there are no groups or new swap starts.
+		if len(groups) == 0 || s.swapHash != groups[index][0].swapHash {
+			groups = append(groups, []sweep{})
+			index++
+		}
+
+		groups[index] = append(groups[index], s)
+	}
+
+	// Sanity check: make sure the number of groups is the same as the
+	// number of distinct swaps.
+	swapsSet := make(map[lntypes.Hash]struct{}, len(groups))
+	for _, s := range orderedSweeps {
+		swapsSet[s.swapHash] = struct{}{}
+	}
+	if len(swapsSet) != len(groups) {
+		return nil, fmt.Errorf("batch %d: there are %d groups of "+
+			"sweeps and %d distinct swaps", b.id, len(groups),
+			len(swapsSet))
+	}
+
+	return groups, nil
+}
+
+// presign tries to presign batch sweep transactions composed of this batch and
+// the sweep. In addition to that it presigns sweep transactions for any subset
+// of sweeps that could remain if one of the sweep transactions gets confirmed.
+// This can be done efficiently, since we keep track of the order in which
+// sweeps are added and the associated swap hashes. So we presign transactions
+// sweeping all the sweeps starting at some past sweeps group. For each inputs
+// layout it presigns many transactions with different fee rates.
+func (b *batch) presign(ctx context.Context, newSweeps []*sweep) error {
+	if b.cfg.presignedHelper == nil {
+		return fmt.Errorf("presignedHelper is not installed")
+	}
+	if len(b.sweeps) == 0 {
+		return fmt.Errorf("presigning should be done when adding to " +
+			"a non-empty batch")
+	}
+
+	// priorityConfTarget defines the confirmation target for quick
+	// inclusion in a block. A value of 2, rather than 1, is used to prevent
+	// fee estimator from failing.
+	// See https://github.com/lightninglabs/loop/issues/898
+	const priorityConfTarget = 2
+
+	// Find the feerate needed to get into next block.
+	nextBlockFeeRate, err := b.wallet.EstimateFeeRate(
+		ctx, priorityConfTarget,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to get nextBlockFeeRate: %w", err)
+	}
+
+	b.Infof("nextBlockFeeRate is %v", nextBlockFeeRate)
+
+	// We need to restore previously added groups. We can do it by reading
+	// all the sweeps from DB (they must be ordered) and grouping by swap.
+	groups, err := b.getSweepsGroups(ctx)
+	if err != nil {
+		return fmt.Errorf("getSweepsGroups failed: %w", err)
+	}
+	if len(groups) == 0 {
+		return fmt.Errorf("getSweepsGroups returned no sweeps groups")
+	}
+
+	// Now presign a transaction spending a suffix of groups as well as new
+	// sweeps. Any non-empty suffix of groups may remain non-swept after
+	// some past tx is confirmed.
+	for len(groups) != 0 {
+		// Create the list of sweeps from the remaining groups and new
+		// sweeps.
+		sweeps := make([]sweep, 0, len(b.sweeps)+len(newSweeps))
+		for _, group := range groups {
+			sweeps = append(sweeps, group...)
+		}
+		for _, sweep := range newSweeps {
+			sweeps = append(sweeps, *sweep)
+		}
+
+		// The primarySweepID is the first sweep from the list of
+		// remaining sweeps if previous groups are confirmed.
+		primarySweepID := sweeps[0].outpoint
+
+		// Cache the destination address.
+		destAddr, err := getPresignedSweepsDestAddr(
+			ctx, b.cfg.presignedHelper, b.primarySweepID,
+			b.cfg.chainParams,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to find destination "+
+				"address: %w", err)
+		}
+
+		err = presign(
+			ctx, b.cfg.presignedHelper, destAddr, primarySweepID,
+			sweeps, nextBlockFeeRate,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to presign a transaction "+
+				"of %d sweeps: %w", len(sweeps), err)
+		}
+
+		// Cut a group to proceed to next suffix of original groups.
+		groups = groups[1:]
+	}
+
+	// Ensure that a batch spending new sweeps only has been presigned by
+	// PresignSweepsGroup.
+	const allowNonEmptyBatch = true
+	err = b.ensurePresigned(ctx, newSweeps, allowNonEmptyBatch)
+	if err != nil {
+		return fmt.Errorf("new sweeps were not presigned; this means "+
+			"that PresignSweepsGroup was not called prior to "+
+			"AddSweep for the group: %w", err)
+	}
+
+	return nil
+}
+
+// presigner tries to presign a batch transaction.
+type presigner interface {
+	// Presign tries to presign a batch transaction. If the method returns
+	// nil, it is guaranteed that future calls to SignTx on this set of
+	// sweeps return valid signed transactions.
+	Presign(ctx context.Context, primarySweepID wire.OutPoint,
+		tx *wire.MsgTx, inputAmt btcutil.Amount) error
+}
+
+// presign tries to presign batch sweep transactions of the sweeps. It signs
+// multiple versions of the transaction to make sure there is a transaction to
+// be published if minRelayFee grows. If feerate is high, then a presigned tx
+// gets LockTime equal to timeout minus 50 blocks, as a precautionary measure.
+// A feerate is considered high if it is at least 100 sat/vbyte AND is at least
+// 10x of the current next block feerate.
+func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
+	primarySweepID wire.OutPoint, sweeps []sweep,
+	nextBlockFeeRate chainfee.SatPerKWeight) error {
+
+	if presigner == nil {
+		return fmt.Errorf("presigner is not installed")
+	}
+
+	if len(sweeps) == 0 {
+		return fmt.Errorf("there are no sweeps")
+	}
+
+	if nextBlockFeeRate == 0 {
+		return fmt.Errorf("nextBlockFeeRate is not set")
+	}
+
+	// Keep track of the total amount this batch is sweeping back.
+	batchAmt := btcutil.Amount(0)
+	for _, sweep := range sweeps {
+		batchAmt += sweep.value
+	}
+
+	// Find the sweep with the earliest expiry.
+	timeout := sweeps[0].timeout
+	for _, sweep := range sweeps[1:] {
+		timeout = min(timeout, sweep.timeout)
+	}
+	if timeout <= 0 {
+		return fmt.Errorf("timeout is invalid: %d", timeout)
+	}
+
+	// Go from the floor (1.01 sat/vbyte) to 2k sat/vbyte with step of 1.2x.
+	const (
+		start            = chainfee.FeePerKwFloor
+		stop             = chainfee.AbsoluteFeePerKwFloor * 2_000
+		factorPPM        = 1_200_000
+		timeoutThreshold = 50
+	)
+
+	// Calculate the locktime value to use for high feerate transactions.
+	// If timeout <= timeoutThreshold, don't set LockTime (keep value 0).
+	var highFeeRateLockTime uint32
+	if timeout > timeoutThreshold {
+		highFeeRateLockTime = uint32(timeout - timeoutThreshold)
+	}
+
+	// Calculate which feerate to consider high. At least 100 sat/vbyte and
+	// at least 10x of current nextBlockFeeRate.
+	highFeeRate := max(100*chainfee.FeePerKwFloor, 10*nextBlockFeeRate)
+
+	// Set LockTime to 0. It is not critical.
+	const currentHeight = 0
+
+	for fr := start; fr <= stop; fr = (fr * factorPPM) / 1_000_000 {
+		// Construct an unsigned transaction for this fee rate.
+		tx, _, feeForWeight, fee, err := constructUnsignedTx(
+			sweeps, destAddr, currentHeight, fr,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to construct unsigned tx "+
+				"for feeRate %v: %w", fr, err)
+		}
+
+		// If the feerate is high enough, set locktime to prevent
+		// broadcasting such a transaction too early by mistake.
+		if fr >= highFeeRate {
+			tx.LockTime = highFeeRateLockTime
+		}
+
+		// Try to presign this transaction.
+		err = presigner.Presign(ctx, primarySweepID, tx, batchAmt)
+		if err != nil {
+			return fmt.Errorf("failed to presign unsigned tx %v "+
+				"for feeRate %v: %w", tx.TxHash(), fr, err)
+		}
+
+		// If fee was clamped, stop here, because fee rate won't grow.
+		if fee < feeForWeight {
+			break
+		}
+	}
+
+	return nil
+}
+
+// publishPresigned creates sweep transaction using a custom transaction signer
+// and publishes it. It returns the fee of the transaction, and an error (if
+// signing and/or publishing failed) and a boolean flag indicating signing
+// success. This mode is incompatible with an external address.
+func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
+	bool) {
+
+	// Sanity check, there should be at least 1 sweep in this batch.
+	if len(b.sweeps) == 0 {
+		return 0, fmt.Errorf("no sweeps in batch"), false
+	}
+
+	// Make sure that no external address is used.
+	for _, sweep := range b.sweeps {
+		if sweep.isExternalAddr {
+			return 0, fmt.Errorf("external address was used with " +
+				"a custom transaction signer"), false
+		}
+	}
+
+	// Cache current height and desired feerate of the batch.
+	currentHeight := b.currentHeight
+	feeRate := b.rbfCache.FeeRate
+
+	// Append this sweep to an array of sweeps. This is needed to keep the
+	// order of sweeps stored, as iterating the sweeps map does not
+	// guarantee same order.
+	sweeps := make([]sweep, 0, len(b.sweeps))
+	for _, sweep := range b.sweeps {
+		sweeps = append(sweeps, sweep)
+	}
+
+	// Cache the destination address.
+	address, err := getPresignedSweepsDestAddr(
+		ctx, b.cfg.presignedHelper, b.primarySweepID,
+		b.cfg.chainParams,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("failed to find destination address: %w",
+			err), false
+	}
+
+	// Construct unsigned batch transaction.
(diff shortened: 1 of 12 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -566,6 +580,54 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
...
+			if err := b.presign(ctx, sweep); err != nil {
+				b.log().Infof("failed to add sweep %x to the "+
+					"batch, because failed to presign new "+
+					"version of batch tx: %v",
+					sweep.swapHash[:6], err)
+
+				return false, nil
+			}
+		} else {
+			if err := b.ensurePresigned(ctx, sweep); err != nil {
```
hieblmi, line 609: if `err == nil` here, don't we then have to call `b.presign` for the sweep here?

sweepbatcher/sweep_batch.go
```diff
@@ -566,6 +580,54 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
 		}
 	}
 
+	// If presigned mode is enabled, we should first presign the new version
+	// of batch transaction. Also ensure that all the sweeps in the batch
+	// use the same mode (presigned or regular).
+	if sweep.presigned {
+		// Ensure that all the sweeps in the batch use presigned mode.
+		for _, s := range b.sweeps {
```
hieblmi, line 588: instead of iterating, do you think it makes sense to add a `presigned` flag to the batch struct that we can check before adding a sweep to a batch?

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,441 @@
...
+		return fmt.Errorf("nextBlockFeerate is not set")
+	}
+
+	// Keep track of the total amount this batch is sweeping back.
+	batchAmt := btcutil.Amount(0)
+	for _, sweep := range sweeps {
+		batchAmt += sweep.value
+	}
+
+	// Find when at least one sweep expires.
```
hieblmi, line 149: nit: "find the sweep with the earliest expiry"

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,441 @@
...
+	if b.cfg.presignedHelper == nil {
+		return fmt.Errorf("presignedHelper is not installed")
+	}
+	if len(b.sweeps) == 0 {
+		return fmt.Errorf("presigning is done when adding to a " +
+			"non-empty batch")
+	}
+
+	// Find the feerate needed to get into next block. Use conf_target=2,
+	nextBlockFeerate, err := b.wallet.EstimateFeeRate(ctx, 2)
```
hieblmi, line 87: why is it 2? Maybe create a constant?

hieblmi (commented):
Amazing work @starius. I've left a couple of questions in my first pass.

sweepbatcher/sweep_batch.go
```diff
@@ -524,6 +560,54 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
...
+			if !s.presigned {
+				b.log().Infof("failed to add sweep %x to the "+
+					"batch, because the batch has "+
```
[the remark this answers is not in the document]

sweepbatcher/sweep_batch.go
```diff
@@ -524,6 +560,54 @@ func (b *batch) addSweep(ctx context.Context, sweep *sweep) (bool, error) {
...
+					"batcher.PresignSweep was not called "+
+					"prior to AddSweep for this input: %w",
+					sweep.swapHash[:6], err)
+			}
+		}
+	} else {
+		// Ensure that all the sweeps in the batch don't use presigned.
+		for _, s := range b.sweeps {
+			if s.presigned {
+				b.log().Infof("failed to add sweep %x to the "+
```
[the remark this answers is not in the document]

sweepbatcher/sweep_batcher.go
```diff
@@ -564,8 +628,40 @@ func (b *Batcher) Run(ctx context.Context) error {
 	}
 }
 
+// PresignSweep creates and stores presigned 1:1 transactions for the sweep.
+// This method must be called prior to AddSweep if presigned mode is enabled.
```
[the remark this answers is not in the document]
starius: I updated `ensurePresigned` and SignTx interface (added argument `presignedOnly`), so `ensurePresigned` actually ensured that it is presigned, not tried to sign (become read-only). Also added a test making sure that the batcher crashes if AddSweep is called before calling PresignSweep (for a presigned sweep). (AddSweep can't return that error, since it only pushes a sweep to the event loop.)

sweepbatcher/sweep_batch.go
starius: If `ensurePresigned` returned no error, this means, it has been presigned already. No need to presign again. If not presigned, then crash the batcher (this is only possible because of a bug).

sweepbatcher/presigned.go
starius: Fixed

sweepbatcher/presigned.go
starius: I updated this place:
```go
        // priorityConfTarget defines the confirmation target for quick
        // inclusion in a block. A value of 2, rather than 1, is used to prevent
        // overestimation caused by temporary anomalies, such as periods without
        // blocks.
        const priorityConfTarget = 2

        // Find the feerate needed to get into next block.
        nextBlockFeeRate, err := b.wallet.EstimateFeeRate(
                ctx, priorityConfTarget,
        )
```

This fee rate is only used to define a high fee rate (10x of this value and at least 100 sats/vbyte).

starius: BTW the fee estimator fails with conf_target=1: https://github.com/lightninglabs/loop/issues/898
Maybe it worth making this a global constant in Loop and use it in all such places.

starius: Updated the comment:
```go
        // priorityConfTarget defines the confirmation target for quick
        // inclusion in a block. A value of 2, rather than 1, is used to prevent
        // fee estimator from failing.
        // See https://github.com/lightninglabs/loop/issues/898
        const priorityConfTarget = 2
```

sweepbatcher/sweep_batch.go
```diff
@@ -1960,7 +2116,21 @@ func (b *batch) persist(ctx context.Context) error {
 
 // getBatchDestAddr returns the batch's destination address. If the batch
 // has already generated an address then the same one will be returned.
+// The method must not be used in presigned mode. Use getSweepsDestAddr instead.
```
hieblmi, line 2119: should we call this method in the comment `getPresignedSweepsDestAddr` to clarify?

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+	// purposes.
+	b.batchTxid = &txHash
+	b.batchPkScript = tx.TxOut[0].PkScript
+
+	return fee, nil, true
+}
+
+// getSweepsDestAddr returns the destination address used by a group of sweeps.
+// The method must be used in presigned mode only.
+func (b *batch) getSweepsDestAddr(ctx context.Context,
```
hieblmi, line 345: could call it `getPresignedSweepsDestAddr` to make it clearer that this only makes sense in the context of presigned batches.

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ensurePresigned checks that we can sign a 1:1 transaction sweeping the input.
+func (b *batch) ensurePresigned(ctx context.Context, newSweeps []*sweep) error {
+	if b.cfg.presignedHelper == nil {
+		return fmt.Errorf("presignedHelper is not installed")
+	}
+	if len(b.sweeps) != 0 {
+		return fmt.Errorf("ensurePresigned is done when adding to an " +
```
hieblmi, line 22: nit: shoud be done...

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+	var batchAmt btcutil.Amount
+	for _, sweep := range newSweeps {
+		batchAmt += sweep.value
+	}
+	const presignedOnly = true
+	signedTx, err := b.cfg.presignedHelper.SignTx(
+		ctx, tx, batchAmt, feeRate, feeRate, presignedOnly,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to sign unsigned tx %v "+
```
hieblmi, line 65: nit: failed to load a presigned tx?

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+		return fmt.Errorf("failed to construct unsigned tx "+
+			"for feeRate %v: %w", feeRate, err)
+	}
+
+	// Check of a presigned transaction exists.
+	var batchAmt btcutil.Amount
+	for _, sweep := range newSweeps {
+		batchAmt += sweep.value
+	}
+	const presignedOnly = true
```
hieblmi, line 60: maybe we could call this flag `loadOnly` or something similar?

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ensurePresigned checks that we can sign a 1:1 transaction sweeping the input.
+func (b *batch) ensurePresigned(ctx context.Context, newSweeps []*sweep) error {
```
hieblmi, line 17: does `ensurePresigned` imply that there should be already be a transaction with given inputs that has a populated witness? I don't see a check anywhere in this method or sub-method that checks for a witness.

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+		}
+
+		// If the feerate is high enough, set locktime to prevent
+		// broadcasting such a transaction too early by mistake.
+		if fr >= highFeeRate {
+			tx.LockTime = highFeeRateLockTime
+		}
+
+		// Try to presign this transaction.
+		err = presigner.Presign(ctx, tx, batchAmt)
```
hieblmi, line 208: Just a side question: Currently the result is not yet used. Is it intended to then pass this pre-signed transaction into our store? I guess once the batch is confirmed we could clean up all the presigned fee versions.

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	// Get a signed transaction. It may be either new transaction or a
+	// pre-signed one.
+	const presignedOnly = false
+	signedTx, err := b.cfg.presignedHelper.SignTx(
+		ctx, tx, batchAmt, minRelayFee, feeRate, presignedOnly,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("failed to sign tx: %w", err),
```
hieblmi, line 294: is it trying to sign when `presignedOnly` is passed in?

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,457 @@
...
+		return nil, fmt.Errorf("pkScript.Address failed for "+
+			"pkScript %x returned for inputs %v: %w", pkScriptBytes,
+			inputs, err)
+	}
+
+	return address, nil
+}
+
+// CheckSignedTx makes sure that signedTx matches the unsignedTx. It checks
+// according to criteria specified in the description of PresignedHelper.SignTx.
```
hieblmi, line 389: looks like the input sequence numbers aren't checked as described in the godoc of `SignTx`.

hieblmi (commented):
It's great work @starius, I've left a couple of more comments, looks almost good to me.

sweepbatcher/sweep_batcher.go
```diff
@@ -156,6 +156,53 @@ type SignMuSig2 func(ctx context.Context, muSig2Version input.MuSig2Version,
...
+
+	// Presign tries to presign a batch transaction. If the method returns
+	// nil, it is guaranteed that future calls to SignTx on this set of
+	// sweeps return valid signed transactions.
+	Presign(ctx context.Context, tx *wire.MsgTx,
+		inputAmt btcutil.Amount) error
+
+	// DestPkScript returns destination pkScript used in a presigned
+	// transaction sweeping the inputs. Returns an error, if such tx
+	// doesn't exist. If there are many such transactions, returns any
```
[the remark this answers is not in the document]
starius: Yes, all of the transactions should have the same destination address, by construction. However this is not enforced by `PresignedHelper`. In theory someone can call `Presign` on transactions with the same list of inputs and different destination addresses and all of the addresses would be considered valid. We can demand `PresignedHelper` to check that address is always the same, but this would make queries heavier. (Checking in `Presign` that destination address is the same in all presigned transactions with such inputs.)

sweepbatcher/sweep_batch.go
```diff
@@ -587,6 +602,80 @@ func (b *batch) addSweeps(ctx context.Context, sweeps []*sweep) (bool, error) {
...
+			if err := b.presign(ctx, sweeps); err != nil {
+				b.Warnf("Failed to add sweep %x to the batch, "+
+					"because failed to presign new version"+
+					" of batch tx: %v",
+					sweeps[0].swapHash[:6], err)
+
+				return false, nil
+			}
+
+		// If this is new batch being formed, make sure we already have
```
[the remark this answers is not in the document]

sweepbatcher/sweep_batch.go
starius: Renamed. Thanks!

sweepbatcher/presigned.go
starius: Renamed

sweepbatcher/presigned.go
starius: Fixed

sweepbatcher/presigned.go
starius: Fixed:

```go
                return fmt.Errorf("failed to find a presigned transaction "+
                        "for feeRate %v, txid of the template is %v: %w",
                        feeRate, tx.TxHash(), err)
```

sweepbatcher/presigned.go
starius: Renamed to `loadOnly`.

sweepbatcher/presigned.go
starius: I added a check to `CheckSignedTx` which is called by this method:

```go
// Make sure all inputs of signedTx have a non-empty witness.
for _, txIn := range signedTx.TxIn {
	if len(txIn.Witness) == 0 {
		return fmt.Errorf("input %s of signed tx is not signed",
			txIn.PreviousOutPoint)
	}
}
```

sweepbatcher/presigned.go
starius: `PresignHelper` is responsible for storing the transactions. SweepBatcher relies on `helper.SignTx` to be able to load transactions stored by previous `helper.Presign` and `helper.SignTx` calls.

> I guess once the batch is confirmed we could clean up all the presigned fee versions.

This is done. There is the method `CleanupTransactions` which is called by `batch.handleConf` method.

sweepbatcher/presigned.go
starius: `presignedOnly = false` in this call, so it tries to sign interactively and if it fails, goes to the store of presigned transactions.

sweepbatcher/presigned.go
starius: Sequence numbers are compared using `unsignedMap`. Also there is a test case to verify this.

starius:
Found a couple of edge cases. Converting to a draft until the fix is ready.

starius:
Fixed a couple of edge cases:

1. The primary sweep ID (UTXO) is now passed along with presigned transactions to help `PresignedHelper` identify the batch. The primary sweep is always spent when the batch is spent. Other sweeps may remain unspent if an intermediate version of the batch is confirmed.

2. If some sweeps in a batch remain unspent after the batch is confirmed, we need to handle them properly. I updated the code to prepare presigned transactions in advance for the sweeps that can remain unspent.

**Example for the second issue:**  
`swap1` is added to the batch. We presign transaction for [`swap1`].

Then `swap2` is added. At this point, we already have a presigned transaction for [`swap2`] created by `PresignSweepsGroup`. When adding `swap2`, we also generate a presigned transaction for [`swap1`, `swap2`].

Then `swap3` is added. Which transactions should we presign?
- [`swap1`, `swap2`, `swap3`] — the full transaction  
- [`swap2`, `swap3`] — needed in case `swap1` is confirmed on its own  
- [`swap3`] — already created by `PresignSweepsGroup`, needed if [`swap1`, `swap2`] is confirmed  

I've added unit tests for all possible cases.

---

Changes since previous review round:
- added commit "sweepbatcher: auto-increment sweep.ID in StoreMock"
- many changes in the last commit, [the diff is here](https://github.com/starius/loop/compare/v0.30.0-presigned-beta..presigned-readd-fixed2)

starius:
Rebased and updated the last commit. In the `handleConf` function, we now use the transaction from the confirmation notification to retrieve the list of inputs for presigned storage cleanup, instead of relying on `b.sweeps`. This is a precaution because `b.sweeps` should contain the same set of inputs, as non-confirmed sweeps are removed in `handleSpend`. However, using the confirmed transaction is more reliable and clearer.

Also added a check to ensure that the transaction from the confirmation notification matches the one from the spend notification. If they differ, a warning is logged.

starius:
Rebased

sweepbatcher/sweep_batch.go
```diff
@@ -1814,41 +1952,92 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+		}
+
+		// In presigned mode we should form a SweepRequest per swap
+		// (i.e. per group) and keep them ordered.
+		L := len(purgeList)
+		if presigned && L != 0 &&
+			purgeList[L-1].SwapHash == newSweep.swapHash {
+
+			// Add the input to existing SweepRequest for this swap.
+			purgeList[L-1].Inputs = append(
```
hieblmi, line 2024: is it correct that if we have 3 pruged sweeps in presigned mode A, B, C, then `purgeList` will have slices [A], [A,B], [A,B,C]?

sweepbatcher/sweep_batch.go
```diff
@@ -1814,41 +1952,92 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+		} else {
 			purgeList = append(purgeList, SweepRequest{
 				SwapHash: newSweep.swapHash,
-				Inputs: []Input{
-					{
-						Outpoint: newSweep.outpoint,
-						Value:    newSweep.value,
-					},
-				},
+				Inputs:   []Input{newInput},
```
hieblmi, line 2030: so when we are in `presigned` mode and the purged list is empty we will land here the first time, right.

sweepbatcher/sweep_batch.go
```diff
@@ -1917,7 +2106,40 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+
+	// If the batch is in presigned mode, cleanup presignedHelper.
+	presigned, err := b.isPresigned()
+	if err != nil {
+		return fmt.Errorf("failed to determine if the batch %d uses "+
+			"presigned mode: %w", b.id, err)
+	}
+
+	if presigned {
+		b.Infof("Cleaning up presigned store")
```
hieblmi, line 2129: maybe add this new functionality to the godoc of this method.

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,602 @@
...
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ensurePresigned checks that we can sign a transaction sweeping the inputs of
+// this group only. This method is called when a group of sweeps of new swap is
+// being added to the batch. If the batch can be either empty or existing. This
+// is know to the caller and affects allowNonEmptyBatch flag.
```
hieblmi, line 20: nit: known

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,602 @@
...
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ensurePresigned checks that we can sign a transaction sweeping the inputs of
+// this group only. This method is called when a group of sweeps of new swap is
+// being added to the batch. If the batch can be either empty or existing. This
```
hieblmi, line 19: nit: comment seems off

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,602 @@
...
+				"unknown sweep %v", b.id, dbSweep.Outpoint)
+		}
+		orderedSweeps[i] = s
+	}
+
+	return orderedSweeps, nil
+}
+
+// getSweepsGroups returns groups in which sweeps were added to the batch.
+// All the sweeps are sorted by addition order and groupped by swap.
```
hieblmi, line 133: nit: groupped

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,602 @@
...
+	groups := [][]sweep{}
+	for _, s := range orderedSweeps {
+		// Start new group if there are no groups or new swap starts.
+		if len(groups) == 0 ||
+			s.swapHash != groups[len(groups)-1][0].swapHash {
+
+			groups = append(groups, []sweep{})
+		}
+
+		index := len(groups) - 1
```
hieblmi, line 151: move this to the start of the for loop and use it in the `if`.

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,602 @@
...
+		// remaining sweeps if previous groups are confirmed.
+		primarySweepID := sweeps[0].outpoint
+
+		// Cache the destination address.
+		destAddr, err := getPresignedSweepsDestAddr(
+			ctx, b.cfg.presignedHelper, b.primarySweepID,
+			b.cfg.chainParams,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to find destination address: %w", err)
```
hieblmi, line 236: nit: lll

hieblmi (commented):
Left some comments and questions for the latest set of changes.

sweepbatcher/presigned.go
starius: Fixed

sweepbatcher/presigned.go
starius: Fixed.

```go
groups := [][]sweep{}
for _, s := range orderedSweeps {
		index := len(groups) - 1

		// Start new group if there are no groups or new swap starts.
		if len(groups) == 0 || s.swapHash != groups[index][0].swapHash {
				groups = append(groups, []sweep{})
				index++
		}

		groups[index] = append(groups[index], s)
}
```

sweepbatcher/presigned.go
starius: Fixed

sweepbatcher/presigned.go
starius: Fixed.

```go
// ensurePresigned checks there there is a presigned transaction spending the
// inputs of this group only. If allowNonEmptyBatch is false, the batch must be
// empty.
func (b *batch) ensurePresigned(ctx context.Context, newSweeps []*sweep,
        allowNonEmptyBatch bool) error {
```

Also I factored out pure function `ensurePresigned` from the method and added a unit test for it.

sweepbatcher/presigned.go
starius: This part of the godoc was removed in the updated version.

sweepbatcher/sweep_batch.go
starius: Done.

```go
// batch. Here we signal to the batcher that this batch was completed. We also
// cleanup up presigned transactions whose primarySweepID is one of the sweeps
// that were spent and fully confirmed: such a transaction can't be broadcasted
// since it is either in a block or double-spends one of spent outputs.
func (b *batch) handleConf(ctx context.Context,
        conf *chainntnfs.TxConfirmation) error {
```

sweepbatcher/sweep_batch.go
starius: We can reach this line either in regular mode or in presigned mode in the beginning or when new swap starts.

I added a comment:

```go
} else {
		// Add the current sweep as a new element to purgeList.
		// This is possible either in regular mode or in
		// presigned mode in the beginning or on new swap.
		purgeList = append(purgeList, SweepRequest{
				SwapHash: newSweep.swapHash,
				Inputs:   []Input{newInput},
				Notifier: newSweep.notifier,
		})
}
```

sweepbatcher/sweep_batch.go
starius: `purgeList` won't have the same sweep in multiple slices. The code groups sweeps by swap hash. `purgeList` reproduces the arguments and the order of the original external `sweepbatcher.AddSweep` calls.

Check the example in `testPresigned_purging`. It adds multiple swaps, each of two sweeps. It adds the swaps to the batcher, then all endpoints go offline (form this point it relies only on presigned transactions). Then it confirmes sweeps of some swaps (sequentially from the first to some swap - because they are published in such order). The sweeps of remaining swaps are added one by one (swap by swap) back to the batcher. The first of the remaining swaps is added to new batch, then other swaps are appended to it. All the transactions for intermediate steps and to the final version were presigned.

I added the following comment:

```go
// In presigned mode we should form a SweepRequest per swap
// (i.e. per group) and keep them ordered. It should reproduce
// the arguments and the order of the original external AddSweep
// calls.
L := len(purgeList)
```

sweepbatcher/sweep_batch.go
```diff
@@ -1814,41 +1952,92 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
 		found := false
 
 		for _, txIn := range spendTx.TxIn {
 			if txIn.PreviousOutPoint == sweep.outpoint {
 				found = true
 				totalSweptAmt += sweep.value
 				notifyList = append(notifyList, sweep)
+				confirmedSweeps = append(
+					confirmedSweeps, sweep.outpoint,
+				)
```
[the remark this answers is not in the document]
starius: Fixed

sweepbatcher/sweep_batch.go
```diff
@@ -1814,41 +1952,92 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+	presigned, err := b.isPresigned()
+	if err != nil {
+		return fmt.Errorf("failed to determine if the batch %d uses "+
+			"presigned mode: %w", b.id, err)
+	}
+
+	// Sort sweeps by the addition order if this is presigned mode.
+	var allSweeps []sweep
+	if presigned {
+		allSweeps, err = b.getOrderedSweeps(ctx)
```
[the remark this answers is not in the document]
starius: Fixed. I added a comment:

```go
// Sort sweeps by the addition order. This is important in presigned
// mode to pass them in correct order to purger (AddSweep) so the
// primary sweep is determined correctly and the presigned transaction
// is found. In regular mode the order doesn't matter, but we do it the
// same way for simplicity.
allSweeps, err = b.getOrderedSweeps(ctx)
```

sweepbatcher/sweep_batcher.go
```diff
@@ -589,8 +658,51 @@ func (b *Batcher) Run(ctx context.Context) error {
...
+	if b.presignedHelper == nil {
+		return fmt.Errorf("presignedHelper is not installed")
+	}
+
+	// Find the feerate needed to get into next block. Use conf_target=2,
+	nextBlockFeeRate, err := b.wallet.EstimateFeeRate(ctx, 2)
+	if err != nil {
+		return fmt.Errorf("failed to get nextBlockFeeRate: %w", err)
+	}
+	infof("PresignSweepsGroup. NextBlockFeeRate is %v", nextBlockFeeRate)
```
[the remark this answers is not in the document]
starius: I updated the log message. I want to log the fact that we are in PresignSweepsGroup and also log the value of `nextBlockFeeRate`.

```go
infof("PresignSweepsGroup: nextBlockFeeRate is %v", nextBlockFeeRate)
```

sweepbatcher/sweep_batcher.go
```diff
@@ -589,8 +658,51 @@ func (b *Batcher) Run(ctx context.Context) error {
 	}
 }
 
+// PresignSweepsGroup creates and stores presigned transactions for the sweeps
+// group. This method must be called prior to AddSweep if presigned mode is
```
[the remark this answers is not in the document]
starius: I updated `AddSweep` to do the check.

Added new commit: `sweepbatcher: load sweeps in AddSweep`.

`AddSweep` now prepares sweeps (loads from DB and fee rate source). It also needs `ctx` to work. Another small fix for ctx is in commit `sweepbatcher: use WithoutCancel for notifications`.

`AddSweep` now fails if not `PresignSweepsGroup` wasn't called before. I could just put whole presigning into `AddSweep`, but I want to put presigning is a different action so we can presign and only if presigning succeeds, switch to the action which makes `AddSweep` calls.

sweepbatcher/sweep_batcher.go
```diff
@@ -589,8 +658,51 @@ func (b *Batcher) Run(ctx context.Context) error {
...
+
+	return presign(
+		ctx, b.presignedHelper, destAddress, primarySweepID, sweeps,
+		nextBlockFeeRate,
+	)
+}
+
 // AddSweep adds a sweep request to the batcher for handling. This will either
-// place the sweep in an existing batch or create a new one.
+// place the sweep in an existing batch or create a new one. In presigned mode
```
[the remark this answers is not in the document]
starius: Fixed. See the above comment.

starius:
@hieblmi Thanks for comments! 🎉
I fixed them.

Also found a small edge case for function `constructUnsignedTx`. Fixed it in commit `sweepbatcher: fix rounding in constructUnsignedTx`.

starius:
Updated the last commit `sweepbatcher: add mode with presigned transactions`. Skipped `ensurePresigned` check in `AddSweep` if the sweep is fully confirmed (the presigned transaction is cleaned up by this time).

sweepbatcher/sweep_batch.go
hieblmi: Thanks for the explanation!

hieblmi (approved):
Really amazing work on this @starius. LGTM!

starius:
@hieblmi One more change. I replaces `helper.IsPresigned()` method with a field in `SweepInfo`. This makes things easier on naut side. I pushed it in a separate temp commit.

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -0,0 +1,1577 @@
...
+	_ lntypes.Hash, _ wire.OutPoint) (*SweepInfo, error) {
+
+	return &SweepInfo{
+		// Set Timeout to prevent warning messages about timeout=0.
+		Timeout: sweepTimeout,
+	}, nil
+}
+
+// testPresigned_forgotten_presign checks that adding sweeps causes the batcher
+// to fail of the sweeps were not presigned with PresignSweepsGroup. In addition
```
bhandras, line 246: nit: of the => if the

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -0,0 +1,1577 @@
...
+	gotOps := make([]wire.OutPoint, 0, len(tx2.TxIn))
+	for _, txIn := range tx2.TxIn {
+		gotOps = append(gotOps, txIn.PreviousOutPoint)
+	}
+
+	require.ElementsMatch(t, wantOps, gotOps)
+}
+
+// TestPresigned tests presigned mode. Most sub-tests doesn't use loopdb.
+func TestPresigned(t *testing.T) {
```
bhandras, line 1614: Really nice tests and coverage!!

bhandras (commented):
Looks really good! My only complaint is the possible timeout overflow, otherwise LGTM!

sweepbatcher/sweep_batch_test.go
```diff
@@ -0,0 +1,318 @@
...
+	p2trAddress, err := btcutil.DecodeAddress(p2trAddr, nil)
+	require.NoError(t, err)
+	p2trPkScript, err := txscript.PayToAddrScript(p2trAddress)
+	require.NoError(t, err)
+
+	serializedPubKey := []byte{
+		0x02, 0x19, 0x2d, 0x74, 0xd0, 0xcb, 0x94, 0x34, 0x4c, 0x95,
+		0x69, 0xc2, 0xe7, 0x79, 0x01, 0x57, 0x3d, 0x8d, 0x79, 0x03,
+		0xc3, 0xeb, 0xec, 0x3a, 0x95, 0x77, 0x24, 0x89, 0x5d, 0xca,
+		0x52, 0xc6, 0xb4}
```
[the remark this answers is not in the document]
starius: Fixed. Thanks!

sweepbatcher/sweep_batcher.go
```diff
@@ -589,11 +594,32 @@ func (b *Batcher) Run(ctx context.Context) error {
...
 }
 
-// AddSweep adds a sweep request to the batcher for handling. This will either
-// place the sweep in an existing batch or create a new one.
-func (b *Batcher) AddSweep(sweepReq *SweepRequest) error {
+// AddSweep loads information about sweeps from the store and fee rate source,
+// and adds them to the batcher for handling. This will either place the sweep
+// in an existing batch or create a new one. The method can be called multiple
+// times, but the sweeps (including the order of them) must be the same. If
+// notifier is provided, the batcher sends back sweeping results though it.
```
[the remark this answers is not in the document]
starius: Fixed

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,625 @@
...
+	// Go from the floor (1.01 sat/vbyte) to 2k sat/vbyte with step of 1.2x.
+	const (
+		start            = chainfee.FeePerKwFloor
+		stop             = chainfee.AbsoluteFeePerKwFloor * 2_000
+		factorPPM        = 1_200_000
+		timeoutThreshold = 50
+	)
+
+	// Calculate the locktime value to use for high feerate transactions.
+	highFeeRateLockTime := uint32(timeout - timeoutThreshold)
```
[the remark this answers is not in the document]
starius: I changed this to fix this edge case. Thanks! Added a test case.

```go
// Calculate the locktime value to use for high feerate transactions.
// If timeout <= timeoutThreshold, don't set LockTime (keep value 0).
var highFeeRateLockTime uint32
if timeout > timeoutThreshold {
	highFeeRateLockTime = uint32(timeout - timeoutThreshold)
}
```

sweepbatcher/presigned.go
```diff
@@ -0,0 +1,625 @@
...
+	// Find actual fee rate of the signed transaction. It may differ from
+	// the desired fee rate, because SignTx may return a presigned tx.
+	output := btcutil.Amount(tx.TxOut[0].Value)
+	fee = batchAmt - output
+	signedFeeRate := chainfee.NewSatPerKWeight(fee, realWeight)
+
+	numSweeps := len(tx.TxIn)
+	b.Infof("attempting to publish custom signed tx=%v, desiredFeerate=%v,"+
+		" signedFeeRate=%v, weight=%v, fee=%v, sweeps=%d, destAddr=%s",
+		txHash, feeRate, signedFeeRate, weight, fee, numSweeps, address)
```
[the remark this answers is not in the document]
starius: Fixed.

sweepbatcher/sweep_batch.go
```diff
@@ -586,6 +601,84 @@ func (b *batch) addSweeps(ctx context.Context, sweeps []*sweep) (bool, error) {
...
+					"was not called prior to AddSweep for"+
+					" this input: %v",
+					sweeps[0].swapHash[:6], err)
+
+				return false, nil
+			}
+		}
+	} else {
+		// Ensure that all the sweeps in the batch don't use presigned.
+		for _, s := range b.sweeps {
```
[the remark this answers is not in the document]
starius: Removed that `else` branch.

sweepbatcher/sweep_batcher_presigned_test.go
starius: Fixed

starius:
@bhandras I pushed two commits to squash them later. They have the fixes for your comments.
Please take another look!

bhandras (approved):
Really nice additions @starius 🥇 LGTM! 🎉

PR #892, opened 2025-02-28
staticaddr: allow rbf'ing withdrawal transactions

hieblmi:
This PR adds the ability to request signatures for higher fee versions of previously published withdrawal transactions.

### DIFF
--- a/staticaddr/withdraw/manager.go
+++ b/staticaddr/withdraw/manager.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 	"sync/atomic"
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
@@ -18,6 +19,7 @@ import (
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/loop/staticaddr/deposit"
 	staticaddressrpc "github.com/lightninglabs/loop/swapserverrpc"
+	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnrpc/walletrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
@@ -26,10 +28,21 @@ import (
 )
 
 var (
-	// ErrWithdrawingInactiveDeposits is returned when the user tries to
-	// withdraw inactive deposits.
-	ErrWithdrawingInactiveDeposits = errors.New("deposits to be " +
-		"withdrawn are unknown or inactive")
+	// ErrWithdrawingMixedDeposits is returned when a withdrawal is
+	// requested for deposits in different states.
+	ErrWithdrawingMixedDeposits = errors.New("need to withdraw deposits " +
+		"having the same state, either all deposited or all " +
+		"withdrawing")
+
+	// ErrDiffPreviousWithdrawalTx signals that the user selected new
+	// deposits that have different previous withdrawal transactions.
+	ErrDiffPreviousWithdrawalTx = errors.New("can't bump fee for " +
+		"deposits with different previous withdrawal tx hash")
+
+	// ErrMissingPreviousWithdrawn is returned when the user tries to bump
+	// the fee for a subset of previously selected deposits to withdraw.
+	ErrMissingPreviousWithdrawn = errors.New("can't bump fee for subset " +
+		"of clustered deposits")
 
 	// MinConfs is the minimum number of confirmations we require for a
 	// deposit to be considered withdrawn.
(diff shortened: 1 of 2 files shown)


bhandras (approved):
LGTM, thanks for the iterations on the design 🎉

staticaddr/withdraw/manager.go
```diff
@@ -274,14 +289,65 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
+		for i := 1; i < len(deposits); i++ {
+			if deposits[i].FinalizedWithdrawalTx.TxHash() != hash {
+				return "", "", ErrDiffPreviousWithdrawalTx
+			}
+		}
+
+		// We also avoid that the user selects a subset of previously
+		// clustered deposits for a fee bump. This would result in a
+		// different transaction shape.
+		allDeposits, err := m.cfg.DepositManager.GetActiveDepositsInState(
```
starius, line 333: We can avoid this call. Previously we got a transaction in `deposits[i].FinalizedWithdrawalTx`.
Can't we verify that the transaction has the same set of deposits as `deposits`?

If we make sure that all `deposits` are unique, we can just compare `len(deposits) == len(prevTx.TxIn)`.

staticaddr/withdraw/manager.go
```diff
@@ -315,6 +381,41 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
+	if !published {
+		return "", "", nil
+	}
+
+	// If this is the first time this cluster of deposits is withdrawn, we
+	// start a goroutine that listens for the spent of the first input of
+	// the withdrawal transaction.
+	// Since we ensure above that the same ensemble of deposits is
+	// republished in case of a fee bump, it suffices if only one spent
+	// notifier is run, since if the first input is spent.
```
starius, line 399: What happens if `loopd` restarts after first withdrawal tx is published? Will it listen for spending after the restart?

staticaddr/withdraw/manager.go
```diff
@@ -325,37 +426,29 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
-	err = m.handleWithdrawal(
-		ctx, deposits, finalizedTx.TxHash(), withdrawalPkScript,
-	)
-	if err != nil {
-		return "", "", err
+	// Update the deposits in the database.
+	for _, d := range deposits {
+		err = m.cfg.DepositManager.UpdateDeposit(ctx, d)
+		if err != nil {
+			return "", "", err
```
starius, line 449: I propose to use `fmt.Errorf` to attach an error message to simplify debugging.

staticaddr/withdraw/manager.go
```diff
@@ -452,44 +545,66 @@ func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
...
+	staticAddress, err := m.cfg.AddressManager.GetStaticAddress(ctx)
+	if err != nil {
+		log.Errorf("error retrieving taproot address %w", err)
+
+		return fmt.Errorf("withdrawal failed")
+	}
 
-	confChan, errChan, err := m.cfg.ChainNotifier.RegisterConfirmationsNtfn(
-		ctx, &txHash, withdrawalPkScript, MinConfs,
-		int32(m.initiationHeight.Load()),
```
starius, line 484: MinConfs is now not used. What if a reorg happens? Shouldn't we run `RegisterConfirmationsNtfn` after a successful spending (1 conf) before transitioning the deposit to `Withdrawn` state?

staticaddr/withdraw/manager.go
```diff
@@ -557,11 +558,17 @@ func (m *Manager) publishFinalizedWithdrawalTx(ctx context.Context,
 	// Publish the withdrawal sweep transaction.
 	err := m.cfg.WalletKit.PublishTransaction(ctx, tx, txLabel)
 	if err != nil {
-		if !strings.Contains(err.Error(), "output already spent") &&
-			!strings.Contains(err.Error(), "insufficient fee") {
+		if !strings.Contains(err.Error(), chain.ErrSameNonWitnessData.Error()) &&
+			!strings.Contains(err.Error(), "output already spent") &&
+			!strings.Contains(err.Error(), chain.ErrInsufficientFee.Error()) {
```
starius, line 563: I propose to factor out this check to a function and reuse everywhere (e.g. in sweepbatcher package).

Also this commit should be squashed into commit `staticaddr: allow rbf'ing withdrawal transactions`.

starius (approved):
LGTM!
Few notes.

staticaddr/withdraw/manager.go
hieblmi: thank you, fixed

staticaddr/withdraw/manager.go
hieblmi: yes, after a restart `recoverWithdrawals` will republish the latest version of the withdrawal tx and monitor the spending.

staticaddr/withdraw/manager.go
hieblmi: done

PR #893, opened 2025-03-03
staticaddr: fix withdraw backwards compat

hieblmi:
This PR restores a previously removed parameter of `ServerWithdrawRequest`, `TxFeeRate`, to allow withdrawing with a fee rate without change.


bhandras (approved):
LGTM, given the old protocol is considered unreleased. 👍

PR #902, opened 2025-03-20
staticaddr: lock mutex for active deposits access

hieblmi:
Fixes race conditions in the static address withdrawal manager.

### DIFF
--- a/staticaddr/deposit/fsm.go
+++ b/staticaddr/deposit/fsm.go
@@ -141,6 +141,11 @@ type FSM struct {
 
 	blockNtfnChan chan uint32
 
+	// quitChan stops after the FSM stops consuming blockNtfnChan.
+	quitChan chan struct{}
+
+	// finalizedDepositChan is used to signal that the deposit has been
+	// finalized and the FSM can be removed from the manager's memory.
 	finalizedDepositChan chan wire.OutPoint
 }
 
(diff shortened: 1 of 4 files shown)


staticaddr/deposit/manager.go
```diff
@@ -144,11 +145,14 @@ func (m *Manager) Run(ctx context.Context, currentHeight uint32) error {
 					return ctx.Err()
 				}
 			}
+			m.mu.Unlock()
```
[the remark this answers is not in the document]
hieblmi: re-structured the code.

staticaddr/deposit/fsm.go
```diff
@@ -203,7 +211,7 @@ func NewFSM(ctx context.Context, deposit *Deposit, cfg *ManagerConfig,
 				return
 			}
 		}
-	}()
+	}(depoFsm)
```
starius, line 214: Why is this passing-as-an-argument needed? Previous version seems also ok.

starius (approved):
LGTM!
Added a nit.

staticaddr/deposit/fsm.go
hieblmi: usually this pattern is better within loops, but I also find in more clear, so keeping it.

PR #908, opened 2025-03-26
staticaddr: non-concurrent deposit recovery

hieblmi:
Previously the deposit recovery took place in a goroutine per deposit. Since the withdrawal manager recovery depends on a completed deposits recovery, this PR removes the goroutine around the deposits recovery to block until all deposits are fully recovered.


bhandras (approved):
LGTM 🌮

PR #910, opened 2025-03-28
 sweepbatcher: customize initialDelay per sweep

starius:
Option `WithInitialDelay` now accepts a function returning initialDelay depending on sweep's data.

This is needed to be able to wait longer for sweeps with low priority, but still sweeping high priority sweeps soon.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -151,11 +151,13 @@ type batchConfig struct {
 	// clock provides methods to work with time and timers.
 	clock clock.Clock
 
-	// initialDelay is the delay of first batch publishing after creation.
-	// It only affects newly created batches, not batches loaded from DB,
-	// so publishing does happen in case of a daemon restart (especially
-	// important in case of a crashloop).
-	initialDelay time.Duration
+	// initialDelayProvider provides the delay of first batch publishing
+	// after creation. It only affects newly created batches, not batches
+	// loaded from DB, so publishing does happen in case of a daemon restart
+	// (especially important in case of a crashloop). If a sweep is about to
+	// expire (time until timeout is less that 2x initialDelay), then
+	// waiting is skipped.
+	initialDelayProvider InitialDelayProvider
 
 	// batchPublishDelay is the delay between receiving a new block or
 	// initial delay completion and publishing the batch transaction.
@@ -650,6 +652,7 @@ func (b *batch) Run(ctx context.Context) error {
 
 	// Cache clock variable.
 	clock := b.cfg.clock
+	startTime := clock.Now()
 
 	blockChan, blockErrChan, err :=
 		b.chainNotifier.RegisterBlockEpochNtfn(runCtx)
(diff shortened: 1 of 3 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -150,11 +150,13 @@ type batchConfig struct {
...
-	// so publishing does happen in case of a daemon restart (especially
-	// important in case of a crashloop).
-	initialDelay time.Duration
+	// initialDelayProvider provides the delay of first batch publishing
+	// after creation. It only affects newly created batches, not batches
+	// loaded from DB, so publishing does happen in case of a daemon restart
+	// (especially important in case of a crashloop). If a sweep is about to
+	// expire (time until timeout is less that 2x initialDelay), then
+	// waiting is skipped.
+	initialDelayProvider InitialDelayProvider
```
bhandras, line 159: nit: as this isn't really about init, but the first publish, for clarity we could rename it to `firstPublishDelayProvider` (and the `firstPublishDelay` below in the runloop). wdyt?

bhandras (approved):
LGTM, nice! I think this greedy logic will help batches a lot! 🎉

sweepbatcher/sweep_batch.go
starius: I prefer the current name. We also have another option called PublishDelay. It could be confused with firstPublishDelayProvider if it is renamed. Also if we grep by PublishDelay, we would find both options.

sweepbatcher/sweep_batcher.go
```diff
@@ -162,6 +162,21 @@ type VerifySchnorrSig func(pubKey *btcec.PublicKey, hash, sig []byte) error
 type FeeRateProvider func(ctx context.Context,
 	swapHash lntypes.Hash) (chainfee.SatPerKWeight, error)
 
+// InitialDelayProvider returns the duration after new batch creation before it
```
hieblmi, line 165: nit: `InitialDelayProvider returns the duration after which a newly created batch is first published.`

sweepbatcher/sweep_batch.go
```diff
@@ -698,6 +699,42 @@ func (b *batch) Run(ctx context.Context) error {
 		b.primarySweepID[0:6], len(b.sweeps))
 
 	for {
+		// If the batch if not empty, find earliest initialDelay.
```
hieblmi, line 702: nit: is not empty

sweepbatcher/sweep_batch.go
```diff
@@ -698,6 +699,42 @@ func (b *batch) Run(ctx context.Context) error {
...
+			totalSweptAmt += sweep.value
+		}
+
+		skipBeforeUpdated := false
+		if totalSweptAmt != 0 {
+			initialDelay, err := b.cfg.initialDelayProvider(
+				ctx, len(b.sweeps), totalSweptAmt,
+			)
+			if err != nil {
+				b.Warnf("initialDelayProvider failed: %v", err)
```
hieblmi, line 714: add to the warning that we use no initial delay in publishing this batch

sweepbatcher/sweep_batch.go
```diff
@@ -698,6 +699,42 @@ func (b *batch) Run(ctx context.Context) error {
...
+				initialDelay = 0
+			}
+			if initialDelay < 0 {
+				b.Warnf("negative delay: %v", initialDelay)
```
hieblmi, line 718: same here

hieblmi (approved):
LGTM!

sweepbatcher/sweep_batcher.go
starius: Fixed! Thanks!

sweepbatcher/sweep_batch.go

sweepbatcher/sweep_batch.go
starius: Done

PR #912, opened 2025-04-02
add simple pagination to the ListSwaps command

### LINKED ISSUE #797: Pagination on ListSwaps
**Is your feature request related to a problem? Please describe.**

At some point will have a lot of swaps, and current filters do not allow one to filter by date range or limit output.

**Describe the solution you'd like**

Implement pagination pattern similar to ListPayments and other LND List gRPC calls. Also having a filter based on index like Payments in LND or datetime would be very helpful.

**Describe alternatives you've considered**

I will query all the history from Loop, then compare Ids with my audit tables, and update rows accordingly. Perfectly functional at current scale, will get nasty at some point.
**Additional context**

kornpow:
This commit adds the ability to set a `max_swaps` and an `index_offset` flag to the ListSwaps command. These new fields are applied AFTER the initial filtering step. The response also now passes additional information about the `last_index_offset` count and `total_filtered_swaps`, which can be used together with the `--max_swaps` and `index_offset` flags to iterate over your set of swaps in chunks.

`$ loop listswaps --loop_out_only --max_swaps 1 --index_offset 3`
```
{
    "swaps":  [
        {
            "amt":  "3000000",
            "id":  "xxx",
            "id_bytes":  "xxx",
            "type":  "LOOP_OUT",
            "state":  "SUCCESS",
            "failure_reason":  "FAILURE_REASON_NONE",
            "initiation_time":  "1743479436945030630",
            "last_update_time":  "1743487638197810121",
            "htlc_address":  "xxx",
            "htlc_address_p2wsh":  "",
            "htlc_address_p2tr":  "xxx",
            "cost_server":  "3043",
            "cost_onchain":  "120",
            "cost_offchain":  "7680",
            "last_hop":  "",
            "outgoing_chan_set":  [],
            "label":  "",
            "asset_info":  null
        }
    ],
    "last_index_offset":  3,
    "total_filtered_swaps":  4
}
```

#### Pull Request Checklist

#### Closes issues
This will close #797


kornpow:
I'm going to start looking at some tests now, but wanted to open the PR to get some initial feedback on the approach first.

kornpow:
Added the test which tests the behavior of both the filtering and the pagination.

One thing I am still trying to decide on is when starting `startIndex` to an out of bounds number. Currently it'll look like this:

```
swaps: [] 
first_index_offset:5 
last_index_offset:2
total_filtered_swaps:3
```
Maybe return an error instead of just having weird behavior?

loopd/swapclient_server.go
```diff
@@ -579,15 +579,42 @@ func (s *swapClientServer) ListSwaps(ctx context.Context,
...
 		rpcSwap, err := s.marshallSwap(ctx, &swp)
 		if err != nil {
 			return nil, err
 		}
 		rpcSwaps = append(rpcSwaps, rpcSwap)
 		idx++
 	}
-	return &looprpc.ListSwapsResponse{Swaps: rpcSwaps}, nil
+
+	filteredSwapCount := len(rpcSwaps)
```
bhandras, line 590: Ideally the pagination could be interleaved with the above filtering loop.

loopd/swapclient_server.go
```diff
@@ -579,15 +579,42 @@ func (s *swapClientServer) ListSwaps(ctx context.Context,
 		if !filterSwap(&swp, req.ListSwapFilter) {
 			continue
 		}
-
```
bhandras, line 582: nit: the extra new line could stay as it makes readability slightly better.

loopd/swapclient_server_test.go
```diff
@@ -595,3 +599,170 @@ func TestHasBandwidth(t *testing.T) {
 		})
 	}
 }
+
+func TestListSwapsFilterAndPagination(t *testing.T) {
```
bhandras, line 603: nit: could you pls add a godoc?

loopd/swapclient_server_test.go
```diff
@@ -595,3 +599,170 @@ func TestHasBandwidth(t *testing.T) {
...
+			State: loopdb.StateInitiated,
+			Cost:  loopdb.SwapCost{},
+		},
+		LastUpdate:       time.Now(),
+		SwapHash:         lntypes.Hash{3},
+		SwapType:         swap.Type(swap.TypeOut),
+		HtlcAddressP2WSH: testnetAddr,
+		HtlcAddressP2TR:  testnetAddr,
+	}
+	// Create test cases with different filter/pagination parameters
```
bhandras, line 639: nit: please close comments with `.`

loopd/swapclient_server_test.go
```diff
@@ -595,3 +599,170 @@ func TestHasBandwidth(t *testing.T) {
...
+			},
+			expectedReturnedSwapCount:  1,
+			expectedFirstIdx:           0,
+			expectedLastIdx:            1,
+			expectedFilteredTotalCount: 1,
+		},
+		{
+			name:                       "fetch swaps-outs no pagination",
+			mockSwaps:                  []loop.SwapInfo{swapIn1, swapOut1, swapOut2},
+			req:                        &looprpc.ListSwapsRequest{ListSwapFilter: &looprpc.ListSwapsFilter{SwapType: looprpc.ListSwapsFilter_LOOP_OUT}},
```
bhandras, line 676: Make sure that lines are not very long. We always try to format code such that we fit 80 columns (unless not possible). Please refer to our style guide: https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md#additional-style-constraints

bhandras (commented):
Concept ACK! Thank you @kornpow, nice addition! Left a few comments.

loopd/swapclient_server.go
kornpow: Ok, I will look into this. It does seem ideal if I could do it that way.

kornpow:
Updated the response a bit. Also, now we only unmarshal the swaps we intend to return!

starius:
> ```
> swaps: [] 
> first_index_offset:5 
> last_index_offset:2
> total_filtered_swaps:3
> ```

I can't find where `first_index_offset` is set. I think we should add this field, to be consistent with LND. E.g. `lncli listinvoices` returns both `first_index_offset` and `last_index_offset`.

starius (commented):
Nice improvement! 🎉

My main comment is that the order in the map is not stable and offsets themselves are not stable. I propose to use a timestamp instead of an offset and to use an opaque pagination tokens for future compatibility. Also to order the swaps by the initiation time to make the order predictable.

kornpow:
> > ```
> > swaps: [] 
> > first_index_offset:5 
> > last_index_offset:2
> > total_filtered_swaps:3
> > ```
> 
> I can't find where `first_index_offset` is set. I think we should add this field, to be consistent with LND. E.g. `lncli listinvoices` returns both `first_index_offset` and `last_index_offset`.

I have returned `first_index_offset` to the response. I forgot that I copied the LND api response initially. But then I realized that `first_index_offset` in the response always equals `index_offset` in request, so I removed it, but it is useful to have so added it back.

starius:
What do you think about switching from indices to timestamps as page identifiers? Timestamps do not depend on filtering, so this would work better in case some swaps appear or disappear in the middle of iteration.

kornpow:
> What do you think about switching from indices to timestamps as page identifiers? Timestamps do not depend on filtering, so this would work better in case some swaps appear or disappear in the middle of iteration.

Similar to the API for `listpayments` in LND? (This LND api has both filtering on `creation_date` and the index based pagination that is currently implemented)
```
kornpow@kornpow:~/Documents/repos/loop listswaps_paginate$ lncli listpayments -h
   --creation_date_start value  timestamp in seconds, if set, filter payments with creation date greater than or equal to it (default: 0)
   --creation_date_end value    timestamp in seconds, if set, filter payments with creation date less than or equal to it (default: 0)
```

Timestamps do depend on filtering though, its just filtering based on a timestamp?

starius:
I suggest replacing `--index_offset` with `--initiation-time-after`.  
The API would return a `"last_initiation_time"` field in the response, which the user could then pass back as the value for `--initiation-time-after`. The API would return only swaps initiated after (`>`) that timestamp.  
The page size would still be controlled by the `--max_swaps` parameter (i.e., the number of swaps to return).  

Unlike indices, timestamps remain stable even if some swaps are added or removed from the dataset. Index shifts can lead to duplicates being returned. Using timestamps avoids that issue.

kornpow:
Here is the current state. It is getting closer. I appreciate your feedback @starius on the API design. You've definitely helped make it a lot simpler.

Here is an example of paging through.

`$ loop listswaps --start_time_ns=0`
```
{
    "swaps":  [
        {
...
            "initiation_time": "1729012176223580685",
...
        },
        {
...
            "initiation_time": "1743457082654026082",
...
        },
        {
...
            "initiation_time": "1743457172366700842",
...
        },
        {
...
            "initiation_time": "1743479436945030630",
...
        }
    ],
    "next_start_time": "0"
}
```

`loop  listswaps --start_time_ns=0 --max_swaps=2`         

```
{
    "swaps": [
        {
...
            "initiation_time": "1729012176223580685",
...
        },
        {
...
            "initiation_time": "1743457082654026082",
...
        }
    ],
    "next_start_time": "1743457082654026083"

```


`listswaps --start_time_ns=1743457082654026083 --max_swaps=2`
```
{
    "swaps": [
        {
...
            "initiation_time": "1743457172366700842",
...
        },
        {
...
            "initiation_time": "1743479436945030630",
...
        }
    ],
    "next_start_time": "0"
}
```

loopd/swapclient_server_test.go
```diff
@@ -595,3 +599,255 @@ func TestHasBandwidth(t *testing.T) {
...
+					StartTimestampNs: unixTime.Add(5 * time.Minute).UnixNano(),
+				},
+			},
+			expectedReturnedSwaps: []lntypes.Hash{
+				swapInOrder0.SwapHash,
+				swapOutOrder1.SwapHash,
+				swapOutOrder2.SwapHash,
+			},
+			expectedNextStartTime: 0,
+		},
```
starius, line 758: Test edge cases:
 - StartTimestampNs = secondSwapStartTime
 - StartTimestampNs = secondSwapStartTime + 1
 - StartTimestampNs = secondSwapStartTime - 1

starius (approved):
LGTM, pending comments 🏆

loopd/swapclient_server.go
```diff
@@ -563,31 +564,68 @@ func (s *swapClientServer) ListSwaps(ctx context.Context,
...
+		maxSwaps      = int(req.MaxSwaps)
+		nextStartTime = int64(0)
+		canPage       = false
 	)
 
 	s.swapsLock.Lock()
 	defer s.swapsLock.Unlock()
 
+	// Convert the swaps map into a slice which can be ordered.
+	var swapList []loop.SwapInfo
```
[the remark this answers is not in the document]
kornpow: I reworked it and now I don't need to preallocate the slice, but let me know if I still do.

loopd/swapclient_server.go
```diff
@@ -563,31 +564,68 @@ func (s *swapClientServer) ListSwaps(ctx context.Context,
...
+	for _, swp := range s.swaps {
+		swapList = append(swapList, swp)
+	}
+
+	// Sort all swaps by a consistent field, InitiationTime.
+	slices.SortFunc(swapList, func(a, b loop.SwapInfo) int {
+		// For descending order (newest first)
```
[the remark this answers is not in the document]
kornpow: Adjusted the comment to match that correct order is ascending. But Im willing to swap the order if you prefer as well.

loopd/swapclient_server.go
```diff
@@ -563,8 +565,11 @@ func (s *swapClientServer) ListSwaps(ctx context.Context,
 	req *looprpc.ListSwapsRequest) (*looprpc.ListSwapsResponse, error) {
 
 	var (
-		rpcSwaps = []*looprpc.SwapStatus{}
-		idx      = 0
+		rpcSwaps      = []*looprpc.SwapStatus{}
+		rpcSwapInfo   = []*loop.SwapInfo{}
```
starius, line 569: I propose to rename `rpcSwapInfo` to `swapInfos`. It is not a proto object (hence no "rpc") and it is a list (hence "s" in the end).

loopd/swapclient_server.go
```diff
@@ -580,14 +585,40 @@ func (s *swapClientServer) ListSwaps(ctx context.Context,
 			continue
 		}
 
-		rpcSwap, err := s.marshallSwap(ctx, &swp)
+		rpcSwapInfo = append(rpcSwapInfo, &swp)
+	}
+
+	// Sort the swaps by initiation time in ascending order (oldest first).
+	slices.SortFunc(rpcSwapInfo, func(a, b *loop.SwapInfo) int {
+		return cmp.Compare(a.InitiationTime.UnixNano(), b.InitiationTime.UnixNano())
```
starius, line 593: This line is too long. Please format it like this:
```go
return cmp.Compare(
  a.InitiationTime.UnixNano(),
  b.InitiationTime.UnixNano(),
)
```

starius (approved):
LGTM!
Added few comments.

loopd/swapclient_server.go
kornpow: sorry, thanks for continuing to point these out.

loopd/swapclient_server.go
```diff
@@ -617,6 +651,11 @@ func filterSwap(swapInfo *loop.SwapInfo, filter *looprpc.ListSwapsFilter) bool {
 		return false
 	}
 
+	// If timestamp filters are set, only return swaps within the specified time range.
+	if filter.StartTimestampNs > 0 && swapInfo.InitiationTime.UnixNano() < filter.StartTimestampNs {
```
starius, line 655: Also please format this line to avoid too long line:

```go
if filter.StartTimestampNs > 0 &&
  swapInfo.InitiationTime.UnixNano() < filter.StartTimestampNs {

  return false
}
```

We use 80 as a limit, counting tab as 8.

starius (approved):

loopd/swapclient_server.go
kornpow: Thanks again

kornpow:
I'm sorry for the delay, and all the churn. It should now properly pass the lint test.

bhandras (approved):
LGTM, nice work @kornpow 🎉

PR #914, opened 2025-04-04
build+loop: add the commit hash to `GetInfoResponse` and change version semantics

bhandras:
With this commit we extend `GetinfoResponse` with the strict commit hash (which also includes whether the build tree was clean or dirty).


bhandras:
cc: @wetfish, @ellemouton

Makefile
```diff
@@ -15,7 +15,7 @@ GOBUILD := GO111MODULE=on go build -v
 GOINSTALL := GO111MODULE=on go install -v
 GOMOD := GO111MODULE=on go mod
 
-COMMIT := $(shell git describe --abbrev=40 --dirty)
+COMMIT := $(shell git describe --abbrev=40 --dirty --tags --long)
```
ellemouton, line 18: another option could be to add a separate `COMMIT_HASH` and then set that separately to the COMMIT one. 

but that of course would result in needing a new global variable & then a new `commit_hash` rpc field in GetInfo. 

So I guess we just need to decide if we want to keep the COMMIT working as it does today where if the "pure" tagged version is being run, then no commit hash is included in this variable. 

But if it doesnt matter, then this is probs the easiest solution. 

wdyt?

bhandras: Great idea, PTAL.

Makefile
```diff
@@ -15,7 +15,7 @@ GOBUILD := GO111MODULE=on go build -v
 GOINSTALL := GO111MODULE=on go install -v
 GOMOD := GO111MODULE=on go mod
 
-COMMIT := $(shell git describe --abbrev=40 --dirty)
+COMMIT := $(shell git describe --abbrev=40 --dirty | sed -E 's/.*-g([0-9a-f]{40})(-dirty)?/\1\2/')
```
ellemouton, line 18: nice!

ellemouton (approved):
nice - lgtm!!

PR #919, opened 2025-04-09
sweepbatcher: notify caller about confirmations

starius:
Add fields `ConfChan` and `ConfErrChan` to `sweepbatcher.SpendNotifier` type which is a part of `SweepRequest` passed to `AddSweep` method.

This is needed to reuse confirmation notifications on the calling side the same way it is done for spending notifications.

Also fixed many small bugs and issues and added missing tests for notifications sent through channels of `SpendNotifier`.

#### Pull Request Checklist


sweepbatcher/sweep_batch.go
```diff
@@ -1908,11 +1910,40 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
 
 // handleConf handles a confirmation notification. This is the final step of the
 // batch. Here we signal to the batcher that this batch was completed.
-func (b *batch) handleConf(ctx context.Context) error {
+func (b *batch) handleConf(ctx context.Context,
+	conf *chainntnfs.TxConfirmation) error {
+
```
bhandras, line 1915: Since this function iterates on sweeps, we should also schedule it to the main eventloop.

sweepbatcher/store.go
```diff
@@ -213,8 +213,8 @@ type dbBatch struct {
 	// ID is the unique identifier of the batch.
```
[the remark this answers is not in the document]
starius: I updated the commit message:

```
sweepbatcher: align dbBatch type with DB schema

Previously, dbBatch had a State field (enum: Open, Closed, Confirmed), but in
the database it is represented as a boolean Confirmed. The Closed state was
stored the same way as Open. This wasn't an issue in practice, since an Open
batch is quickly transitioned to Closed after startup.

However, the in-memory mock stores plain dbBatch instances, leading to
inconsistent behavior between the mock and the real DB-backed store. This
commit updates dbBatch to match the database representation by replacing
```

Also added a note to the description of Closed const:

```go
// Closed is the state in which the batch is no longer able to accept
// new sweeps. NOTE: this state exists only in-memory. In the database
// it is stored as Open and converted to Closed after a spend
// notification arrives (quickly after start of Batch.Run).
Closed batchState = 1
```

sweepbatcher/sweep_batch.go
```diff
@@ -2025,6 +2027,43 @@ func (b *batch) writeToErrChan(err error) {
...
+		done()
+
+		return
+	}
+	notifiers := make([]*SpendNotifier, 0, len(b.sweeps))
+	for _, s := range b.sweeps {
+		// If the sweep's notifier is empty then this means that a swap
+		// is not waiting to read an update from it, so we can skip
+		// the notification part.
+		if s.notifier == nil || *s.notifier == (SpendNotifier{}) {
```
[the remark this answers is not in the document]
starius: It was a copy-paste. I updated this line:
```go
if s.notifier == nil || s.notifier.SpendErrChan == nil {
```

sweepbatcher/store.go
```diff
@@ -259,7 +259,7 @@ func convertBatchRow(row sqlc.SweepBatch) *dbBatch {
 	}
 
 	if row.Confirmed {
-		batch.State = batchOpen
+		batch.State = batchConfirmed
```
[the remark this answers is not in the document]
starius: Yes, and then after confirmation notification (after 3 confs).

```
Open -(spend notif)-> Closed -(conf notif)-> Confirmed
```

sweepbatcher/sweep_batcher.go
```diff
@@ -951,11 +951,12 @@ func (b *Batcher) monitorSpendAndNotify(ctx context.Context, sweep *sweep,
 	parentBatchID int32, notifier *SpendNotifier) error {
 
 	spendCtx, cancel := context.WithCancel(ctx)
-	defer cancel()
 
 	// Then we get the total amount that was swept by the batch.
 	totalSwept, err := b.store.TotalSweptAmount(ctx, parentBatchID)
 	if err != nil {
+		cancel()
```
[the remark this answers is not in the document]
starius: This code path is executed only when a fully confirmed sweep is added again to the sweeper. This may happen if the daemon was down when the sweep was fully confirmed (got 3rd confirmation). This is not the main code path.

sweepbatcher/sweep_batcher.go
```diff
@@ -916,15 +904,18 @@ func (b *Batcher) FetchUnconfirmedBatches(ctx context.Context) ([]*batch,
...
+		} else {
+			// We don't store Closed state separately in DB.
+			// If the batch is closed (included into a block, but
+			// not fully confirmed), it is now considered Open
+			// again. It will receive a spending notification as
+			// soon as it starts, so it is not an issue. If a sweep
+			// manages to be added during this time, it will be
+			// detected as missing when analyzing the spend
+			// notification and will be added to new batch.
+			batch.state = Open
```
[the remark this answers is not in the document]
starius: Avoiding this would require a DB migration, which doesn’t seem worthwhile just to improve this edge case.

sweepbatcher/sweep_batch.go
starius: `handleConf` is already called from the main event loop, so it can freely access `batch.sweeps` - nobody else should access it at the same time.

starius:
Converting to draft. I'm working to add more commits here and to rebase it on top of https://github.com/lightninglabs/loop/pull/891

sweepbatcher/sweep_batch.go
bhandras: Indeed, you're right.

sweepbatcher/sweep_batch.go
```diff
@@ -1949,7 +1948,106 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+		// the notification part.
+		if sweep.notifier == nil ||
+			*sweep.notifier == (SpendNotifier{}) {
+
+			continue
+		}
+
+		spendDetail := SpendDetail{
+			Tx: spendTx,
+			OnChainFeePortion: getFeePortionPaidBySweep(
```
bhandras: Alternatively we could communicate the fee portion once the sweep is fully-confirmed. wdyt?

bhandras (approved):
Really nice set of fixes! Making sweep-batcher more robust one step at a time 🚀 
LGTM 🎉

litbot-9000:
@hieblmi: review reminder
@sputn1ck: review reminder

sweepbatcher/sweep_batch.go
starius: I added `OnChainFeePortion` to type `ConfDetail`, which is send through `ConfChan`.

I did it in a separate commit to make it clear what changed. I plan to squash it with the previous commit before merging.

starius:
I removed commits related to presigned mode from this PR. I'll open another PR.

sweepbatcher/sweep_batcher.go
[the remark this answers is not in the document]
hieblmi: +1, great catch

sweepbatcher/sweep_batch.go
```diff
@@ -1822,29 +1822,22 @@ func (b *batch) monitorSpend(ctx context.Context, primarySweep sweep) error {
 		b.Infof("monitoring spend for outpoint %s",
 			primarySweep.outpoint.String())
 
-		for {
+		select {
```
hieblmi, line 1827: please fix typo in commit message.

sweepbatcher/sweep_batch.go
```diff
@@ -2099,16 +2099,20 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
-	if err != nil {
-		return err
-	}
-
 	// We are no longer able to accept new sweeps, so we mark the batch as
 	// closed and persist on storage.
 	b.state = Closed
 
-	return b.persist(ctx)
+	if err := b.persist(ctx); err != nil {
```
hieblmi: nit: can do `err = b.persist...`

hieblmi: another great catch!

sweepbatcher/sweep_batch.go
```diff
@@ -2159,7 +2161,37 @@ func (b *batch) handleConf(ctx context.Context,
...
+			select {
+			// Try to write the confirmation to the notification
+			// channel.
+			case notifier.ConfChan <- conf:
+
+			// If a quit signal was provided by the swap,
+			// continue.
+			case <-notifier.QuitChan:
+			}
+		}(s.notifier)
```
hieblmi, line 2208: 👍

hieblmi (approved):
LGTM, great set of fixes!

sweepbatcher/sweep_batch.go
```diff
@@ -1822,29 +1822,22 @@ func (b *batch) monitorSpend(ctx context.Context, primarySweep sweep) error {
...
-
-				return
+			b.writeToErrChan(
+				fmt.Errorf("spend error: %w", err),
+			)
 
-			case <-ctx.Done():
-				return
-			}
+		case <-ctx.Done():
```
sputn1ck, line 1842: prior we returned here and in all other ctx.done cases, is this intended? We just fall through to the last return right?

sweepbatcher/sweep_batch.go
starius: Fixed. Thanks!

sweepbatcher/sweep_batch.go
starius: Fixed

sweepbatcher/sweep_batch.go
starius: Prior we returned in all the cases of the `select`, so the `for` loop always used to have exactly one iteration. That is why the `for` loop is not needed and `return` statements can be removed, because this is the end of the function anyway.

PR #929, opened 2025-04-25
mod: add replace directive to compile tapd

guggero:
Creating this mostly as a reminder and FYI that once https://github.com/lightninglabs/taproot-assets/pull/1198 is in, this `replace` directive will be required for `tapd` to compile.

Because tapd now uses a fork of the SQL migration library, which can only be used through a replace directive which isn't populated transitively to other dependent projects, we now need to copy that directive to every project.


guggero:
Dependent PR was merged, we now need this to build things in `litd`.

bhandras (approved):
LGTM 🎉

guggero:
> LGTM! Does that mean we can also use go code migrations in loop?

Yes, absolutely!

PR #930, opened 2025-04-28
perms: move perms.go into looprpc

ffranr:
Overall plan and context: https://github.com/lightninglabs/lightning-node-connect/issues/113#issuecomment-2832804329

---

The Lightning Node Connect (LNC) WASM client currently imports both `looprpc` and the entire `loop` module to access `perms.go`. This creates problems because `loop` imports `taproot-assets`, resulting in too many symbols for the WASM build.  

To fix this, we will move `perms.go`—which defines RPC endpoint permissions—into `looprpc`. After this change, the LNC WASM client will import only `looprpc`, while the rest of `loop` can continue importing `taproot-assets` without affecting the LNC WASM build.


ffranr:
More changes will be necessary to fix the LNC WASM build issue, but they depend on taproot-assets changes that are not yet merged.

bhandras (approved):
Changes LGTM 🎉 (still waiting for go mod check to run)

starius:
I ran `go mod tidy` in both `loop` and `loop/looprpc`. Also reverted https://github.com/lightninglabs/loop/pull/926/commits/d7d9e2d24961fb2fe00c6613fabe6a945645ef03 which was causing a mod check failure.

starius (approved):

PR #942, opened 2025-05-13
staticaddr: fix crash in loop-in manager shutdown

starius:
If the notification manager shuts down, sweepReqs channel is closed and sweepReq object received from it becomes nil, causing panic in handleLoopInSweepReq. Now this is fixed by checking if the channel was closed and if it was shutting down loop-in manager itself.

#### Pull Request Checklist


bhandras (approved):
LGTM, thank you!

hieblmi (approved):
Thanks for the fix!

PR #946, opened 2025-05-22
Update commit_hash field to GetInfo response

ViktorTigerstrom:
This PR updates the `GetInfo` response to adhere to https://github.com/lightninglabs/lightning-terminal/pull/1034.

In general it updates the output of the `commit_hash` field in the GetInfoResponse.
The `commit_hash` field will now contain the most recent commit_hash that the build was based on. If the build had uncommitted changes, this field will contain the most recent commit hash, suffixed by "-dirty".

We also change the output of `commit` field in the `--version` command. The `Commit` field will now contain most recent git commit tag.

Note that both of these changes are breaking changes!

Additionally we also add a `commit_hash` part of the output of the `--version` output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,10 @@ GOBUILD := GO111MODULE=on go build -v
 GOINSTALL := GO111MODULE=on go install -v
 GOMOD := GO111MODULE=on go mod
 
-COMMIT := $(shell git describe --abbrev=40 --dirty | sed -E 's/.*-g([0-9a-f]{40})(-dirty)?/\1\2/')
-LDFLAGS := -ldflags "-X $(PKG).Commit=$(COMMIT)"
+COMMIT := $(shell git describe --abbrev=40 --dirty --tags)
+COMMIT_HASH := $(shell git rev-parse HEAD)
+DIRTY := $(shell git diff-index --quiet HEAD -- || echo dirty)
+LDFLAGS := -ldflags "-X $(PKG).Commit=$(COMMIT) -X $(PKG).CommitHash=$(COMMIT_HASH) -X $(PKG).Dirty=$(DIRTY)"
 DEV_TAGS = dev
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -15,8 +15,10 @@ GOBUILD := GO111MODULE=on go build -v
 GOINSTALL := GO111MODULE=on go install -v
 GOMOD := GO111MODULE=on go mod
 
-COMMIT := $(shell git describe --abbrev=40 --dirty | sed -E 's/.*-g([0-9a-f]{40})(-dirty)?/\1\2/')
-LDFLAGS := -ldflags "-X $(PKG).Commit=$(COMMIT)"
+COMMIT := $(shell git describe --abbrev=40 --dirty --tags)
+COMMIT_HASH := $(shell git rev-parse HEAD)
+DIRTY := $(shell git diff-index --quiet HEAD -- || echo -dirty)
```
bhandras, line 20: Instead of `-dirty` we should store it as `dirty`.

version.go
```diff
@@ -12,10 +12,21 @@ import (
...
+// build, the Commit field stores the most recent tag suffixed by the commit
+// hash, and/or "-dirty". This should be set using the -ldflags during
+// compilation.
 var Commit string
 
+// CommitHash stores the current git commit hash of this build. This should be
+// set using the -ldflags during compilation.
+var CommitHash string
+
+// Dirty stores a "-dirty" string, if the build had uncommitted changes when
```
bhandras, line 26: ... so then here it'd be `dirty` too (and where we concatenate we add the dash).

bhandras (commented):
Looks good, just one comment!

Makefile
ViktorTigerstrom: Ah good idea! Addressed with the latest push :).

hieblmi (approved):
tACK, thanks @ViktorTigerstrom!

PR #951, opened 2025-06-11
swapserverrpc: arbitrary static swap amount

hieblmi:
This PR adds an `amount` field to the `ServerStaticAddressLoopInRequest`.


bhandras (approved):
LGTM 🎉

swapserverrpc/staticaddr.proto
```diff
@@ -135,6 +135,15 @@ message ServerStaticAddressLoopInRequest {
     // swap payment. If the timeout is reached the swap will be aborted on the
     // server side and the client can retry the swap with different parameters.
     uint32 payment_timeout_seconds = 8;
+
+    /*
+    The optional swap amount the client is attempting to swap. If specified the
+    server will take out this amount from the total value of provided
+    deposit_outpoints and will send the change back to the static address. If
+    not specified the server will use the total amount of the deposit_outpoints
+    as swap amount.
```
starius: What happens, if the amount is too close to the total amount (so the change would be dust)? I think it makes sense if it fails and suggests to remove the flag to use the whole amount.
Could you add the actual expected behavior in this case to the comment, please?

Also, if the amount is equal to the total amount, it makes sense just to succeed, right? And if higher - fail.

starius (approved):
**LGTM!** 🌴
One comment.

swapserverrpc/staticaddr.proto
hieblmi: I expanded the comment to explain. 

Note, there is no flag to use the whole amount of provided deposits for backwards compatibility. If deposits are specified without an amount the whole amount is used.

hieblmi:
Rebased master

litbot-9000:
@sputn1ck: review reminder

swapserverrpc/staticaddr.proto
```diff
@@ -135,6 +137,17 @@ message ServerStaticAddressLoopInRequest {
...
+    /*
+    The optional swap amount the client is attempting to swap. If specified the
+    server will take out this amount from the total value of provided
+    deposit_outpoints and will send the change back to the static address. If
+    this results in dust change the server will reject the swap request. If the
+    amount is not specified the server will use the total amount of the
+    deposit_outpoints as swap amount without providing an additional flag - this
+    is to maintain backwards compatibility.
+    */
+    uint64 amount = 9;
```
starius, line 150: I propose to specify in the field description that it is measured in sats.

starius (approved):

swapserverrpc/staticaddr.proto
hieblmi: fixed and rebased.

PR #952, opened 2025-06-12
sweepbatcher: fixes for presigned mode

starius:
Fixed TODO left from https://github.com/lightninglabs/loop/pull/891 : re-add sweeps to new batches only after fully confirmed.
In case of a reorg sweeps should not go to another batch but stay in the current batch until it is fully confirmed. Only after that the remaining sweeps are re-added to another batch. Field `sweep.completed` is now set to true only for fully-confirmed sweeps.

Fixed OnChainFeePortion values. There were two mistakes. In case of a swap with multiple sweeps only the fee of the first sweep of a swap was accounted. Another issue is that  rounding diff (the remainder) was attributed to all the sweeps rather than to the first (primary) sweep of the batch. The sweep to attribute the remainder was chosen by comparing SignatureScript which is 
always empty. New approach is to find the primary sweep and to compare its outpoint directly.

#### Pull Request Checklist

### DIFF
--- a/loopout.go
+++ b/loopout.go
@@ -1147,7 +1147,7 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmedV2(globalCtx context.Context,
 	quitChan := make(chan bool, 1)
 
 	defer func() {
-		quitChan <- true
+		close(quitChan)
 	}()
 
 	notifier := sweepbatcher.SpendNotifier{
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -407,9 +407,16 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 		}
 	}
 
+	// Determine the current minimum relay fee based on our chain backend.
+	minRelayFee, err := b.wallet.MinRelayFee(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get minRelayFee: %w", err),
+			false
+	}
+
 	// Cache current height and desired feerate of the batch.
 	currentHeight := b.currentHeight
-	feeRate := b.rbfCache.FeeRate
+	feeRate := max(b.rbfCache.FeeRate, minRelayFee)
 
 	// Append this sweep to an array of sweeps. This is needed to keep the
 	// order of sweeps stored, as iterating the sweeps map does not
@@ -445,13 +452,6 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 	batchAmt := btcutil.Amount(0)
 	for _, sweep := range sweeps {
 		batchAmt += sweep.value
-	}
-
-	// Determine the current minimum relay fee based on our chain backend.
-	minRelayFee, err := b.wallet.MinRelayFee(ctx)
-	if err != nil {
-		return 0, fmt.Errorf("failed to get minRelayFee: %w", err),
-			false
 	}
 
 	// Get a pre-signed transaction.
@@ -507,6 +507,9 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 	// purposes.
 	b.batchTxid = &txHash
 	b.batchPkScript = tx.TxOut[0].PkScript
+
+	// Update cached FeeRate not to broadcast a tx with lower feeRate.
+	b.rbfCache.FeeRate = max(b.rbfCache.FeeRate, signedFeeRate)
 
 	return fee, nil, true
 }
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -604,7 +604,7 @@ func CheckSignedTx(unsignedTx, signedTx *wire.MsgTx, inputAmt btcutil.Amount,
 	unsignedOut := unsignedTx.TxOut[0]
 	signedOut := signedTx.TxOut[0]
 	if !bytes.Equal(unsignedOut.PkScript, signedOut.PkScript) {
-		return fmt.Errorf("mismatch of output pkScript: %v, %v",
+		return fmt.Errorf("mismatch of output pkScript: %x, %x",
 			unsignedOut.PkScript, signedOut.PkScript)
 	}
 
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -36,13 +36,7 @@ func (b *batch) ensurePresigned(ctx context.Context, newSweeps []*sweep,
 // presignedTxChecker has methods to check if the inputs are presigned.
 type presignedTxChecker interface {
 	destPkScripter
-
-	// SignTx signs an unsigned transaction or returns a pre-signed tx.
-	// It is only called with loadOnly=true by ensurePresigned.
-	SignTx(ctx context.Context, primarySweepID wire.OutPoint,
-		tx *wire.MsgTx, inputAmt btcutil.Amount,
-		minRelayFee, feeRate chainfee.SatPerKWeight,
-		loadOnly bool) (*wire.MsgTx, error)
+	presigner
 }
 
 // ensurePresigned checks that there is a presigned transaction spending the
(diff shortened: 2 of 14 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -1949,7 +1948,106 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+			Tx: spendTx,
+			OnChainFeePortion: getFeePortionPaidBySweep(
+				spendTx, feePortionPaidPerSweep,
+				roundingDifference, &sweep,
+			),
+		}
+
+		// Dispatch the sweep notifier, we don't care about the outcome
+		// of this action so we don't wait for it.
+		go func() {
```
hieblmi, line 2030: we have to pass in the `sweep` and `spendDetail`, otherwise we might launch the goroutine for the same sweep.

sweepbatcher/sweep_batch.go
```diff
@@ -1949,7 +1948,106 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+
+	return nil
+}
+
+// handleConf handles a confirmation notification. This is the final step of the
+// batch. Here we signal to the batcher that this batch was completed.
+func (b *batch) handleConf(ctx context.Context,
+	conf *chainntnfs.TxConfirmation) error {
+
+	spendTx := conf.Tx
```
hieblmi, line 2061: nit: not likely, but `conf.Tx` could be `null`.

litbot-9000:
@bhandras: review reminder
@sputn1ck: review reminder

sweepbatcher/sweep_batch.go
starius: This is not necessary after Go 1.22: https://go.dev/wiki/LoopvarExperiment

sweepbatcher/sweep_batch.go
starius: Fixed in a separate commit.

sweepbatcher/sweep_batch.go
```diff
@@ -1953,7 +1952,106 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+				roundingDifference, &sweep,
+			),
+		}
+
+		// Dispatch the sweep notifier, we don't care about the outcome
+		// of this action so we don't wait for it.
+		go func() {
+			// Make sure this context doesn't expire so we
+			// successfully notify the caller.
+			ctx := context.WithoutCancel(ctx)
```
bhandras, line 2033: nit: alternatively we could just create a new `context.Background()` and use that.

sweepbatcher/sweep_batch.go
```diff
@@ -1953,7 +1952,106 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
 	}
 
-	// Determine if we should use presigned mode for the batch.
+	// Make a set of confirmed sweeps.
+	confirmedSet := make(map[wire.OutPoint]struct{}, len(spendTx.TxIn))
+	for _, txIn := range spendTx.TxIn {
+		confirmedSet[txIn.PreviousOutPoint] = struct{}{}
+	}
+
+	// As a previous version of the batch transaction may get confirmed,
```
bhandras, line 1965: nit: I think this comment is a bit misleading as we don't really collect the non-confirmed sweeps here to feed them back to the batcher, we just notify those that confirmed and then start the monitoring for 3 conf after which we will do the collection and re-feeding.

sweepbatcher/presigned.go
```diff
@@ -604,8 +605,9 @@ func CheckSignedTx(unsignedTx, signedTx *wire.MsgTx, inputAmt btcutil.Amount,
 	unsignedOut := unsignedTx.TxOut[0]
 	signedOut := signedTx.TxOut[0]
 	if !bytes.Equal(unsignedOut.PkScript, signedOut.PkScript) {
-		return fmt.Errorf("mismatch of output pkScript: %v, %v",
-			unsignedOut.PkScript, signedOut.PkScript)
+		return fmt.Errorf("mismatch of output pkScript: %s, %s",
```
bhandras: nit: can just use %x

sweepbatcher/sweep_batcher.go
```diff
@@ -728,7 +730,14 @@ func (b *Batcher) PresignSweepsGroup(ctx context.Context, inputs []Input,
...
 	}
-	infof("PresignSweepsGroup: nextBlockFeeRate is %v", nextBlockFeeRate)
+	destPkscript, err := txscript.PayToAddrScript(destAddress)
+	if err != nil {
+		return fmt.Errorf("txscript.PayToAddrScript failed: %w", err)
+	}
+	infof("PresignSweepsGroup: nextBlockFeeRate is %v, inputs: %v, "+
+		"destAddress: %v, destPkscript: %v sweepTimeout: %d",
+		nextBlockFeeRate, inputs, destAddress,
+		hex.EncodeToString(destPkscript), sweepTimeout)
```
bhandras: nit: can just use %x

bhandras (approved):
LGTM! 🎉

sweepbatcher/sweep_batch.go
```diff
@@ -1953,7 +1952,106 @@ func (b *batch) handleSpend(ctx context.Context, spendTx *wire.MsgTx) error {
...
+
+			sweep.notifySweepSpend(ctx, &spendDetail)
+		}()
+	}
+
+	b.Infof("spent, confirmed sweeps: %v", confirmedSweeps)
+
+	// We are no longer able to accept new sweeps, so we mark the batch as
+	// closed and persist on storage.
+	b.state = Closed
```
hieblmi, line 2043: Just checking, this couldn't this result in races because of `b.scheduleNextCall()`, right? E.g. in https://github.com/lightninglabs/loop/blob/38bf7642d5d734be5f3681a5fee10edc6d63f889/sweepbatcher/sweep_batch.go#L515

sweepbatcher/sweep_batcher.go
starius: Fixed

sweepbatcher/presigned.go
starius: Fixed

sweepbatcher/sweep_batch.go
starius: The comment was outdated. I updated it:
```go
// As a previous version of the batch transaction may get confirmed,
// which does not contain the latest sweeps, we need to detect which
// sweeps are in the transaction to correctly calculate fee portions
// and notify proper sweeps.
```

sweepbatcher/sweep_batch.go
starius: I prefer `context.WithoutCancel`, because it inherits key-values of the parent context. This is used for new logging.

sweepbatcher/store.go
```diff
@@ -236,7 +236,7 @@ type dbSweep struct {
 	// Amount is the amount of the sweep.
 	Amount btcutil.Amount
 
-	// Completed indicates whether this sweep is completed.
+	// Completed indicates whether this sweep is fully-confirmed.
```
hieblmi, line 231: nit: to clarify maybe say "... this sweep is considered reorg-safely confirmed."

sweepbatcher/sweep_batcher.go
```diff
@@ -274,7 +274,7 @@ type addSweepsRequest struct {
 	notifier *SpendNotifier
 
 	// completed is set if the sweep is spent and the spending transaction
-	// is confirmed.
+	// is fully confirmed.
```
hieblmi: nit: to clarify maybe say "... the spending transaction is considered reorg-safely confirmed."

loopout.go
```diff
@@ -1147,7 +1147,7 @@ func (s *loopOutSwap) waitForHtlcSpendConfirmedV2(globalCtx context.Context,
 	quitChan := make(chan bool, 1)
 
 	defer func() {
-		quitChan <- true
+		close(quitChan)
```
hieblmi, line 1150: so before this change, only one of multiple selects would have ended?

sweepbatcher/sweep_batcher.go
```diff
@@ -202,8 +202,8 @@ type VerifySchnorrSig func(pubKey *btcec.PublicKey, hash, sig []byte) error
 
 // FeeRateProvider is a function that returns min fee rate of a batch sweeping
 // the UTXO of the swap.
-type FeeRateProvider func(ctx context.Context,
-	swapHash lntypes.Hash) (chainfee.SatPerKWeight, error)
+type FeeRateProvider func(ctx context.Context, swapHash lntypes.Hash,
+	utxo wire.OutPoint) (chainfee.SatPerKWeight, error)
```
hieblmi, line 198: why is this needed?

starius: This is needed to check if the UTXO is a static address deposit (function `getIsPresigned`) and to load static loop-in swap data (function `getSweepInfoAndFeeRatePresigned`). We can't do it solely by swah-hash, the UTXO is also needed.

loopout.go
starius: Yes. However normally there is just one `select` simultaneously. But it is better to close so it works correctly if there are multiple `select`s in edge cases.

sweepbatcher/sweep_batcher.go
starius: Fixed. Also adjusted other places.

sweepbatcher/store.go
starius: Fixed.

sweepbatcher/sweep_batch.go
starius: Yes, all such methods are called from the event loop or using `scheduleNextCall` as a form of mutex with the event loop. So they do not overlap in time.

PR #953, opened 2025-06-12
staticaddr: wait for address manager initiation

hieblmi:
This PR reverts a regression that was introduced in 042fe9f3253601fa2093abcbb2c3b42c76d6eb66 by making sure that the address manager is initialized before other managers begin their initialization.

It also applies the init pattern of reservations to staticaddr-managers.

### DIFF
--- a/staticaddr/address/manager.go
+++ b/staticaddr/address/manager.go
@@ -67,13 +67,17 @@ func NewManager(cfg *ManagerConfig, currentHeight int32) *Manager {
 }
 
 // Run runs the address manager.
-func (m *Manager) Run(ctx context.Context) error {
+func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
 	newBlockChan, newBlockErrChan, err :=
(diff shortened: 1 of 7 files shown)


staticaddr/address/manager.go
```diff
@@ -90,6 +99,14 @@ func (m *Manager) Run(ctx context.Context) error {
 	}
 }
 
+// WaitInitComplete waits until the address manager has completed its setup.
+func (m *Manager) WaitInitComplete() {
```
sputn1ck: I'd add waiting for a timeout here, either via passing in a context or just listening for a timer here and then returning an error.

We also have a different pattern already https://github.com/lightninglabs/loop/blob/59f67de3064ec282f03c164da80380f867397965/loopd/daemon.go#L778 but that is more locs in the main function.

hieblmi: I went with your suggestion and applied your pattern to all static services. Will also run this over integration tests.

PR #954, opened 2025-06-13
staticaddr: fix deposit recovery deadlock

hieblmi:
Fixes a deadlock in the deposit recovery that arises if deposits expired while the client was offline.
Since `recoverDeposits` blocks on `fsm.SendEvent(ctx, OnRecover, nil)` the main run loop never gets a chance to read from `m.finalizedDepositChan`.

### DIFF
--- a/staticaddr/deposit/manager.go
+++ b/staticaddr/deposit/manager.go
@@ -191,10 +191,15 @@ func (m *Manager) recoverDeposits(ctx context.Context) error {
 		}
 
 		// Send the OnRecover event to the state machine.
-		err = fsm.SendEvent(ctx, OnRecover, nil)
-		if err != nil {
-			log.Errorf("Error sending OnStart event: %v", err)
-		}
+		// TODO(hieblmi): Add a unit test that would fail with the
+		//  dead-lock before the fsm was passed in.
+		go func(fsm *FSM) {
+			err := fsm.SendEvent(ctx, OnRecover, nil)
+			if err != nil {
+				log.Errorf("Error sending OnStart event: %v",
+					err)
+			}
+		}(fsm)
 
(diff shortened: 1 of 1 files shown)


staticaddr/deposit/manager.go
```diff
@@ -191,10 +191,13 @@ func (m *Manager) recoverDeposits(ctx context.Context) error {
 		}
 
 		// Send the OnRecover event to the state machine.
-		err = fsm.SendEvent(ctx, OnRecover, nil)
-		if err != nil {
-			log.Errorf("Error sending OnStart event: %v", err)
-		}
+		go func(fsm *FSM) {
```
sputn1ck, line 196: I wonder what the tradeoff is between having this in a go func or more asynchronous fsm actions? from what I can tell in effect it is the same?

sputn1ck: But with most recent projects I tend to force tha actions to be asynchronous as this means  we can go events only vs listening for channels everywhere where we need to interact with a blocking action

staticaddr/deposit/manager.go
```diff
@@ -191,10 +191,13 @@ func (m *Manager) recoverDeposits(ctx context.Context) error {
...
+			err := fsm.SendEvent(ctx, OnRecover, nil)
+			if err != nil {
+				log.Errorf("Error sending OnStart event: %v",
+					err)
+			}
+		}(fsm)
```
starius, line 202: I propose to add a unit test that would fail with the dead-lock before this change and to make sure this goroutine doesn't leak in that test.

hieblmi: Given that is is a pattern we use frequently I only left a unit-test todo here.

PR #956, opened 2025-06-16
staticaddr: minor fixes

hieblmi:
_depends on https://github.com/lightninglabs/loop/pull/954_


staticaddr/deposit/manager.go
```diff
@@ -202,7 +208,13 @@ func (m *Manager) recoverDeposits(ctx context.Context) error {
 		m.mu.Lock()
 		m.activeDeposits[d.OutPoint] = fsm
 		m.mu.Unlock()
+
+		numRecoveredDeposits++
 	}
+	wg.Wait()
```
sputn1ck: I don't think this commit makes sense. If we recover into the action `waitforexpirysweptaction` this recover will block until the tx is confirmed, making loop unusable for uncertain block times.

sputn1ck: A solution would be to make the action non-blocking (e.g. by moving receiving from the notifiers into a goroutine and sending an event to the fsm)

hieblmi: Thank you for the catch @sputn1ck. I removed this commit.

staticaddr/loopin/manager.go
```diff
@@ -473,16 +473,15 @@ func (m *Manager) recoverLoopIns(ctx context.Context) error {
...
+		go func(fsm *FSM, swapHash lntypes.Hash) {
+			err := fsm.SendEvent(ctx, OnRecover, nil)
 			if err != nil {
 				log.Errorf("Error sending OnStart event: %v",
 					err)
 			}
 
 			m.activeLoopIns[swapHash] = fsm
-		}()
+		}(fsm, loopIn.SwapHash)
```
starius, line 484: Why is this needed? `fsm` and `swapHash` are a local variable and they can't change outside of the goroutine.

starius (approved):
**LGTM!**
Not sure if the second commit is needed.

staticaddr/loopin/manager.go
hieblmi: `loopIn` is a iteration variable which might change before the goroutine launches, so this is needed.

PR #958, opened 2025-06-17
loopout: ignore saved HTLC txid, add an option to skip some HTLC transactions

starius:
If a reorg happens, the saved txid prevents loop-out from discoveing new txid. So do not use this data after restart.

Added option WithSkippedTxns, which has one historical problematic tx by default. Sweeps originating from these transactions are omitted when reading from DB.

loopdb: add column sweep_batches.cancelled and replaced DropBatch with CancelBatch. It is needed, because sweep.batch_id is a foreign key to batch.

Changed StoreMock.InsertSweepBatch not to reuse batch_id. This is needed by the test, which checks that new batch has fresh ID.

#### Pull Request Checklist


client.go
```diff
@@ -146,6 +147,10 @@ type ClientConfig struct {
 	// be attempted.
 	LoopOutMaxParts uint32
 
+	// SkippedTxs is the list of existing HTLC txids to skip when starting
+	// Loop. This should only be used if affected by the historical bug.
```
hieblmi, line 151: nit: maybe mention reorg here, so `the historical reorg bug`. Should we also include reorg here to make it easier to understand, like `SkippedReorgdTxns`?

client.go
```diff
@@ -265,8 +266,29 @@ func NewClient(dbDir string, loopDB loopdb.SwapStore,
 		// delay time to sweepbatcher's handling. The delay used in
 		// loopout.go is repushDelay.
 		sweepbatcher.WithPublishDelay(
-			repushDelay+additionalDelay,
+			repushDelay + additionalDelay,
 		),
+	}
+
+	if len(cfg.SkippedTxs) != 0 {
+		skippedTxs := make(map[chainhash.Hash]struct{})
```
hieblmi: nit: I think usually we name it `txns` instead of `txs`.

loopd/config.go
```diff
@@ -184,6 +184,8 @@ type Config struct {
 
 	LoopOutMaxParts uint32 `long:"loopoutmaxparts" description:"The maximum number of payment parts that may be used for a loop out swap."`
 
+	SkippedTxs []string `long:"skippedtxs" description:"The list of existing HTLC txids to skip when starting Loop. This should only be used if affected by the historical bug."`
```
hieblmi: Does this only need to be provided once on a restart and can then be skipped for future runs of loopd. Maybe we document the requirement here.

Could we also add a section for this new value in the `sample-loopd.conf`?

alexbosworth (approved):

client.go
starius: Fixed.

loopd/config.go
starius: This option is intended for testing mostly. It is not expected to be set in prod. I added `hidden:"true"` to it.

sweepbatcher/sweep_batcher.go
```diff
@@ -574,6 +591,14 @@ func NewBatcher(wallet lndclient.WalletKitClient,
 	store BatcherStore, sweepStore SweepFetcher,
 	opts ...BatcherOption) *Batcher {
 
+	badTx1, err := chainhash.NewHashFromStr(
+		"7028bdac753a254785d29506f311abcda323706b531345105f38999" +
+			"aecd6f3d1",
+	)
+	if err != nil {
+		panic(err)
```
sputn1ck, line 599: nit: i think we should return error here instead of panicing to allow the main function to close properly

starius: I don't want to change the interface of `NewBatcher`, it is used in many places. This won't panic, because the input is a constant.

PR #959, opened 2025-06-17
staticaddr: method to fetch deposits by outpoints

hieblmi:
This PR adds method `DepositsForOutpoints` to the deposit manager and required sub-methods to the deposit store.

This PR was extracted from and is required by https://github.com/lightninglabs/loop/pull/887 to ease review.

### DIFF
--- a/staticaddr/deposit/interface.go
+++ b/staticaddr/deposit/interface.go
@@ -26,6 +26,10 @@ type Store interface {
 	// GetDeposit retrieves a deposit with depositID from the database.
 	GetDeposit(ctx context.Context, depositID ID) (*Deposit, error)
 
+	// DepositForOutpoint retrieves the deposit with the given outpoint.
+	DepositForOutpoint(ctx context.Context, outpoint string) (*Deposit,
+		error)
+
 	// AllDeposits retrieves all deposits from the store.
 	AllDeposits(ctx context.Context) ([]*Deposit, error)
 }
--- a/staticaddr/deposit/manager.go
+++ b/staticaddr/deposit/manager.go
@@ -572,3 +572,36 @@ func (m *Manager) toActiveDeposits(outpoints *[]wire.OutPoint) ([]*FSM,
 
 	return fsms, deposits
 }
+
+// DepositsForOutpoints returns all deposits that are behind the given
+// outpoints.
+func (m *Manager) DepositsForOutpoints(ctx context.Context,
+	outpoints []string) ([]*Deposit, error) {
+
+	// Check for duplicates.
+	existingOutpoints := make(map[string]struct{}, len(outpoints))
+	for i, o := range outpoints {
+		if _, ok := existingOutpoints[o]; ok {
+			return nil, fmt.Errorf("duplicate outpoint %s "+
+				"at index %d", o, i)
+		}
+		existingOutpoints[o] = struct{}{}
+	}
+
+	deposits := make([]*Deposit, 0, len(outpoints))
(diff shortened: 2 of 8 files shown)


staticaddr/deposit/manager.go
```diff
@@ -572,3 +572,38 @@ func (m *Manager) toActiveDeposits(outpoints *[]wire.OutPoint) ([]*FSM,
...
 	return fsms, deposits
 }
+
+// DepositsForOutpoints returns all deposits that are behind the given
+// outpoints.
+func (m *Manager) DepositsForOutpoints(ctx context.Context,
+	outpoints []string) ([]*Deposit, error) {
+
+	// Check for duplicates.
+	duplicates := make(map[string]struct{})
```
bhandras: nit: i'd call this `existingOutpoints`.

staticaddr/deposit/manager.go
```diff
@@ -572,3 +572,38 @@ func (m *Manager) toActiveDeposits(outpoints *[]wire.OutPoint) ([]*FSM,
 
 	return fsms, deposits
 }
+
+// DepositsForOutpoints returns all deposits that are behind the given
+// outpoints.
+func (m *Manager) DepositsForOutpoints(ctx context.Context,
```
bhandras, line 578: Where would we use the function? Could you please add coverage?

bhandras (approved):

staticaddr/deposit/manager.go
hieblmi: Yes sure, it is used to restore deposits of a swap after fetching it from the db, see https://github.com/hieblmi/loop/commit/349b9309b6b15bdd3f7e01b9e9abf66f0c42368b#diff-2c26627ef62d3f06764cb007a83924ef8abb36066a183b4d05e7f42e7c6a7ac9R23-R283

Ideally this would all be handled by `GetLoopInByHash`, but it isn't currently.

staticaddr/deposit/interface.go
```diff
@@ -26,6 +27,10 @@ type Store interface {
 	// GetDeposit retrieves a deposit with depositID from the database.
 	GetDeposit(ctx context.Context, depositID ID) (*Deposit, error)
 
+	// DepositForOutpoint retrieves the deposit with the given outpoint.
+	DepositForOutpoint(ctx context.Context, txHash chainhash.Hash,
+		idx uint32) (*Deposit, error)
```
starius: I propose to change the signature to more high level:
```go
DepositForOutpoint(ctx context.Context, outpoint string) (*Deposit, error)
```

staticaddr/deposit/manager.go
starius: I propose to pre-allocate the map. We now the max size.

staticaddr/deposit/sql_store.go
```diff
@@ -149,6 +149,44 @@ func (s *SqlStore) GetDeposit(ctx context.Context, id ID) (*Deposit, error) {
...
+			if err != nil {
+				return err
+			}
+
+			latestUpdate, err := q.GetLatestDepositUpdate(
+				ctx, row.DepositID,
+			)
+			if err != nil {
+				return err
+			}
```
starius, line 178: We can do it with SQL join and avoid making an explicit DB transaction (ExecTx).

starius (approved):
LGTM! 💾
Few comments.

hieblmi:
Rebased master

hieblmi:
Rebased master

PR #967, opened 2025-06-23
staticaddr: neutrino fixes

starius:
Recovered withdrawal transactions are now published in parallel to speed-up starting Loop.
Also increased the timeout given for this startup (including the withdrawals' recovery) from 10s to 1m.

Also added logging and removed an unused method in an interface.

#### Pull Request Checklist

### DIFF
--- a/staticaddr/deposit/fsm.go
+++ b/staticaddr/deposit/fsm.go
@@ -399,7 +399,7 @@ func (f *FSM) updateDeposit(ctx context.Context,
 
 	err := f.cfg.Store.UpdateDeposit(ctx, f.deposit)
 	if err != nil {
-		f.Errorf("unable to update deposit: %w", err)
+		f.Errorf("unable to update deposit: %v", err)
 	}
 }
 


staticaddr/withdraw/manager.go
```diff
@@ -577,7 +580,7 @@ func (m *Manager) publishFinalizedWithdrawalTx(ctx context.Context,
 			return false, nil
 		}
 	} else {
-		log.Debugf("published deposit withdrawal with txid: %v",
+		log.Debugf("Published deposit withdrawal with txid: %v!",
```
hieblmi: nit: when copying the txid from the log the `...` and `!` might be copied too, not sure if we want that.

loopd/daemon.go
```diff
@@ -922,10 +922,15 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 			}
 		}()
 
+		// We need a higher timeout here, because withdrawalManager
+		// publishes transactions and each PublishTransaction call can
+		// wait for getting inv messages from a peer (neutrino).
```
hieblmi, line 927: 🙏

staticaddr/withdraw/manager.go
```diff
@@ -238,32 +239,45 @@ func (m *Manager) recoverWithdrawals(ctx context.Context) error {
...
-		tx := finalizedWithdrawalTx
-		err = m.cfg.DepositManager.TransitionDeposits(
-			ctx, deposits, deposit.OnWithdrawInitiated,
-			deposit.Withdrawing,
-		)
-		if err != nil {
-			return err
-		}
+	for tx, deposits := range depositsByWithdrawalTx {
+		eg.Go(func() error {
```
hieblmi, line 251: We should pass in `tx` and `deposits` to the goroutine to make sure each goroutine operates on the right variable.

hieblmi (commented):
Thanks for the improvement. Only see one issue, otherwise looks good to go.

staticaddr/withdraw/manager.go
starius: I removed `!` and added a space before `...`.

staticaddr/withdraw/manager.go
starius: We can't pass additional arguments to [Go](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go). Modern Go has `loopvar` always enabled, so this should work as expected.

starius:
I pushed another commit `staticaddr: use tx hash not *wire.MsgTx as a key`. Noticed that `recoverWithdrawals` relies on GetActiveDepositsInState reusing the same *wire.MsgTx pointer for a unique transaction. Reworked this using txid to remove this assumption.

staticaddr/withdraw/manager.go
hieblmi: I think we then should capture `txid` and `deposits` as suggested here: https://go.dev/doc/faq#closures_and_goroutines, like
```
for _, v := range values {
        v := v // create a new 'v'.
        go func() {
            fmt.Println(v)
            done <- true
        }()
    }
```

staticaddr/withdraw/manager.go
```diff
@@ -227,15 +227,18 @@ func (m *Manager) recoverWithdrawals(ctx context.Context) error {
 	}
 
 	// Group the deposits by their finalized withdrawal transaction.
-	depositsByWithdrawalTx := make(map[*wire.MsgTx][]*deposit.Deposit)
+	depositsByWithdrawalTx := make(map[chainhash.Hash][]*deposit.Deposit)
```
hieblmi, line 230: thanks for the fix!

staticaddr/withdraw/manager.go
starius: This is no longer needed. See the end of that section:

> This behavior of the language, not defining a new variable for each iteration, was considered a mistake in retrospect, and has been addressed in [Go 1.22](https://go.dev/wiki/LoopvarExperiment), which does indeed create a new variable for each iteration, eliminating this issue.

starius:
Rebased and pushed another commit with logging fix "staticaddr: fix error formatting in the log".

PR #975, opened 2025-07-09
sweepbatcher: fix reorg detection

starius:
The reorg channel is now passed to `RegisterSpendNtfn`, and waiting for spend notifications remains active even after the transaction receives its first confirmation. The dedicated goroutine previously used to wait for the spend is no longer needed, as we now handle both spend and potential reorg events in the main event loop while the batch is running.

Following this change, `RegisterConfirmationsNtfn` runs without a reorg channel, as it would only detect deep reorgs that undo the final confirmation - something we can't handle anyway. We can't track fully confirmed swaps indefinitely to guard against such rare reorgs; instead, we can mitigate the risk by increasing the required confirmation depth.

Also changed sweepbatcher to update feerate independently of `AddSweep` as well. `AddSweep` may not be called after getting the first confirmation, but feerate updates are still needed in case of reorg. Update test `TestFeeRateGrows` not to call `AddSweep` again and make sure feerate is updated itself.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -169,9 +169,10 @@ type batchConfig struct {
 	// initial delay completion and publishing the batch transaction.
 	batchPublishDelay time.Duration
 
-	// noBumping instructs sweepbatcher not to fee bump itself and rely on
-	// external source of fee rates (FeeRateProvider).
-	noBumping bool
+	// customFeeRate provides custom min fee rate per swap. The batch uses
+	// max of the fee rates of its swaps. In this mode confTarget is
+	// ignored and fee bumping by sweepbatcher is disabled.
+	customFeeRate FeeRateProvider
 
 	// txLabeler is a function generating a transaction label. It is called
 	// before publishing a batch transaction. Batch ID is passed to it.
@@ -723,6 +724,9 @@ func (b *batch) addSweeps(ctx context.Context, sweeps []*sweep) (bool, error) {
 			// lower that minFeeRate of other sweeps (so it is
 			// applied).
 			if b.rbfCache.FeeRate < s.minFeeRate {
+				b.Infof("Increasing feerate of the batch "+
+					"from %v to %v", b.rbfCache.FeeRate,
+					s.minFeeRate)
 				b.rbfCache.FeeRate = s.minFeeRate
 			}
 		}
@@ -769,6 +773,9 @@ func (b *batch) addSweeps(ctx context.Context, sweeps []*sweep) (bool, error) {
 		// Update FeeRate. Max(s.minFeeRate) for all the sweeps of
 		// the batch is the basis for fee bumps.
 		if b.rbfCache.FeeRate < s.minFeeRate {
+			b.Infof("Increasing feerate of the batch "+
+				"from %v to %v", b.rbfCache.FeeRate,
+				s.minFeeRate)
 			b.rbfCache.FeeRate = s.minFeeRate
 			b.rbfCache.SkipNextBump = true
 		}
@@ -968,6 +975,12 @@ func (b *batch) Run(ctx context.Context) error {
 				continue
 			}
 
+			// Update feerate of sweeps. This is normally done by
+			// AddSweep, but it may not be called after the sweep
+			// is confirmed, but fresh feerate is still needed to
+			// keep publishing in case of reorg.
+			b.updateFeeRate(ctx)
+
 			err := b.publish(ctx)
 			if err != nil {
 				return fmt.Errorf("publish error: %w", err)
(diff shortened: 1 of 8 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -979,23 +979,26 @@ func (b *batch) Run(ctx context.Context) error {
 				return fmt.Errorf("handleSpend error: %w", err)
 			}
 
+		case err := <-b.spendErrChan:
```
hieblmi, line 995: unrelated to this, but since the `Run` method is quite long, could we move the code at the beginning `for` around `initialDelay` and `initialDelayChan` into a separate function?

sweepbatcher/sweep_batch.go
```diff
@@ -1812,44 +1815,31 @@ func (b *batch) updateRbfRate(ctx context.Context) error {
...
+	if b.spendChan != nil || b.spendErrChan != nil || b.reorgChan != nil {
+		return fmt.Errorf("an attempt to run monitorSpend multiple " +
+			"times per batch")
+	}
 
-	spendChan, spendErr, err := b.chainNotifier.RegisterSpendNtfn(
-		spendCtx, &primarySweep.outpoint, primarySweep.htlc.PkScript,
+	reorgChan := make(chan struct{}, 1)
+
+	spendChan, spendErrChan, err := b.chainNotifier.RegisterSpendNtfn(
```
hieblmi, line 1873: Will `RegisterSpendNtfn` keep `spendChan` and `spendErrChan` open and signal spends/errors after the reorg chan fired?
I assume yes, so this is a nice simplification of the spend-handling flow.

sweepbatcher/sweep_batcher.go
```diff
@@ -1567,11 +1551,53 @@ func (b *Batcher) loadSweep(ctx context.Context, swapHash lntypes.Hash,
...
 
+// feeRateEstimator determines feerate by confTarget.
+type feeRateEstimator interface {
+	// EstimateFeeRate returns feerate corresponding to the confTarget.
+	EstimateFeeRate(ctx context.Context,
+		confTarget int32) (chainfee.SatPerKWeight, error)
+}
+
+// minimumSweepFeeRate determines minimum feerate for a sweep.
+func minimumSweepFeeRate(ctx context.Context, customFeeRate FeeRateProvider,
```
hieblmi, line 1562: great that you ensured that `AddSweep` and `Run` in the reorg case call the same fee rate estimates.

hieblmi (approved):
This LGTM! I just have a question about the behavior of `RegisterSpendNtfn` in case a reorg happens.

sweepbatcher/sweep_batch.go
starius: `initialDelay`, `initialDelayChan` and related variables in the beginning of `Run` are too connected to each other. In the current version they are all local and it is easy to follow the logic of waiting. How to factor out these variables without sacrificing readability?

I updated the last commit: factored out feerate updates to `updateFeeRate` method.

sweepbatcher/sweep_batch.go
starius: Yes, I added an itest for LND to check this: https://github.com/lightningnetwork/lnd/pull/10083

test/chainnotifier_mock.go
```diff
@@ -10,7 +11,6 @@ import (
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/lnrpc/chainrpc"
-	"golang.org/x/net/context"
```
bhandras, line 13: Ugh, nice catch!

sweepbatcher/sweep_batch.go
```diff
@@ -1812,44 +1815,31 @@ func (b *batch) updateRbfRate(ctx context.Context) error {
...
-			b.writeToSpendErrChan(ctx, err)
-
-			b.writeToErrChan(
-				fmt.Errorf("spend error: %w", err),
-			)
-
-		case <-ctx.Done():
-		}
-	}()
+	b.spendChan = spendChan
```
bhandras, line 1890: Nice simplification. I'd add a comment that this is safe to do as we're always calling `monitorSpend` from the runloop's goroutine.

sweepbatcher/sweep_batcher.go
```diff
@@ -1567,11 +1551,53 @@ func (b *Batcher) loadSweep(ctx context.Context, swapHash lntypes.Hash,
...
+		}
+		if minFeeRate < chainfee.AbsoluteFeePerKwFloor {
+			return 0, fmt.Errorf("min fee rate too low (%v) for "+
+				"%x", minFeeRate, swapHash[:6])
+		}
+
+		return minFeeRate, nil
+	}
+
+	if sweepConfTarget == 0 {
```
bhandras: LND's walletkit will fail with conftarget of 0 or 1. A way around would be override it to 2 if smaller?

bhandras (approved):
LGTM 🎉 (just one comment regarding conf target!)

sweepbatcher/sweep_batcher.go
starius: I added this check in a separate commit. Note that this check is unlikely to fire, since this is already fixed in GetConfTargetAndFeeRate.

sweepbatcher/sweep_batch.go
starius: Thanks! I added the comment.

PR #976, opened 2025-07-09
sweepbatcher: batch change outputs

hieblmi:
The batcher receives the ability to sign and publish batches with change outputs.

Change is created when the clients choose fractional swap amounts from the input values to be swapped. Usually the change is sent back to client's addresses.

If change is provided to `ensurePresigned`, `presign` and `publishPresigned`, it is expected to be placed within the passed `[]sweep`.
Each `sweep` optionally holds a pointer to a change output. Only the first sweep in a sweep group called primary deposit will hold a pointer to the respective swap's change output.
- [x] unit test coverage for change outputs

### DIFF
--- a/sweepbatcher/greedy_batch_selection.go
+++ b/sweepbatcher/greedy_batch_selection.go
@@ -210,6 +210,13 @@ func estimateBatchWeight(batch *batch) (feeDetails, error) {
 			err)
 	}
 
+	// Add change output weights.
+	for _, s := range batch.sweeps {
+		if s.change != nil {
+			weight.AddOutput(s.change.PkScript)
+		}
+	}
+
 	// Add inputs.
 	for _, sweep := range batch.sweeps {
 		if sweep.nonCoopHint || sweep.coopFailed {
--- a/sweepbatcher/greedy_batch_selection.go
+++ b/sweepbatcher/greedy_batch_selection.go
@@ -210,11 +210,21 @@ func estimateBatchWeight(batch *batch) (feeDetails, error) {
 			err)
 	}
 
-	// Add change output weights.
+	// Add change output weights. Change outputs with identical pkscript
+	// will be consolidated into a single output.
+	changeOutputs := make(map[string]struct{})
 	for _, s := range batch.sweeps {
-		if s.change != nil {
-			weight.AddOutput(s.change.PkScript)
+		if s.change == nil {
+			continue
 		}
+
+		pkScriptString := string(s.change.PkScript)
+		if _, has := changeOutputs[pkScriptString]; has {
+			continue
+		}
+
+		weight.AddOutput(s.change.PkScript)
+		changeOutputs[pkScriptString] = struct{}{}
 	}
 
 	// Add inputs.
--- a/sweepbatcher/greedy_batch_selection_test.go
+++ b/sweepbatcher/greedy_batch_selection_test.go
@@ -36,6 +36,7 @@ const (
 
 	coopTwoSweepBatchWeight          = coopNewBatchWeight + coopInputWeight
 	coopSingleSweepChangeBatchWeight = coopInputWeight + batchOutputWeight + changeOutputWeight
+	coopDoubleSweepChangeBatchWeight = 2*coopInputWeight + batchOutputWeight + changeOutputWeight
 	nonCoopTwoSweepBatchWeight       = coopTwoSweepBatchWeight + 2*nonCoopPenalty
 	v2v3BatchWeight                  = nonCoopTwoSweepBatchWeight - 25
 	mixedTwoSweepBatchWeight         = coopTwoSweepBatchWeight + nonCoopPenalty
@@ -322,6 +323,35 @@ func TestEstimateBatchWeight(t *testing.T) {
 				BatchId: 1,
 				FeeRate: lowFeeRate,
 				Weight:  coopSingleSweepChangeBatchWeight,
+			},
+		},
+
+		{
+			name: "two sweeps regular batch with identical change",
+			batch: &batch{
+				id: 1,
+				rbfCache: rbfCache{
+					FeeRate: lowFeeRate,
+				},
+				sweeps: map[wire.OutPoint]sweep{
+					outpoint1: {
+						htlcSuccessEstimator: se3,
+						change: &wire.TxOut{
+							PkScript: changePkscript,
+						},
+					},
+					outpoint2: {
+						htlcSuccessEstimator: se3,
+						change: &wire.TxOut{
+							PkScript: changePkscript,
+						},
+					},
+				},
+			},
+			wantBatchFeeDetails: feeDetails{
+				BatchId: 1,
+				FeeRate: lowFeeRate,
+				Weight:  coopDoubleSweepChangeBatchWeight,
 			},
 		},
 
--- a/sweepbatcher/greedy_batch_selection_test.go
+++ b/sweepbatcher/greedy_batch_selection_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/loop/swap"
 	"github.com/lightningnetwork/lnd/input"
@@ -16,24 +17,28 @@ import (
 
 // Useful constants for tests.
 const (
-	lowFeeRate  = chainfee.FeePerKwFloor
-	highFeeRate = chainfee.SatPerKWeight(30000)
+	lowFeeRate    = chainfee.FeePerKwFloor
+	mediumFeeRate = lowFeeRate + 200
+	highFeeRate   = chainfee.SatPerKWeight(30000)
 
 	coopInputWeight       = lntypes.WeightUnit(230)
+	batchOutputWeight     = lntypes.WeightUnit(343)
 	nonCoopInputWeight    = lntypes.WeightUnit(393)
 	nonCoopPenalty        = nonCoopInputWeight - coopInputWeight
 	coopNewBatchWeight    = lntypes.WeightUnit(444)
 	nonCoopNewBatchWeight = coopNewBatchWeight + nonCoopPenalty
+	changeOutputWeight    = lntypes.WeightUnit(input.P2TROutputSize)
 
 	// p2pkhDiscount is weight discount P2PKH output has over P2TR output.
 	p2pkhDiscount = lntypes.WeightUnit(
 		input.P2TROutputSize-input.P2PKHOutputSize,
 	) * 4
 
-	coopTwoSweepBatchWeight    = coopNewBatchWeight + coopInputWeight
-	nonCoopTwoSweepBatchWeight = coopTwoSweepBatchWeight + 2*nonCoopPenalty
-	v2v3BatchWeight            = nonCoopTwoSweepBatchWeight - 25
-	mixedTwoSweepBatchWeight   = coopTwoSweepBatchWeight + nonCoopPenalty
+	coopTwoSweepBatchWeight          = coopNewBatchWeight + coopInputWeight
+	coopSingleSweepChangeBatchWeight = coopInputWeight + batchOutputWeight + changeOutputWeight
+	nonCoopTwoSweepBatchWeight       = coopTwoSweepBatchWeight + 2*nonCoopPenalty
+	v2v3BatchWeight                  = nonCoopTwoSweepBatchWeight - 25
+	mixedTwoSweepBatchWeight         = coopTwoSweepBatchWeight + nonCoopPenalty
 )
 
 // testHtlcV2SuccessEstimator adds weight of non-cooperative input to estimator
@@ -265,6 +270,13 @@ func TestEstimateBatchWeight(t *testing.T) {
 	se3 := testHtlcV3SuccessEstimator
 	trAddr := (*btcutil.AddressTaproot)(nil)
 
+	changeAddr := "bc1pdx9ggvtjjcpaqfqk375qhdmzx9xu8dcu7w94lqfcxhh0rj" +
+		"lwyyeq5ryn6r"
+	changeAddress, err := btcutil.DecodeAddress(changeAddr, nil)
+	require.NoError(t, err)
+	changePkscript, err := txscript.PayToAddrScript(changeAddress)
+	require.NoError(t, err)
+
 	cases := []struct {
 		name                string
 		batch               *batch
@@ -287,6 +299,29 @@ func TestEstimateBatchWeight(t *testing.T) {
 				BatchId: 1,
 				FeeRate: lowFeeRate,
 				Weight:  coopNewBatchWeight,
+			},
+		},
+
+		{
+			name: "one sweep regular batch with change",
+			batch: &batch{
+				id: 1,
+				rbfCache: rbfCache{
+					FeeRate: lowFeeRate,
+				},
+				sweeps: map[wire.OutPoint]sweep{
+					outpoint1: {
+						htlcSuccessEstimator: se3,
+						change: &wire.TxOut{
+							PkScript: changePkscript,
+						},
+					},
+				},
+			},
+			wantBatchFeeDetails: feeDetails{
+				BatchId: 1,
+				FeeRate: lowFeeRate,
+				Weight:  coopSingleSweepChangeBatchWeight,
 			},
 		},
 
@@ -774,6 +809,47 @@ func TestSelectBatches(t *testing.T) {
 			},
 			oneSweepBatch: feeDetails{
 				FeeRate: highFeeRate,
+				Weight:  coopNewBatchWeight,
+			},
+			wantBestBatchesIds: []int32{1, newBatchSignal},
+		},
+
+		{
+			name: "low fee change sweep, placed in new batch",
+			batches: []feeDetails{
+				{
+					BatchId: 1,
+					FeeRate: mediumFeeRate,
+					Weight:  coopNewBatchWeight,
+				},
+			},
+			sweep: feeDetails{
+				FeeRate: lowFeeRate,
+				Weight:  coopInputWeight + changeOutputWeight,
+			},
+			oneSweepBatch: feeDetails{
+				FeeRate: lowFeeRate,
+				Weight:  coopNewBatchWeight,
+			},
+			wantBestBatchesIds: []int32{newBatchSignal, 1},
+		},
+
+		{
+			name: "low fee sweep without change, placed in " +
+				"existing batch",
+			batches: []feeDetails{
+				{
+					BatchId: 1,
+					FeeRate: mediumFeeRate,
+					Weight:  coopNewBatchWeight,
+				},
+			},
+			sweep: feeDetails{
+				FeeRate: lowFeeRate,
+				Weight:  coopInputWeight,
+			},
+			oneSweepBatch: feeDetails{
+				FeeRate: lowFeeRate,
 				Weight:  coopNewBatchWeight,
 			},
 			wantBestBatchesIds: []int32{1, newBatchSignal},
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -78,7 +78,7 @@ func ensurePresigned(ctx context.Context, newSweeps []*sweep,
 	const currentHeight = 0
 
 	// Check if we can sign with minimum fee rate.
-	const feeRate = chainfee.FeePerKwFloor
+	feeRate := minRelayFeeRate
 
 	tx, _, _, _, err := constructUnsignedTx(
 		sweeps, destAddr, currentHeight, feeRate, minRelayFeeRate,
@@ -330,6 +330,10 @@ func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
 		return fmt.Errorf("nextBlockFeeRate is not set")
 	}
 
+	if minRelayFeeRate == 0 {
+		return fmt.Errorf("minRelayFeeRate is not set")
+	}
+
 	// Keep track of the total amount this batch is sweeping back.
 	batchAmt := btcutil.Amount(0)
 	for _, sweep := range sweeps {
@@ -345,9 +349,9 @@ func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
 		return fmt.Errorf("timeout is invalid: %d", timeout)
 	}
 
-	// Go from the floor (1.01 sat/vbyte) to 2k sat/vbyte with step of 1.2x.
+	// Go from minRelayFeeRate to 2k sat/vbyte with step of 1.2x.
+	var start = minRelayFeeRate
 	const (
-		start            = chainfee.FeePerKwFloor
 		stop             = chainfee.AbsoluteFeePerKwFloor * 2_000
 		factorPPM        = 1_200_000
 		timeoutThreshold = 50
@@ -384,12 +388,9 @@ func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
 		}
 
 		// Try to presign this transaction.
-		const (
-			loadOnly    = false
-			minRelayFee = chainfee.AbsoluteFeePerKwFloor
-		)
+		const loadOnly = false
 		_, err = presigner.SignTx(
-			ctx, primarySweepID, tx, batchAmt, minRelayFee, fr,
+			ctx, primarySweepID, tx, batchAmt, minRelayFeeRate, fr,
 			loadOnly,
 		)
 		if err != nil {
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -51,6 +51,7 @@ func ensurePresigned(ctx context.Context, newSweeps []*sweep,
 			outpoint:  s.outpoint,
 			value:     s.value,
 			presigned: s.presigned,
+			change:    s.change,
 		}
 	}
 
@@ -493,10 +494,12 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 	signedFeeRate := chainfee.NewSatPerKWeight(fee, realWeight)
 
 	numSweeps := len(tx.TxIn)
+	numChange := len(tx.TxOut) - 1
 	b.Infof("attempting to publish custom signed tx=%v, desiredFeerate=%v,"+
-		" signedFeeRate=%v, weight=%v, fee=%v, sweeps=%d, destAddr=%s",
+		" signedFeeRate=%v, weight=%v, fee=%v, sweeps=%d, "+
+		"changeOutputs=%d, destAddr=%s",
 		txHash, feeRate, signedFeeRate, realWeight, fee, numSweeps,
-		address)
+		numChange, address)
 	b.debugLogTx("serialized batch", tx)
 
 	// Publish the transaction.
@@ -593,23 +596,31 @@ func CheckSignedTx(unsignedTx, signedTx *wire.MsgTx, inputAmt btcutil.Amount,
 	}
 
 	// Compare outputs.
-	if len(unsignedTx.TxOut) != 1 {
-		return fmt.Errorf("unsigned tx has %d outputs, want 1",
-			len(unsignedTx.TxOut))
-	}
-	if len(signedTx.TxOut) != 1 {
-		return fmt.Errorf("the signed tx has %d outputs, want 1",
+	if len(unsignedTx.TxOut) != len(signedTx.TxOut) {
+		return fmt.Errorf("unsigned tx has %d outputs, signed tx has "+
+			"%d outputs, should be equal", len(unsignedTx.TxOut),
 			len(signedTx.TxOut))
 	}
-	unsignedOut := unsignedTx.TxOut[0]
-	signedOut := signedTx.TxOut[0]
-	if !bytes.Equal(unsignedOut.PkScript, signedOut.PkScript) {
-		return fmt.Errorf("mismatch of output pkScript: %x, %x",
-			unsignedOut.PkScript, signedOut.PkScript)
+	for i, o := range unsignedTx.TxOut {
+		if !bytes.Equal(o.PkScript, signedTx.TxOut[i].PkScript) {
+			return fmt.Errorf("mismatch of output pkScript: %x, %x",
+				o.PkScript, signedTx.TxOut[i].PkScript)
+		}
+		if i != 0 && o.Value != signedTx.TxOut[i].Value {
+			return fmt.Errorf("mismatch of output value: %d, %d",
+				o.Value, signedTx.TxOut[i].Value)
+		}
+	}
+
+	// Calculate the total value of all outputs to help determine the
+	// transaction fee.
+	totalOutputValue := btcutil.Amount(0)
+	for _, o := range signedTx.TxOut {
+		totalOutputValue += btcutil.Amount(o.Value)
 	}
 
 	// Find the feerate of signedTx.
-	fee := inputAmt - btcutil.Amount(signedOut.Value)
+	fee := inputAmt - totalOutputValue
 	weight := lntypes.WeightUnit(
 		blockchain.GetTransactionWeight(btcutil.NewTx(signedTx)),
 	)
--- a/sweepbatcher/presigned.go
+++ b/sweepbatcher/presigned.go
@@ -28,8 +28,14 @@ func (b *batch) ensurePresigned(ctx context.Context, newSweeps []*sweep,
 			"adding to an empty batch")
 	}
 
+	minRelayFeeRate, err := b.wallet.MinRelayFee(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get minRelayFee: %w", err)
+	}
+
 	return ensurePresigned(
-		ctx, newSweeps, b.cfg.presignedHelper, b.cfg.chainParams,
+		ctx, newSweeps, b.cfg.presignedHelper, minRelayFeeRate,
+		b.cfg.chainParams,
 	)
 }
 
@@ -43,6 +49,7 @@ type presignedTxChecker interface {
 // inputs of this group only.
 func ensurePresigned(ctx context.Context, newSweeps []*sweep,
 	presignedTxChecker presignedTxChecker,
+	minRelayFeeRate chainfee.SatPerKWeight,
 	chainParams *chaincfg.Params) error {
 
 	sweeps := make([]sweep, len(newSweeps))
@@ -74,7 +81,7 @@ func ensurePresigned(ctx context.Context, newSweeps []*sweep,
 	const feeRate = chainfee.FeePerKwFloor
 
 	tx, _, _, _, err := constructUnsignedTx(
-		sweeps, destAddr, currentHeight, feeRate,
+		sweeps, destAddr, currentHeight, feeRate, minRelayFeeRate,
 	)
 	if err != nil {
 		return fmt.Errorf("failed to construct unsigned tx "+
@@ -218,6 +225,14 @@ func (b *batch) presign(ctx context.Context, newSweeps []*sweep) error {
 
 	b.Infof("nextBlockFeeRate is %v", nextBlockFeeRate)
 
+	// Find the minRelayFeeRate.
+	minRelayFeeRate, err := b.wallet.MinRelayFee(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get minRelayFeeRate: %w", err)
+	}
+
+	b.Infof("minRelayFeeRate is %v", minRelayFeeRate)
+
 	// We need to restore previously added groups. We can do it by reading
 	// all the sweeps from DB (they must be ordered) and grouping by swap.
 	groups, err := b.getSweepsGroups(ctx)
@@ -258,7 +273,7 @@ func (b *batch) presign(ctx context.Context, newSweeps []*sweep) error {
 
 		err = presign(
 			ctx, b.cfg.presignedHelper, destAddr, primarySweepID,
-			sweeps, nextBlockFeeRate,
+			sweeps, nextBlockFeeRate, minRelayFeeRate,
 		)
 		if err != nil {
 			return fmt.Errorf("failed to presign a transaction "+
@@ -300,7 +315,8 @@ type presigner interface {
 // 10x of the current next block feerate.
 func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
 	primarySweepID wire.OutPoint, sweeps []sweep,
-	nextBlockFeeRate chainfee.SatPerKWeight) error {
+	nextBlockFeeRate chainfee.SatPerKWeight,
+	minRelayFeeRate chainfee.SatPerKWeight) error {
 
 	if presigner == nil {
 		return fmt.Errorf("presigner is not installed")
@@ -354,7 +370,7 @@ func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
 	for fr := start; fr <= stop; fr = (fr * factorPPM) / 1_000_000 {
 		// Construct an unsigned transaction for this fee rate.
 		tx, _, feeForWeight, fee, err := constructUnsignedTx(
-			sweeps, destAddr, currentHeight, fr,
+			sweeps, destAddr, currentHeight, fr, minRelayFeeRate,
 		)
 		if err != nil {
 			return fmt.Errorf("failed to construct unsigned tx "+
@@ -411,15 +427,15 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 	}
 
 	// Determine the current minimum relay fee based on our chain backend.
-	minRelayFee, err := b.wallet.MinRelayFee(ctx)
+	minRelayFeeRate, err := b.wallet.MinRelayFee(ctx)
 	if err != nil {
-		return 0, fmt.Errorf("failed to get minRelayFee: %w", err),
+		return 0, fmt.Errorf("failed to get minRelayFeeRate: %w", err),
 			false
 	}
 
 	// Cache current height and desired feerate of the batch.
 	currentHeight := b.currentHeight
-	feeRate := max(b.rbfCache.FeeRate, minRelayFee)
+	feeRate := max(b.rbfCache.FeeRate, minRelayFeeRate)
 
 	// Append this sweep to an array of sweeps. This is needed to keep the
 	// order of sweeps stored, as iterating the sweeps map does not
@@ -441,7 +457,7 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 
 	// Construct unsigned batch transaction.
 	tx, weight, _, fee, err := constructUnsignedTx(
-		sweeps, address, currentHeight, feeRate,
+		sweeps, address, currentHeight, feeRate, minRelayFeeRate,
 	)
 	if err != nil {
 		return 0, fmt.Errorf("failed to construct tx: %w", err),
@@ -460,7 +476,7 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 	// Get a pre-signed transaction.
 	const loadOnly = false
 	signedTx, err := b.cfg.presignedHelper.SignTx(
-		ctx, b.primarySweepID, tx, batchAmt, minRelayFee, feeRate,
+		ctx, b.primarySweepID, tx, batchAmt, minRelayFeeRate, feeRate,
 		loadOnly,
 	)
 	if err != nil {
@@ -470,7 +486,7 @@ func (b *batch) publishPresigned(ctx context.Context) (btcutil.Amount, error,
 
 	// Run sanity checks to make sure presignedHelper.SignTx complied with
 	// all the invariants.
-	err = CheckSignedTx(tx, signedTx, batchAmt, minRelayFee)
+	err = CheckSignedTx(tx, signedTx, batchAmt, minRelayFeeRate)
 	if err != nil {
 		return 0, fmt.Errorf("signed tx doesn't correspond the "+
 			"unsigned tx: %w", err), false
--- a/sweepbatcher/presigned_test.go
+++ b/sweepbatcher/presigned_test.go
@@ -15,6 +15,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	minRelayFeeRate = chainfee.FeePerKwFloor
+)
+
 // TestOrderedSweeps checks that methods batch.getOrderedSweeps and
 // batch.getSweepsGroups works properly.
 func TestOrderedSweeps(t *testing.T) {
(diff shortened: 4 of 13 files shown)


sweepbatcher/presigned.go
```diff
@@ -557,6 +579,46 @@ func getPresignedSweepsDestAddr(ctx context.Context, helper destPkScripter,
...
+// getChangeOutputs retrieves the change output references of each sweep and
+// de-duplicates them. The function must be used in presigned mode only.
+func getChangeOutputs(sweeps []sweep, chainParams *chaincfg.Params) (
+	map[*wire.TxOut]btcutil.Address, error) {
+
+	changeOutputs := make(map[*wire.TxOut]btcutil.Address)
+	for _, sweep := range sweeps {
+		// If the sweep has a change output, add it to the changeOutputs
+		// map to avoid duplicates.
+		if sweep.change != nil {
```
hieblmi: flatten depth by `if == nil { continue }`

sweepbatcher/presigned.go
```diff
@@ -593,23 +655,23 @@ func CheckSignedTx(unsignedTx, signedTx *wire.MsgTx, inputAmt btcutil.Amount,
...
-		return fmt.Errorf("unsigned tx has %d outputs, want 1",
-			len(unsignedTx.TxOut))
-	}
-	if len(signedTx.TxOut) != 1 {
-		return fmt.Errorf("the signed tx has %d outputs, want 1",
+	if len(unsignedTx.TxOut) != len(signedTx.TxOut) {
+		return fmt.Errorf("unsigned tx has %d outputs, signed tx has "+
+			"%d outputs, should be equal", len(unsignedTx.TxOut),
 			len(signedTx.TxOut))
 	}
```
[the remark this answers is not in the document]
hieblmi: is the fee calculation now correct after subtracting the sum of change outputs?

starius:
> Multiple sweeps holding the same change reference will produce a single change output in the batch transaction.

This seems to be outdated.

sweepbatcher/greedy_batch_selection.go
```diff
@@ -210,6 +210,13 @@ func estimateBatchWeight(batch *batch) (feeDetails, error) {
 			err)
 	}
 
+	// Add change output weights.
+	for _, s := range batch.sweeps {
+		if s.change != nil {
+			weight.AddP2TROutput()
+		}
+	}
+
```
starius, line 229: I propose to add test cases for change outputs into unit tests in file `greedy_batch_selection_test.go`

sweepbatcher/greedy_batch_selection.go
```diff
@@ -210,6 +210,13 @@ func estimateBatchWeight(batch *batch) (feeDetails, error) {
 			err)
 	}
 
+	// Add change output weights.
+	for _, s := range batch.sweeps {
+		if s.change != nil {
+			weight.AddP2TROutput()
```
starius:
```go
weight.AddOutput(s.change.PkScript)
```

Just in case we allow non-taproot change outputs in the future.

sweepbatcher/presigned.go
```diff
@@ -593,23 +596,30 @@ func CheckSignedTx(unsignedTx, signedTx *wire.MsgTx, inputAmt btcutil.Amount,
...
-	unsignedOut := unsignedTx.TxOut[0]
-	signedOut := signedTx.TxOut[0]
-	if !bytes.Equal(unsignedOut.PkScript, signedOut.PkScript) {
-		return fmt.Errorf("mismatch of output pkScript: %x, %x",
-			unsignedOut.PkScript, signedOut.PkScript)
+	for i, o := range unsignedTx.TxOut {
+		if !bytes.Equal(o.PkScript, signedTx.TxOut[i].PkScript) {
+			return fmt.Errorf("mismatch of output pkScript: %x, %x",
+				o.PkScript, signedTx.TxOut[i].PkScript)
+		}
```
starius, line 625: Compare satoshi values of changes.
```go
if i != 0 && o.Value != signedTx.TxOut[i].Value {
  error
```

sweepbatcher/presigned.go
```diff
@@ -593,23 +596,30 @@ func CheckSignedTx(unsignedTx, signedTx *wire.MsgTx, inputAmt btcutil.Amount,
...
+	}
+
+	// The first output is always the batch output.
+	batchOutput := signedTx.TxOut[0]
+
+	// Calculate the total value of change outputs to help determine the
+	// transaction fee.
+	totalChangeValue := btcutil.Amount(0)
+	for i := 1; i < len(signedTx.TxOut); i++ {
+		totalChangeValue += btcutil.Amount(signedTx.TxOut[i].Value)
```
starius: We are not interested in `batchOutput` itself, it is needed only to calculate the fee. This code can be simplified: just decrease inputAmt by the sum of all outputs and it will be the fee.

sweepbatcher/presigned_test.go
```diff
@@ -1517,7 +1518,8 @@ func TestCheckSignedTx(t *testing.T) {
 			},
 			inputAmt:    3_000_000,
 			minRelayFee: 253,
-			wantErr:     "the signed tx has 2 outputs, want 1",
+			wantErr: "unsigned tx has 1 outputs, signed tx " +
+				"has 2 outputs, should be equal",
```
starius, line 1606: Add error cases where pkscripts and amouts are wrong, please.

sweepbatcher/sweep_batch.go
```diff
@@ -1297,6 +1310,11 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
 			"failed: %w", err)
 	}
 
+	// Add the optional change outputs to weight estimates.
+	for _, _ = range changeOutputs {
+		weightEstimate.AddP2TROutput()
```
starius:
```go
weight.AddOutput(s.change.PkScript)
```

sweepbatcher/sweep_batch.go
```diff
@@ -1318,11 +1336,50 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
 	fee := clampBatchFee(feeForWeight, batchAmt)
 
 	// Add the batch transaction output, which excludes the fees paid to
-	// miners.
-	batchTx.AddTxOut(&wire.TxOut{
-		PkScript: batchPkScript,
-		Value:    int64(batchAmt - fee),
-	})
+	// miners. Reduce the amount by the sum of change outputs, if any.
+	if len(changeOutputs) == 0 {
```
starius: We don't need this special case, IMHO.
If there is no change, sumChange is 0 and the `else` branch works well.

sweepbatcher/sweep_batch.go
```diff
@@ -1318,11 +1336,50 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
...
+		// Then add change outputs.
+		for _, txOut := range changeOutputs {
+			batchTx.AddTxOut(&wire.TxOut{
+				PkScript: txOut.PkScript,
+				Value:    txOut.Value,
+			})
+		}
+	}
+
+	// Check that for each swap, inputs exceed the change outputs.
```
starius, line 1463: This check is only needed when there is change.
```go
if len(changeOutputs) != 0 {
```
We can save a couple of `map` allocations with this `if`.

sweepbatcher/sweep_batch.go
```diff
@@ -1396,7 +1453,9 @@ func (b *batch) publishMixedBatch(ctx context.Context) (btcutil.Amount, error,
 			attempt)
 
 		// Construct unsigned batch transaction.
-		var err error
+		var (
+			err error
+		)
```
starius: Why is this change needed?

sweepbatcher/sweep_batch_test.go
```diff
@@ -275,6 +288,97 @@ func TestConstructUnsignedTx(t *testing.T) {
 			wantFee:          789,
 		},
 
+		{
+			name: "all sweeps same change output",
```
starius: Why "same"? There is just a single sweep.

sweepbatcher/sweep_batch_test.go
```diff
@@ -40,6 +44,15 @@ func TestConstructUnsignedTx(t *testing.T) {
 	p2trPkScript, err := txscript.PayToAddrScript(p2trAddress)
 	require.NoError(t, err)
 
+	change1 := &wire.TxOut{
+		Value:    100_000,
+		PkScript: p2trPkScript,
```
starius: Please use another pkscript here, because `p2trPkScript` is used as the main output by the tests, so they won't be distinguished by the test if they are confused (e.g. the order is broken).

sweepbatcher/sweep_batch_test.go
```diff
@@ -40,6 +44,15 @@ func TestConstructUnsignedTx(t *testing.T) {
...
+	}
+	change2 := &wire.TxOut{
+		Value:    200_000,
+		PkScript: p2trPkScript,
```
starius: And here please use another pkscript as well so we can distinguish `change1` from `change2`.

sweepbatcher/sweep_batch_test.go
```diff
@@ -275,6 +288,97 @@ func TestConstructUnsignedTx(t *testing.T) {
...
+					{
+						Value:    change2.Value,
+						PkScript: change2.PkScript,
+					},
+				},
+			},
+			wantWeight:       1248,
+			wantFeeForWeight: 1248,
+			wantFee:          1248,
+		},
```
starius, line 407: I propose to add test cases:
 - error: change output exceeds inputs of its swap
 - error: main output is dust (change outputs too large).
 - a case where `clampBatchFee` kicks in with change outputs. E.g. total amount is 1 BTC, change outputs are 0.9 BTC, our output is 0.1 BTC. In this case we want to constrain the fee by 0.02 BTC, not but 0.2 BTC.

sweepbatcher/sweep_batch.go
```diff
@@ -1236,10 +1239,13 @@ func (b *batch) createPsbt(unsignedTx *wire.MsgTx, sweeps []sweep) ([]byte,
 }
 
 // constructUnsignedTx creates unsigned tx from the sweeps, paying to the addr.
-// It also returns absolute fee (from weight and clamped).
+// It also returns absolute fee (from weight and clamped). The main output is
+// the first output of the transaction, followed by an optional list of change
+// outputs.
```
starius: I propose to add the following check (and a comment to godoc, and a test case): if main output is dust or even negative, the function fails with an error. Previously function `clampBatchFee` partly protected from this (though I think was is still possible to get a dust output if swap amout < 1.2 * dust).

I propose to add explicit checks and tests for this. Also we should be careful with `-` operation since it may render negative numbers into very high positive numbers in case the type is unsigned. We should compare first, then apply `-`.

sweepbatcher/sweep_batch.go
```diff
@@ -1318,11 +1336,50 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
...
+	} else {
+		// Reduce the batch output by the sum of change outputs.
+		var sumChange int64
+		for _, change := range changeOutputs {
+			sumChange += change.Value
+		}
+		// Add the main output first.
+		batchTx.AddTxOut(&wire.TxOut{
+			PkScript: batchPkScript,
+			Value:    int64(batchAmt-fee) - sumChange,
```
starius: Here I propose to add the following check:
```go
if fee + sumChange + lnwallet.DustLimitForSize(len(batchPkScript)) < batchAmt {
  error
}
```

sweepbatcher/sweep_batch.go
```diff
@@ -1318,11 +1336,50 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
 	fee := clampBatchFee(feeForWeight, batchAmt)
```
starius: I think we want to pass here `batchAmt - sumChange`. We should also check that `batchAmt > sumChange` first.

The idea of `clampBatchFee` is to cap by 20% the fee that we pay for sweeping. Since change outputs are not swept to us, but added on top of our cut, they should not be taken into account when calculating this upper bound. Otherwise we may actually pay more than 20% if change outputs constitutes a significant portion of the amount. E.g. total amount is 1 BTC, change outputs are 0.9 BTC, our output is 0.1 BTC. In this case we want to constrain the fee by 0.02 BTC, not but 0.2 BTC.

sweepbatcher/sweep_batcher.go
```diff
@@ -713,7 +720,8 @@ func (b *Batcher) Run(ctx context.Context) error {
 // swap. The order of sweeps is important. The first sweep serves as
 // primarySweepID if the group starts a new batch.
 func (b *Batcher) PresignSweepsGroup(ctx context.Context, inputs []Input,
-	sweepTimeout int32, destAddress btcutil.Address) error {
+	sweepTimeout int32, destAddress btcutil.Address,
+	changeOutput *wire.TxOut) error {
```
starius, line 725: Add to `godoc` that `changeOutput` may be nil.

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1215,9 +1223,9 @@ func testPresigned_presigned_and_regular_sweeps(t *testing.T, store testStore,
 
 	setFeeRate(feeRateLow)
 
-	/////////////////////////////////////
+	// ///////////////////////////////////
```
starius: This breaks pseudographics :-)

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1584,11 +1592,11 @@ func testPresigned_purging(t *testing.T, numSwaps, numConfirmedSwaps int,
 
 		// An attempt to presign must succeed.
 		err := batcher.PresignSweepsGroup(
-			ctx, group, sweepTimeout, destAddr,
+			ctx, group, sweepTimeout, destAddr, nil,
 		)
 		require.NoError(t, err)
 
-		// Add the sweep, triggering the publish attempt.
+		// Add the sweep, triggering the publishing attempt.
```
starius, line 2140: I propose to add a test case of change outputs to TestPresigned. It can be done as a copy of the basic test case, removing all insignificant code from it and adding change output and verifying that it is present in the broadcasted transaction and the pkscript and value are as expected.

sweepbatcher/greedy_batch_selection.go
hieblmi: fixed

sweepbatcher/presigned.go
hieblmi: fixed

sweepbatcher/presigned.go
hieblmi: good simplification, fixed.

sweepbatcher/presigned_test.go
hieblmi: added

sweepbatcher/sweep_batch.go
hieblmi: fixed

sweepbatcher/sweep_batch.go
hieblmi: yes this is simpler, fixed

sweepbatcher/sweep_batch.go
hieblmi: added

sweepbatcher/sweep_batch.go
hieblmi: This is a remainder of a previous change, removed.

sweepbatcher/sweep_batch_test.go
hieblmi: a left over from the previous version, fixed

sweepbatcher/sweep_batch_test.go
hieblmi: fixed

sweepbatcher/sweep_batch_test.go
hieblmi: fixed

sweepbatcher/sweep_batch_test.go
hieblmi: Thank you for these improvements. I've added the testcases. I've also added an additional dustlimit check for all outputs of a batch tx, see:
```
// Ensure that each output is above dust limit.
	for _, txOut := range batchTx.TxOut {
		// copied from lnwallet.DustLimitForSize
		txout := &wire.TxOut{PkScript: txOut.PkScript}
		dustlimit := btcutil.Amount(mempool.GetDustThreshold(txout))
		if txOut.Value < int64(dustlimit) {
			return nil, 0, 0, 0, fmt.Errorf("output %v is below "+
				"dust limit %v", btcutil.Amount(txOut.Value),
				dustlimit)
		}
	}
```

sweepbatcher/sweep_batch.go
hieblmi: See https://github.com/lightninglabs/loop/pull/976#discussion_r2207450552

sweepbatcher/sweep_batch_test.go
hieblmi: I realized that we shouldn't do that, otherwise swap change can grief batch creation.

sweepbatcher/sweep_batch.go
hieblmi: This makes total sense, thank you for the catch. I added a unit-test `clamp fee to max fee to swap amount ration` to test this. I've added also a check for `batchAmt > sumChange`.

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: I fixed my ide setting to remove those.

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: I added this test case which revealed a bug in `PresignSweepsGroup` where we added the change output to every sweep of the group, but should only add it to the primary sweep of a group.

sweepbatcher/sweep_batch.go
```diff
@@ -1314,15 +1335,80 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
...
-	fee := clampBatchFee(feeForWeight, batchAmt)
+	fee := clampBatchFee(feeForWeight, batchAmt-btcutil.Amount(sumChange))
 
-	// Add the batch transaction output, which excludes the fees paid to
-	// miners.
+	// Ensure that batch amount exceeds the sum of change outputs and the
+	// fee, and that it is also greater than dust limit for the main
+	// output.
+	if fee+btcutil.Amount(sumChange)+
+		lnwallet.DustLimitForSize(len(batchPkScript)) > batchAmt {
```
starius: `lnwallet.DustLimitForSize` calls `mempool.GetDustThreshold` under the hood (which we also use below).
Let's use `mempool.GetDustThreshold` in all places.
We can make a wrapper function like this to avoid copy-paste:
```go
func dustLimitForPkScript(pkscript []byte) btcutil.Amount
```
It would call `mempool.GetDustThreshold` under the hood. It only depends on pkscript, not on the value of TxOut.

And we will always use it in SB, e.g. in `checkSweepToAdd` to guarantee uniformity.

sweepbatcher/sweep_batch_test.go
```diff
@@ -2,6 +2,7 @@ package sweepbatcher
 
 import (
 	"fmt"
+	"github.com/lightningnetwork/lnd/lnwallet"
```
starius: Move this import to the second group.

sweepbatcher/sweep_batch_test.go
```diff
@@ -275,6 +304,207 @@ func TestConstructUnsignedTx(t *testing.T) {
...
+					value:    btcutil.Amount(change1.Value),
+					change:   change1,
+				},
+			},
+			address:       p2trAddress,
+			currentHeight: 800_000,
+			feeRate:       1,
+			wantErr: "output 0.00000329 BTC is below dust " +
+				"limit 0.00000330 BTC",
+		},
```
starius, line 582: "main output is dust" test-case has two exact copies.

sweepbatcher/sweep_batch_test.go
```diff
@@ -275,6 +304,207 @@ func TestConstructUnsignedTx(t *testing.T) {
...
+
+		{
+			name: "clamp fee to max fee to swap amount ration",
```
starius: s/ration/ratio/ ?

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1161,111 @@ func testPresigned_presigned_group(t *testing.T,
 	require.Equal(t, batchPkScript, tx.TxOut[0].PkScript)
 }
 
+// testPresigned_presigned_group_with_change tests passing multiple sweeps to
+// the method PresignSweepsGroup.
+func testPresigned_presigned_group_with_change(t *testing.T,
+	batcherStore testBatcherStore) {
+
+	//defer test.Guard(t)()
```
starius: Should `defer test.Guard(t)()` be uncommented?

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -59,6 +63,16 @@ func (h *mockPresignedHelper) SetOutpointOnline(op wire.OutPoint, online bool) {
 	h.onlineOutpoints[op] = online
 }
 
+// SetOutpointOnline changes the online status of an outpoint.
+func (h *mockPresignedHelper) SetChangeForPrimaryDeposit(op wire.OutPoint,
```
starius: I propose to make this method private so it is clear that it is not a part of public interface but used for testing purposes.

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -59,6 +63,16 @@ func (h *mockPresignedHelper) SetOutpointOnline(op wire.OutPoint, online bool) {
 	h.onlineOutpoints[op] = online
 }
 
+// SetOutpointOnline changes the online status of an outpoint.
```
starius: godoc is outdated

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -211,6 +225,9 @@ func (h *mockPresignedHelper) FetchSweep(_ context.Context,
 	// Find IsPresigned.
 	_, isPresigned := h.onlineOutpoints[utxo]
 
+	// Find change.
+	change, _ := h.changeOutputs[utxo]
```
starius:
```go
change := h.changeOutputs[utxo]
```

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1161,111 @@ func testPresigned_presigned_group(t *testing.T,
...
+	require.Len(t, tx.TxOut, 2)
+	require.ElementsMatch(
+		t, []wire.OutPoint{op1, op2},
+		[]wire.OutPoint{
+			tx.TxIn[0].PreviousOutPoint,
+			tx.TxIn[1].PreviousOutPoint,
+		},
+	)
+	require.Equal(t, int64(2993740), tx.TxOut[0].Value)
+	require.Equal(t, int64(500_000), tx.TxOut[1].Value)
```
starius: `s/int64(500_000)/change.Value/`

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1161,111 @@ func testPresigned_presigned_group(t *testing.T,
...
+	require.Equal(t, batchPkScript, tx.TxOut[0].PkScript)
+	require.Equal(t, []byte{0xaf, 0xfe}, tx.TxOut[1].PkScript)
```
starius: `s/[]byte{0xaf, 0xfe}/change.PkScript/`

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1161,111 @@ func testPresigned_presigned_group(t *testing.T,
 	require.Equal(t, batchPkScript, tx.TxOut[0].PkScript)
 }
 
+// testPresigned_presigned_group_with_change tests passing multiple sweeps to
+// the method PresignSweepsGroup.
```
starius: Mention adding a change in the godoc.

sweepbatcher/sweep_batcher_presigned_test.go
starius: I propose to undo this change (restore the pseudographics).

sweepbatcher/sweep_batch.go
hieblmi: I created function `dustLimitForPkScript` which we call in `checkSweepToAdd` to check dust size on optional change outputs of sweeps, and in `constructUnsignedTx`.

sweepbatcher/sweep_batch_test.go
hieblmi: Added a check to `checkSweepToAdd ` to prevent dust change from entering a batch, so the proposed check here can still be applied.

sweepbatcher/sweep_batch_test.go
hieblmi: fixed

sweepbatcher/sweep_batch_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/greedy_batch_selection.go
hieblmi: I added test case `one sweep regular batch with change`.

sweepbatcher/sweep_batch.go
```diff
@@ -2495,3 +2588,10 @@ func clampBatchFee(fee btcutil.Amount,
 
 	return fee
 }
+
+// dustLimitForPkScript returns the dust limit for a given pkScript.
+func dustLimitForPkScript(pkscript []byte) btcutil.Amount {
```
starius: I propose to add a small unit test for a couple of address types (e.g. p2wpkh and p2tr).

sweepbatcher/sweep_batch.go
```diff
@@ -495,6 +499,16 @@ func (b *batch) checkSweepToAdd(_ context.Context, sweep *sweep) (bool, error) {
...
+	// If the sweep has a change output it must carry more than dust value.
+	if sweep.change != nil &&
+		sweep.value < dustLimitForPkScript(sweep.change.PkScript) {
+
+		return false, fmt.Errorf("sweep %x has a change output "+
+			"with value %d below dust limit %d",
+			sweep.swapHash[:6], sweep.change.Value,
+			dustLimitForPkScript(sweep.change.PkScript))
+	}
+
```
starius: Actually, this seems to be unneeded. We have this check in `constructUnsignedTx` which is called from `presign` which is called by `addSweeps`. So if the change is dust, it won't enter anyway.

I propose to remove this check and to add a test similar to `testPresigned_presigned_group_with_change` testing that trying to add too large or too small change (making one of the outputs dust) fails in both `PresignSweepsGroup` and subsequent `AddSweep`. Maybe we can even check it in the existing test as failed attempts before the succeeded attempt. Or just duplicate the test.

starius (approved):
**LGTM!** 🥇

Added a couple of proposals.

sweepbatcher/greedy_batch_selection.go
starius: Please also add a couple of test cases to `TestSelectBatches` to demonstrate that a change can affect batch selection (one test, another - control). The idea is that a change increases weight. Create one existing batch with 1 sweep without a change with a slightly higher fee rate. The sweep being added has lower feerate and it has a change. Without a change it would be prefered to add it to the existing batch. With a change it is better to create new batch, because of the change size. You can adjust feerates to achieve such decision change.

sweepbatcher/sweep_batch.go
```diff
@@ -2495,3 +2588,10 @@ func clampBatchFee(fee btcutil.Amount,
 
 	return fee
 }
+
+// dustLimitForPkScript returns the dust limit for a given pkScript.
```
starius: Please add to godoc: "An output must be greater or equal to this value."

starius: We can also move this function to loop/utils and make it public.

hieblmi: done

hieblmi: done

sweepbatcher/sweep_batch.go
hieblmi: Added `TestDustLimitForPkScript`.

sweepbatcher/sweep_batch.go
hieblmi: added a separate test `testPresigned_presigned_group_with_dust_change`.

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1163,195 @@ func testPresigned_presigned_group(t *testing.T,
...
+
+	// Enable only one of the sweeps.
+	presignedHelper.SetOutpointOnline(op1, true)
+	presignedHelper.SetOutpointOnline(op2, true)
+
+	// An attempt to presign must fail.
+	err := batcher.PresignSweepsGroup(
+		ctx, group1, sweepTimeout, destAddr, change,
+	)
+	require.Error(t, err, "failed to construct unsigned tx for feeRate "+
```
starius: s/Error/EqualError/

`require.Error` only checks that `err != nil`, it doesn't check the text of the error.

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1799,6 +2016,14 @@ func TestPresigned(t *testing.T) {
 		testPresigned_presigned_group(t, NewStoreMock())
 	})
 
+	t.Run("presigned_group", func(t *testing.T) {
```
starius: s/presigned_group/change/

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1799,6 +2016,14 @@ func TestPresigned(t *testing.T) {
...
+		testPresigned_presigned_group_with_change(t, NewStoreMock())
+	})
+
+	t.Run("presigned_group", func(t *testing.T) {
```
starius: s/presigned_group/dust_change/

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1163,195 @@ func testPresigned_presigned_group(t *testing.T,
...
+		"0.00000477 BTC")
+
+	// Add the sweep, triggering the publishing attempt.
+	err = batcher.AddSweep(ctx, &SweepRequest{
+		SwapHash: swapHash1,
+		Inputs:   group1,
+		Notifier: &dummyNotifier,
+	})
+	require.ErrorContains(t, err, "were not presigned")
+}
```
starius, line 1685: Can you add one more test, in which the change is almost as large as the input and the main output is dust, please?

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: fixed

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: added.

sweepbatcher/greedy_batch_selection.go
hieblmi: I added the described testcases.

sweepbatcher/greedy_batch_selection_test.go
```diff
@@ -778,6 +813,47 @@ func TestSelectBatches(t *testing.T) {
...
+			batches: []feeDetails{
+				{
+					BatchId: 1,
+					FeeRate: mediumFeeRate,
+					Weight:  coopNewBatchWeight,
+				},
+			},
+			sweep: feeDetails{
+				FeeRate: highFeeRate,
+				Weight:  coopInputWeight + changeOutputWeight,
```
starius: This test case passes even if `changeOutputWeight` is not added. Also its name is confusing, because the sweep is not placed in "existing medium batch", but it also added to a new batch.

I propose to replace this test case which another one which works as control for "low fee change sweep, placed in new batch":

```diff
diff --git a/sweepbatcher/greedy_batch_selection_test.go b/sweepbatcher/greedy_batch_selection_test.go
index 33b0d5a..0a819f9 100644
--- a/sweepbatcher/greedy_batch_selection_test.go
+++ b/sweepbatcher/greedy_batch_selection_test.go
@@ -835,8 +835,8 @@ func TestSelectBatches(t *testing.T) {
                },
 
                {
-                       name: "high fee change sweep, placed in existing " +
-                               "medium batch",
+                       name: "low fee sweep without change, placed in "+
+                               "existing batch",
                        batches: []feeDetails{
                                {
                                        BatchId: 1,
@@ -845,14 +845,14 @@ func TestSelectBatches(t *testing.T) {
                                },
                        },
                        sweep: feeDetails{
-                               FeeRate: highFeeRate,
-                               Weight:  coopInputWeight + changeOutputWeight,
+                               FeeRate: lowFeeRate,
+                               Weight:  coopInputWeight,
                        },
                        oneSweepBatch: feeDetails{
-                               FeeRate: highFeeRate,
+                               FeeRate: lowFeeRate,
                                Weight:  coopNewBatchWeight,
                        },
-                       wantBestBatchesIds: []int32{newBatchSignal, 1},
+                       wantBestBatchesIds: []int32{1, newBatchSignal},
                },
        }
 
```

sweepbatcher/sweep_batch.go
```diff
@@ -1314,15 +1334,78 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
...
+	// Ensure that the batch amount is greater than the sum of change.
+	if batchAmt <= btcutil.Amount(sumChange) {
+		return nil, 0, 0, 0, fmt.Errorf("batch amount %v is <= the "+
+			"sum of change outputs %v", batchAmt,
+			btcutil.Amount(sumChange))
+	}
+
 	// Clamp the calculated fee to the max allowed fee amount for the batch.
-	fee := clampBatchFee(feeForWeight, batchAmt)
+	fee := clampBatchFee(feeForWeight, batchAmt-btcutil.Amount(sumChange))
```
starius: In test `TestPresigned/dust_main_output` I saw:

> fee 0.00000065 BTC

Such a fee is too low (below relay fee).

I think we should fail here if the fee is below relay fee rate. The function should accept min relay fee rate as an argument, we already have the value in many places, where the function is called.

Feerate can be found with `chainfee.NewSatPerKWeight(fee, weight)`.

Please add a unit test for `constructUnsignedTx` for this error and adjust test `TestPresigned/dust_main_output` not to crash on this error. Additionally, you can add another test case for TestPresigned to reproduce this new error (fee below relay fee).

sweepbatcher/sweep_batch_test.go
```diff
@@ -70,6 +97,8 @@ func TestConstructUnsignedTx(t *testing.T) {
 		return fmt.Errorf("weight estimator test failure")
 	}
 
+	dustLimit := lnwallet.DustLimitForSize(input.P2TRSize)
```
starius: `destAddr` and `batchPkScript` used as batch destination address in most cases are a p2wkh address, not taproot.

I propose to use our new function `utils.DustLimitForPkScript(batchPkScript)` here instead of `lnwallet.DustLimitForSize(input.P2TRSize)` to get the same values. The numeric values are different. Taproot is 330, p2wkh is 294.

I discovered this by changing change amounts by 2 satoshis in the test cases:

```diff
diff --git a/sweepbatcher/sweep_batcher_presigned_test.go b/sweepbatcher/sweep_batcher_presigned_test.go
index fd28bc7..2ad90ab 100644
--- a/sweepbatcher/sweep_batcher_presigned_test.go
+++ b/sweepbatcher/sweep_batcher_presigned_test.go
@@ -1318,7 +1318,7 @@ func testPresigned_presigned_group_with_dust_main_output(t *testing.T,
        }
        dustLimit := int64(lnwallet.DustLimitForSize(input.P2TRSize))
        change := &wire.TxOut{
-               Value:    inputValue - dustLimit + 1,
+               Value:    inputValue - dustLimit - 1,
                PkScript: []byte{0xaf, 0xfe},
        }
 
@@ -1401,7 +1401,7 @@ func testPresigned_presigned_group_with_dust_change(t *testing.T,
        }
        dustLimit := lnwallet.DustLimitForSize(input.P2TRSize)
        change := &wire.TxOut{
-               Value:    int64(dustLimit - 1),
+               Value:    int64(dustLimit + 1),
                PkScript: []byte{0xaf, 0xfe},
        }
 
```

My expectation was that inputs would be added successfully (and the tests would fail, as they test failures). However it was not the case, because these `dustLimit` variables used in `sweep_batch_test.go` are higher than the actual limit.

Please use `utils.DustLimitForPkScript(batchPkScript)` here below in this file.

sweepbatcher/greedy_batch_selection_test.go
hieblmi: replaced

sweepbatcher/sweep_batch.go
hieblmi: I've threaded the min relay fee through to `clampBatchFee` which now checks it the tx fee is below the min relay fee, and errors if so.

I've also added a test case for that scenario and fixed `TestPresigned/dust_main_output` by lowering lnd's min relay fee in that test.

sweepbatcher/sweep_batch_test.go
hieblmi: done

sweepbatcher/presigned.go
```diff
@@ -300,7 +315,8 @@ type presigner interface {
 // 10x of the current next block feerate.
 func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
 	primarySweepID wire.OutPoint, sweeps []sweep,
-	nextBlockFeeRate chainfee.SatPerKWeight) error {
+	nextBlockFeeRate chainfee.SatPerKWeight,
+	minRelayFeeRate chainfee.SatPerKWeight) error {
```
starius, line 319: Below in the function there are variables `start` and `minRelayFee`. I think we should use `minRelayFeeRate` as a value for them. Currently they just default to `chainfee.FeePerKwFloor` and `chainfee.AbsoluteFeePerKwFloor`.

If `minRelayFeeRate` is higher, then fewer transactions will be presigned (since `start` will be higher). We can verify this with a unit test as well.

starius (approved):
**LGTM!** 💾

I sent PR https://github.com/hieblmi/loop/pull/1 with some changes that I propose, mostly for tests, but also most importantly to use minRelayFeeRate in `presign` and `ensurePresigned` as start value for presigning and when passing to SignTx. Please review the changes.

sweepbatcher/presigned.go
hieblmi: Thanks for the adjustments, I kept your commit as is.

sweepbatcher/greedy_batch_selection.go
```diff
@@ -210,6 +210,13 @@ func estimateBatchWeight(batch *batch) (feeDetails, error) {
 			err)
 	}
 
+	// Add change output weights.
+	for _, s := range batch.sweeps {
+		if s.change != nil {
```
sputn1ck: Would it make sense to add a check for if a change pkscript repeats that we add to the other outputs value instead of adding a new output with the same pkscript? Is that something that can happen (e.g. on static loop in if change is the static addr change would be the same?)

sweepbatcher/sweep_batch.go
```diff
@@ -1255,6 +1260,13 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
 		LockTime: uint32(currentHeight),
 	}
 
+	var changeOutputs []*wire.TxOut
+	for _, sweep := range sweeps {
```
sputn1ck: I think here a similiar comment to my other comment, what happens if the pkscript is the same between multiple sweeps?

sweepbatcher/sweep_batch_test.go
```diff
@@ -275,6 +304,208 @@ func TestConstructUnsignedTx(t *testing.T) {
...
+					},
+				},
+			},
+			wantWeight:       616,
+			wantFeeForWeight: 616,
+			wantFee:          616,
+		},
+
+		{
+			name: "all sweeps different change outputs",
```
sputn1ck, line 355: I think a testcase "all sweeps same change output" makes sense and the result should be a single change output right?

sweepbatcher/greedy_batch_selection.go
hieblmi: Very good suggestion!

sweepbatcher/sweep_batch.go
hieblmi: It totally makes sense to consolidate values of change outputs with identical pkscript. I've changed the workings to
do so here and in the greedy batch selection. I will also add unit tests for this.

sweepbatcher/sweep_batch_test.go
hieblmi: added `"all sweeps identical change pkscripts"`

hieblmi:
I've added commit `sweepbatcher: consolidate identical change pkscripts` to tally up the values of identical change pkscripts.

sweepbatcher/greedy_batch_selection.go
```diff
@@ -210,11 +210,21 @@ func estimateBatchWeight(batch *batch) (feeDetails, error) {
...
+			continue
 		}
+
+		pkScriptString := string(s.change.PkScript)
+		if _, has := changeOutputs[pkScriptString]; has {
+			continue
+		}
+
+		weight.AddOutput(s.change.PkScript)
+		changeOutputs[pkScriptString] = struct{}{}
```
starius, line 227: :+1: 

Please add a test case to `TestEstimateBatchWeight`.

sweepbatcher/sweep_batch_test.go
```diff
@@ -395,6 +395,57 @@ func TestConstructUnsignedTx(t *testing.T) {
 			wantFee:          1248,
 		},
 
+		{
+			name: "all sweeps identical change pkscripts",
```
starius: s/all sweeps identical change pkscripts/identical change pkscripts/

sweepbatcher/sweep_batch_test.go
```diff
@@ -395,6 +395,57 @@ func TestConstructUnsignedTx(t *testing.T) {
...
+					{
+						Value:    2 * change1.Value,
+						PkScript: change1.PkScript,
+					},
+				},
+			},
+			wantWeight:       1076,
+			wantFeeForWeight: 1076,
+			wantFee:          1076,
+		},
```
starius, line 458: Please add another test case with equal change pkscripts, but different satoshi values.

sweepbatcher/sweep_batcher_presigned_test.go
```diff
@@ -1135,6 +1162,400 @@ func testPresigned_presigned_group(t *testing.T,
...
+		},
+	)
+	require.Equal(t, int64(2_493_300), tx.TxOut[0].Value)
+	require.Equal(t, change.Value, tx.TxOut[1].Value)
+	require.Equal(t, batchPkScript, tx.TxOut[0].PkScript)
+	require.Equal(t, change.PkScript, tx.TxOut[1].PkScript)
+
+	// Mine a blocks to trigger republishing.
+	require.NoError(t, lnd.NotifyHeight(601))
+}
```
starius, line 1269: I propose to add a unit test with two swaps sharing the same change output pkscript.

starius (approved):
**LGTM!** 🌴
The code of last commit is great. I propose to add few tests to validate it.

sweepbatcher/greedy_batch_selection.go
hieblmi: Added `"two sweeps regular batch with identical change"`.

sweepbatcher/sweep_batch_test.go
hieblmi: changed

sweepbatcher/sweep_batch_test.go
hieblmi: added `"identical change pkscripts different values"`.

sweepbatcher/sweep_batcher_presigned_test.go
hieblmi: added `testPresigned_presigned_group_with_identical_change_pkscript`.

hieblmi:
@starius I had to sort the change outputs in `constructUnsignedTx` after they've been deduplicated in a map in order to make the order deterministic for the tests, see https://github.com/lightninglabs/loop/pull/976/commits/29161bbc554387f05715182647e4584e8af51eca#diff-355a81f8dab9aa0faaaa04740ca44f425ff0dd18dea7240e6a28e1697f222dd1R1441

sweepbatcher/sweep_batch.go
```diff
@@ -1365,15 +1398,96 @@ func constructUnsignedTx(sweeps []sweep, address btcutil.Address,
...
 	})
+	// Then add change outputs. Sort the keys first to make tests
+	// deterministic.
+	keys := make([]string, 0, len(changeOutputs))
+	for k := range changeOutputs {
+		keys = append(keys, k)
+	}
+
+	// Sort the keys
+	sort.Strings(keys)
```
starius: Let's use [BIP-0069](https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki) for outputs sorting (change only, so main output is 0).
We have it implemented for inputs in `utils/get_prevout_info.go`

hieblmi: I've added a bip69 compatible utils method to sort tx outputs. I've added it in a separate commit 245a6661.

starius (approved):
**LGTM!** 🏆

PR #978, opened 2025-07-12
make: set CGO_ENABLED=0

starius:
Make `loop` and `loopd` binaries static, more portable.

Binaries built with dynamic dependencies can't be freely moved across machines with different distributions (e.g. NixOS -> Debian breaks because it can't find some libraries).

#### Pull Request Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,8 @@ GOIMPORTS_PKG := github.com/rinchsan/gosimports/cmd/gosimports
 GO_BIN := ${GOPATH}/bin
 GOIMPORTS_BIN := $(GO_BIN)/gosimports
 
-GOBUILD := GO111MODULE=on go build -v
-GOINSTALL := GO111MODULE=on go install -v
+GOBUILD := CGO_ENABLED=0 GO111MODULE=on go build -v
+GOINSTALL := CGO_ENABLED=0 GO111MODULE=on go install -v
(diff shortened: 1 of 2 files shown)


litbot-9000:

Makefile
```diff
@@ -11,8 +11,8 @@ GOIMPORTS_PKG := github.com/rinchsan/gosimports/cmd/gosimports
 GO_BIN := ${GOPATH}/bin
 GOIMPORTS_BIN := $(GO_BIN)/gosimports
 
-GOBUILD := GO111MODULE=on go build -v
-GOINSTALL := GO111MODULE=on go install -v
+GOBUILD := CGO_ENABLED=0 GO111MODULE=on go build -v
+GOINSTALL := CGO_ENABLED=0 GO111MODULE=on go install -v
```
hieblmi, line 15: I read that certain packages might silently disable functionality without c bindings, not sure if that's a concern for us?

starius: CGO is already disabled in itests, so if some functionality is missing, we would notice this.

bhandras: I think the only package that may cause issues is our sqlite library, but fortunately modernc sqlite is cgo free.

bhandras (approved):
LGTM 🎉

starius:
Rebased

PR #990, opened 2025-08-12
loopdb: add deposit indices

hieblmi:
Add db indices for deposits and deposit updates.


loopdb/sqlc/migrations/000010_static_address_deposits.up.sql
```diff
@@ -45,4 +45,6 @@ CREATE TABLE IF NOT EXISTS deposit_updates (
 
     -- update_timestamp is the timestamp of the update.
     update_timestamp TIMESTAMP NOT NULL
-);
\ No newline at end of file
+);
+
+CREATE INDEX IF NOT EXISTS deposit_updates_id_timestamp_index ON deposit_updates(deposit_id, update_timestamp DESC);
```
starius: This code won't be executed by existing Loop clients, since it is an old migration. I propose to make new migration to add this index. And also add a corresponding code to a `.down.sql` file.

Could you add a new line, please? GitHub shows that it is missing in the end of the file.

bhandras: +1

hieblmi: apologies, fixed!

PR #993, opened 2025-08-19
gomod: update to Go 1.24.6 and `golangci-lint` v2 

hieblmi:
This PR upgrades loop's go version to `1.24.6`, upgrades the linter to `v2.4.0` and migrates the `.golangci.yml` file and fixes lint issues.

### DIFF
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -70,8 +70,7 @@ type Daemon struct {
 	// same process.
 	swapClientServer
 
-	// To be used atomically.
-	started int32
+	started atomic.Bool
 
 	// ErrChan is an error channel that users of the Daemon struct must use
 	// to detect runtime errors and also whether a shutdown is fully
@@ -128,7 +127,7 @@ func (d *Daemon) Start() error {
 	// There should be no reason to start the daemon twice. Therefore,
 	// return an error if that's tried. This is mostly to guard against
 	// Start and StartAsSubserver both being called.
-	if atomic.AddInt32(&d.started, 1) != 1 {
+	if !d.started.CompareAndSwap(false, true) {
 		return errOnlyStartOnce
 	}
 
@@ -190,7 +189,7 @@ func (d *Daemon) StartAsSubserver(lndGrpc *lndclient.GrpcLndServices,
 	// There should be no reason to start the daemon twice. Therefore,
 	// return an error if that's tried. This is mostly to guard against
 	// Start and StartAsSubserver both being called.
-	if atomic.AddInt32(&d.started, 1) != 1 {
+	if !d.started.CompareAndSwap(false, true) {
 		return errOnlyStartOnce
 	}
 


.golangci.v1.bck.yml
```diff
@@ -0,0 +1,149 @@
+run:
```
hieblmi: Should we keep the v1 file around as a backup?

bhandras: I don't think we need them, we can always just restore right?

starius: We would be able to recover it from Git if needed.

loopd/daemon.go
```diff
@@ -64,15 +64,15 @@ type ListenerCfg struct {
...
-	started int32
-
 	// swapClientServer is the embedded RPC server that satisfies the client
 	// RPC interface. We embed this struct so the Daemon itself can be
 	// registered to an existing grpc.Server to run as a subserver in the
 	// same process.
 	swapClientServer
 
+	// To be used atomically.
+	started int32
```
starius: Can we use `atomic.Bool`? (In a separate commit.)

.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ env:
 
   # If you change this value, please change it in the following files as well:
   # /Dockerfile
-  GO_VERSION: 1.24.0
+  GO_VERSION: 1.24.6
```
starius, line 23: There is a typo in the commit message: "upgarde"

tools/Dockerfile
```diff
@@ -1,16 +1,18 @@
...
 
-RUN cd /tmp \
-  && mkdir -p /tmp/build/.cache \
-  && mkdir -p /tmp/build/.modcache \
-  && cd /tmp/tools \
-  && go install -trimpath github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/
+RUN mkdir -p /tmp/build/.cache /tmp/build/.modcache \
+ && cd /tmp/tools \
+ && go install -trimpath github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.4.0 \
```
starius: `@v2.4.0` is not needed, because the version is fixed in go.mod.

.golangci.v1.bck.yml
hieblmi: deleted the file

executor.go
```diff
@@ -45,14 +45,13 @@ type executorConfig struct {
 //
 // TODO(roasbeef): rename to SubSwapper.
```
starius: I think we this TODO is outdated and can be removed.

liquidity/liquidity.go
```diff
@@ -272,7 +272,7 @@ type Manager struct {
 
 // Run periodically checks whether we should automatically dispatch a loop out.
 // We run this loop even if automated swaps are not currently enabled rather
-// than managing starting and stopping the ticker as our parameters are updated.
+// than managing to start and stop the ticker as our parameters are updated.
```
starius: We run this loop even if automated swaps are not currently enabled, instead of starting and stopping the ticker whenever our parameters are updated.

staticaddr/deposit/deposit.go
```diff
@@ -29,15 +29,17 @@ func (r *ID) FromByteSlice(b []byte) error {
 // Deposit bundles an utxo at a static address together with manager-relevant
 // data.
 type Deposit struct {
+	// Outpoint of the deposit.
+	wire.OutPoint
+
+	sync.Mutex
```
starius: I would move the mutex to the beginning. Maybe it also makes sense to make wire.OutPoint non-anonymous.

loopd/daemon.go
hieblmi: added commit f97c18da8a9e7a67b631f64874b3bcaf739b6ed5

.github/workflows/main.yml
hieblmi: fixed

tools/Dockerfile
hieblmi: fixed

executor.go
hieblmi: removed

liquidity/liquidity.go
hieblmi: better, updated

staticaddr/deposit/deposit.go
hieblmi: done

PR #1000, opened 2025-08-28
staticaddr: serialize static address creation

hieblmi:
_last 2 commits are relevant_ 

In order to remove the mutex in `address/manager.go` we serialize address creation through a channel in the Run loop.

### DIFF
--- a/staticaddr/address/manager.go
+++ b/staticaddr/address/manager.go
@@ -54,6 +54,21 @@ type Manager struct {
 	cfg *ManagerConfig
 
 	currentHeight atomic.Int32
+
+	// addrRequest is a channel used to request new static addresses from
+	// the manager. The manager employs a go worker routine that handles the
+	// requests.
+	addrRequest chan request
+}
+
+type request struct {
+	ctx      context.Context
+	respChan chan response
+}
+
+type response struct {
+	parameters *Parameters
+	err        error
 }
 
 // NewManager creates a new address manager.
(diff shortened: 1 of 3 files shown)


staticaddr/address/manager.go
```diff
@@ -54,19 +54,91 @@ type Manager struct {
...
+			case req.respChan <- response{err: err}:
+
+			case <-req.ctx.Done():
+
+			case <-managerCtx.Done():
+			}
+		}
+	}()
+
+	addr, expiry, e := m.newAddress(req.ctx)
```
starius: Why `e`, not `err`?

staticaddr/address/manager.go
```diff
@@ -54,19 +54,91 @@ type Manager struct {
...
+
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// handleAddrRequest is responsible for processing a single address request.
+func (m *Manager) handleAddrRequest(managerCtx context.Context, req request) {
+	// If processing this request panics, we want to recover to process
```
starius: Panic means a bug in the code. I think it's better to crash the process so this doesn't go unnoticed?

staticaddr/address/manager.go
hieblmi: fixed

staticaddr/address/manager.go
hieblmi: Yes I actually agree, I've removed the recovery here.

hieblmi:
Rebased

hieblmi:
closing in favor of a rework PR.

PR #1003, opened 2025-09-08
staticaddr: poll deposits on block arrival

hieblmi:
In this commit we remove the ticker-based polling
of new deposits as this is wasteful, and we replace
it by polling on the arrival of new blocks.

### DIFF
--- a/staticaddr/deposit/manager.go
+++ b/staticaddr/deposit/manager.go
@@ -19,10 +19,6 @@ import (
 )
 
 const (
-	// PollInterval is the interval in which we poll for new deposits to our
-	// static address.
-	PollInterval = 10 * time.Second
-
 	// MinConfs is the minimum number of confirmations we require for a
 	// deposit to be considered available for loop-ins, coop-spends and
 	// timeouts.
@@ -74,16 +70,16 @@ type ManagerConfig struct {
 type Manager struct {
 	cfg *ManagerConfig
 
-	// mu guards access to activeDeposits map.
+	// mu guards access to the activeDeposits map.
 	mu sync.Mutex
 
 	// activeDeposits contains all the active static address outputs.
 	activeDeposits map[wire.OutPoint]*FSM
 
-	// deposits contains all the deposits that have ever been made to the
+	// deposits contain all the deposits that have ever been made to the
 	// static address. This field is used to store and recover deposits. It
-	// also serves as basis for reconciliation of newly detected deposits by
-	// matching them against deposits in this map that were already seen.
+	// also serves as a basis for reconciliation of newly detected deposits
+	// by matching them against deposits in this map that were already seen.
 	deposits map[wire.OutPoint]*Deposit
 
 	// finalizedDepositChan is a channel that receives deposits that have
@@ -119,8 +115,13 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
 		return err
 	}
 
-	// Start the deposit notifier.
-	m.pollDeposits(ctx)
+	// Initially reconcile new deposits after a restart, so we catch up with
+	// missed deposits while we were offline.
+	if err = m.reconcileDeposits(ctx); err != nil {
+		log.Errorf("unable to reconcile deposits: %v", err)
+
+		return err
+	}
 
 	// Communicate to the caller that the address manager has completed its
 	// initialization.
@@ -147,6 +148,15 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
 				case <-ctx.Done():
 					return ctx.Err()
 				}
+			}
+
+			// Reconcile new deposits that might have just gotten
+			// confirmed.
+			if err = m.reconcileDeposits(ctx); err != nil {
+				log.Errorf("unable to reconcile deposits: %v",
+					err)
+
+				return err
 			}
 
 		case outpoint := <-m.finalizedDepositChan:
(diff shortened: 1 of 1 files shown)


staticaddr/deposit/manager.go
```diff
@@ -119,17 +119,18 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
...
 	// Communicate to the caller that the address manager has completed its
 	// initialization.
 	close(initChan)
 
 	for {
 		select {
 		case height := <-newBlockChan:
-			// Inform all active deposits about a new block arrival.
+			// Poll for new deposits and spin up fsms for them.
+			go m.pollDeposits(ctx)
```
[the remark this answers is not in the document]
hieblmi: I've added a semaphore gate to ensure that only one `pollDeposit` is running at a time, skipping any concurrent run attempt.

staticaddr/deposit/manager.go
```diff
@@ -119,17 +123,18 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
 		return err
 	}
 
-	// Start the deposit notifier.
-	m.pollDeposits(ctx)
-
 	// Communicate to the caller that the address manager has completed its
 	// initialization.
 	close(initChan)
```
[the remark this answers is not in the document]
hieblmi: fixed

staticaddr/deposit/manager.go
```diff
@@ -213,28 +222,25 @@ func (m *Manager) recoverDeposits(ctx context.Context) error {
...
-				if err != nil {
-					log.Errorf("unable to reconcile "+
-						"deposits: %v", err)
-				}
-
-			case <-ctx.Done():
-				return
-			}
+	select {
+	case m.pollingGate <- struct{}{}:
```
sputn1ck: Why not a simple mutex instead pf this chan construct? Why does this need to run in a goroutine?

staticaddr/deposit/manager.go
```diff
@@ -129,7 +134,11 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
 	for {
 		select {
 		case height := <-newBlockChan:
-			// Inform all active deposits about a new block arrival.
+			// Poll for new deposits and spin up fsms for them.
+			go m.pollDeposits(ctx)
```
bhandras: I'd say the the only blocking call here is `ListUnspent` so instead of spanning on a new goroutine, I'd suggest keeping `pollDeposits` synchronous with a timeout context to guard against `ListUnspent` blocking forever if there's any network/systemic error.

staticaddr/deposit/manager.go
hieblmi: the polling starts new fsms which could take time, so the go routine here ensures that the following notifications reach the deposits without the delay. A mutex would do it too I guess.

staticaddr/deposit/manager.go
hieblmi: the timeout ctx makes sense. Will change this.

staticaddr/deposit/manager.go
```diff
@@ -149,6 +151,16 @@
 				}
 			}
 
+			// Poll for new deposits.
+			func() {
+				timeoutCtx, cancel = context.WithTimeout(
+					ctx, PollTimeout,
+				)
```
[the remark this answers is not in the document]
hieblmi: fixed

staticaddr/deposit/manager.go
```diff
@@ -119,8 +118,11 @@
 		return err
 	}
 
-	// Start the deposit notifier.
-	m.pollDeposits(ctx)
+	// Initially poll for new deposits after a restart, so we catch up with
+	// missed deposits while we were offline.
+	timeoutCtx, cancel := context.WithTimeout(ctx, PollTimeout)
+	defer cancel()
+	m.pollDeposits(timeoutCtx)
```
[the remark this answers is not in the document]
hieblmi: fixed

staticaddr/deposit/manager.go
```diff
@@ -216,25 +225,18 @@ func (m *Manager) recoverDeposits(ctx context.Context) error {
...
-		defer ticker.Stop()
-		for {
-			select {
-			case <-ticker.C:
-				err := m.reconcileDeposits(ctx)
-				if err != nil {
-					log.Errorf("unable to reconcile "+
-						"deposits: %v", err)
-				}
+	select {
```
bhandras: I think we could even remove this select altogether (and with that `pollDeposits` and just call `m.reconcileDeposits` from `Run()`.

hieblmi: perfect, as `pollDeposits` got smaller and smaller I didn't realize we can do without it.

staticaddr/deposit/manager.go
```diff
@@ -149,6 +148,13 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
 				}
 			}
 
+			// Reconcile new deposits that might have just gotten
+			// confirmed.
+			if err = m.reconcileDeposits(ctx); err != nil {
+				log.Errorf("unable to reconcile deposits: %v",
```
sputn1ck, line 156: shouldn't we maybe actually handle/return the error here? the error cases seem that lnd is actually unresponsive or am i missing something?

hieblmi: that's right. Fixed.

PR #1009, opened 2025-09-30
staticaddr: fast-flag

hieblmi:
The fast flag enables the client to access change of a fractional amount swap faster. Since batching can delay the publishing of the on-chain swap transaction the client can pay a fee for faster access to optional change. A few examples of how this PR changes the command line arguments for fractional swaps.
```
loop static in 500000                                                 // Loop in exactly 500000 sats; auto-select deposits to cover the amount.
loop static in 500000 --fast                                         // Same as above, but request fast publication behavior (may cost more).
loop static in --all                                                 // Loop in the full balance of all eligible static-address deposits.
loop static in 500000 --all                                         // Loop in exactly 500000, select all eligible static-address deposits.
loop static in --utxo txid:vout                                     // Loop in exactly the specified deposit UTXO.
loop static in --utxo txid:vout --utxo txid:vout ...                 // Loop in the union of the listed deposit UTXOs.
loop static in 500000 --utxo txid:vout --utxo txid:vout ... --fast     // 500000 from selected deposits with fast publication behavior.
```


sweepbatcher/sweep_batcher.go
```diff
@@ -259,6 +259,10 @@ type SweepRequest struct {
 	// Notifier is a notifier that is used to notify the requester of this
 	// sweep that the sweep was successful.
 	Notifier *SpendNotifier
+
+	// Fast is set by the client if the sweep should be published
+	// immediately.
+	Fast bool
```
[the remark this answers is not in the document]
hieblmi: called by a package-external call site.

looprpc/client.proto
```diff
@@ -845,6 +845,12 @@ message QuoteRequest {
     This option only pertains to loop in swaps.
     */
     bool auto_select_deposits = 10;
+
+    /*
+    If set to true the server will immediately publish the swap in exchange for
+    a higher fee. This can be useful if the client expects change from a swap.
```
hieblmi, line 851: only available for swaps with change...

swapserverrpc/server.proto
```diff
@@ -294,6 +294,9 @@ message ServerLoopInQuoteRequest {
     // service fee. This is to cover for the increased on-chain fee the server
     // has to pay when the sweeping transaction is broadcast.
     uint32 num_static_address_deposits = 7;
+
+    // If set, request the server to use fast publication behavior.
```
hieblmi: add description

cmd/loop/staticaddr.go
```diff
@@ -506,6 +499,18 @@ func staticAddressLoopIn(ctx *cli.Context) error {
 		return cli.ShowCommandHelp(ctx, "in")
 	}
 
+	var selectedAmount int64
```
bhandras, line 509: It's sort of a breaking change. I'd suggest keeping as is?

bhandras (approved):
LGTM 🎉

cmd/loop/staticaddr.go
hieblmi: The client hasn't been released yet and this feature isn't deployed yet.

looprpc/client.proto
```diff
@@ -845,6 +845,12 @@ message QuoteRequest {
     This option only pertains to loop in swaps.
     */
     bool auto_select_deposits = 10;
+
+    /*
+    If set to true the server will immediately publish the swap in exchange for
```
starius, line 850: Add a note that this field is only allowed for static loop-ins. If it is used in a regular loop-in, `loopd` should return an error.

swapserverrpc/server.proto
starius: Add a note that this field is only allowed for static loop-ins. If it is used in a regular loop-in, loop server is expected to return an error.

loopdb/sqlc/migrations/000020_static_address_fast_flag.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Drop 'fast' flag from static_address_swaps
+ALTER TABLE static_address_swaps DROP COLUMN fast;
```
starius: Line end is missing.

loopdb/sqlc/migrations/000020_static_address_fast_flag.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add 'fast' flag to static_address_swaps, default false
+ALTER TABLE static_address_swaps ADD COLUMN fast BOOLEAN NOT NULL DEFAULT false;
```
starius: Line end is missing.

sweepbatcher/sweep_batcher.go
```diff
@@ -210,11 +210,11 @@ type FeeRateProvider func(ctx context.Context, swapHash lntypes.Hash,
 // faster. If the function returns an error, no delay is used and the error is
 // logged as a warning.
 type InitialDelayProvider func(ctx context.Context, numSweeps int,
-	value btcutil.Amount) (time.Duration, error)
+	value btcutil.Amount, fast bool) (time.Duration, error)
```
starius, line 213: IIUC, `true` is never passed to this callback. If `fast=true`, there is a separate code path in `spinUpBatch`:
```
if cfg.initialDelayProvider == nil || fast {
		cfg.initialDelayProvider = zeroInitialDelay
	}
```

I propose to use this new argument instead of making a special case in `spinUpBatch`.

sweepbatcher/sweep_batcher.go
```diff
@@ -968,6 +981,12 @@ func (b *Batcher) handleSweeps(ctx context.Context, sweeps []*sweep,
 		}
 	}
 
+	// If fast is set, we spin up a new batch which is published
+	// immediately.
+	if fast {
+		return b.spinUpNewBatch(ctx, sweeps, true)
+	}
```
starius, line 988: This prevents batching fast sweeps. It makes sense to batch them to save on fees.

I propose to remove this `if` and track fastness per `sweep`. We can add the flag to `sweep` and modify `batch.Run` method to pass `fast` to `b.cfg.initialDelayProvider` call (the only place where `initialDelayProvider` is called).

It makes sense to add method `isFast()` for batch similar to `isUrgent` and call if from `batch.Run` for convenience.

And we don't need to pass `fast` to `spinUpBatch` anymore, because we can always deduce it from `sweeps`.

loopd/swapclient_server.go
```diff
@@ -1025,6 +1025,7 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 		Private:        req.Private,
 		Initiator:      defaultLoopdInitiator,
 		NumDeposits:    uint32(numDeposits),
+		Fast:           req.Fast,
```
starius, line 1036: I propose to check here that if `req.Fast` is set, we make sure that the swap is static. Otherwise fail. Just as a sanity check measure.

cmd/loop/staticaddr.go
```diff
@@ -485,6 +485,12 @@ var staticAddressLoopInCommand = cli.Command{
 				"is change it is sent back to the static " +
 				"address.",
 		},
+		cli.BoolFlag{
+			Name: "fast",
+			Usage: "expedited publishing of the change output if " +
+				"the swap creates one. This results in a " +
+				"higher swap fee.",
+		},
```
starius, line 493: I propose to rephrase it to make the purpose clearer:

```
Usage: "complete the swap faster by paying a higher fee, so the change output is available sooner",
```

cmd/loop/staticaddr.go
```diff
@@ -478,13 +478,6 @@ var staticAddressLoopInCommand = cli.Command{
 				"The client can retry the swap with adjusted " +
 				"parameters after the payment timed out.",
 		},
-		cli.Uint64Flag{
-			Name: "amount",
-			Usage: "the number of satoshis that should be " +
-				"swapped from the selected deposits. If there" +
-				"is change it is sent back to the static " +
-				"address.",
-		},
```
starius, line 487: Can we keep previous flag as well? So it is possible to set it via `--amount`. Some people may prefer this, e.g. in scripts where each argument is on a separate line. Positional argument is good for manual use.

starius:
CLI is broken:

```
$ ~/fast-flag/loop --network signet static in --utxo b7c34d293e816cdcb102f1e2dcfaba9c4f553356a881fcc042ad6522da972176:0
On-chain fees for static address loop-ins are not included.
They were already paid when the deposits were created.

Previously deposited on-chain:                  0 sat
Receive off-chain:                          -1568 sat
Estimated total fee:                         1568 sat

CONTINUE SWAP? (y/n): n
```

Also:

```
$ ~/fast-flag/loop --network signet static in --all
On-chain fees for static address loop-ins are not included.
They were already paid when the deposits were created.

Previously deposited on-chain:                  0 sat
Receive off-chain:                          -6399 sat
Estimated total fee:                         6399 sat

CONTINUE SWAP? (y/n):
```

I think this also exists in `master`.

It seems to pass `QuoteRequest{Amt: 0}` in `displayInDetails`. We have variable `allDeposits` in that function (`staticAddressLoopIn`), we can deduce the amount from it if user doesn't specify anything.

hieblmi:
> CLI is broken:
> 

Yes thanks, I noticed this bug a while ago and have a fix on my todo list. Its because the quoting part doesn't have access to the utxo amount yet.

looprpc/client.proto
hieblmi: done

swapserverrpc/server.proto
hieblmi: done

loopdb/sqlc/migrations/000020_static_address_fast_flag.down.sql
hieblmi: fixed

loopdb/sqlc/migrations/000020_static_address_fast_flag.up.sql
hieblmi: fixed

sweepbatcher/sweep_batcher.go
hieblmi: Thanks for the suggestion to briefly delay the publication of a fast batch to gather more fast swaps. This will be addressed in a separate PR.

loopd/swapclient_server.go
hieblmi: added a sanity check.

cmd/loop/staticaddr.go
hieblmi: added this description

cmd/loop/staticaddr.go
hieblmi: That makes sense, I re-added the flag.

cmd/loop/staticaddr.go
hieblmi: I reverted this change, it was requested here as well: https://github.com/lightninglabs/loop/pull/1009#discussion_r2407800833

sweepbatcher/sweep_batcher.go
hieblmi: `fast=true` is passed on the delay provider on the server side.

PR #1013, opened 2025-10-07
cmd/loop: update urfave/cli to v3, auto-generate loop CLI reference manual and man file

starius:
## Summary

  - Upgraded the Loop CLI to `urfave/cli/v3`, adjusted every command/action to the new API
  - Hardened user input across commands by parsing numeric flags with their native helpers, rejecting negative amounts, and keeping static-address labels consistent.
  - Added `loop man` / `loop markdown` helpers plus Make/Ci steps that regenerate manpages/markdown and verify docs stay up to date.
  - Hid the development-only `forceautoloop` command from end users.

## Testing

  - `make build`
  - `loop man`
  - `loop markdown`
  - `make docs-check`
  - Manually testing `loop` to make sure it works properly

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,25 @@ jobs:
 
       - name: run check
         run: make rpc-check
-  
+
+  ########################
+  # SQL compile and check
+  ########################
+  sqlc-check:
+    name: SQL compilation check
+    runs-on: ubuntu-latest
+    steps:
+      - name: git checkout
+        uses: actions/checkout@v2
+
+      - name: setup go ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: '~${{ env.GO_VERSION }}'
+
+      - name: run check
+        run: make sqlc-check
+
   ########################
   # go mod check
   ########################


cmd/loop/debug.go
```diff
@@ -22,6 +22,7 @@ var forceAutoloopCmd = cli.Command{
 	autoloop timer. THIS MUST NOT BE USED IN A PROD ENVIRONMENT.
 	`,
 	Action: forceAutoloop,
+	Hidden: true,
```
sputn1ck, line 25: This command is already hidden behind the dev build flag, but nothing wrong with adding extra hidden flag.

starius: The man and the markdown are built using `loop-debug` binary built by `make build`. That target builds with "dev" tag.

cmd/loop/markdown_tabular.md.gotmpl
```diff
@@ -0,0 +1,80 @@
+{{ define "flags" }}
```
bhandras, line 1: This is really cool!

bhandras (approved):
Thank you! LGTM!!

starius:
Rebased

starius:
Updated static loop in usage:
```
USAGE:
   loop static in [amt] [--all | --utxo xxx:xx]
```

Updated docs for `--fast` flag and fixed `make docs-check` so it actually catches if docs is outdated.

starius:
Fixed documentation generation to put paths like `~/.loop` instead of expanded paths like `/home/user/.loop` which were breaking reproducibility.

CI was catching this with diffs like this:

```diff
-\fB--loopdir\fP="": path to loop's base directory (default: /home/user/.loop)
+\fB--loopdir\fP="": path to loop's base directory (default: /home/runner/.loop)
```

PR #1015, opened 2025-10-07
staticaddr: signet testing fixes

hieblmi:
- fixes a display issue with the quoted receive amount
- fixes auto-selection of expired deposits

### DIFF
--- a/staticaddr/loopin/manager.go
+++ b/staticaddr/loopin/manager.go
@@ -862,8 +862,25 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
 // are needed to cover the amount requested without leaving a dust change. It
 // returns an error if the sum of deposits minus dust is less than the requested
 // amount.
-func SelectDeposits(targetAmount btcutil.Amount, deposits []*deposit.Deposit,
-	csvExpiry uint32, blockHeight uint32) ([]*deposit.Deposit, error) {
+func SelectDeposits(targetAmount btcutil.Amount,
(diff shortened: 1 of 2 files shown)


staticaddr/loopin/manager.go
```diff
@@ -862,8 +862,31 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
+	for _, d := range unfilteredDeposits {
+		depositExpiryHeight := d.ConfirmationHeight + int64(csvExpiry)
+
+		// The htlc expiry height is the current height plus the htlc
+		// cltv delta.
+		htlcExpiryHeight := blockHeight + DefaultLoopInOnChainCltvDelta
+
+		// Ensure that the deposit doesn't expire before the htlc.
+		if depositExpiryHeight < int64(htlcExpiryHeight+
+			DepositHtlcDelta) {
```
starius: I propose to factor out a function:
```go
func IsSwappable(confirmationHeight, blockHeight, csvExpiry uint32) bool
```

We can cover this logic with unit tests and maybe reuse in other places.

starius (approved):
**LGTM!**
I tested it in regtest. Works great!
```
$ loop static in 800000
Send on-chain:                             800000 sat
Receive off-chain:                         796723 sat
Estimated total fee:                         3277 sat

CONTINUE SWAP? (y/n):
```

staticaddr/loopin/manager.go
hieblmi: I've added a unit test in `manager_test.go`.

PR #1022, opened 2025-10-15
sweepbatcher: presign transactions in parallel

starius:
If a `SignTx` call is slow, the whole `presign()` function could timeout if all the calls are done sequentially. In this commit each call is done in a separate goroutine to reduce total latency.

CC @GeorgeTsagk for taproot-assets commit fix (second commit of this PR).

#### Pull Request Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,8 +23,8 @@ require (
 	github.com/lightninglabs/lndclient v0.20.0-1
 	github.com/lightninglabs/loop/looprpc v1.0.7
 	github.com/lightninglabs/loop/swapserverrpc v1.0.14
-	github.com/lightninglabs/taproot-assets v0.7.0-rc1.0.20250925100956-c44e078c33e0
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.10-0.20250925100956-c44e078c33e0
+	github.com/lightninglabs/taproot-assets v0.7.0-rc1.0.20251014172227-e6ae082c0b4b
+	github.com/lightninglabs/taproot-assets/taprpc v1.0.10-0.20251014172227-e6ae082c0b4b
 	github.com/lightningnetwork/lnd v0.20.0-beta.rc1
 	github.com/lightningnetwork/lnd/cert v1.2.2
 	github.com/lightningnetwork/lnd/clock v1.1.1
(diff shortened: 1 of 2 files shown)


sweepbatcher/presigned.go
```diff
@@ -389,21 +392,31 @@ func presign(ctx context.Context, presigner presigner, destAddr btcutil.Address,
...
-			return fmt.Errorf("failed to presign unsigned tx %v "+
-				"for feeRate %v: %w", tx.TxHash(), fr, err)
-		}
+		eg.Go(func() error {
+			_, err := presigner.SignTx(
+				grCtx, primarySweepID, tx, batchAmt,
+				minRelayFeeRate, fr, loadOnly,
+			)
+			if err != nil {
+				return fmt.Errorf("failed to presign unsigned "+
```
hieblmi, line 401: before we return, should we cancel `grCtx` so the other signing requests are stopped?

go.mod
```diff
@@ -23,8 +23,8 @@ require (
 	github.com/lightninglabs/lndclient v0.20.0-1
 	github.com/lightninglabs/loop/looprpc v1.0.7
 	github.com/lightninglabs/loop/swapserverrpc v1.0.14
-	github.com/lightninglabs/taproot-assets v0.7.0-rc1.0.20250925100956-c44e078c33e0
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.10-0.20250925100956-c44e078c33e0
+	github.com/lightninglabs/taproot-assets v0.7.0-rc1.0.20251014172227-e6ae082c0b4b
```
GeorgeTsagk, line 26: :+1: 
We'll soon tag our rc2 so eventually you might want to bump again to a tagged version, if preferred.

sweepbatcher/presigned.go
GeorgeTsagk: +1, if you're signing with sighash_all then I believe you should terminate early. If signatures in an otherwise "failed" batch can be re-used (i.e sighash_single?) then you should still sign them. Although I believe the former is the case.

starius: grCtx is canceled automatically if an error is returned.

https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext

> The derived Context is canceled the first time a function passed to Go returns a non-nil error or the first time Wait returns, whichever occurs first.

starius: @GeorgeTsagk Each goroutine signs a separate transaction (they are the same except that they have different fee rates). We need all of them to proceed. If some of them fail to sign, we cancel the whole process.

PR #1025, opened 2025-10-26
cli: rename --amount to --amt in static

starius:
```
loop static in --amt
loop static withdraw --amt
```

This is done to make it consistent with other commands and products.

#### Pull Request Checklist

### DIFF
--- a/cmd/loop/staticaddr.go
+++ b/cmd/loop/staticaddr.go
@@ -158,7 +158,8 @@ var withdrawalCommand = &cli.Command{
 				"the transaction",
 		},
 		&cli.Uint64Flag{
-			Name: "amount",
+			Name:    "amt",
+			Aliases: []string{"amount"},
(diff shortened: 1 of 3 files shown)


hieblmi (commented):
To remain compatible to current automation scripts should we just add the amt flag rather than replacing the amount flag with it?

starius:
@hieblmi I think we can just rename it, because the `--amount` flag is broken in release 0.31.4 and 0.31.5 is not released yet, so it is unlikely that somebody already uses it in automation scripts.

bhandras:
Actually `github.com/urfave/cli/v2` does support flag aliases, ie you can do something like this:

```
&cli.IntFlag{
  Name:        "amount",
  Aliases:     []string{"amt"},
}
```

starius:
I added `--amount` as an alias.

bhandras (approved):

PR #1038, opened 2025-11-10
loopout: don't print htlc address

hieblmi:


starius (approved):
**LGTM!** 🏆

If someone really needs this info, it can be found in `loop listswaps` output. (Maybe it makes sense to hide it there as well under `--verbose` option or something.)

Also `loop in` still prints HTLC address. I propose to remove it as well.

hieblmi:
For loop ins with --external we want to print the htlc.

bhandras (approved):

PR #1040, opened 2025-11-11
staticaddr: channel funding with deposits

hieblmi:
_depends on https://github.com/lightninglabs/loop/pull/1073_

This PR introduces a `openchannel` subcommand to static addresses.

It provides the same experience as `lncli openchannel`. 
Exmples:

Open a channel with all available deposits:
```
loop static openchannel --fundmax --sat_per_vbyte XXX --private ...
```

Open a channel with specified local funding amount, coin-selected from available deposits under fee and dust considerations.
```
loop static openchannel --local_amt XXX --sat_per_vbyte YYY --private ...
```

Open a channel from two deposits AAA and BBB while taking their combined value(fundmax) as funding amount and considering fees and dust limit.
```
loop static openchannel --utxo AAA --utxo BBB --fundmax --sat_per_vbyte YYY --private ...
```

Open a channel from two deposits AAA and BBB while taking a specified amount(local_amt) as funding amount and considering fees and dust limit.
```
loop static openchannel --utxo AAA --utxo BBB --local_amt --sat_per_vbyte YYY --private ...
```

TODOs:

- [x] Integration tests


cmd/loop/openchannel.go
```diff
@@ -0,0 +1,351 @@
...
+		},
+	},
+	Action: openChannel,
+}
+
+func openChannel(ctx context.Context, cmd *cli.Command) error {
+	var (
+		args      = cmd.Args()
+		remaining []string
+		ctxb      = context.Background()
```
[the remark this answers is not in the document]
hieblmi: fixed

cmd/loop/openchannel.go
```diff
@@ -0,0 +1,351 @@
...
+		req.Outpoints = outpoints
+	}
+
+	// The fundmax flag is NOT allowed to be combined with local_amt above.
+	// It is allowed to be combined with push_amt, but only if explicitly
+	// set.
+	if cmd.Bool("fundmax") && req.LocalFundingAmount != 0 {
+		return fmt.Errorf("local amount cannot be set if attempting " +
+			"to commit the maximum amount out of the wallet")
+	}
```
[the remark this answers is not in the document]
hieblmi: fixed

cmd/loop/openchannel.go
```diff
@@ -0,0 +1,351 @@
...
+		req.CommitmentType = lnrpc.CommitmentType_SIMPLE_TAPROOT
+	default:
+		return fmt.Errorf("unsupported channel type %v", channelType)
+	}
+
+	wrappedReq := &looprpc.StaticOpenChannelRequest{
+		OpenChannelRequest: req,
+	}
+
+	resp, err := client.StaticOpenChannel(ctxb, wrappedReq)
```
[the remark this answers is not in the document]
hieblmi: fixed

cmd/loop/openchannel.go
```diff
@@ -0,0 +1,351 @@
...
+		RemoteChanReserveSat:       cmd.Uint64("remote_reserve_sats"),
+		Memo:                       cmd.String("memo"),
+	}
+
+	switch {
+	case cmd.IsSet("node_key"):
+		nodePubHex, err := hex.DecodeString(cmd.String("node_key"))
+		if err != nil {
+			return fmt.Errorf("unable to decode node public key: "+
+				"%v", err)
```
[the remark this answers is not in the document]
hieblmi: fixed

hieblmi:
@gemini-code-assist review

staticaddr/deposit/fsm.go
```diff
@@ -366,6 +387,17 @@ func (f *FSM) DepositStatesV0() fsm.States {
...
+		OpeningChannel: fsm.State{
+			Transitions: fsm.Transitions{
+				fsm.OnError:        Deposited,
+				OnChannelPublished: ChannelPublished,
+				OnRecover:          OpeningChannel,
+			},
+			Action: fsm.NoOpAction,
+		},
+		ChannelPublished: fsm.State{
+			Action: fsm.NoOpAction,
```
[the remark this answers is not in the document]
hieblmi: @gemini-code-assist `f.FinalizeDepositAction` removes the fsm from the set of active fsms. Isn't it needed later once the channel is active?

hieblmi:
@gemini-code-assist review

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,659 @@
...
+		deposits, err = m.cfg.DepositManager.GetActiveDepositsInState(
+			deposit.Deposited,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		if req.LocalFundingAmount != 0 {
+			deposits, err = staticutil.SelectDeposits(
+				deposits, req.LocalFundingAmount,
```
sputn1ck, line 334: SelectDeposits isn’t fee-aware. We lock deposits based on local_amt + dust, then may fail once fees are applied. Worth a fee-aware selection or pre-check before OnOpeningChannel.

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,659 @@
...
+		}
+	}
+
+	// We need to transition the deposits to the opening channel state
+	// before we start the channel open process. This is important to
+	// ensure that the deposits are not used for other purposes while we
+	// are opening the channel.
+	err = m.cfg.DepositManager.TransitionDeposits(
+		ctx, deposits, deposit.OnOpeningChannel, deposit.OpeningChannel,
+	)
```
sputn1ck, line 365: Once deposits move to OpeningChannel there’s no recovery path here. On restart, these can stay locked indefinitely unless something else reconciles. Should we reconcile with lnd pending chans / on-chain spends?

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,659 @@
...
+		chanCommitmentType = lnrpc.CommitmentType_ANCHORS
+
+	case lnrpc.CommitmentType_SIMPLE_TAPROOT:
+		chanCommitmentType = lnrpc.CommitmentType_SIMPLE_TAPROOT
+	}
+
+	if req.SatPerVbyte == 0 {
+		// Get the fee rate for the withdrawal sweep.
+		feeRate, err = m.cfg.WalletKit.EstimateFeeRate(
+			ctx, defaultConfTarget,
```
sputn1ck: We estimate fee rate here, but CreateFinalizedWithdrawalTx estimates again when sat_per_vbyte is unset. That can skew funding amt vs PSBT build. Can we thread a single fee rate through?

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,659 @@
...
+
+	openChanRequest := &lnrpc.OpenChannelRequest{
+		NodePubkey:                 req.NodePubkey,
+		LocalFundingAmount:         int64(chanFundingAmt),
+		PushSat:                    req.PushSat,
+		Private:                    req.Private,
+		MinHtlcMsat:                req.MinHtlcMsat,
+		RemoteCsvDelay:             req.RemoteCsvDelay,
+		MinConfs:                   defaultUtxoMinConf,
+		SpendUnconfirmed:           req.SpendUnconfirmed,
```
sputn1ck: MinConfs is hard-coded to 1. If caller sets MinConfs !=1, we silently ignore it. Should we validate/reject unsupported MinConfs before locking deposits?

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,659 @@
...
+		Memo:                       req.Memo,
+	}
+
+	chanTxHash, err := m.openChannelPsbt(
+		ctx, openChanRequest, deposits, int64(req.SatPerVbyte),
+	)
+	if err != nil {
+		log.Infof("error opening channel: %v", err)
+		err2 := m.cfg.DepositManager.TransitionDeposits(
+			ctx, deposits, fsm.OnError, deposit.Deposited,
```
sputn1ck: Any error after PSBT finalize currently transitions deposits back to Deposited. If finalize already handed the tx to lnd (NoPublish=false), we can reuse spent deposits. Can we gate rollback once finalize is sent or persist txid/reconcile?

staticaddr/openchannel/managert_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCalculateFundingTxValues tests the calculateFundingTxValaues function
+// with various scenarios.
+func TestCalculateFundingTxValues(t *testing.T) {
```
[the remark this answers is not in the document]
hieblmi: fixed

staticaddr/openchannel/manager.go
hieblmi: I made `SelectDeposits` fee aware.

staticaddr/openchannel/manager.go
hieblmi: Good point. This needs reconciliation on startup to recover deposits stuck in OpeningChannel state. Will address in a follow-up PR that checks lnd pending channels and on-chain spends to transition deposits back to Deposited or forward to ChannelPublished.

staticaddr/openchannel/manager.go
hieblmi: Fixed

staticaddr/openchannel/manager.go
hieblmi: Gated rollback after we send out the finalize.

staticaddr/openchannel/manager.go
hieblmi: done

litbot-9000:
@sputn1ck: review reminder

hieblmi:
Hi @sputn1ck, I addressed some more findings, namely:
- Added crash recovery for OpeningChannel deposits.
- Upgraded permissions to swap:execute for both StaticOpenChannel and StaticAddressLoopIn.
- Aligned channel-open fee/value calculation paths by threading commitment type into tx construction.
- Hardcoded forwarded SpendUnconfirmed to false.
- Added unit tests for openchannel recovery logic.

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,806 @@
...
+			// If the remote peer canceled on us, the reservation
+			// has already been deleted. We don't need to try to
+			// remove it again, this would just produce another
+			// error.
+			cancelErr := chanfunding.ErrRemoteCanceled.Error()
+			if err != nil && strings.Contains(
+				err.Error(), cancelErr,
+			) {
+
+				shimPending = false
```
starius, line 587: Data race in shimPending variable. It is accessed from multiple goroutines without synchronization.

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,806 @@
...
+			// interrupts now, we don't need to clean up anything.
+			shimPending = false
+
+			hash, err := chainhash.NewHash(
+				update.ChanPending.Txid,
+			)
+			if err != nil {
+				log.Infof("Error creating hash for channel "+
+					"open tx: %v", err)
+			}
```
starius, line 721: I think we should `return error` in this `if`. We use the `hash` variable below as if it is valid, but it could be `nil` and we just log the error and crash later.

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,806 @@
...
+			log.Infof("Please monitor the channel from lnd")
+
+			err = m.cfg.DepositManager.TransitionDeposits(
+				ctx, deposits, deposit.OnChannelPublished,
+				deposit.ChannelPublished,
+			)
+			if err != nil {
+				log.Errorf("error transitioning deposits to "+
+					"ChannelPublished: %v", err)
+			}
```
starius, line 734: `TransitionDeposits` errors are logged but the function still returns success, leaving deposits stuck in `OpeningChannel` while the channel is actually pending. This should be returned to the caller or retried.

staticaddr/deposit/fsm.go
```diff
@@ -366,6 +387,17 @@ func (f *FSM) DepositStatesV0() fsm.States {
...
+			Transitions: fsm.Transitions{
+				fsm.OnError:        Deposited,
+				OnChannelPublished: ChannelPublished,
+				OnRecover:          OpeningChannel,
+			},
+			Action: fsm.NoOpAction,
+		},
+		ChannelPublished: fsm.State{
+			Action: f.FinalizeDepositAction,
+		},
```
starius, line 412: These two states lack OnExpiry transitions from them. OnExpiry may be delivered to each of them.

1. OpeningChannel has Action: fsm.NoOpAction. This means it stops current FSM execution and unlocks the FSM mutex, letting potential SendEvent calls sneak in.
2. FinalizeDepositAction also returns fsm.NoOp and triggers further things via a channel (f.finalizedDepositChan) so there is still a window when SendEvent can sneak in.

`handleBlockNotification` fires OnExpiry on every new block once the deposit is expired (staticaddr/deposit/fsm.go), regardless of state.

So there is a possibility that OnExpiry is delivered to each of these new states. We have to add state self-transitions to each of them.

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,806 @@
...
+	}
+
+	if len(openingDeposits) == 0 {
+		return nil
+	}
+
+	log.Infof("Recovering %d deposits in OpeningChannel state",
+		len(openingDeposits))
+
+	utxos, err := m.cfg.WalletKit.ListUnspent(
```
starius, line 180: What if the deposit is spent, but not confirmed? Shouldn't we still track it? Maybe it is evicted from mempool. Or even confirmed first and then reorged.

staticaddr/openchannel/manager_test.go
```diff
@@ -0,0 +1,314 @@
...
+			if tc.expectedErrSubstr == "" {
+				require.NoError(t, err)
+				require.Equal(t, tc.expectedType, commitmentType)
+				return
+			}
+
+			require.ErrorContains(t, err, tc.expectedErrSubstr)
+		})
+	}
+}
```
starius, line 597: I propose to add test coverage for edge cases:
 - reorg
 - daemon restart during channel opening
 - eviction from mempool
 - rejection from mempool
 - stream errors
 - PSBT finalize then stream abort
 - duplicate outpoints in request

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,806 @@
...
+		newOpenChannelRequestChan: make(chan newOpenChannelRequest),
+		errChan:                   make(chan error),
+	}
+
+	return m
+}
+
+// Run runs the open channel manager.
+func (m *Manager) Run(ctx context.Context) error {
+	err := m.recoverOpeningChannelDeposits(ctx)
```
starius, line 125: Should we recover deposits during daemon running as well, not only during startup time?

If the PSBT finalize succeeds but the stream fails before `ChanPending`, deposits remain in `OpeningChannel` until restart.

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,806 @@
...
+			req.SatPerVbyte * 1000,
+		).FeePerKWeight()
+	}
+
+	// There are three ways in which we select deposits to open a channel
+	// with. 1.) The user manually selects the deposits. 2.) The user only
+	// selects a local channel amount in which case we coin-select deposits
+	// to cover for it. 3.) The user selects the fundmax flag, in which case
+	// we select all deposits to fund the channel.
+	if len(req.Outpoints) > 0 {
```
starius, line 294: What if `req.Outpoints` has duplicates? This leads to fee miscalculation and invalid PSBT building (having the same input twice). I propose to return an error early if we detect duplicates in `req.Outpoints` and add test coverage.

staticaddr/deposit/manager.go
starius: dead code

looprpc/client.proto
```diff
@@ -206,13 +207,33 @@ service SwapClient {
...
+    // Wrap lnd's request so Loop can extend this RPC with Loop-specific fields
+    // without diverging from the upstream API surface.
+    lnrpc.OpenChannelRequest open_channel_request = 1;
+}
+
+message StaticOpenChannelResponse {
+    /*
+    The transaction hash of the channel opening transaction.
+    */
+    string channel_open_tx_hash = 1;
```
starius: Should we also add the output index (always 0 in the current implementation)? Just in case we implement batch open or something in the future. The code which uses this thing shouldn't guess what is the index, it should take it from the API.

starius: Alternatively change `channel_open_tx_hash` to `channel_open_outpoint` and put "txid:index" there.

staticaddr/deposit/manager.go
hieblmi: fixed

staticaddr/openchannel/manager.go
hieblmi: Added a mutex to gate access.

staticaddr/openchannel/manager.go
hieblmi: Good catch, fixed.

staticaddr/openchannel/manager.go
hieblmi: The current approach is to just log the transition error and return the `chanOutpoint` since the transition error might be a recurring error that isn't solved without intervention. 
So the approach is to reconcile these `OpeningChannel` states after a restart in `recoverOpeningChannelDeposits`.

staticaddr/deposit/fsm.go
hieblmi: That is a great catch, thank you!
Fixed it.

staticaddr/openchannel/manager.go
hieblmi: I've fixed that in commit 15bfd345d1a2050ea27d50f39499c97dacd9d60d

staticaddr/openchannel/manager.go
hieblmi: added logic to detect dupes and return in that case.

looprpc/client.proto
hieblmi: I like the second suggestion, added.

staticaddr/openchannel/manager.go
hieblmi: Fixed. We are checking if there's an error but the PSBT was finalized and sent, then we recover remaining deposits.

staticaddr/openchannel/manager_test.go
hieblmi: Added!

staticaddr/openchannel/manager.go
starius: What do we do with TransitionDeposits errors? Can we retry TransitionDeposits several times and if it still fails, run recoverOpeningChannelDeposits? And if it also fails, return hard error.
Which errors TransitionDeposits can return? Don't we hide anything important when skipping the errors?

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,837 @@
...
+		CommitmentType:             chanCommitmentType,
+		ZeroConf:                   req.ZeroConf,
+		ScidAlias:                  req.ScidAlias,
+		BaseFee:                    req.BaseFee,
+		FeeRate:                    req.FeeRate,
+		UseBaseFee:                 req.UseBaseFee,
+		UseFeeRate:                 req.UseFeeRate,
+		RemoteChanReserveSat:       req.RemoteChanReserveSat,
+		Memo:                       req.Memo,
+	}
```
starius: Most of the fields are copied from `req`.
Could we just clone `req` and then adjust few fields (SpendUnconfirmed, MinConfs, etc).

Or even better accept `req` as-is and apply checks on its fields. E.g. demand the correct value of LocalFundingAmount, MinConfs, SpendUnconfirmed) in the input `req`.

Note that some fields are silently ignored or skipped when passing to LND:
  - target_conf: ignored (not used for fee selection, not rejected).
  - sat_per_byte (deprecated): ignored (not used, not rejected).
  - node_pubkey_string (deprecated): ignored by this path (only node_pubkey bytes are used).
  - funding_shim: ignored/overwritten (manager always builds its own PSBT shim).
  - sat_per_vbyte: used locally for withdrawal/funding fee math, but not forwarded to lnd OpenChannel request.
  - fund_max, outpoints: consumed by Loop’s own deposit-selection logic, not forwarded.
  - min_confs: normalized to 1 for PSBT flow (0 accepted as default/unset).

Shouldn't we pass fund_max to LND? It affects channel size selection.

hieblmi: I think your second option is good, I will apply `Or even better accept req as-is and apply checks on its fields`.
However, `fundmax` cannot be passed to lnd in psbt mode as that isn't implemented yet.

hieblmi:
@starius addressed recent comments in commit e0934b6a2a3cc4dead1dbaa059f5af47d8e65ced

staticaddr/openchannel/manager.go
```diff
@@ -0,0 +1,849 @@
...
+					"after failed channel open: %v",
+					err2)
+			}
+		}
+
+		return nil, err
+	}
+
+	return chanOutpoint, nil
+}
```
starius: Can we reduce indentation in this code?

```go
chanOutpoint, err := m.openChannelPsbt(ctx, reqClone, deposits, feeRate)
if err == nil {
  return chanOutpoint, nil
}

log.Infof("error opening channel: %v", err)
...

```

staticaddr/openchannel/manager.go
```diff
@@ -447,6 +447,7 @@ func (m *Manager) openChannelPsbt(ctx context.Context,
 		psbtFinalized bool
 		basePsbtBytes []byte
 		quit          = make(chan struct{})
+		quitErr       = make(chan error, 1)
```
starius: Can we merge these two channels together?

hieblmi: We can remove the chan and just assign the error instead.

staticaddr/openchannel/manager.go
hieblmi: done

PR #1042, opened 2025-11-12
sweepbatcher: handle batch shutdown during re-add

starius:
Started from [CI failure](https://github.com/lightninglabs/loop/actions/runs/19282552307/job/55136597881?pr=1041) of `TestSweepBatcherHandleSweepRace/loopdb` that surfaced `ErrBatcherShuttingDown` when the `AddSweep` loop caught a batch right as it finished. Reproduced via:

```
go test ./sweepbatcher -run TestSweepBatcherHandleSweepRace/loopdb -cpu=12,4,8 -count=1
```

Discovered that `handleSweeps` treated `ErrBatchShuttingDown` from `batch.addSweeps` as fatal, so the batcher exited even though the sweep was already confirmed. Fixed by treating that error as "batch already done" so we fall through to the persisted status/`monitorSpend` path, and added a regression test that
deterministically simulates the shutdown window.

New regression test added:
```
go test ./sweepbatcher -run TestSweepBatcherHandleBatchShutdown
```

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batcher.go
+++ b/sweepbatcher/sweep_batcher.go
@@ -906,7 +906,15 @@ func (b *Batcher) handleSweeps(ctx context.Context, sweeps []*sweep,
 	for _, batch := range b.batches {
 		if batch.sweepExists(sweep.outpoint) {
 			accepted, err := batch.addSweeps(ctx, sweeps)
-			if err != nil && !errors.Is(err, ErrBatchShuttingDown) {
+			if errors.Is(err, ErrBatchShuttingDown) {
+				// The batch finished while we were trying to
+				// re-add the sweep. Leave it in the map for the
+				// lazy cleanup below and fall back to the
+				// monitorSpendAndNotify path below.
+				break
+			}
+
+			if err != nil {
 				return err
 			}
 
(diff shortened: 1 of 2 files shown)


sweepbatcher/sweep_batcher.go
```diff
@@ -906,7 +906,15 @@ func (b *Batcher) handleSweeps(ctx context.Context, sweeps []*sweep,
 	for _, batch := range b.batches {
 		if batch.sweepExists(sweep.outpoint) {
 			accepted, err := batch.addSweeps(ctx, sweeps)
-			if err != nil && !errors.Is(err, ErrBatchShuttingDown) {
+			if errors.Is(err, ErrBatchShuttingDown) {
```
hieblmi, line 909: IIUC there is no way that `addSweeps` still adds the sweep _and_ returns `ErrBatchShuttingDown`, right?

hieblmi (approved):
I have one question, other than that looks good to me!

sweepbatcher/sweep_batcher.go
starius: No, it is not possible. ErrBatchShuttingDown is returned only in the beginning of `addSweeps`, before all the checks and additions.

PR #1043, opened 2025-11-12
staticaddr: psbt withdrawals

hieblmi:
This PR a new rpc for withdrawals via psbts,
```
rpc ServerPsbtWithdrawDeposits (ServerPsbtWithdrawRequest)
        returns (ServerPsbtWithdrawResponse);
```

### DIFF
--- a/staticaddr/withdraw/manager_test.go
+++ b/staticaddr/withdraw/manager_test.go
@@ -4,11 +4,19 @@ import (
 	"context"
 	"testing"
 
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/loop/staticaddr/deposit"
 	"github.com/lightninglabs/loop/swapserverrpc"
 	"github.com/lightninglabs/loop/test"
+	"github.com/lightningnetwork/lnd/funding"
 	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


staticaddr/withdraw/manager.go
```diff
@@ -547,45 +565,40 @@ func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
...
-	prevOutFetcher := txscript.NewMultiPrevOutFetcher(prevOuts)
-	_, err = m.signMusig2Tx(
-		ctx, prevOutFetcher, outpoints, m.cfg.Signer, withdrawalTx,
-		withdrawalSessions, coopServerNonces,
-	)
-	if err != nil {
-		return nil, err
+	if len(sigResp.SigningInfo) != len(deposits) {
+		return nil, nil, errors.New("invalid number of " +
+			"deposit signatures")
```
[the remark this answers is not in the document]
hieblmi: What is your point? The checks are in place. Do you suggest additional checks?

swapserverrpc/staticaddr.proto
```diff
@@ -92,13 +100,41 @@ message ServerWithdrawRequest {
...
     repeated bytes musig2_sweep_sigs = 1;
 
-    // The nonces that the server used to generate the sweepless sweep sigs.
+    // The nonces that the server used to generate the withdrawal sigs.
     repeated bytes server_nonces = 2;
 }
 
+message ServerPsbtWithdrawRequest {
+    // The withdrawal psbt.
+    bytes withdrawal_psbt = 1;
```
hieblmi, line 123: nit: comment that txscript.SigHashDefault will be enforced by default.

litbot-9000:

staticaddr/withdraw/manager.go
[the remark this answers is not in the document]
starius: I propose to check not only the length, but also that they have the same elements (1:1 matching). That each deposit got one record in sigResp.SigningInfo.

Please also add a test checking this situation.

staticaddr/staticutil/outpoints.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// ToWireOutpoints converts lnrpc.OutPoint protos into wire.OutPoint structs so
+// they can be consumed by lower level transaction building code.
+func ToWireOutpoints(outpoints []*lnrpc.OutPoint) ([]wire.OutPoint, error) {
```
starius, line 12: This function is not used and not tested.

staticaddr/staticutil/utils.go
```diff
@@ -0,0 +1,205 @@
...
+		jhash[b], jhash[hashSize-1-b] = jhash[hashSize-1-b], jhash[b]
+	}
+	return bytes.Compare(ihash[:], jhash[:]) == -1
+}
+
+// SelectDeposits sorts the deposits by amount in descending order. It then
+// selects the deposits that are needed to cover the amount requested without
+// leaving a dust change. It returns an error if the sum of deposits minus dust
+// is less than the requested amount.
+func SelectDeposits(deposits []*deposit.Deposit, amount int64) (
```
starius, line 172: The function is not used and not tested.

staticaddr/withdraw/manager.go
```diff
@@ -501,17 +505,27 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
-		ctx, deposits,
+	addrParams, err := m.cfg.AddressManager.GetStaticAddressParameters(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	staticAddress, err := m.cfg.AddressManager.GetStaticAddress(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
```
starius, line 525: This makes two DB queries. It could be one: staticAddress can be built from addrParams. We can factor out a function from `address.Manager.GetStaticAddress` which converts params to address. Or just merge two methods GetStaticAddressParameters and GetStaticAddress together and return both things at once.

staticaddr/staticutil/utils.go
```diff
@@ -0,0 +1,205 @@
...
+		nonces[deposit.String()] = session.PublicNonce[:]
+		depositToIdx[deposit.String()] = i
+	}
+
+	return sessions, nonces, depositToIdx, nil
+}
+
+// CreateMusig2Session creates a musig2 session for the deposit.
+func CreateMusig2Session(ctx context.Context,
+	signer lndclient.SignerClient, addrParams *address.Parameters,
```
starius, line 103: Idea: can we move address.Parameters to `script` package and reuse it and this function.

staticaddr/withdraw/manager.go
```diff
@@ -726,105 +739,87 @@ func toOutpoints(deposits []*deposit.Deposit) []wire.OutPoint {
 // signMusig2Tx adds the server nonces to the musig2 sessions and signs the
 // transaction.
 func (m *Manager) signMusig2Tx(ctx context.Context,
-	prevOutFetcher *txscript.MultiPrevOutFetcher, outpoints []wire.OutPoint,
+	prevOutFetcher *txscript.MultiPrevOutFetcher,
```
starius, line 755: Add tests for signMusig2Tx.

When sigInfo is missing an entry for one of the deposits, the call should error (no partial return). Build a dummy tx with two inputs, provide sessions for both, but only one SigningInfo entry; expect error.

When sigInfo has all expected keys but one maps to the wrong index (mismatched depositsToIdx), ensure it errors.

staticaddr/withdraw/manager.go
```diff
@@ -956,153 +893,187 @@ func (m *Manager) createWithdrawalTx(ctx context.Context,
...
-	// Get the weight of the sweep output.
-	switch sweepAddress.(type) {
-	case *btcutil.AddressWitnessPubKeyHash:
-		weightEstimator.AddP2WKHOutput()
+	return msgTx, psbtBuf.Bytes(), nil
+}
 
-	case *btcutil.AddressTaproot:
-		weightEstimator.AddP2TROutput()
+func CalculateWithdrawalTxValaues(deposits []*deposit.Deposit,
```
starius: typo: "Valaues"

staticaddr/withdraw/manager.go
```diff
@@ -956,153 +893,187 @@ func (m *Manager) createWithdrawalTx(ctx context.Context,
...
+	localAmount btcutil.Amount, feeRate chainfee.SatPerKWeight,
+	withdrawalAddress btcutil.Address,
+	commitmentType lnrpc.CommitmentType) (btcutil.Amount, btcutil.Amount,
```
starius, line 958: Add a table test for CalculateWithdrawalTxValues covering: insufficient funds after dust/fee, negative change, and min-channel-size guard.

staticaddr/withdraw/manager.go
hieblmi: I totally agree. These methods will be removed in the follow-up PR for multi addresses so i think we can leave them for now.

staticaddr/staticutil/outpoints.go
hieblmi: These were factored out from the open-channel PR and will be used in there.

staticaddr/staticutil/utils.go
hieblmi: These were factored out from the open-channel PR and will be used in there.

staticaddr/staticutil/utils.go
hieblmi: I will refactor this in a follow-up PR. Created issue https://github.com/lightninglabs/loop/issues/1056

staticaddr/withdraw/manager.go
hieblmi: Thank you for these suggestions, I've added checks and tests as you recommended. I also added a test for the case when an input of the tx is not present in the deposit map, see "tx outpoint not in deposit index map".

staticaddr/withdraw/manager.go
hieblmi: fixed

staticaddr/withdraw/manager.go
hieblmi: I added comprehensive test coverage for this function.

staticaddr/withdraw/manager.go
```diff
@@ -501,17 +505,27 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
 	return finalizedTx.TxID(), withdrawalAddress.String(), nil
 }
 
-func (m *Manager) createFinalizedWithdrawalTx(ctx context.Context,
+func (m *Manager) CreateFinalizedWithdrawalTx(ctx context.Context,
```
sputn1ck, line 511: Missing godoc

hieblmi: fixed

PR #1044, opened 2025-11-13
sweepbatcher: persist confirmed batches atomically

starius:
Fixes a crash window where `handleConf` updated the batch row to confirmed but failed before marking sweeps complete, so re-added sweeps spawned a duplicate batch or kept retrying. Batch confirmation and sweep completion are now persisted inside a single DB transaction `ConfirmBatchWithSweeps`, and `handleConf` uses the helper to atomically store the batch and the set of confirmed sweeps.

Added `TestSweepBatcherConfirmedBatchIncompleteSweeps` that runs against the real `loopdb` backend, injects a failure mid-transaction, and verifies the database never ends up with confirmed=true batches paired with completed=false sweeps.

#### Pull Request Checklist

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -2224,10 +2224,6 @@ func (b *batch) handleConf(ctx context.Context,
 	b.Infof("confirmed in txid %s", b.batchTxid)
 	b.state = Confirmed
 
-	if err := b.persist(ctx); err != nil {
-		return fmt.Errorf("saving batch failed: %w", err)
-	}
-
 	// If the batch is in presigned mode, cleanup presignedHelper.
 	presigned, err := b.isPresigned()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -2328,8 +2322,15 @@ func (b *batch) handleConf(ctx context.Context,
 		}
 	}
 
-	b.Infof("fully confirmed sweeps: %v, purged sweeps: %v, "+
-		"purged swaps: %v", confirmedSweeps, purgedSweeps, purgedSwaps)
+	b.Infof("Fully confirmed sweeps: %v, purged sweeps: %v, "+
+		"purged swaps: %v. Saving the batch and sweeps to DB",
+		confirmedSweeps, purgedSweeps, purgedSwaps)
+
+	if err := b.persistConfirmedBatch(ctx, dbConfirmed); err != nil {
```
hieblmi, line 2329: Nice, now both batch and sweeps are persisted at the same time!

Can we make sure that in other places where we call `b.persist(ctx)` there doesn't arise this discrepancy?

litbot-9000:
@bhandras: review reminder
@sputn1ck: review reminder

sweepbatcher/sweep_batch.go
starius: I re-checked all other places. This is the only one when a batch and a sweep are mutated simultaneously. Also double-checked with AI - no other findings.

starius:
Rebased

PR #1050, opened 2025-11-24
Add LiT CI job and runner cleanup action

ffranr:
- Add composite cleanup-space action to reclaim runner disk before heavy jobs.
- Wire LiT integration/unit test job into CI, using local Loop replacements for go.mod and LiT setup node action.

#### Pull Request Checklist
- [x] ~~Update `release_notes.md` if your PR contains major features, breaking changes or bugfixes~~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -125,10 +125,10 @@ jobs:
         run: make docs-check
 
   ########################
-  # run unit tests
+  # run unit-test sqlite3 race
   ########################
   unit-test:
-    name: run unit tests
+    name: run unit-test sqlite3 race
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
@@ -139,10 +139,25 @@ jobs:
         with:
           go-version: '~${{ env.GO_VERSION }}'
 
-      - name: run unit tests
+      - name: run unit-test sqlite3 race
         run: make unit-race
 
-      - name: run unit test with postgres
+  ########################
+  # run unit-test postgres race
+  ########################
+  unit-test-postgres:
+    name: run unit-test postgres race
+    runs-on: ubuntu-latest
+    steps:
+      - name: git checkout
+        uses: actions/checkout@v4
+
+      - name: setup go ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: '~${{ env.GO_VERSION }}'
+
+      - name: run unit-test postgres race
         run: make unit-postgres-race
 
   ########################


.github/workflows/main.yml
```diff
@@ -124,22 +125,75 @@ jobs:
         run: make docs-check
 
   ########################
-  # run unit tests
+  # run unit-test race
```
starius: Please add here and below in this section that it is sqlite3.

.github/actions/lit-setup/action.yml
```diff
@@ -0,0 +1,61 @@
...
+      with:
+        node-version: 16.x
+
+    - name: Cache yarn dependencies
+      uses: actions/cache@v4
+      with:
+        path: |
+          ~/.cache/yarn
+          lightning-terminal/app/node_modules
+        key: ${{ runner.os }}-lit-node16-${{ hashFiles('lightning-terminal/app/yarn.lock') }}
```
starius, line 51: Is it time to bump LiT to latest node LTS version? (24)

.github/actions/cleanup-space/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+        shopt -s nullglob
+        directories_to_remove=(
+          /usr/share/dotnet
+          /usr/local/lib/android
+          /opt/ghc
+          /usr/share/swift
+          /usr/local/julia*
+          /opt/hostedtoolcache
+          /usr/local/share/boost
+          "${AGENT_TOOLSDIRECTORY:-}"
```
starius: Do we need this? We don't use dotnet, GHC, swift etc.

.github/actions/cleanup-space/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+          '^llvm-.*'
+          'php.*'
+          '^mongodb-.*'
+          '^mysql-.*'
+          azure-cli
+          google-chrome-stable
+          firefox
+          powershell
+          mono-devel
+          libgl1-mesa-dri
```
starius: This list of packages seems also unrelated to Loop. We don't use mysql and PHP.

.github/actions/cleanup-space/action.yaml
ffranr: I've removed the cleanup action. It wasn't strictly necessary right now anyway. I was attempting to add something very general that we could use across LL projects.

.github/actions/cleanup-space/action.yaml
ffranr: Removed, see: https://github.com/lightninglabs/loop/pull/1050#discussion_r2561147972

.github/actions/lit-setup/action.yml
ffranr: This is the action we basically have in taproot-assets right now. I think there's some value in that consistency. I'll try to upgrade both soon. Thanks for the heads up on this.

.github/actions/cleanup-space/action.yaml
hieblmi: Could we also remove the first commit that introduces the `action.yml`?

hieblmi (approved):
Thanks for the improvements, LGTM ✅

Just one nit about the first commit, I think we can drop it.

PR #1055, opened 2025-12-02
sweepbatcher: threshold for immediate batch publishing

hieblmi:
This PR Implements a configurable threshold to trigger immediate batch publishing when cumulative sweep value exceeds a specified amount, bypassing the initial delay grouping period.

### DIFF
--- a/sweepbatcher/sweep_batch.go
+++ b/sweepbatcher/sweep_batch.go
@@ -200,6 +200,11 @@ type batchConfig struct {
 	// chainParams are the chain parameters of the chain that is used by
 	// batches.
 	chainParams *chaincfg.Params
+
+	// immediatePublishThreshold is the cumulative value threshold above
+	// which a batch should be published immediately without waiting for
+	// the initial delay grouping period. If 0, the feature is disabled.
+	immediatePublishThreshold btcutil.Amount
 }
 
 // rbfCache stores data related to our last fee bump.
@@ -807,6 +812,24 @@ func (b *batch) sweepExists(outpoint wire.OutPoint) bool {
 	return ok
 }
 
+// totalValue returns the sum of all sweep values in the batch.
+func (b *batch) totalValue() btcutil.Amount {
+	var total btcutil.Amount
+	for _, sweep := range b.sweeps {
+		total += sweep.value
+	}
+	return total
+}
+
+// exceedsThreshold checks if the batch value exceeds the immediate publish
+// threshold. Returns false if the threshold is not configured (0).
+func (b *batch) exceedsThreshold() bool {
+	if b.cfg.immediatePublishThreshold == 0 {
+		return false
+	}
+	return b.totalValue() >= b.cfg.immediatePublishThreshold
+}
+
 // Wait waits for the batch to gracefully stop.
 func (b *batch) Wait() {
 	b.Infof("Stopping")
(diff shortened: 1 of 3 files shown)


sweepbatcher/sweep_batch.go
```diff
@@ -807,6 +812,24 @@ func (b *batch) sweepExists(outpoint wire.OutPoint) bool {
...
 }
 
+// totalValue returns the sum of all sweep values in the batch.
+func (b *batch) totalValue() btcutil.Amount {
+	var total btcutil.Amount
+	for _, sweep := range b.sweeps {
+		total += sweep.value
+	}
+	return total
+}
```
starius, line 822: We already have variable `totalSweptAmt` on the call site. We can reuse it instead of adding new method. Or we can keep the method and call it there to find the value of `totalSweptAmt`.

sweepbatcher/sweep_batch.go
```diff
@@ -807,6 +812,24 @@ func (b *batch) sweepExists(outpoint wire.OutPoint) bool {
 	return ok
 }
 
+// totalValue returns the sum of all sweep values in the batch.
```
starius, line 815: Add a note that the method must run in the batch's event loop, so avoid races.

sweepbatcher/sweep_batch.go
```diff
@@ -807,6 +812,24 @@ func (b *batch) sweepExists(outpoint wire.OutPoint) bool {
...
+func (b *batch) totalValue() btcutil.Amount {
+	var total btcutil.Amount
+	for _, sweep := range b.sweeps {
+		total += sweep.value
+	}
+	return total
+}
+
+// exceedsThreshold checks if the batch value exceeds the immediate publish
+// threshold. Returns false if the threshold is not configured (0).
```
starius, line 825: I propose to add the amount as an argument or to add a note that the method must be called from the batch's event loop.

sweepbatcher/sweep_batch.go
```diff
@@ -920,6 +943,25 @@ func (b *batch) Run(ctx context.Context) error {
...
+					b.cfg.immediatePublishThreshold)
+
+				// Set skipBefore to now to bypass initial
+				// delay.
+				now := clock.Now()
+				skipBefore = &now
+
+				// Trigger immediate publish by setting timer to
+				// fire after batchPublishDelay.
+				timerChan = clock.TickAfter(b.cfg.batchPublishDelay)
```
starius, line 963: Simplification idea.

I propose to move the whole branch above, under `if initialDelay < 0 {`
There we already have a couple of situations where we skip waiting. For consistency it is better to have them all in one place. And skipping can also be done by setting `initialDelay = 0`.

I propose also to add an empty line before `delayStop := startTime.Add(initialDelay)`

sweepbatcher/sweep_batcher_test.go
```diff
@@ -5504,3 +5504,235 @@ func runTests(t *testing.T, testFn func(t *testing.T, store testStore,
...
+
+	// Wait for batch publishing to be skipped due to initialDelay.
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		testLogger.mu.Lock()
+		defer testLogger.mu.Unlock()
+
+		assert.Contains(c, testLogger.debugMessages, stillWaitingMsg)
+	}, test.Timeout, eventuallyCheckFrequency)
+
+	// Scenario 2: Add another sweep that pushes total above threshold,
```
starius, line 5646: We also need another test case where the amount is initially below the threshold and then it increases.

sweepbatcher/sweep_batch.go
```diff
@@ -200,6 +200,11 @@ type batchConfig struct {
 	// chainParams are the chain parameters of the chain that is used by
 	// batches.
 	chainParams *chaincfg.Params
+
+	// immediatePublishThreshold is the cumulative value threshold above
+	// which a batch should be published immediately without waiting for
+	// the initial delay grouping period. If 0, the feature is disabled.
+	immediatePublishThreshold btcutil.Amount
```
starius, line 207: IMHO we can implement this inside `initialDelayProvider`. It already has totalSweptAmount argument, so it can have the threshold logic.

PR #1059, opened 2025-12-15
staticaddr: re-register on htlc tx conf error

hieblmi:
We need to re-register for htlc tx confs if the registration fires a conf error in `MonitorInvoiceAndHtlcTxAction`. Otherwise we miss sweeping the htlc timeout-path.

### DIFF
--- a/staticaddr/loopin/actions.go
+++ b/staticaddr/loopin/actions.go
@@ -575,6 +575,10 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
 			f.Errorf("htlc tx conf chan error, re-registering: "+
 				"%v", err)
 
+			// A previous confirmation may no longer be valid if the
+			// subscription failed, so reset and wait for a fresh one.
+			htlcConfirmed = false
+
 			// Re-register for htlc confirmation.
 			htlcConfChan, htlcErrConfChan, err = registerHtlcConf()
 			if err != nil {


staticaddr/loopin/actions.go
```diff
@@ -572,7 +572,15 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
...
-			f.Errorf("htlc tx conf chan error: %v", err)
+			f.Errorf("htlc tx conf chan error, re-registering: "+
+				"%v", err)
+
+			// Re-register for htlc confirmation.
+			htlcConfChan, htlcErrConfChan, err = registerHtlcConf()
+			if err != nil {
+				f.Errorf("unable to monitor htlc tx "+
+					"confirmation: %v", err)
+			}
```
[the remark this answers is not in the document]
hieblmi: fixed

bhandras (approved):
LGTM 🎉

starius:
Thanks! 🏂

I added code coverage and a couple of related fixes here: https://github.com/hieblmi/loop/pull/3 PTAL

staticaddr/loopin/actions.go
```diff
@@ -695,6 +695,8 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
 		case err = <-invoiceErrChan:
 			f.Errorf("invoice subscription error: %v", err)
 
+			return f.HandleError(err)
```
hieblmi: This case occurs if the invoice deadline was reached and the invoice got cancelled. In this case we can't return here, but need to keep monitoring the htlc publication. So I am going to revert this line. @starius

starius: if the invoice subscription (not the invoice!) fails / cancels, we have to re-subscribe, otherwise we won't know that invoice was paid.

This is not critical so let's just add a TODO.

PR #1065, opened 2026-01-09
staticaddr: dynamic deposit conf target

hieblmi:
Update static address loop-in client to support dynamic server-side confirmation requirements.

  Changes:

  - Reduce MinConfs from 6 to 3 to allow faster swap attempts for small amounts while the server enforces risk-based confirmation requirements
  - Update SelectDeposits to prioritize more-confirmed deposits first, increasing the likelihood of server acceptance when dynamic requirements apply
  - Add client-side logging of insufficient confirmation details from server errors, helping users understand when deposits need more confirmations before retrying

  Proto updates:

  - Add DepositConfirmationStatus message with current/required confirmations and blocks to wait
  - Add InsufficientConfirmationsDetails message for structured error responses

  Backwards compatible: Clients with older servers will continue to work normally. The server determines actual confirmation requirements.

### DIFF
--- a/staticaddr/deposit/manager.go
+++ b/staticaddr/deposit/manager.go
@@ -20,7 +20,7 @@ const (
 	// MinConfs is the minimum number of confirmations we require for a
 	// deposit to be considered available for loop-ins, coop-spends and
 	// timeouts.
-	MinConfs = 6
+	MinConfs = 3
 
 	// MaxConfs is unset since we don't require a max number of
 	// confirmations for deposits.
--- a/staticaddr/loopin/actions.go
+++ b/staticaddr/loopin/actions.go
@@ -30,6 +30,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"google.golang.org/grpc/status"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


litbot-9000:

staticaddr/deposit/manager.go
```diff
@@ -23,7 +23,7 @@ const (
 	// MinConfs is the minimum number of confirmations we require for a
 	// deposit to be considered available for loop-ins, coop-spends and
 	// timeouts.
-	MinConfs = 6
+	MinConfs = 3
```
starius, line 23: The PR description says "Reduce MinConfs from 6 to 1", but the commit reduces only to 3.

swapserverrpc/staticaddr.proto
```diff
@@ -279,3 +279,30 @@ message ClientSweeplessSigningInfo {
 
 message PushStaticAddressSweeplessSigsResponse {
 }
+
```
starius, line 282: When we implement the waiting version of dyn-conf, we can drop these RPCs and most of the handling logic on the client side, right? The server will stop returning these errors, so we can clean up the proto files.

starius (approved):
**LGTM!**  💾

staticaddr/deposit/manager.go
hieblmi: Thanks, 1 was originally targeted. Fixed the description.

swapserverrpc/staticaddr.proto
hieblmi: Yes, it is backwards compatible if we remove the proto messages. We can then also remove `logInsufficientConfirmationsDetails(err)`.

PR #1066, opened 2026-01-15
Add a command to manually sweep loop-out HTLC

starius:
Add loopd RPC `SweepHtlc`  and CLI command `loop out sweephtlc`:

```
$ loop out sweephtlc -h
NAME:
   loop out sweephtlc - sweep an HTLC output using the preimage success path

USAGE:
   loop out sweephtlc [options]

OPTIONS:
   --outpoint string  htlc outpoint to sweep (format: txid:vout)
   --htlcaddr string  htlc address corresponding to the outpoint
   --feerate uint     fee rate to use in sat/vbyte
   --destaddr string  optional destination address; defaults to a new wallet address
   --preimage string  optional preimage hex to override stored swap preimage
   --publish          publish the sweep transaction immediately (default: false)
```

#### Pull Request Checklist

### DIFF
--- a/cmd/loop/loopout.go
+++ b/cmd/loop/loopout.go
@@ -121,6 +121,9 @@ var loopOutCommand = &cli.Command{
 		verboseFlag,
 		channelFlag,
 	},
+	Commands: []*cli.Command{
+		sweepHtlcCommand,
+	},
 	Action: loopOut,
 }
 
--- a/cmd/loop/sweephtlc.go
+++ b/cmd/loop/sweephtlc.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"fmt"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/loop/looprpc"
+	"github.com/urfave/cli/v3"
+)
+
+// sweepHtlcCommand exposes HTLC success-path sweeping over loop CLI.
+var sweepHtlcCommand = &cli.Command{
+	Name:  "sweephtlc",
+	Usage: "sweep an HTLC output using the preimage success path",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:     "outpoint",
+			Usage:    "htlc outpoint to sweep (format: txid:vout)",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "htlcaddr",
+			Usage:    "htlc address corresponding to the outpoint",
+			Required: true,
+		},
+		&cli.UintFlag{
+			Name:     "feerate",
+			Usage:    "fee rate to use in sat/vbyte",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name: "destaddr",
+			Usage: "optional destination address; defaults to a " +
+				"new wallet address",
+		},
+		&cli.StringFlag{
+			Name: "preimage",
+			Usage: "optional preimage hex to override stored " +
+				"swap preimage",
+		},
+		&cli.BoolFlag{
+			Name:  "publish",
+			Usage: "publish the sweep transaction immediately",
+			Value: false,
+		},
+	},
+	Hidden: true,
+	Action: sweepHtlc,
+}
+
+// sweepHtlc executes the SweepHtlc RPC and prints the sweep transaction hex.
+func sweepHtlc(ctx context.Context, cmd *cli.Command) error {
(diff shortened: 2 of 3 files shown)


loopd/sweep_htlc.go
```diff
@@ -0,0 +1,399 @@
...
+
+	infof("sweephtlc: start sweep for %v -> %v", req.Outpoint,
+		sweepAddr.EncodeAddress())
+
+	// Locate the loop-out swap whose HTLC script matches the outpoint so
+	// we can obtain keys and the stored preimage.
+	swaps, err := store.FetchLoopOutSwaps(ctx)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
starius: This functionality is not for everyday use, so it is acceptable to do a full scan here.

litbot-9000:
@bhandras: review reminder
@sputn1ck: review reminder
@hieblmi: review reminder

looprpc/client.proto
```diff
@@ -739,6 +745,61 @@ message ListSwapsResponse {
     int64 next_start_time = 2;
 }
 
+// SweepHtlcRequest instructs loopd to sweep a swap HTLC via its success path.
+message SweepHtlcRequest {
+    // Optional override for the sweep destination; defaults to a new address
```
hieblmi, line 750: we should make the field itself optional like `optional string dest_address = 1;`, same for `preimage`.

utils/fees.go
```diff
@@ -0,0 +1,44 @@
...
+	clamped := false
+	if fee > maxFeeAmount {
+		clampedFee = maxFeeAmount
+		clamped = true
+	}
+
+	if minRelayFeeRate > 0 {
+		clampedFeeRate := chainfee.NewSatPerKWeight(clampedFee, weight)
+		if clampedFeeRate < minRelayFeeRate {
+			return 0, clamped, fmt.Errorf("clamped fee rate %v is "+
```
hieblmi, line 37: just wondering if we should return the right fee that would satisfy the min relay fee?

loopd/sweep_htlc.go
```diff
@@ -0,0 +1,399 @@
...
+	select {
+	case conf := <-confChan:
+		fundingTx = conf.Tx
+		infof("sweephtlc: funding confirmed at height %v",
+			conf.BlockHeight)
+
+	case ntfnErr := <-errChan:
+		infof("sweephtlc: conf ntfn error for %v: %v",
+			req.Outpoint, ntfnErr)
+		return nil, status.Errorf(codes.Internal,
```
hieblmi, line 213: i think we should add spaces before the returns

loopd/sweep_htlc_test.go
```diff
@@ -0,0 +1,586 @@
...
+			"sweephtlc: registering conf ntfn for %v hint=%v",
+			"sweephtlc: waiting for confirmation of %v",
+			"sweephtlc: funding confirmed at height %v",
+			"sweephtlc: swap hash validated for %v",
+		},
+	},
+}
+
+// ptrFee is a small helper to take a fee literal by address for table cases.
+func ptrFee(f chainfee.SatPerKWeight) *chainfee.SatPerKWeight {
```
hieblmi: is that really needed?

hieblmi (approved):
Great work @starius, I just have a couple of nits.

You can also decide whether it would make sense to add an integration test for it.

looprpc/client.proto
starius: In proto3 all the fields are optional, so we don't have to mark them explicitly as `optional`.

loopd/sweep_htlc.go
starius: Fixed

loopd/sweep_htlc_test.go
starius: Removed it. I use zero value instead of 0.

utils/fees.go
starius: Previous behavior was to fail, this commit is just a refactoring.

We should consider this change separately. On the one hand, the current behavior protects from mistakes leading to funds burning (too high fee). On the other hand, what if fees spike and become more than 20% of some relatively small coin, then we can't sweep it. So it is a trade-off. I think the current state is preferable (burning a lot of funds in a huge fee sounds a worse and more real risk to me, than failing to sweep a relatively small UTXO).

starius:
> You can also decide whether it would make sense to add an integration test for it.

I added an integration test, it works.

PR #1072, opened 2026-02-05
cmd/loop: CLI session recording and replay tests

starius:
This work adds a repeatable, high-level testing path for the Loop CLI by recording real command runs and replaying them in tests. The goal is to make CLI behavior verifiable and regression-resistant without depending on a live loopd or local environment details. The recorded sessions become fixtures that exercise success and error paths across subcommands and flags, while replay ensures deterministic output checks in CI.

## What changed

- Added a session recorder that captures CLI stdin/stdout/stderr and gRPC traffic when `LOOP_SESSION_RECORD=true` is set.
- Wired a pluggable gRPC transport, stdio hooks, and a test clock so recording and replay use the same code paths as normal CLI execution.
- Normalized help output to show home paths as `~` and normalized JSON output and timestamps during replay so sessions are portable across machines.
- Added replay tests that load fixtures, inject recorded env/stdin, replay the recorded gRPC stream, and compare outputs.
- Recorded a broad set of CLI scenarios into `cmd/loop/testdata/sessions/` and documented how to add more.

## Why this approach

The CLI previously relied on manual testing or live daemons, which makes it hard to detect regressions and impossible to run in CI without complex setup. Recording/replay gives us deterministic, reproducible coverage of real behavior, including error messaging, output formatting, and gRPC request/response handling, while still exercising the production code paths.

## How it works

Recording wraps the CLI process and intercepts stdio and gRPC calls, saving a JSON session file with events and metadata. Replay loads these events, supplies recorded input, and provides a recorded gRPC connection so commands run without a live daemon. Output is normalized where needed so sessions can be replayed across machines.

## Testing

- `go test ./cmd/loop -v`
- To record a session: `LOOP_SESSION_RECORD=true loop <command> --network regtest`

#### Pull Request Checklist


cmd/loop/session_replay_test.go
```diff
@@ -0,0 +1,1139 @@
...
+	}
+
+	return nil
+}
+
+// consume fetches the next matching event with locking.
+func (c *recordedClientConn) consume(method, expected string,
+	alternatives ...string) (*grpcPayload, int, error) {
+
+	return c.consumeLocked(method, expected, alternatives...)
```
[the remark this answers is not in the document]
starius: Fixed

starius:
Rebased, add recorded sessions for newly added commands and flags (the last commit).

cmd/loop/session_recorder.go
```diff
@@ -0,0 +1,625 @@
...
+		msg := runErr.Error()
+		r.metadata.RunError = &msg
+	} else {
+		r.metadata.RunError = nil
+	}
+	r.metadata.Duration = &duration
+}
+
+// finalize writes the recorded session to disk once.
+func (r *sessionRecorder) finalize(runErr error) error {
```
kaldun-tech, line 339: Minor: consider writing session even if stopHooks fails, to preserve debugging data

From Claude, the realistic risk is low. stopHooks() just calls unhook functions that restore original file descriptors. Failures here would be unusual (maybe a closed pipe, or dup2 failure). In normal operation this won't happen.

It's suggestion is:
```
r.finalizeOnce.Do(func() {
    // Still try to stop hooks, but don't abandon ship on error
    hookErr := r.stopHooks()

    r.logExit(runErr)

    // ... write session to disk ...
    // Return hook error only if write succeeded
    if finalizeErr == nil {
        finalizeErr = hookErr
    }
})
```

cmd/loop/session_replay_test.go
```diff
@@ -0,0 +1,1120 @@
...
+	// Allocate a new struct value and copy exported fields.
+	cloned := reflect.New(value.Elem().Type())
+	copyExportedFields(cloned.Elem(), value.Elem())
+
+	return cloned.Interface(), true
+}
+
+// copyExportedFields copies exported fields from src into dst.
+func copyExportedFields(dst, src reflect.Value) {
+	// Iterate fields and copy only exported ones.
```
kaldun-tech, line 892: Moderate concern here: This copies only exported fields from cli.Command, cli.Flag, and cli.Argument structs. Here's the points where it's fragile:

1. Unexported state is silently ignored

If urfave/cli stores important state in unexported fields, the clone won't have it. For example, if
cli.StringFlag had:
```
type StringFlag struct {
    Name  string      // exported - copied
    value string      // unexported - NOT copied
    isSet bool        // unexported - NOT copied
}
```

The test at line 1071 (require.False(t, clonedAlpha.IsSet())) actually relies on this behavior - it expects
IsSet() to return false because the unexported state isn't copied. So currently this works in your favor.

2. Structural assumptions about cli.Command

The code explicitly handles these fields (line 688-692):
- Flags
- MutuallyExclusiveFlags
- Arguments
- Commands

If urfave/cli v3.5 adds a new field like ConditionalFlags []cli.Flag, the cloning code won't deep-clone it - it'll share references with the original.

3. Version pinning mitigates but doesn't eliminate risk

The repo is on urfave/cli v3.4.1. When you upgrade, the test TestCloneCommandForReplayResetsFlagState should catch obvious breaks, but it doesn't comprehensively cover all cli.Command fields.

Suggestions:
1. Add a comment documenting the urfave/cli version dependency:
```
// cloneCommandForReplay deep-clones a command tree for deterministic replays.
// NOTE: This relies on urfave/cli/v3 internal structure. If upgrading cli
// versions, verify TestCloneCommandForReplayResetsFlagState still passes.
```
2. Consider a simpler alternative: Since this is test-only, you could just call newRootCommand() fresh for each test instead of cloning. Is cloning is actually necessary?

cmd/loop/session_stdio.go
```diff
@@ -0,0 +1,138 @@
...
+	setDest(w)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer r.Close()
+
+		writer := composeWriter(forward, onChunk)
+		_, _ = io.Copy(writer, r)
```
kaldun-tech, line 51: Minor/Nit: You have a lot of error swallowing in this file. Consider logging io.Copy errors in session_stdio.go for debugging incomplete recordings, or documenting why they're intentionally ignored.

Claude suggests:
```
go func() {
    defer wg.Done()
    defer r.Close()

    writer := composeWriter(forward, onChunk)
    if _, err := io.Copy(writer, r); err != nil {
        // At minimum, log it for debugging
        log.Debugf("session stdout copy error: %v", err)
    }
}()
```

Or track errors in a struct field and surface them in finalize():

```
type sessionRecorder struct {
    // ...
    copyErr error  // Set if io.Copy fails
}
```

kaldun-tech:
My biggest question about the design for fixture maintenance. With 114 fixture files, what's the plan for bulk updates when CLI output format changes? A few things that would help long-term:

1. A make record-sessions target, even if it requires manual regtest setup
2. An update mode for tests in session_replay_test.go (UPDATE_FIXTURES=true go test ...) to regenerate expected output
3. Consider normalizing metadata.version in comparisons so releases don't invalidate all fixtures

I don't think it's a blocker for this test update. Just want to understand the maintenance story before this grows further. Otherwise seems solid!

cmd/loop/session_replay_test.go
```diff
@@ -0,0 +1,1120 @@
...
+}
+
+// newRootCommandForReplay returns a root command clone with fresh flag state.
+func newRootCommandForReplay() *cli.Command {
+	// Clone the root command tree to avoid shared flag state.
+	return cloneCommandForReplay(newRootCommand())
+}
+
+// cloneCommandForReplay deep-clones a command tree for deterministic replays.
+func cloneCommandForReplay(cmd *cli.Command) *cli.Command {
```
kaldun-tech, line 680: I'll be more concise on this method. Do you need to clone at all vs calling newRootCommand()? If there is a reason cloning is needed I recommend to add a comment to document this.

cmd/loop/main.go
```diff
@@ -92,8 +102,51 @@ var (
...
+
+	// cliClock provides the time source used by CLI commands.
+	cliClock clock.Clock = clock.NewDefaultClock()
+
+	// sessionRec is the active recorder when session capture is enabled.
+	sessionRec *sessionRecorder
+
+	// forceDeterministicJSON is enabled by tests to obtain stable JSON
+	// output.
+	forceDeterministicJSON bool
```
kaldun-tech, line 114: Could these package-level variables be encapsulated in a context or config struct? Looks fine for a CLI. May complicate future refactorings.

cmd/loop/session_replay_test.go
kaldun-tech: I figured out why cloning is necessary. The package-level command pointers (loopOutCommand, etc.) have mutable flag state that would bleed otherwise between test runs. Recommend to add a comment explaining this, something like:

```
// cloneCommandForReplay deep-clones a command tree because the CLI uses
// package-level command pointers with mutable flag state. Without cloning,
// flag state from one replay would bleed into subsequent replays.
//
// NOTE: This uses reflection to copy exported fields only. If upgrading
// urfave/cli versions, verify TestCloneCommandForReplayResetsFlagState
// still passes.
func cloneCommandForReplay(cmd *cli.Command) *cli.Command {
```

cmd/loop/session_replay_test.go
```diff
@@ -0,0 +1,1120 @@
...
+	for _, value := range values {
+		if value == v {
+			return true
+		}
+	}
+
+	return false
+}
+
+// TestRecordedSessions replays all recorded sessions and compares output.
```
kaldun-tech, line 533: Test parallelism safety (minor): The tests modify package globals (sessionRec, sessionTransport, cliClock, forceDeterministicJSON) without synchronization. This is fine since tests run sequentially. A comment could prevent future breakage:

```
// TestRecordedSessions replays all recorded sessions and compares output.
// NOTE: Do not add t.Parallel() - this test modifies package-level globals.
func TestRecordedSessions(t *testing.T) {
```

litbot-9000:
@bhandras: review reminder
@hieblmi: review reminder

starius:
> My biggest question about the design for fixture maintenance. With 114 fixture files, what's the plan for bulk updates when CLI output format changes? A few things that would help long-term:
> 
>     1. A make record-sessions target, even if it requires manual regtest setup
> 
>     2. An update mode for tests in session_replay_test.go (UPDATE_FIXTURES=true go test ...) to regenerate expected output
> 
>     3. Consider normalizing metadata.version in comparisons so releases don't invalidate all fixtures
> 
> 
> I don't think it's a blocker for this test update. Just want to understand the maintenance story before this grows further. Otherwise seems solid!

Thanks for suggestion!

I implemented re-generation of all fixtures when only stdout and stderr changed (see commit "cmd/loop: bless text-only session updates", env var LOOP_UPDATE_RECORDED_SESSIONS). If a broader change was deployed, it makes sense to regenerate the fixtures. Fortunately AI is capable of doing it.

I also changed the format of fixtures to have multi-line stdout and stderr outputs actually stored as multiline (`[]string` in JSON). Then it is easier to spot the changes than when it is all in one long line with \n inside.

I pushed new commits instead of squashing to make it easier to follow the changes. I'll rebase the changes later. Also I'm aware that the CI is broken now - I'll fix it later as well.

cmd/loop/session_replay_test.go
starius: Handled in `cmd/loop: note replay test stays serial`. I added an explicit note above `TestRecordedSessions` that it must not use `t.Parallel()` because the replay harness swaps package-level globals for transport, clock, recorder state, and JSON normalization.

cmd/loop/main.go
starius: I left the package-level CLI hooks as-is in this series. They are scoped to the CLI runtime and the record/replay harness, and moving them behind a config or context object would turn this into a broader refactor than this PR needs.

cmd/loop/session_replay_test.go
starius: Handled in `cmd/loop: document replay cloning`. I kept cloning and added the missing explanation: the CLI keeps package-level command pointers with mutable flag state, so reusing one root would leak `IsSet` and related state across replayed sessions. The comment also calls out that the clone logic depends on the current `urfave/cli` structure and points future upgrades at `TestCloneCommandForReplayResetsFlagState`.

cmd/loop/session_stdio.go
starius: Handled in `cmd/loop: clarify ignored stdio errors`. I kept the current behavior but documented why those background `io.Copy` errors are intentionally ignored. During hook restoration the pipe endpoints are closed on purpose, and that can terminate the copy loops with an expected error. Logging those teardown paths would mostly add noise rather than improve diagnosis.

cmd/loop/session_replay_test.go
starius: Handled in `cmd/loop: document replay cloning`. I kept cloning and added the missing explanation: the CLI keeps package-level command pointers with mutable flag state, so reusing one root would leak `IsSet` and related state across replayed sessions. The comment also calls out that the clone logic depends on the current `urfave/cli` structure and points future upgrades at `TestCloneCommandForReplayResetsFlagState`.

cmd/loop/session_recorder.go
starius: Handled in `cmd/loop: keep sessions on hook errors`. Finalization now still records the exit state and writes the session file even if hook cleanup returns an error, and only reports the cleanup error after the file has been persisted. That keeps the debugging artifact without hiding an unusual teardown failure.

starius:
Rebased - that was the only way to fix CI, because it runs against a merged state with master, and there was a change in CLI texts in master recently. Fixed that in "fixup! testdata: add static address and swaps sessions"

kaldun-tech:
Cool, I'm wrapping up my review. This is a really excellent PR with a lot of strong changes including the deterministic replay and bless mode safety. Have a couple more non-blocking suggestions.

cmd/loop/session_stdio.go
```diff
@@ -0,0 +1,138 @@
...
+		_, _ = io.Copy(w, source)
+	})
+
+	return func() error {
+		os.Stdin = orig
+		_ = r.Close()
+		_ = w.Close()
+		if !useOrig {
+			wg.Wait()
+		}
```
kaldun-tech, line 106: Minor: I'd like to add a comment here to explain the intentional behavior of the goroutine, that it skips waiting for copy. For example:

`
// When reading from the original terminal stdin, we skip waiting
// for the copy goroutine. It blocks on terminal input indefinitely
// and will terminate when the pipe closes or the process exits.
// Waiting here would hang the CLI.
`

cmd/loop/session_recorder.go
```diff
@@ -0,0 +1,657 @@
...
+
+	return sessionDefaultDir, name, nil
+}
+
+// logEvent records a new event with the elapsed timestamp.
+func (r *sessionRecorder) logEvent(kind string, payload any) {
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return
+	}
```
kaldun-tech, line 275: Minor: consider logging marshal errors here to stderr

kaldun-tech (approved):
A couple other things I checked here:
- The testdata fixtures do not contain sensitive data. 
- Did not find any concerns with race conditions in session_recorder
- We could add a Makefile target for recording sessions in a follow-up PR, but this would only wrap a one-liner already documented in AGENTS.md. Don't think it's needed.
- Bless mode is great!

cmd/loop/session_stdio.go
starius: Thanks! I added the comment.

cmd/loop/session_recorder.go
starius: Logging to stderr may be problematic here, since stderr may be hooked.

I fixed it by saving the error and checking it from Finalize() method.

starius:
@kaldun-tech Thanks for review! I addressed the comments and rebased the PR, to make atomic Git commits.

hieblmi:
@claude please review this

starius:
> **1. Silent behavioral regression in abandonSwap (swaps.go)**
>
> The removal of the `cmd.IsSet("id")` case silently changes argument priority. Previously, `--id <value>` took precedence over a positional argument when both were supplied. After this change, a positional argument always wins (since the flag case is gone and only `cmd.NArg() > 0` remains). This could break users who rely on `--id`.
>
> If this is intentional (e.g., the flag is being deprecated in favour of positional args), it should be documented and probably also enforced with a runtime warning when users still pass `--id`.

This one is not a supported CLI behavior change. `loop abandonswap` does not define an `--id` flag, so the removed branch was orphan flag access rather than a user-visible flag priority rule. `loop swapinfo` has an `--id` flag and keeps the existing `--id` precedence; `loop abandonswap` only accepts the positional `ID` plus `--i_know_what_i_am_doing`.

Relevant current shape:

```go
var abandonSwapCommand = &cli.Command{
	Name:      "abandonswap",
	ArgsUsage: "ID",
	Flags: []cli.Flag{
		&cli.BoolFlag{Name: "i_know_what_i_am_doing"},
	},
	Action: abandonSwap,
}
```

I left the code as-is here because reintroducing the branch would keep dead code around for a flag the command cannot parse.

> **2. Hook error silently dropped when file write fails in finalize**
>
> Inside `finalizeOnce.Do`, the function collects `hookErr := r.stopHooks()` early on, but if `os.Create` or `encoder.Encode` fail, the function returns early and `hookErr` is never surfaced. Consider using `errors.Join(err, hookErr)` in those early-return paths so hook errors are not silently lost.

Fixed in `cmd/loop: add session recorder plumbing`. `finalize` now joins write errors with deferred recorder event errors and hook cleanup errors, including the early-return paths for directory creation, file creation, and JSON encoding.

Relevant part:

```go
file, err := os.Create(r.filePath)
if err != nil {
	finalizeErr = errors.Join(err, eventErr, hookErr)
	return
}

encoder := json.NewEncoder(file)
encoder.SetIndent("", "  ")
if err := encoder.Encode(fileContent); err != nil {
	finalizeErr = errors.Join(err, eventErr, hookErr)
	return
}
```

I also updated the finalize test so it asserts both the write failure and hook failure are returned:

```go
err := recorder.Finalize(nil)
require.ErrorContains(t, err, "is a directory")
require.ErrorContains(t, err, "stdout restore failed")
```

> **3. Package-level global state makes tests non-parallelisable**
>
> `sessionRec`, `cliClock`, `sessionTransport`, and `forceDeterministicJSON` are all package-level globals mutated by the test harness. The note in `TestRecordedSessions` acknowledges this. This is understandable given the CLI's design, but it means the entire `session_replay_test.go` suite is inherently sequential. A potential follow-up: wrapping these in a `cliEnv` struct passed through context could eventually allow parallelism.

Leaving this as a follow-up. The replay test already documents that it must stay serial because the CLI uses package-level command/runtime state. Moving this into a `cliEnv` would be a larger CLI plumbing refactor and is not necessary for this PR’s test feature.

Current guardrail:

```go
// NOTE: Do not add t.Parallel() here; the replay harness mutates package-level
// globals such as the active transport, clock, and JSON normalization mode.
func TestRecordedSessions(t *testing.T) {
```

> **4. Reflection-based command cloning is fragile against urfave/cli upgrades**
>
> `cloneCommandForReplay` copies exported fields via reflection and relies on urfave/cli storing all replay-relevant state in exported fields. The existing `TestCloneCommandForReplayResetsFlagState` provides a safety net, but only covers fields explicitly set in that test. Consider filing an upstream issue requesting a `Clone()` method on `cli.Command` for a more stable interface.

Leaving this unchanged in the PR. The code now has an explicit comment explaining why replay clones the command tree and what urfave/cli upgrade risk to watch for, and the regression test covers the mutable flag/argument state that caused the issue here.

Relevant comment:

```go
// NOTE: This relies on urfave/cli storing replay-relevant configuration in
// exported fields. If upgrading urfave/cli, verify that
// TestCloneCommandForReplayResetsFlagState still passes.
func cloneCommandForReplay(cmd *cli.Command) *cli.Command {
```

An upstream `Clone()` API would be cleaner, but that is outside this PR and would not remove the need for the local safety test.

> **5. sessionDefaultDir is a hardcoded relative path**
>
> `sessionDefaultDir = "cmd/loop/testdata/sessions"` only works when the binary is run from the repository root. This is documented in the new AGENTS.md, but a runtime check or warning when the directory does not exist would be more user-friendly.

Fixed in `cmd/loop: add session recorder plumbing`. When session recording is enabled, the recorder now verifies that `cmd/loop/testdata/sessions` exists and is a directory before choosing the output file. If it is missing, recording fails with a clear error telling the caller to run from the repository root instead of silently creating a new tree in the wrong directory.

Relevant code:

```go
func ensureSessionBaseDir(baseDir string) error {
	info, err := os.Stat(baseDir)
	if errors.Is(err, fs.ErrNotExist) {
		return fmt.Errorf("%s does not exist; run session recording "+
			"from the repository root", baseDir)
	}
	if err != nil {
		return err
	}
	if !info.IsDir() {
		return fmt.Errorf("%s is not a directory", baseDir)
	}

	return nil
}
```

I added a unit test for that failure mode:

```go
t.Chdir(t.TempDir())
t.Setenv(sessionEnvVar, "true")

recorder, err := newSessionRecorder([]string{testLoopArg, testOutArg})
require.Nil(t, recorder)
require.EqualError(t, err, "cmd/loop/testdata/sessions does not "+
	"exist; run session recording from the repository root")
```

> **6. Redundant contains wrapper**
>
> The `contains(values []string, v string) bool` function only wraps `slices.Contains` and is called in just two places. Call `slices.Contains` directly and remove the wrapper.

Fixed in `cmd/loop: add session replay tests`. The wrapper is gone and the call site now uses `slices.Contains` directly.

```go
if evt.Event == expected || slices.Contains(alternatives, evt.Event) {
	return &evt, idx, nil
}
```

> **7. Five separate sessionRec nil checks in main()**
>
> The recording setup is scattered across five separate `if sessionRec != nil` blocks. Consider consolidating into a single helper like `installSessionRecorder()` that handles clock, transport, context injection, and signal handler in one place, keeping the normal code path cleaner.

Leaving this unchanged. The separate checks are a little repetitive, but they keep each installed hook close to its lifetime/defer and avoid introducing a broader setup abstraction in this PR. Since all blocks are guarded by `sessionRec != nil`, the normal CLI path remains straightforward and avoids recorder work entirely.

Representative shape:

```go
if sessionRec != nil {
	restoreClock := hookClock(
		clock.NewTestClock(time.Unix(sessionClockStartUnix, 0)),
	)
	defer restoreClock()

	if err := sessionRec.Start(nil, nil, nil); err != nil {
		fatal(err)
	}
}
```

> **8. Double-signal os.Exit bypasses deferred cleanup**
>
> In `installSessionSignalHandler`, a second signal calls `os.Exit(130)` from a goroutine, bypassing all deferred functions in `main()`. This is acceptable for a signal handler since `sessionRec.Finalize` is correctly called before exit, but a comment noting that deferred cleanup will not run would help future maintainers.

Fixed in `cmd/loop: add session recorder plumbing`. I added a short comment before the second-signal `os.Exit(130)` path explaining that the session is finalized first because `os.Exit` skips deferred cleanup.

```go
// A second signal requests immediate shutdown. Finalize the
// session first because os.Exit skips deferred cleanup.
_ = sessionRec.Finalize(fmt.Errorf("signal: %s", sig))
os.Exit(130)
```

> **9. TestUpdateRecordedSessionsEnabled env var setup**
>
> The test saves the existing env var value via `t.Setenv` and then immediately clears it with `os.Unsetenv`. The `t.Setenv` to an empty string is redundant before the immediate manual unset and can be simplified.

Fixed in `cmd/loop: bless text-only session updates`. The test setup is simplified and lint-clean. It still covers the unset, true, and invalid cases.

```go
if current, ok := os.LookupEnv(updateRecordedSessionsEnvVar); ok {
	t.Setenv(updateRecordedSessionsEnvVar, current)
} else {
	t.Setenv(updateRecordedSessionsEnvVar, "false")
}
_ = os.Unsetenv(updateRecordedSessionsEnvVar)

enabled, err := updateRecordedSessionsEnabled()
require.NoError(t, err)
require.False(t, enabled)
```

hieblmi (commented):
Great work @starius, I've found a coupe of issues:

1. cmd/loop/session_replay_test.go:689: normal replay mode does not assert that all recorded gRPC events were consumed. That check only runs in bless mode, so CI can miss regressions where a command stops making a recorded trailing RPC but the visible text still matches.
  2. cmd/loop/main.go:263: live recording replaces the CLI clock with a fixed timestamp. Since that timestamp is January 26, 2026, future loop out / quote out recordings
     send stale swap_publication_deadline values via cmd/loop/loopout.go:251 and cmd/loop/quote.go:209, which can diverge from real CLI behavior or be rejected by a live server.

starius:
@hieblmi Great catches! I fixed both review comments:
1. Normal replay now asserts that the recorded gRPC stream was fully consumed, so CI catches trailing-RPC regressions even when the visible CLI output still matches.
2. Live session recording now captures a per-session `clock_start_unix` and replays against that recorded timestamp, so new `loop out` and `quote out` recordings keep sending realistic publication deadlines while replay stays deterministic. All fixtures added in the PR get this field filled.

hieblmi (approved):
There is one more issue found:

  - cmd/loop/session_replay_test.go:352 replays recorded gRPC errors as plain errors.New(...), losing the original gRPC status code. Code like cmd/loop/liquidity.go:712 uses status.FromError to branch on codes.FailedPrecondition, so replay can exercise a different path than the real CLI. The suggestswaps fixture already shows this risk: cmd/loop/testdata/sessions/liquidity/07_loop-suggestswaps.json:13 expects the raw RPC error instead of the friendly CLI error.

Maybe separately addressed in the future could be a Linux specific path encoding, which fails on other platforms:

- go test ./cmd/loop fails locally on darwin because help fixtures hard-code Linux default paths like ~/.loop: cmd/loop/testdata/sessions/static-loop-in/14_loop-static-
     in.json:141, cmd/loop/testdata/sessions/static-openchannel/01_loop-static-openchannel-help.json:86. The current code derives default help text from platform-specific paths at package init: cmd/loop/main.go:57. CI is green on Linux, but the test suite is not portable across supported developer platforms.

starius:
@hieblmi I addressed the issues found:

 - Recorded gRPC errors now preserve structured status codes/messages, and replay reconstructs them as gRPC status errors.
  - liquidity/07_loop-suggestswaps.json now expects the friendly CLI error.
  - Recorded-session replay pins help default paths to fixture values, avoiding platform-specific path diffs on Darwin.

Could you check that the tests pass on Darwin, please?

hieblmi:
Thanks @starius, I tested two runs on Darwin successfully

```
 ~/projects/go/src/github.com/loop  cli-tests20  go test ./cmd/loop -run 'TestRecordedSessions/liquidity/07_loop-suggestswaps.json$' -v -count=1
=== RUN   TestRecordedSessions
=== RUN   TestRecordedSessions/liquidity/07_loop-suggestswaps.json
--- PASS: TestRecordedSessions (0.00s)
    --- PASS: TestRecordedSessions/liquidity/07_loop-suggestswaps.json (0.00s)
PASS
ok      github.com/lightninglabs/loop/cmd/loop  0.821s
 ~/projects/go/src/github.com/loop  cli-tests20  go test ./cmd/loop -run 'Test(LogGRPCMessageRecordsStatus|ReplayGRPCErrorPreservesStatusCode|HookRecordedHelpDefaults|DefaultPathText)$' -v
  -count=1
=== RUN   TestDefaultPathText
=== RUN   TestDefaultPathText/empty_value
=== RUN   TestDefaultPathText/nil_homedir_func
=== RUN   TestDefaultPathText/homedir_error
=== RUN   TestDefaultPathText/exact_home
=== RUN   TestDefaultPathText/home_prefix
=== RUN   TestDefaultPathText/non-home_path
=== RUN   TestDefaultPathText/prefix_but_not_path_segment
--- PASS: TestDefaultPathText (0.00s)
    --- PASS: TestDefaultPathText/empty_value (0.00s)
    --- PASS: TestDefaultPathText/nil_homedir_func (0.00s)
    --- PASS: TestDefaultPathText/homedir_error (0.00s)
    --- PASS: TestDefaultPathText/exact_home (0.00s)
    --- PASS: TestDefaultPathText/home_prefix (0.00s)
    --- PASS: TestDefaultPathText/non-home_path (0.00s)
    --- PASS: TestDefaultPathText/prefix_but_not_path_segment (0.00s)
=== RUN   TestLogGRPCMessageRecordsStatus
--- PASS: TestLogGRPCMessageRecordsStatus (0.00s)
=== RUN   TestReplayGRPCErrorPreservesStatusCode
=== PAUSE TestReplayGRPCErrorPreservesStatusCode
=== RUN   TestHookRecordedHelpDefaults
--- PASS: TestHookRecordedHelpDefaults (0.00s)
=== CONT  TestReplayGRPCErrorPreservesStatusCode
--- PASS: TestReplayGRPCErrorPreservesStatusCode (0.00s)
PASS
```

hieblmi (approved):
LGTM!

PR #1073, opened 2026-02-25
looprpc: preparatory commits for static open channel

hieblmi:
looprpc: prepare for lnrpc usage

In successive commits we want to access lnrpc specifc methods
and objects, so here we import github.com/lightningnetwork/lnd v0.20.1-beta
into looprpc and update relevant versions.

### DIFF
--- a/looprpc/perms.go
+++ b/looprpc/perms.go
@@ -136,6 +136,13 @@ var RequiredPermissions = map[string][]bakery.Op{
 		Entity: "loop",
 		Action: "in",
 	}},
+	"/looprpc.SwapClient/StaticOpenChannel": {{
+		Entity: "deposit",
+		Action: "spend",
+	}, {
+		Entity: "loop",
+		Action: "in",
+	}},
 	"/looprpc.SwapClient/GetLsatTokens": {{
 		Entity: "auth",
 		Action: "read",


looprpc/gen_protos_docker.sh
```diff
@@ -5,9 +5,13 @@ set -e
 # Directory of the script file, independent of where it's called from.
 DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
 
+# Get versions from looprpc go.mod
 PROTOBUF_VERSION=$(go list -f '{{.Version}}' -m google.golang.org/protobuf)
 GRPC_GATEWAY_VERSION=$(go list -f '{{.Version}}' -m github.com/grpc-ecosystem/grpc-gateway/v2)
 
+# Get lnd directory from parent go.mod to use the correct version
+lnd_dir=$(cd .. && go list -f '{{.Dir}}' -m github.com/lightningnetwork/lnd)
```
bhandras, line 13: Should we print out `lnd_dir`? Could help to reassure we're indeed using the right one.

looprpc/client.proto
```diff
@@ -206,13 +207,34 @@ service SwapClient {
...
     StaticAddressLoopIn initiates a static address loop-in swap.
     */
     rpc StaticAddressLoopIn (StaticAddressLoopInRequest)
         returns (StaticAddressLoopInResponse);
+
+    /* loop:`static openchannel`
+    StaticOpenChannel opens a channel funded by selected static address
+    deposits.
+    */
+    rpc StaticOpenChannel (StaticOpenChannelRequest)
```
bhandras, line 220: I don't think the commit message is correct, we're extending the proto, not just using this type.

looprpc/client.proto
```diff
@@ -1810,23 +1832,6 @@ message WithdrawDepositsResponse {
     string address = 2;
 }
 
-message OutPoint {
```
bhandras, line 1813: Maybe also mention in the commit message that the replace is safe as the lnrpc types are the same.

hieblmi: good point

looprpc/client.proto
```diff
@@ -1,6 +1,7 @@
 syntax = "proto3";
 
 import "swapserverrpc/common.proto";
+import "lnrpc/lightning.proto";
```
starius, line 4: I propose to squash commits "looprpc: use lnrpc.OutPoint and lnrpc.OpenChannelRequest" and "loop: use lnrpc.OutPoint instead of looprpc.OutPoint". Otherwise commit "looprpc: use lnrpc.OutPoint and lnrpc.OpenChannelRequest" does not compile.

looprpc/perms.go
```diff
@@ -131,7 +131,14 @@ var RequiredPermissions = map[string][]bakery.Op{
...
 		Entity: "swap",
-		Action: "read",
+		Action: "execute",
+	}, {
+		Entity: "loop",
+		Action: "in",
+	}},
+	"/looprpc.SwapClient/StaticOpenChannel": {{
+		Entity: "swap",
+		Action: "execute",
```
starius: StaticOpenChannel and Withdrawal are not swaps. Do we need swap:execute permission for them? Can we make another permission to spend a deposit?

notifications/manager_test.go
```diff
@@ -133,11 +133,12 @@ func TestManager_ReservationNotification(t *testing.T) {
...
 
-	// Wait a bit to ensure manager is running and has subscribed
+	// Wait a bit to ensure manager is running and has subscribed to the
+	// server notifications stream.
 	require.Eventually(t, func() bool {
-		mgr.Lock()
-		defer mgr.Unlock()
-		return len(mgr.subscribers[NotificationTypeReservation]) > 0
+		mockClient.Lock()
+		defer mockClient.Unlock()
```
starius, line 140: Add an empty line before `return`

looprpc/client.proto
hieblmi: fixed

notifications/manager_test.go
hieblmi: fixed

looprpc/perms.go
hieblmi: good point. We could use `deposit`/`spend`. But unfortunately adding new caveats to existing calls breaks old clients since they need all listed permissions, so they'd have to re-bake their macaroons.

So I'll leave the current permissions untouched, but add a new deposit/spend caveat for the open channel rpc.

looprpc/client.proto
hieblmi: That's a good clarification, I fixed the comment.

looprpc/gen_protos_docker.sh
hieblmi: Done!

PR #1078, opened 2026-02-28
staticaddr: various fixes

hieblmi:
A ~small~ collection of fixes for issues that agents identified.

  1. Fixes inconsistency where createHtlcSweepTx tried to dynamically find the HTLC output index but then unconditionally read index 0, making the search dead code. Asserts the index-0 invariant explicitly.               
  2. Fixes a busy-loop where SweepHtlcTimeoutAction logged context cancellation but continued retrying instead of returning.
  3. Fixes a data race where recoverLoopIns wrote to a shared map from a goroutine without synchronization.                                                                                                                  
  4. Adds missing error check so the function returns early instead of spawning a goroutine that blocks forever on a nil channel.
  5. Fixes garbled log output caused by using %w (error wrapping verb) in log.Errorf where only fmt.Errorf supports it.                                                                                                      
  6. Adds error handling for RegisterConfirmationsNtfn so failures don't silently proceed to read from nil channels.
  7. Fixes incorrect/stale godoc comments from copy-paste across multiple packages.
  8. Fixes data races on SendEvent calls during deposit/loopin manager startup.
  9. Removes unused errChan fields, stale interfaces, unused config fields, and dead map declarations across loopin, openchannel, withdraw, deposit, and address managers.
  10. Uses context.WithoutCancel for shim cleanup so the cancellation RPC doesn't fail when the original request context is already canceled.
  11. Same pattern — uses a live context for ReleaseRoutingPlugin teardown so deferred cleanup doesn't fail with a canceled context.
  12. Same pattern — uses a live context to cancel probe invoices so cleanup succeeds after the monitoring context is canceled.
  13. Bumps golangci-lint to a newer version.
  14. Fixes all lint violations surfaced by the updated linter.
  15. Updates ansi and related deps (cellbuf, colorprofile, term) for API compatibility.
  16. Fixes test goroutine leak by sending batch confirmation after AssertRegisterConf and using a cancellable context.
  17. Adds a replace directive to redirect lukechampine.com/uint128 to its GitHub mirror.
  18. Aligns looprpc/go.mod dependency versions and adds the golang-migrate replace directive.
  19. Adds a replace directive to redirect git.sr.ht/~sbinet/gg to its GitHub mirror.


staticaddr/loopin/loopin.go
```diff
@@ -402,7 +402,7 @@ func (l *StaticAddressLoopIn) createHtlcSweepTx(ctx context.Context,
 
 	fee := feeRate.FeeForWeight(weightEstimator.Weight())
 
-	htlcOutValue := htlcTx.TxOut[0].Value
+	htlcOutValue := htlcTx.TxOut[htlcInputIndex].Value
```
starius, line 407: Please add a regression test, so this or a similar bug does not re-appear again in the future.

staticaddr/loopin/actions.go
```diff
@@ -718,14 +718,13 @@ func (f *FSM) SweepHtlcTimeoutAction(ctx context.Context,
 		}
 
 		f.Errorf("unable to create and publish htlc timeout sweep "+
-			"tx: %v, retrying in  %v", err, time.Hour.String())
+			"tx: %v, retrying in %v", err, time.Hour.String())
```
starius: 1. Make a const instead of using `time.Hour` directly.
2. Reuse the const below in `case <-time.After`
3. When logging, you can pass it directly, no `.String()` call is needed. %v and %s do t automatically

staticaddr/loopin/actions.go
starius: s/break/return OnHtlcTimeoutSweepPublished/

and remove `return OnHtlcTimeoutSweepPublished` in the end of the function

staticaddr/loopin/actions.go
```diff
@@ -718,14 +718,13 @@ func (f *FSM) SweepHtlcTimeoutAction(ctx context.Context,
...
 
 		select {
 		case <-ctx.Done():
-			f.Errorf("%v", ctx.Err())
+			return f.HandleError(ctx.Err())
```
starius, line 732: Add a comment, explaining when this ctx may be cancelled and why it is ok to return an error here (giving up the broadcasting attempts).

staticaddr/loopin/manager.go
```diff
@@ -575,16 +575,18 @@ func (m *Manager) recoverLoopIns(ctx context.Context) error {
 			return err
 		}
 
+		// Add the FSM to the active loop-ins map before sending
+		// the recover event to avoid a data race.
+		m.activeLoopIns[loopIn.SwapHash] = fsm
+
 		// Send the OnRecover event to the state machine.
-		go func(fsm *FSM, swapHash lntypes.Hash) {
+		go func(fsm *FSM) {
```
starius: `s/fsm *FSM//`

staticaddr/loopin/manager.go
```diff
@@ -575,16 +575,18 @@ func (m *Manager) recoverLoopIns(ctx context.Context) error {
 			return err
 		}
 
+		// Add the FSM to the active loop-ins map before sending
+		// the recover event to avoid a data race.
+		m.activeLoopIns[loopIn.SwapHash] = fsm
```
starius: The map `m.activeLoopIns` is never used. We can just remove it completely, then we don't need to fix the race.

Can you confirm that the map is not needed, please? Maybe we just forgot to use it somewhere?

---

I also found the following dead code in staticaddr with AI. Seems valid. Can you also fix this, please?

## 2. `withdraw.Store` interface is stale and mismatched

- Interface declares:
  - `CreateWithdrawal(ctx, tx, confirmationHeight, deposits, changePkScript)`
  - `GetAllWithdrawals(ctx)`
  - File: `staticaddr/withdraw/interface.go:15`.
- Actual manager uses concrete `*SqlStore` API:
  - `CreateWithdrawal(ctx, deposits)` at `staticaddr/withdraw/manager.go:459`
  - `UpdateWithdrawal(...)` at `staticaddr/withdraw/manager.go:724`
  - Concrete methods in `staticaddr/withdraw/sql_store.go:73` and `:118`.

Impact:
- Interface is not a valid abstraction for current manager code.
- Likely leftover from pre-refactor API.

## 3. Unused `errChan` fields in managers

Declared and initialized, never used:

- `staticaddr/loopin/manager.go:137`, `:157`
- `staticaddr/openchannel/manager.go:107`, `:117`
- `staticaddr/withdraw/manager.go:133`, `:156`

Impact:
- Dead fields/comments imply behavior (“forwards errors”) that does not exist.

## 4. Unused config fields in `openchannel.Config`

Declared but not used in `openchannel/manager.go` implementation:

- `Server` (`:52`)
- `AddressManager` (`:56`)
- `ChainNotifier` (`:76`)
- `Signer` (`:79`)

Impact:
- Stale config surface increases cognitive overhead and maintenance burden.

## 5. Unused config fields in `deposit.ManagerConfig`

Declared but not used in deposit manager/FSM logic:

- `AddressClient` (`staticaddr/deposit/manager.go:45`)
- `SwapClient` (`:52`)
- `ChainParams` (`:64`)

Impact:
- Unused wiring in daemon setup and dead config parameters.

## 6. Possibly dead address-store methods/path

- Store method `GetStaticAddress(ctx, pkScript)` appears unused by runtime code:
  - Interface: `staticaddr/address/interface.go:20`
  - SQL impl: `staticaddr/address/sql_store.go:45`
- Current manager path uses `GetAllStaticAddresses` instead.
- `SqlStore.Close()` at `staticaddr/address/sql_store.go:79` also appears unused in-repo.

Impact:
- Stale API surface; low risk but adds maintenance noise.

staticaddr/address/interface.go
```diff
@@ -33,7 +33,7 @@ type Parameters struct {
 	// timeout path.
 	ClientPubkey *btcec.PublicKey
 
-	// ClientPubkey is the client's pubkey for the static address. It is
+	// ServerPubkey is the server's pubkey for the static address. It is
```
starius, line 31: I found and fixed more cases like this:

https://gist.githubusercontent.com/starius/fa7ce4f88bc2aa06b6b7f817c88967fe/raw/b1b3148d62d290545cc7d7761cd4b9f777f1f6c2/godoc.diff

Can you check it and `git apply` this diff, please?

staticaddr/withdraw/manager.go
```diff
@@ -687,13 +687,21 @@ func (m *Manager) handleWithdrawal(ctx context.Context,
 			// If the transaction received one confirmation, we
 			// ensure re-org safety by waiting for some more
 			// confirmations.
-			var confChan chan *chainntnfs.TxConfirmation
-			confChan, errChan, err =
+			var (
+				confChan    chan *chainntnfs.TxConfirmation
+				confErrChan chan error
+			)
+			confChan, confErrChan, err =
```
starius: You can remove `var` and use `:=`

A new `err` will be created - this is better. Currently we use `err` across goroutines, which is risky.

staticaddr/withdraw/manager.go
```diff
@@ -687,13 +687,21 @@ func (m *Manager) handleWithdrawal(ctx context.Context,
...
 				m.cfg.ChainNotifier.RegisterConfirmationsNtfn(
 					ctx, spentTx.SpenderTxHash,
 					withdrawalPkscript, MinConfs,
 					int32(m.initiationHeight.Load()),
 				)
+			if err != nil {
+				log.Errorf("Error registering confirmation "+
+					"notification: %v", err)
+				return
```
starius, line 697: new line before `return` and after `}`

starius, line 697: I think we should rework the code watching for withdrawal confirmation. We can do it in a follow-up PR.

1. If Register* calls fail, not just log an error and stop, but retry again at next block detection. We use this retry approach in other parts: retrying on next block.
2. Handle reorgs. Currently if it the withdrawal tx gets one confirmation, but gets reorged, we do not automatically handle this situation.

staticaddr/openchannel/manager.go
starius: Can remove `maybeCancelShim` name and just do this?
```go
defer func() {
  // code of maybeCancelShim
}()
```

staticaddr/openchannel/manager.go
```diff
@@ -497,7 +497,7 @@ func (m *Manager) openChannelPsbt(ctx context.Context,
 				},
 			}
 			_, err := m.cfg.LightningClient.FundingStateStep(
-				ctx, cancelMsg,
+				context.WithoutCancel(ctx), cancelMsg,
```
starius, line 479: I found more cases like this. Can you check the commits and `git am` them into the branch, please?

https://gist.githubusercontent.com/starius/c5f80259f09996e48b251a6acee58ac6/raw/c658bc3760908cbff2a493893c37bf124d43c50c/without-cancel.patch

staticaddr/loopin/loopin.go
hieblmi: done

staticaddr/loopin/manager.go
hieblmi: done

staticaddr/withdraw/manager.go
hieblmi: done

staticaddr/openchannel/manager.go
hieblmi: done

staticaddr/loopin/manager.go
hieblmi: I'll run integration tests over these removals, thanks for flagging them.

staticaddr/withdraw/manager.go
hieblmi: Added: https://github.com/lightninglabs/loop/issues/1087

staticaddr/openchannel/manager.go
hieblmi: thank you, applied it.

staticaddr/address/interface.go
hieblmi: thank you, done

staticaddr/loopin/actions.go
hieblmi: done

hieblmi:
Thanks for your review @starius. Please not that I upgraded the linter version and successively fixed all newly discovered linter issues, see commit: https://github.com/lightninglabs/loop/pull/1078/changes/b6d86f1526697a84ead751a1006ef804e1854101

loopdb/postgres.go
```diff
@@ -33,7 +33,7 @@ type PostgresConfig struct {
 	Host               string `long:"host" description:"Database server hostname."`
 	Port               int    `long:"port" description:"Database server port."`
 	User               string `long:"user" description:"Database user."`
-	Password           string `long:"password" description:"Database user's password."`
+	Password           string `long:"password" description:"Database user's password."` //nolint:gosec
```
starius, line 36: _Nit for a follow-up PR if you like it._

I wish flags' parsing libraries provided a convenient way to pass password as files. Like:

```go
Password  string `long:"password" description:"Database user's password." from:"file"`
```

Unfortunately it is not supported, but we can make a custom field type:

```go
  type Secret string

  func (s *Secret) UnmarshalFlag(v string) error {
        // Example convention: @/path/to/file => read from file, else raw value
        if strings.HasPrefix(v, "@") {
                b, err := os.ReadFile(strings.TrimPrefix(v, "@"))
                if err != nil {
                        return err
                }
                *s = Secret(strings.TrimRight(string(b), "\r\n"))
                return nil
        }
        *s = Secret(v)
        return nil
  }

  type Opts struct {
        Password   Secret `long:"password" description:"Database user's password or @file to read the password from it."`
  }
```

staticaddr/withdraw/manager.go
starius: Could you add a TODO here and below under `case err := <-confErrChan:` referencing this issue, please?

staticaddr/loopin/loopin_test.go
```diff
@@ -0,0 +1,159 @@
...
+	// distinct values when a selected amount creates change.
+	htlcTx, err := loopIn.createHtlcTx(network, feeRate, maxFeePercentage)
+	require.NoError(t, err)
+	require.Len(t, htlcTx.TxOut, 2, "expected HTLC + change outputs")
+
+	// Identify which output is change and which is HTLC.
+	var htlcIdx int
+	if bytes.Equal(htlcTx.TxOut[0].PkScript, pkScript) {
+		htlcIdx = 1
+	}
```
starius, line 123: This test always gets `htlcIdx=0`, by construction of `createHtlcTx`. To actually test that `createHtlcSweepTx` tolerates HTLC output being at both indices, we need to add another subtest in which we edit `htlcTx` after getting it from `createHtlcTx` to swap its outputs, so that the change is at index 0 and the HTLC is at index 1. Then we can actually cover the regression being fixed.

staticaddr/loopin/manager.go
```diff
@@ -807,7 +809,7 @@ func (m *Manager) startLoopInFsm(ctx context.Context,
 
 	// Send the start event to the state machine.
 	go func() {
-		err = loopInFsm.SendEvent(ctx, OnInitHtlc, nil)
+		err := loopInFsm.SendEvent(ctx, OnInitHtlc, nil)
 		if err != nil {
 			log.Errorf("Error sending OnNewRequest event: %v", err)
 		}
```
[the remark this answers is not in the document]
starius: This comment is not fixed. The log message says "Error sending OnNewRequest event", but the event being sent is actually OnInitHtlc

staticaddr/withdraw/manager.go
```diff
@@ -676,6 +671,9 @@ func (m *Manager) handleWithdrawal(ctx context.Context,
 		ctx, &d.OutPoint, addrParams.PkScript,
 		int32(d.ConfirmationHeight),
 	)
+	if err != nil {
+		return fmt.Errorf("unable to register spend ntfn: %w", err)
+	}
```
starius, line 676: Found one more issue.

If `handleWithdrawal` fails (because GetStaticAddressParameters or RegisterSpendNtfn fails), then the call site (WithdrawDeposits) return the error and doesn't update the internal state and doesn't persist the state change to Withdrawing. It is a correctness/state-tracking failure. Restart recovery path for withdrawals may not pick it up (because recovery sources on Withdrawing deposits).

Recommended shape:

  1. After publish success, immediately persist/transition withdrawal state (FinalizedWithdrawalTx, finalizedWithdrawalTxns, Withdrawing, DB update).
  2. Start handleWithdrawal as best-effort; if it fails, log + schedule retry, but do not return user-facing failure for already-published tx.

Secondarily, improve `handleWithdrawal` itself:

  1. Stop fetching address params inside it; pass required pkScript from caller.
  2. The pkScript [of the static address] value can be calculated upfront, even before we publish anything, at the very beginning of FSM operation.

routing_plugin.go
```diff
@@ -124,9 +124,9 @@ func ReleaseRoutingPlugin(ctx context.Context) {
 		return
 	}
 
-	if err := routingPluginInstance.Done(ctx); err != nil {
-		log.Errorf("Error while releasing routing plugin: %v",
-			err)
+	err := routingPluginInstance.Done(context.WithoutCancel(ctx))
```
starius: If `routingPluginInstance.Done` hangs (for example, if the lnd call to `ImportMissionControl` hangs), `ReleaseRoutingPlugin` can run indefinitely while holding the `routingPluginMx` mutex.

This means other loop-outs can also get stuck in `AcquireRoutingPlugin`.

Two ideas to address this:
- add a timeout to the context (for example, 30s);
- call `Done` outside the mutex (though I’m not sure whether this is safe).

hieblmi:
@starius Created #1088 to track your suggestion about the `Secret` type for reading passwords from files.

loopdb/postgres.go
hieblmi: I added an follow-up issue for that: https://github.com/lightninglabs/loop/issues/1088

staticaddr/withdraw/manager.go
hieblmi: I added a follow-up issue for that: https://github.com/lightninglabs/loop/issues/1087

routing_plugin.go
hieblmi: Thanks for the suggestions, lets go for the safe 30s timeout context !

staticaddr/withdraw/manager.go
hieblmi: done

staticaddr/loopin/manager.go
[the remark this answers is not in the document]
hieblmi: Apologies, now its fixed.

staticaddr/loopin/loopin_test.go
hieblmi: Thanks for catching this, it is fixed in commit: fb6f157c

looprpc/go.mod
```diff
@@ -191,3 +191,5 @@ replace gonum.org/v1/plot => github.com/gonum/plot v0.10.1
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
 
 replace lukechampine.com/uint128 => github.com/lukechampine/uint128 v1.2.0
+
+replace git.sr.ht/~sbinet/gg => ../third_party/gg
```
starius: There is GitHub mirror which we can reuse. It has the same hash in go.sum as the original one for the tag used (0.3.1):

```diff
diff --git a/go.mod b/go.mod
index 0bd78ab2..0c402985 100644
--- a/go.mod
+++ b/go.mod
@@ -221,7 +221,7 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
 
 replace lukechampine.com/uint128 => github.com/lukechampine/uint128 v1.2.0
 
-replace git.sr.ht/~sbinet/gg => ./third_party/gg
+replace git.sr.ht/~sbinet/gg => github.com/diiyw/gg v0.3.1
 
 replace github.com/lightninglabs/loop/swapserverrpc => ./swapserverrpc
 
diff --git a/go.sum b/go.sum
index e1545344..745ef137 100644
--- a/go.sum
+++ b/go.sum
@@ -757,6 +757,7 @@ github.com/decred/dcrd/lru v1.1.2 h1:KdCzlkxppuoIDGEvCGah1fZRicrDH36IipvlB1ROkFY
 github.com/decred/dcrd/lru v1.1.2/go.mod h1:gEdCVgXs1/YoBvFWt7Scgknbhwik3FgVSzlnCcXL2N8=
 github.com/dhui/dktest v0.4.5 h1:uUfYBIVREmj/Rw6MvgmqNAYzTiKOHJak+enB5Di73MM=
 github.com/dhui/dktest v0.4.5/go.mod h1:tmcyeHDKagvlDrz7gDKq4UAJOLIfVZYkfD5OnHDwcCo=
+github.com/diiyw/gg v0.3.1/go.mod h1:KGYtlADtqsqANL9ueOFkWymvzUvLMQllU5Ixo+8v3pc=
 github.com/distribution/reference v0.6.0 h1:0IXCQ5g4/QMHHkarYzh5l+u8T3t73zM5QvfrDyIgxBk=
 github.com/distribution/reference v0.6.0/go.mod h1:BbU0aIcezP1/5jX/8MP0YiH4SdvB5Y4f/wlDRiLyi3E=
 github.com/docker/cli v28.1.1+incompatible h1:eyUemzeI45DY7eDPuwUcmDyDj1pM98oD5MdSpiItp8k=
diff --git a/looprpc/go.mod b/looprpc/go.mod
index c8b16e39..9830a660 100644
--- a/looprpc/go.mod
+++ b/looprpc/go.mod
@@ -192,4 +192,4 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
 
 replace lukechampine.com/uint128 => github.com/lukechampine/uint128 v1.2.0
 
-replace git.sr.ht/~sbinet/gg => ../third_party/gg
+replace git.sr.ht/~sbinet/gg => github.com/diiyw/gg v0.3.1

```

Then we don't need the third_party/gg dummy.

staticaddr/loopin/loopin_test.go
```diff
@@ -0,0 +1,176 @@
...
+	changeValue := htlcTx.TxOut[1-htlcIdx].Value
+	require.NotEqual(t, htlcValue, changeValue,
+		"HTLC and change values must differ for this test to be "+
+			"meaningful")
+
+	// assertSweepUsesHtlcValue calls createHtlcSweepTx and verifies that
+	// the sweep output is derived from the HTLC value, not the change.
+	assertSweepUsesHtlcValue := func(t *testing.T) {
+		sweepTx, err := loopIn.createHtlcSweepTx(
+			context.Background(), signer, sweepAddr, feeRate,
```
starius: t.Context()

staticaddr/loopin/loopin_test.go
```diff
@@ -0,0 +1,176 @@
...
+	t.Run("natural output order", func(t *testing.T) {
+		assertSweepUsesHtlcValue(t)
+	})
+
+	// Swap the outputs so the change is at index 0 and the HTLC is at
+	// index 1. This covers the regression where createHtlcSweepTx
+	// unconditionally read TxOut[0].Value.
+	t.Run("swapped output order", func(t *testing.T) {
+		htlcTx.TxOut[0], htlcTx.TxOut[1] = htlcTx.TxOut[1],
+			htlcTx.TxOut[0]
```
starius: This swap does not affect `createHtlcSweepTx`, because that function calls `createHtlcTx` internally and builds its own `htlcTx`. The locally swapped `htlcTx` in the test is therefore irrelevant.

So this fix cannot actually be tested with that subtest. Sorry for my earlier suggestion. The test itself is still useful, and we can keep it, but its name and comments should be adjusted to reflect that it is a generic sweep-value test, not a test of output-index permutations.

Because both functions are private and in the same package, we can define an explicit output-layout invariant for the HTLC tx. `createHtlcSweepTx` can enforce this invariant and fail fast if it is violated. Since `createHtlcTx` is its dependency, the invariant should always hold unless there is a bug.

looprpc/go.mod
hieblmi: Thank you, that worked.

staticaddr/loopin/loopin_test.go
hieblmi: fixed

staticaddr/loopin/loopin_test.go
hieblmi: I ensure that the invariant is enforced, also fixed the test and test description.

sputn1ck:
@claude can you please review this, focus on if any change is changing any non-superficial logic

looprpc/go.mod
```diff
@@ -191,3 +191,5 @@ replace gonum.org/v1/plot => github.com/gonum/plot v0.10.1
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
 
 replace lukechampine.com/uint128 => github.com/lukechampine/uint128 v1.2.0
+
+replace git.sr.ht/~sbinet/gg => github.com/diiyw/gg v0.3.1
```
sputn1ck: do we need this commit? feels wonky.

hieblmi: `git.sr.ht/~sbinet/gg` is flaky, but `gonum.org/v1/gonum` depends on it. So we need to point it to the stable `github.com/diiyw/gg v0.3.1`.

sputn1ck: might be super nitpicky and random, but somehow I'm not comfortable with this replace. If `gonum.org/v1/gonum`  is not doing the replace imho we shouldn't do it either.

hieblmi: Good call! The upstream dependency seems to have been fixed. I removed the commit in question here.

PR #1089, opened 2026-03-07
fix: use grpc.Dial instead of grpc.NewClient

Olexandr88:
Replace grpc.NewClient with grpc.Dial, which is the standard gRPC API for creating client connections.

### DIFF
--- a/assets/client.go
+++ b/assets/client.go
@@ -305,8 +305,8 @@ func getClientConn(config *TapdConfig) (*grpc.ClientConn, error) {
 		grpc.WithDefaultCallOptions(maxMsgRecvSize),
 	}
 
-	// Dial the gRPC server.
-	conn, err := grpc.Dial(config.Host, opts...)
+	// Create the gRPC client connection.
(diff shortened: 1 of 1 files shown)


assets/client.go
```diff
@@ -306,7 +306,7 @@ func getClientConn(config *TapdConfig) (*grpc.ClientConn, error) {
 	}
 
 	// Dial the gRPC server.
-	conn, err := grpc.NewClient(config.Host, opts...)
+	conn, err := grpc.Dial(config.Host, opts...)
```
starius: Why is this needed?

Note that [grpc.Dial](https://pkg.go.dev/google.golang.org/grpc#Dial) is deprecated and they propose to use grpc.NewClient instead:

> Deprecated: use NewClient instead.

Olexandr88: I reverted the change back to grpc.NewClient and updated the comment to match the code

hieblmi:
Could you squash the two commits, @Olexandr88?

PR #1097, opened 2026-03-16
staticaddr: dynamic deposit conf target

hieblmi:
Update static address loop-in client to support dynamic server-side confirmation requirements.

  Changes:

  - Reduce MinConfs from 6 to 3 to allow faster swap attempts for small amounts while the server enforces risk-based confirmation requirements
  - Update SelectDeposits to prioritize more-confirmed deposits first, increasing the likelihood of server acceptance when dynamic requirements apply
  - Add client-side logging of insufficient confirmation details from server errors, helping users understand when deposits need more confirmations before retrying

  Proto updates:

  - Add DepositConfirmationStatus message with current/required confirmations and blocks to wait
  - Add InsufficientConfirmationsDetails message for structured error responses

  Backwards compatible: Clients with older servers will continue to work normally. The server determines actual confirmation requirements.

### DIFF
--- a/staticaddr/deposit/manager.go
+++ b/staticaddr/deposit/manager.go
@@ -20,7 +20,7 @@ const (
 	// MinConfs is the minimum number of confirmations we require for a
 	// deposit to be considered available for loop-ins, coop-spends and
 	// timeouts.
-	MinConfs = 6
+	MinConfs = 3
 
 	// MaxConfs is unset since we don't require a max number of
 	// confirmations for deposits.
--- a/staticaddr/loopin/actions.go
+++ b/staticaddr/loopin/actions.go
@@ -30,6 +30,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"google.golang.org/grpc/status"
 )
 
 const (
@@ -146,6 +147,10 @@ func (f *FSM) InitHtlcAction(ctx context.Context,
 		ctx, loopInReq,
 	)
 	if err != nil {
+		// Check if this is an insufficient confirmations error and log
+		// the details to help the user understand what's needed.
+		logInsufficientConfirmationsDetails(err)
+
 		err = fmt.Errorf("unable to initiate the loop-in with the "+
 			"server: %w", err)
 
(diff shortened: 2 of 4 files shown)


staticaddr/loopin/manager.go
```diff
@@ -875,17 +877,28 @@ func SelectDeposits(targetAmount btcutil.Amount,
...
+			jConfs = blockHeight -
+				uint32(deposits[j].ConfirmationHeight)
 		}
+		if iConfs != jConfs {
+			return iConfs > jConfs
+		}
+
+		// Secondary: larger amounts first.
 		return deposits[i].Value > deposits[j].Value
 	})
```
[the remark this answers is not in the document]
hieblmi: Re-introduced the Teriary tie-breaker.

hieblmi:
@claude review this

staticaddr/deposit/manager.go
```diff
@@ -20,7 +20,7 @@ const (
 	// MinConfs is the minimum number of confirmations we require for a
 	// deposit to be considered available for loop-ins, coop-spends and
 	// timeouts.
-	MinConfs = 6
+	MinConfs = 3
```
starius, line 23: We need to hold this PR until we deploy the server side. Otherwise a new client might send its deposits too early to an old server.

staticaddr/loopin/manager.go
```diff
@@ -850,11 +850,14 @@ func (m *Manager) GetAllSwaps(ctx context.Context) ([]*StaticAddressLoopIn,
...
 
-// SelectDeposits sorts the deposits by amount in descending order, then by
-// blocks-until-expiry in ascending order. It then selects the deposits that
-// are needed to cover the amount requested without leaving a dust change. It
-// returns an error if the sum of deposits minus dust is less than the requested
-// amount.
+// SelectDeposits sorts the deposits to optimize for successful swaps with
+// dynamic confirmation requirements: 1) more confirmations first (higher chance
+// of server acceptance), 2) larger amounts first (to minimize number of deposits
+// used), 3) expiring sooner first (prioritize time-sensitive deposits).
```
starius, line 856: `SelectDeposits` can now turn a valid amount-only swap into a more expensive or failing one because it switched to a confirmations-first greedy prefix instead of actually optimizing the chosen subset. The new logic sorts primarily by confirmations and then keeps taking the first deposits until the target is covered in `staticaddr/loopin/manager.go:853-930`.

That means a set like `500k@10 conf`, `500k@9 conf`, `1,000k@8 conf` for a `1,000k` target now selects two inputs instead of the exact one-input match. This is not just a UX issue: the chosen input count feeds the quote path in `staticaddr/loopin/manager.go:729-742`, and the later HTLC fee guards scale with `f.loopIn.htlcWeight(hasChange)` in `staticaddr/loopin/actions.go:221-270`. So the new order can produce avoidable `ErrSwapFeeTooHigh` or `ErrFeeTooHigh` failures, or simply worse fees, even when a slightly younger single deposit would already have been acceptable to the server.

If the client wants to prefer older deposits, they need a subset selection strategy that still minimizes input count and fee impact, or at least a second pass that prefers exact or lower-input covers before falling back to the confirmations-first prefix.

staticaddr/loopin/actions.go
```diff
@@ -146,6 +147,10 @@ func (f *FSM) InitHtlcAction(ctx context.Context,
 		ctx, loopInReq,
 	)
 	if err != nil {
+		// Check if this is an insufficient confirmations error and log
+		// the details to help the user understand what's needed.
+		logInsufficientConfirmationsDetails(err)
```
starius, line 152: The new normal insufficient-confirmation path leaks an invoice. `InitHtlcAction` creates the swap invoice before it calls `ServerStaticAddressLoopIn` in `staticaddr/loopin/actions.go:109-148`. If the server rejects the request, the new branch only logs the structured details and returns `HandleError` in `staticaddr/loopin/actions.go:149-157`.

The only invoice cancel path I found is much later in the payment-timeout monitor in `staticaddr/loopin/actions.go:558-569`, which is never reached for this early failure.

Early `ServerStaticAddressLoopIn` failures, and the follow-on fee-validation failures in `staticaddr/loopin/actions.go:231-270`, should cancel the freshly created invoice before unlocking the deposits.

PR #1104, opened 2026-03-24
staticaddr: surface mempool deposits immediately

hieblmi:
This PR does the following:

  - It changes static-address deposit handling so deposits are tracked as soon as they appear in the mempool, not only after they confirm.
  - It exposes those unconfirmed deposits through the static-address RPC/CLI flows, including summary totals and expiry-related metadata.
  - It allows static loop-ins to consider unconfirmed deposits, while still preferring confirmed deposits first during auto-selection.
  - It adds clearer user messaging for low-confirmation deposits, warning that swap execution may be delayed by the server’s confirmation-risk policy.
  - It keeps the safety boundary for other spend paths: withdrawals and static-address channel opens still require confirmed deposits.
  - It makes withdraw --all stricter so it fails instead of silently skipping unconfirmed deposits.
  - It preserves deposit identity across RBF/replacement scenarios, so a logical deposit can survive an outpoint change without losing associated loop-in state.
  - It hides replaced/vanished deposits from normal listings while still retaining them internally as historical records.
  - It hardens loop-in lifecycle handling by canceling orphan invoices on early failures and by aborting cleanly if selected deposits are replaced before payment
    settles.
  - It improves recovery/cleanup behavior around deposit finalization and restart handling, reducing cases where state can get stuck or become inconsistent.
  - It adds substantial test coverage around confirmation tracking, reconciliation, rebinding, loop-in selection, RPC behavior, and CLI warnings.


staticaddr/deposit/manager.go
```diff
@@ -351,6 +396,117 @@ func (m *Manager) getBlockHeight(ctx context.Context,
...
+		if deposit.ConfirmationHeight > 0 ||
+			!deposit.IsInStateNoLock(Deposited) {
+
+			deposit.Unlock()
+			continue
+		}
+
+		// Persist the replacement marker before removing the deposit from the
+		// active set so restarted clients and RPC consumers see the same outcome.
+		deposit.SetStateNoLock(Replaced)
```
starius, line 582: Persisting `Replaced` after a single missing `ListUnspent` observation looks too strong. If an unconfirmed outpoint temporarily disappears from lnd's wallet view and later reappears or confirms with the same outpoint, this code drops the FSM, `Replaced` becomes final, and `filterNewDeposits` will never recreate it because the outpoint is already in `m.deposits`. That would strand a valid deposit until manual repair. Can we make this reversible or require stronger evidence before finalizing the record?

staticaddr/loopin/manager.go
starius, line 901: Now that `ConfirmationHeight == 0` is considered swappable below, keeping amount-first ordering here means auto-selection can pick a larger unconfirmed deposit over a slightly smaller confirmed one, even when the confirmed set would let the swap proceed immediately. That seems like the opposite of what users want from auto-selection. Should this prefer confirmed deposits before unconfirmed ones, or otherwise optimize for earliest likely server acceptance?

staticaddr/loopin/manager.go
```diff
@@ -914,20 +919,33 @@ func SelectDeposits(targetAmount btcutil.Amount,
 // IsSwappable checks if a deposit is swappable. It returns true if the deposit
 // is not expired and the htlc is not too close to expiry.
 func IsSwappable(confirmationHeight, blockHeight, csvExpiry uint32) bool {
+	if confirmationHeight == 0 {
+		return true
+	}
```
starius, line 931: This change makes low-/0-conf deposits a normal initiation path. One thing that still worries me is early cleanup: `InitHtlcAction` creates the private invoice before `ServerStaticAddressLoopIn`, and on early server rejection or HTLC fee-guard failure it returns through `HandleError` without cancelling that invoice. Repeated retries while waiting for confirmations will accumulate orphan invoices. I'd fix that in this PR.

staticaddr/loopin/manager.go
hieblmi: Amount-first ordering prefers larger unconfirmed over smaller confirmed. SelectDeposits sort now prefers confirmed deposits first, then falls back to amount/expiry.

staticaddr/deposit/manager.go
hieblmi: 1. invalidateVanishedUnconfirmedDeposits strands valid deposits 
2. two consecutive missing observations required before marking Replaced.
3. new reviveReappearedDeposits reactivates a deposit back to Deposited and restarts its FSM if the same outpoint reappears.

staticaddr/loopin/manager.go
hieblmi: fixed. InitHtlcAction uses a deferred cancelSwapInvoice that fires on fsm.OnError after the invoice is created
  but before the swap is persisted.

hieblmi:
@claude review this

hieblmi:
@claude please review again.

hieblmi:
@claude review this again

litbot-9000:
@starius: review reminder

staticaddr/deposit/manager.go
```diff
@@ -269,12 +340,70 @@ func (m *Manager) reconcileDeposits(ctx context.Context) error {
...
+
+		if beforeHeight == afterHeight {
+			m.currentHeight.Store(uint32(afterHeight))
+			return utxos, afterHeight, nil
+		}
+	}
+
+	return nil, 0, errors.New("unable to get stable best block while " +
+		"listing deposits")
+}
```
starius, line 420: Commit "staticaddr: track unconfirmed deposits" is not formatted.

staticaddr/deposit/manager.go
```diff
@@ -666,7 +665,6 @@ func (m *Manager) invalidateVanishedDeposits(ctx context.Context,
 
 		m.missingDeposits[outpoint]++
 		if m.missingDeposits[outpoint] < vanishedDepositThreshold {
-
```
starius: This formatting change seems to be a left-over from a previous commit. Please apply it in the source commit.

starius:
There are some findings which seem to be valid.

We don't have to fix everything at this point, but it is better to assess them. If they confirm as valid, we need to track them with a TODO to acknowledge and to plan a follow-up.

### F-01 [MEDIUM] Risk decisions are not durable across loopd restart

Where:

- `staticaddr/loopin/actions.go:735` starts the payment deadline from an in-memory event.
- Risk accepted starts it at `staticaddr/loopin/actions.go:821`.
- Legacy min-conf fallback starts it at `staticaddr/loopin/actions.go:856`.
- Risk rejection cleanup is also only driven by the in-memory notification at `staticaddr/loopin/actions.go:836`.

Problem: the Loop server replays pending risk notifications only until a notification has been sent on the gRPC stream, and the Loop client only caches accepted/rejected notifications in process memory. If `loopd` receives a risk decision and restarts before the loop-in FSM handles it durably, recovery can lose that decision.

Impact:

- Risk accepted: the swap can recover with no payment deadline until a later legacy block fallback, or can start a fresh full timeout instead of preserving elapsed time since acceptance.
- Risk rejected: the swap can miss immediate invoice cancellation/deposit cleanup and only clear through later timeout behavior.

Fix direction: persist a risk decision state/timestamp or deadline before acting on the notification, then reconstruct the remaining timeout or rejection path on recovery.

### F-02 [MEDIUM] Mutable `ConfirmationHeight` is read without synchronization or snapshots

Where:

- Writes happen under the deposit mutex in `staticaddr/deposit/manager.go:535` and `staticaddr/deposit/manager.go:610`.
- Unsynchronized reads happen in loop-in selection at `staticaddr/loopin/manager.go:873`, `staticaddr/loopin/manager.go:889`, and `staticaddr/loopin/manager.go:897`.
- RPC paths also read it directly at `loopd/swapclient_server.go:1778`, `loopd/swapclient_server.go:2032`, `loopd/swapclient_server.go:2180`, and `loopd/swapclient_server.go:2272`.

Problem: this PR makes `deposit.Deposit.ConfirmationHeight` mutable after deposit creation. Before this branch, confirmation height was effectively immutable after creation, so returning live deposit pointers was less risky.

Impact: RPC handlers, CLI-driven selection, and loop-in initiation can race the deposit manager while it backfills confirmation height. Callers can observe inconsistent snapshots while deciding whether a deposit is confirmed, whether withdraw-all should fail, how to sort auto-selected deposits, and what confirmation height to return to users.

Fix direction: avoid returning live mutable deposit pointers for read-only flows, or add locked accessors/snapshot structs and make selection/RPC paths consume those snapshots.


### F-03 [MEDIUM] Blocking notification fan-out can stall risk notifications

Where:

- Static loop-in sweep notifications send while holding `Manager`'s lock and have no `default` case at `notifications/manager.go:411`.
- Unfinished swap notifications do the same at `notifications/manager.go:516`.
- Risk accepted/rejected notifications are handled by the same receive loop and lock at `notifications/manager.go:428` and `notifications/manager.go:472`.

Problem: a slow or wedged sweep/unfinished-swap subscriber with a full buffered channel can block `handleNotification` while the manager lock is held. While blocked, the notification receive loop cannot process later risk accepted/rejected notifications.

Impact: the new confirmation-risk path depends on prompt risk notifications to start the payment deadline or cancel on rejection. A stalled unrelated subscriber can delay those actions indefinitely, which compounds F-01's restart/recovery problem.

Fix direction: copy subscribers under lock and send outside the lock, use per-subscriber queues/workers, or make non-droppable notification types durable without blocking the global notification manager.

This is mostly pre-existing notification-manager behavior, but this PR makes it newly relevant to static loop-in payment deadlines.

hieblmi:
@starius thank you for the review and findings. I addressed F-01 and F-03 by persisting the risk notification and loading it post restart.

PR #1108, opened 2026-04-04
fix: correct loop iteration syntax in MuSig2Sign

Olexandr88:
Fix invalid loop iteration over slice length in MuSig2Sign


utils/musig.go
```diff
@@ -31,10 +31,10 @@ func MuSig2Sign(version input.MuSig2Version, privKeys []*btcec.PrivateKey,
 
 	// Next we'll pass around all public nonces to all MuSig2 sessions so
 	// that they become usable for creating the partial signatures.
-	for i := range len(privKeys) {
+	for i := range privKeys {
```
starius, line 34: This change is not needed. The code is already correct. See https://go.dev/doc/go1.22#language

Olexandr88: Thanks for the clarification! Closing the PR.

PR #1109, opened 2026-04-04
fix: handle single-signer case in MuSig2Sign

Olexandr88:
Fix single-signer case in MuSig2Sign

### DIFF
--- a/utils/musig.go
+++ b/utils/musig.go
@@ -62,7 +62,7 @@ func MuSig2Sign(version input.MuSig2Version, privKeys []*btcec.PrivateKey,
 
 	// Now that we have all partial sigs we can just combine them to
 	// get the final signature.
-	var haveAllSigs bool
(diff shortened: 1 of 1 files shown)


utils/musig.go
```diff
@@ -62,7 +62,7 @@ func MuSig2Sign(version input.MuSig2Version, privKeys []*btcec.PrivateKey,
 
 	// Now that we have all partial sigs we can just combine them to
 	// get the final signature.
-	var haveAllSigs bool
+	haveAllSigs := len(signatures) == 1
```
starius, line 65: `lnd/input` does not allow a single signer use-case in musig2, so this patch is wrong.

I addressed the finding another way in https://github.com/lightninglabs/loop/pull/1110

Olexandr88: Thank you for the clarification

PR #1111, opened 2026-04-06
staticaddr: deflake deposit manager test

starius:
The sqlite and postgres race jobs were both hanging in deposit.TestManager. The test was observing the manager through implementation details that were not safe to share with the manager itself:

- it replaced the manager's internal finalizedDepositChan and then waited on the same channel the manager consumes
- it reused package-level block and confirmation channels across runs
- it treated confirmationHeight+expiry as the last pre-expiry block even though the production IsExpired check uses >=
- it relied on scheduler timing when asserting that no sign request had happened yet

Make the test assert on stable effects instead of internal channel ownership:

- create per-test notifier channels in the test context
- run the manager from a cancellable t.Context-derived context and assert clean shutdown
- send the actual last pre-expiry height, then the expiry height
- wait for the expiry sign and publish steps with bounded timeouts
- verify finalization by waiting for the manager to remove the deposit from activeDeposits instead of racing its private finalization channel

This keeps the test aligned with the production expiry semantics and removes the race that only showed up reliably under -race.

#### Pull Request Checklist

### DIFF
--- a/staticaddr/deposit/manager_test.go
+++ b/staticaddr/deposit/manager_test.go
@@ -33,16 +33,6 @@ var (
 	defaultExpiry = uint32(100)
 
 	defaultDepositConfirmations = uint32(3)
-
-	confChan = make(chan *chainntnfs.TxConfirmation)
-
-	confErrChan = make(chan error)
-
-	blockChan = make(chan int32)
-
-	blockErrChan = make(chan error)
-
-	finalizedDepositChan = make(chan wire.OutPoint)
 )
 
 type mockStaticAddressClient struct {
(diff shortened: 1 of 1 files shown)


hieblmi (approved):
LGTM!

staticaddr/deposit/manager_test.go
```diff
@@ -229,48 +219,87 @@ func (m *MockChainNotifier) RegisterSpendNtfn(ctx context.Context,
...
+		defaultDepositConfirmations + defaultExpiry,
+	)
 
-	// Ensure that the deposit state machine didn't sign for the expiry tx.
-	<-testContext.mockLnd.SignOutputRawChannel
+	// Ensure that the deposit state machine signed the expiry tx.
+	select {
+	case <-testContext.mockLnd.SignOutputRawChannel:
+
+	case <-time.After(test.Timeout):
```
bhandras: Instead of using `test.Timeout` here and elsewhere we could use a shorter one, eg. defaultTimeout (30 secs)?

bhandras (approved):
LGTM, thank you! 🎉

staticaddr/deposit/manager_test.go
starius: Fixed

PR #1119, opened 2026-04-14
autoloop: support static loop-ins

starius:
This adds static-address-backed loop-ins to the current autoloop engine.

## What changes

- Liquidity params now include an explicit loop-in source, so operators can choose wallet-funded loop-ins or static-address-backed loop-ins.
- The static-address subsystem now exposes an autoloop preparation helper that selects deposits, quotes the swap, and returns a ready-to-dispatch static loop-in request.
- The liquidity planner can now produce `static_loop_in` suggestions, include them in `forceautoloop`, and enforce `MaxAutoInFlight` across all suggested swap types.
- Persisted static loop-ins now count in budget accounting, in-flight accounting, and peer traffic backoff.
- `SuggestSwaps` now exposes static loop-in suggestions through a dedicated RPC field and includes the static no-candidate disqualification reason.
- The missing public mapping for `AUTO_REASON_CUSTOM_CHANNEL_DATA` is added, so custom asset channels are reported as disqualified instead of breaking the RPC response.
- Static loop-in label validation now happens at the external RPC boundary, while internal autoloop dispatch can still use reserved autoloop labels.

## Important design choices

- This is intentionally a narrow first step: static autoloop uses full deposits only.
- No partial static loop-ins are attempted, so no change output is created.
- Static autoloop always uses `fast=false`.
- There is no silent fallback from static-address loop-ins to wallet-funded loop-ins.
- If a peer deficit can only be satisfied by a partial static loop-in, the planner returns no static candidate for that peer.
- Static loop-ins are returned separately from legacy `loop_in` suggestions.
- Failed static loop-ins are still accounted pessimistically because exact realized static swap cost is not persisted yet.
- The main planner still excludes channels with custom channel data; this series just makes that exclusion visible and safe at the RPC layer.

## Validation

- Unit coverage was added for loop-in source params, static autoloop preparation, no-change deposit selection, static fee/weight helpers, accounting with persisted static loop-ins, RPC encoding, and planner behavior.
- The feature was also validated end to end in regtest: a confirmed `1,000,000 sat` static deposit was suggested through `suggestswaps`, dispatched through `forceautoloop`, completed successfully, moved the deposit to `LOOPED_IN`, increased outbound liquidity on the target BTC peer, and was not suggested again afterward.

#### Pull Request Checklist


hieblmi:
@claude please review

hieblmi:
There is a finding regarding time consuming deposit selection:

[P1] staticaddr/loopin/autoloop.go:198 can do exponential subset search over deposits while building a static loop-in candidate. This is on the autoloop suggestion path, so a static address with many small deposits can make the liquidity manager stall when trying to decide that no candidate exists. The simple high-impact guard is to sum filtered swappable deposits and return ErrNoAutoloopCandidate if total < minAmount; I’d also bound the number of candidates or switch away from raw DFS for impossible [min,max] ranges.

I am currently looking to tAck the PR.

hieblmi (approved):
✅ tACK (pending the P1 item found in https://github.com/lightninglabs/loop/pull/1119#issuecomment-4387527907)

I've tested the new mode with 
```
loop --network=regtest setparams \
  --autoloop \
  --loopinsource=static-address
```
and then 
```
loop --network=regtest setparams \
  --autoloop \
  --loopinsource=static-address \
  --maxswapfee=10 \
  --maxroutingfee=10 \
  --maxprepayfee=1 \
  --maxprepay=30000 \
  --maxminer=1500000 \
  --sweeplimit=3 \
  --autobudget=200000 \
  --autoinflight=2 \
  --minamt=250000 \
  --maxamt=5000000
```
while setting the ticker duration to 10s. 

It only fired once the minimum amount of deposits was available. It then selected up to maxamt.

Very nice work!

litbot-9000:
@sputn1ck: review reminder

starius:
> There is a finding regarding time consuming deposit selection:
> 
> [P1] staticaddr/loopin/autoloop.go:198 can do exponential subset search over deposits while building a static loop-in candidate. This is on the autoloop suggestion path, so a static address with many small deposits can make the liquidity manager stall when trying to decide that no candidate exists. The simple high-impact guard is to sum filtered swappable deposits and return ErrNoAutoloopCandidate if total < minAmount; I’d also bound the number of candidates or switch away from raw DFS for impossible [min,max] ranges.
> 
> I am currently looking to tAck the PR.

Great catch!

I replaced the exponential search with a DP over deposits. It uses up to 128 MiB of RAM for the candidate table. When the number of deposits is modest, it returns exact results, i.e. the same results the exponential search would have returned. With a very large number of deposits, it gradually reduces accuracy by increasing the bucket size from 1 satoshi to 2 satoshis, and so on.

I also changed the objective function. The selector first finds the maximum total within the [min, max] range, using fewer deposits as a tie-breaker. It then looks at the band [best - (best - min) / 4, best] and chooses the best subset by expiry within that band. The goal is to keep larger swaps preferred while still trying to consume older, earlier-expiring deposits when the amount trade-off is small.

I also applied these earlier Claude findings:
- fix zero-HTLC-feerate fallback
- improve the CLI error for an invalid --loopinsource value
- reorder label validation

starius:
Fixed lint findings, rebased.

hieblmi (approved):
LGTM!

starius:
Rebased, added 3 commits:
 - `liquidity: gate static autoloop` puts this feature behind `--experimental` opt-in flag
 - `cmd/loop: record static autoloop CLI` adds CLI tests (recorded sessions for static-address autoloop params, `getparams`, `suggestswaps`, and the no-`--experimental` rejection)
 - `cmd/loop: avoid escaped session text` - cleanup after https://github.com/lightninglabs/loop/pull/1072

Also updated recorded CLI sessions in these commits:
 - `looprpc: add autoloop loop-in source` - added the loop_in_source updates to existing liquidity sessions
 - `looprpc: expose static autoloop output` - added the empty static_loop_in output update to the existing suggestswaps liquidity session

PR #1120, opened 2026-04-14
loopout: handle htlc confirmed notification

starius:
Add a HTLC-confirmed notification which triggers sweeping HTLC to the local wallet, reusing the code of the sweephtlc command.

#### Pull Request Checklist

### DIFF
--- a/loopd/htlc_confirmed_recovery.go
+++ b/loopd/htlc_confirmed_recovery.go
@@ -1,0 +1,112 @@
+package loopd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/loop/looprpc"
+	"github.com/lightninglabs/loop/swapserverrpc"
+)
+
+// htlcConfirmedSubscriber exposes the HTLC-confirmed notification stream used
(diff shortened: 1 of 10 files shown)


loopd/htlc_confirmed_recovery.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/loop/looprpc"
+	"github.com/lightninglabs/loop/swapserverrpc"
+)
+
+const (
+	// htlcConfirmedRecoverySatPerVByte is the fixed fee rate used by the
+	// HTLC recovery worker.
+	htlcConfirmedRecoverySatPerVByte = 10
```
hieblmi: should this be configurable from the notification source, and here we cap at a configured max fee rate?

hieblmi (approved):
This LGTM, pending clarification on fee limit as per comment.

Let's also cover this by an integration test.

litbot-9000:
@bhandras: review reminder

loopd/htlc_confirmed_recovery.go
starius: Fixed. Also removed the fee check here, since it is already checked by sweephtlc code itself later.

PR #1125, opened 2026-04-20
cmd/loop: drop misleading static loop-in fee warning

hieblmi:


alexbosworth (approved):

cmd/loop/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	require.NotContains(t, output,
+		"On-chain fees for static address loop-ins are not included.")
+	require.NotContains(t, output,
+		"They were already paid when the deposits were created.")
+	require.Contains(t, output,
+		"Previously deposited on-chain:             250000 sat")
+	require.Contains(t, output,
+		"Estimated total fee:                          386 sat")
+}
```
starius: Do we really need a test for this specific change? We don't have CLI tests overall yet. When https://github.com/lightninglabs/loop/pull/1072 is ready, we can just add it there as a recorded session.

starius (approved):
**LGTM!** 💯

cmd/loop/main_test.go
hieblmi: Removed it.

PR #1141, opened 2026-05-15
staticaddr: dyn-conf-tracker

hieblmi:
This PR surfaces static-address deposits as soon as they appear in the wallet, including mempool deposits, instead of waiting for the old confirmation readiness
  threshold.

  It also updates the static-address flows around those low-confirmation deposits:

  - Track, reconcile, and list unconfirmed static-address deposits.
  - Preserve logical deposit identity across RBF/replacement while hiding replaced deposits from normal listings.
  - Allow static loop-ins to select low-confirmation deposits, while keeping withdrawals and channel opens restricted to confirmed inputs.
  - Add CLI/RPC warnings and metadata so users can see when low-confirmation deposits may delay swap execution.
  - Harden static loop-in lifecycle handling around orphan invoices, replaced inputs, failed swaps, and server confirmation-risk decisions.
  - Persist confirmation-risk decisions so restart recovery preserves accepted/rejected risk state and payment deadline timing.
  - Prevent notification-manager stalls by queueing blocking notification fan-out paths.
  - Add regression coverage for deposit tracking, selection, replacement, recovery, notification delivery, and risk-decision persistence.


hieblmi:
@claude please review

hieblmi:
@claude please re-review

hieblmi:
On the two priority items:

1. `invoiceCanceledForNonPayment`: this is already covered on the current branch via the shared cancellation paths. `cancelInvoice()` and `handleRiskRejected()` both set the invoice state to canceled and set `invoiceCanceledForNonPayment = true`, so a later HTLC confirmation moves the deposits back to `SweepHtlcTimeout` instead of leaving them reusable.

2. Legacy fallback `startedAt`: using `time.Time{}` is intentional. The legacy MinConfs path is the compatibility substitute for a risk-accepted notification; before that point the client is still waiting for the deposits to become acceptable. Charging the payment timeout from swap initiation would make slow-confirming deposits consume the server payment window before the server is expected to pay, which regresses the dyn-conf behavior.

hieblmi:
@claude please review the latest set of changes. The `Replaced` deposit state was removed in favor of just-in-time reconciliation before withdraw/swap/openchannel.

loopd/swapclient_server.go
```diff
@@ -1693,58 +1714,35 @@ func (s *swapClientServer) ListUnspentDeposits(ctx context.Context,
...
-			// Confirmed deposits need to be checked.
-			outpoints = append(outpoints, utxo.OutPoint.String())
-			confirmedToCheck[utxo.OutPoint] = struct{}{}
-		}
+		outpoints = append(outpoints, utxo.OutPoint.String())
 	}
 
 	// Check the spent status of the deposits by looking at their states.
-	ignoreUnknownOutpoints := false
+	ignoreUnknownOutpoints := true
```
starius, line 1760: A brand-new mempool/confirmed deposit visible to lnd but not yet reconciled is silently omitted for up to `PollInterval`. Since this endpoint already hits lnd and is used by clients to pick deposits, I’d add `EnsureDepositsFresh(ctx)` before `ListUnspentRaw` or before `DepositsForOutpoints`.

loopd/swapclient_server.go
```diff
@@ -2065,7 +2090,7 @@ func (s *swapClientServer) GetStaticAddressSummary(ctx context.Context,
 	_ *looprpc.StaticAddressSummaryRequest) (
 	*looprpc.StaticAddressSummaryResponse, error) {
 
-	allDeposits, err := s.depositManager.GetAllDeposits(ctx)
+	allDeposits, err := s.depositManager.GetVisibleDeposits(ctx)
```
starius, line 2124: Here it might also return stale information. Should we call `EnsureDepositsFresh` here?

loopd/swapclient_server.go
```diff
@@ -1830,7 +1854,7 @@ func (s *swapClientServer) ListStaticAddressDeposits(ctx context.Context,
 			"outpoints")
 	}
 
-	allDeposits, err := s.depositManager.GetAllDeposits(ctx)
+	allDeposits, err := s.depositManager.GetVisibleDeposits(ctx)
```
starius, line 1887: Here it might also return stale information. Should we call `EnsureDepositsFresh` here?

staticaddr/deposit/manager.go
```diff
@@ -318,37 +458,156 @@ func (m *Manager) createNewDeposit(ctx context.Context,
...
+		if err != nil {
+			return err
+		}
+
+		deposit.Lock()
+		if deposit.ConfirmationHeight == confirmationHeight {
+			deposit.Unlock()
+			continue
+		}
+
```
starius: Should `ConfirmationHeight` now be treated as mutex-protected deposit state?

This PR makes `ConfirmationHeight` mutable during reconciliation, but several callers still read it directly from `*deposit.Deposit` values, for example in `staticaddr/loopin`, `staticaddr/openchannel`, and `loopd/swapclient_server.go`. It is not obvious where the boundary is between deposit-manager-owned mutable state and fields that outside packages may read directly.

Could we either:
1. guard all `ConfirmationHeight` reads/writes with deposit methods, or
2. return an immutable read-only snapshot type for list/summary/selection paths?

I think the snapshot approach would make the boundary clearer: FSM/transition paths can keep using live `*Deposit` pointers, while read-only consumers get a copy like `DepositSnapshot`/`DepositView` that does not expose the mutex-backed mutable struct.

Either way, it would help to document on `Deposit` and its mutex which fields are protected by the lock. Since `ConfirmationHeight` is now updated after creation, direct field access outside the deposit package is easy to misuse.

notifications/manager.go
```diff
@@ -271,6 +314,80 @@ func (m *Manager) SubscribeStaticLoopInSweepRequests(ctx context.Context,
...
+	m.subscribers[NotificationTypeStaticLoopInRiskAccepted] = append(
+		m.subscribers[NotificationTypeStaticLoopInRiskAccepted], sub,
+	)
+	if ntfn, ok := m.staticLoopInRiskAccepted[swapHash]; ok {
+		notifChan <- ntfn
+		delete(m.staticLoopInRiskAccepted, swapHash)
+	}
+	m.Unlock()
+
+	context.AfterFunc(ctx, func() {
```
starius, line 353: AI found a bug here. Here is the patch: https://gist.github.com/starius/316e644ea82b9a76625ade9512345016

Bug: if the Loop server sent a confrisk accept/reject before the swap FSM was persisted, `PersistStaticLoopInRiskDecision` could fail. The manager still cached the decision for replay, but if the current subscriber context was canceled before the FSM consumed the buffered notification, the cleanup path deleted the cached decision. After restart/resubscribe, the decision was gone, so the swap could miss the server’s risk response and stall.

Fix: cached risk decisions now track whether they were successfully persisted. The manager only deletes cached decisions on subscriber replay/cancel when persistence succeeded. If persistence failed, the cache entry is retained as the only recovery path until a later subscriber/FSM can receive and persist/consume it.

How this can happen in real `loopd`:

1. `loopd` receives a static loop-in risk accepted/rejected notification from the Loop server.
2. The notification manager tries to persist it through `RecordStaticAddressRiskDecision`.
3. Persistence can fail in real code, for example if the swap row is not visible/stored yet (`ErrLoopInNotFound`), or because of a transient DB/context error.
4. The manager still forwards/caches the notification for the FSM.
5. If the FSM action context is then canceled before it reads the buffered notification, old code deleted the cached notification even though it had not been persisted.
6. On the next same-process FSM subscription, there was no durable decision and no cached decision, so the client behaved as if the Loop server never sent it.

Real cancellation cases include daemon shutdown, monitor action restart/recovery, or an error path while `MonitorInvoiceAndHtlcTxAction` is active.

Important nuance: this fix does **not** make an unpersisted decision survive a hard process death. It fixes the in-process loss where the notification manager already has the decision, persistence failed, and the current subscriber goes away before consuming it.

litbot-9000:

loopd/swapclient_server.go
hieblmi: `StaticAddressSummary` might get called frequently by UX so I'd avoid the lnd lookup here and just show the current memory view.

loopd/swapclient_server.go
hieblmi: Same as the comment above, the in-mem view is good enough I think to avoid frequent lnd calls by UI.

staticaddr/deposit/manager.go
hieblmi: Yep, `ConfirmationHeight` is now treated as mutable deposit state protected by the deposit lock.

I went with the guarded-access approach rather than adding a separate snapshot type:

- `Deposit` now documents that both `state` and `ConfirmationHeight` are protected by the deposit lock.
- Normal external reads use `GetConfirmationHeight()`.
- The reconciliation write path locks the deposit before reading/updating `ConfirmationHeight`, and restores the previous value if the store update fails.
- Code that needs to inspect multiple deposits first locks them in canonical outpoint order with `lockDeposits`, then uses the `NoLock` helpers only while those locks are held.
- Store persistence now reads through the locked accessor, except for `NoLock` variants used from already-locked paths.

So the intended boundary is: callers may keep using live `*deposit.Deposit` values, but mutable fields should be accessed through deposit methods unless the caller already holds the deposit lock.

notifications/manager.go
hieblmi: Yes, this is addressed now.

The notification manager keeps a cached risk decision together with whether that decision was durably persisted. `handleStaticLoopInRiskDecision` attempts persistence first, then caches the accepted/rejected notification with the resulting persisted flag. Replay and subscriber-cancel cleanup only delete the cached notification if `staticLoopInRiskPersisted[swapHash]` is true, so if `PersistStaticLoopInRiskDecision` failed because the swap row was not available yet, the in-memory decision remains replayable for a later subscriber/FSM.

I also added comments on the `Manager` maps to make the intended lifecycle explicit:

- accepted/rejected maps cache risk decisions by swap hash for later matching subscribers,
- `staticLoopInRiskPersisted` records whether the cached decision was actually persisted,
- unpersisted decisions remain cached after subscriber cancellation so they can be replayed.

This is covered by `TestManager_StaticLoopInRiskDecisionReplayOnPersistFailure` and `TestManager_StaticLoopInRiskDecisionReplaysAfterSubscriberCancel`; `go test ./notifications` passes.

One caveat remains the same as in your note: if persistence failed, this protects the in-process loss path, but it does not make that unpersisted decision survive a hard process death.

loopd/swapclient_server.go
hieblmi: Addressed: `ListUnspentDeposits` now calls `s.depositManager.EnsureDepositsFresh(ctx)` before `DepositsForOutpoints`, so deposits already visible to lnd are reconciled before this endpoint maps wallet UTXOs to active deposit records.

staticaddr/loopin/actions.go
```diff
@@ -1077,47 +1204,12 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
 				continue
 			}
 
-			cancelInvoiceSubscription()
-			f.cancelSwapInvoice()
-
-			return f.HandleError(errors.New(
-				"server rejected confirmation risk wait",
-			))
+			return handleRiskRejected("risk rejection")
```
starius: Finding 1: risk rejection can abandon a confirmed HTLC timeout sweep

This is related, but not the same as https://github.com/lightninglabs/loop/issues/1168

`handleRiskRejected` always returns `fsm.OnError`, and `MonitorInvoiceAndHtlcTx` exits immediately on a risk rejection. At this point the client has already pushed HTLC signatures to the Loop server (`staticaddr/loopin/actions.go:662-684`), and the monitor's own contract says the server can publish the HTLC and the client must monitor it for timeout sweep (`staticaddr/loopin/actions.go:770-776`). If the HTLC has already confirmed, or confirms after the rejection, the safer available action is to keep/enter the `OnSweepingHtlcTimeout` path; instead the current `OnError` path routes `LoopingIn` deposits back to `Deposited` (`staticaddr/deposit/fsm.go:384-390`) and stops the loop-in monitor, so the client can miss the required timeout sweep and may show the spent deposits as reusable.

Suboptimal decision: after receiving risk rejection, the component chooses the generic unlock/fail path even when the HTLC timeout-sweep path is at least as safe and strictly more correct once HTLC publication is possible.

<details>
<summary>Failing regression test</summary>

```go
package loopin

import (
	"context"
	"testing"
	"time"

	"github.com/btcsuite/btcd/btcec/v2"
	"github.com/btcsuite/btcd/chaincfg/chainhash"
	"github.com/btcsuite/btcd/wire"
	"github.com/lightninglabs/loop/fsm"
	"github.com/lightninglabs/loop/staticaddr/deposit"
	"github.com/lightninglabs/loop/staticaddr/script"
	"github.com/lightninglabs/loop/staticaddr/version"
	"github.com/lightninglabs/loop/swapserverrpc"
	"github.com/lightninglabs/loop/test"
	"github.com/lightningnetwork/lnd/invoices"
	"github.com/lightningnetwork/lnd/lntypes"
	"github.com/stretchr/testify/require"
)

func TestMonitorInvoiceAndHtlcTxRiskRejectedAfterHtlcConfirmedLocksDeposits(t *testing.T) {
	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Second)
	defer cancel()

	mockLnd := test.NewMockLnd()

	clientKey, err := btcec.NewPrivateKey()
	require.NoError(t, err)
	serverKey, err := btcec.NewPrivateKey()
	require.NoError(t, err)

	swapHash := lntypes.Hash{5, 6, 8}
	depositOutpoint := wire.OutPoint{Hash: chainhash.Hash{10}, Index: 0}

	loopIn := &StaticAddressLoopIn{
		SwapHash:              swapHash,
		HtlcCltvExpiry:        mockLnd.Height + 100,
		InitiationHeight:      uint32(mockLnd.Height),
		InitiationTime:        time.Now(),
		ProtocolVersion:       version.ProtocolVersion_V0,
		ClientPubkey:          clientKey.PubKey(),
		ServerPubkey:          serverKey.PubKey(),
		PaymentTimeoutSeconds: 3_600,
		DepositOutpoints:      []string{depositOutpoint.String()},
		Deposits: []*deposit.Deposit{{
			OutPoint: depositOutpoint,
		}},
	}
	loopIn.SetState(MonitorInvoiceAndHtlcTx)

	mockLnd.SetInvoice(&lndclient.Invoice{
		Hash:  swapHash,
		State: invoices.ContractOpen,
	})

	notificationMgr := &mockNotificationManager{
		riskRejected: make(chan *swapserverrpc.ServerStaticLoopInRiskRejectedNotification, 1),
	}
	depositMgr := &recordingDepositManager{
		transitionChan: make(chan depositTransition, 1),
	}

	cfg := &Config{
		AddressManager: &mockAddressManager{params: &script.Parameters{
			ClientPubkey:    clientKey.PubKey(),
			ServerPubkey:    serverKey.PubKey(),
			ProtocolVersion: version.ProtocolVersion_V0,
		}},
		ChainNotifier:       mockLnd.ChainNotifier,
		DepositManager:      depositMgr,
		InvoicesClient:      mockLnd.LndServices.Invoices,
		LndClient:           mockLnd.Client,
		ChainParams:         mockLnd.ChainParams,
		NotificationManager: notificationMgr,
		Store: &mockStore{loopIns: map[lntypes.Hash]*StaticAddressLoopIn{
			swapHash: {},
		}},
	}

	f, err := NewFSM(ctx, loopIn, cfg, false)
	require.NoError(t, err)

	resultChan := make(chan fsm.EventType, 1)
	go func() {
		resultChan <- f.MonitorInvoiceAndHtlcTxAction(ctx, nil)
	}()

	select {
	case <-mockLnd.SingleInvoiceSubcribeChannel:
	case <-ctx.Done():
		t.Fatalf("invoice subscription not registered: %v", ctx.Err())
	}

	var confReg *test.ConfRegistration
	select {
	case confReg = <-mockLnd.RegisterConfChannel:
	case <-ctx.Done():
		t.Fatalf("htlc conf registration not received: %v", ctx.Err())
	}

	htlc, err := loopIn.getHtlc(mockLnd.ChainParams)
	require.NoError(t, err)
	htlcTx := wire.NewMsgTx(2)
	htlcTx.AddTxOut(&wire.TxOut{PkScript: htlc.PkScript})
	confReg.ConfChan <- &chainntnfs.TxConfirmation{Tx: htlcTx}
	require.Eventually(t, func() bool {
		return len(confReg.ConfChan) == 0
	}, time.Second, 10*time.Millisecond)

	notificationMgr.riskRejected <- &swapserverrpc.ServerStaticLoopInRiskRejectedNotification{
		SwapHash: swapHash[:],
	}

	select {
	case transition := <-depositMgr.transitionChan:
		require.Equal(t, deposit.OnSweepingHtlcTimeout, transition.event)
		require.Equal(t, deposit.SweepHtlcTimeout, transition.state)
	case event := <-resultChan:
		t.Fatalf("monitor exited with %v before locking confirmed HTLC deposits for timeout sweep", event)
	case <-ctx.Done():
		t.Fatalf("deposits were not locked for timeout sweep: %v", ctx.Err())
	}
}
```

Reproduced failure:

```
GOFLAGS='-p=1' GOMAXPROCS=1 go test -run TestMonitorInvoiceAndHtlcTxRiskRejectedAfterHtlcConfirmedLocksDeposits -count=1 -parallel=1 -v ./staticaddr/loopin
=== RUN   TestMonitorInvoiceAndHtlcTxRiskRejectedAfterHtlcConfirmedLocksDeposits
    zz_risk_reject_order_test.go:122: monitor exited with OnError before locking confirmed HTLC deposits for timeout sweep
--- FAIL: TestMonitorInvoiceAndHtlcTxRiskRejectedAfterHtlcConfirmedLocksDeposits (0.01s)
FAIL
FAIL	github.com/lightninglabs/loop/staticaddr/loopin	0.048s
FAIL
```

</details>

Suggested direction: risk rejection should cancel the invoice and stop waiting for payment, but it must not abandon HTLC monitoring after signatures were handed to the Loop server; if `htlcConfirmed` is already true, transition deposits with `deposit.OnSweepingHtlcTimeout`, and otherwise keep watching until the code can safely conclude the HTLC will not need a timeout sweep.

starius (commented):
Found one related `fsm.OnError` issue.

Could you streamline Git history in the PR, please? I think ae7065f1 "cmd/loop: update static loop-in replay fixtures" can be squashed into 5cbc3a3d "cmd/loop: warn for low-confirmation static deposits".

staticaddr/loopin/actions.go
hieblmi: Yes, this is a known existing issue rather than something specific to the risk rejection path in this PR. Once the client has handed out HTLC tx signatures, the server can still publish an HTLC tx for an aborted swap, so the client should keep monitoring those armed aborted swaps until they are safe.

I opened https://github.com/lightninglabs/loop/issues/1171 for the follow-up. The intended fix is to add an HTLC tx monitor for aborted swaps for which the client handed out signatures. It should watch for possible HTLC publication and wait for the timeout path to be swept. If a monitored HTLC tx has not been seen yet and one of its deposits is spent, the monitor for that HTLC can stop because the all-input HTLC tx can no longer be mined.

hieblmi:
Follow-up on the two new risk-watcher commits:

- 276ef1e1219a9aee49ae19c1b32c498747b54164 adds a small `confirmationRiskWatcher` helper for static loop-in confirmation-risk notifications. The helper keeps the notification plumbing in one place: it subscribes to accepted/rejected notifications for a single swap, filters out nil or wrong-hash updates, normalizes both server messages into one internal update type, and restores or records the durable risk-decision timestamp through the store. The commit also adds focused watcher tests for hash filtering and timestamp restoration/replay handling.

- 156fb86e4898a93ff461b692ce2dce414c65bb38 wires that helper into `MonitorInvoiceAndHtlcTxAction`. The intent is cleanup rather than behavior change: the FSM action still owns the swap-state side effects, invoice cancellation, payment deadline handling, HTLC monitoring, and rejection/error path. The watcher only replaces the duplicated accepted/rejected channel handling and the inline decision-time restoration code. This commit also adds an action-level regression test that wrong-hash risk notifications do not persist a decision, cancel the invoice, or exit the monitor loop, while the correct accepted notification is still handled.

High level, this keeps the blast radius small while making the monitor action easier to read. Risk notification management is isolated, but the existing loop semantics and timestamp restoration behavior remain in the same places where they matter.

staticaddr/deposit/manager.go
```diff
@@ -340,37 +375,72 @@ func (m *Manager) createNewDeposit(ctx context.Context,
...
+		deposit.ConfirmationHeight = confirmationHeight
+
+		err = m.cfg.Store.UpdateDeposit(ctx, deposit)
+		if err != nil {
+			deposit.ConfirmationHeight = previousConfirmationHeight
+			deposit.Unlock()
+			return err
+		}
+
+		deposit.Unlock()
```
starius: Manual Unlock is a footgun. If we add more `continue`, `break` or `return` statements in the future, this will result in a deadlock.

Let's rewrite in using a closure:

```go
err := func() error {
  deposit.Lock()
  defer deposit.Unlock()

  ...
}()
if err != nil {
  return err
}
```

staticaddr/deposit/manager_height_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfirmationHeightForUtxo(t *testing.T) {
```
starius, line 14: godoc

staticaddr/deposit/manager_test.go
```diff
@@ -141,6 +142,21 @@ func (m *mockAddressManager) ListUnspent(ctx context.Context,
...
+// listUnspentOverride delegates all address-manager methods except
+// ListUnspent to another implementation.
+type listUnspentOverride struct {
+	AddressManager
+
+	listUnspent func(context.Context, int32, int32) ([]*lnwallet.Utxo,
+		error)
+}
+
+func (l *listUnspentOverride) ListUnspent(ctx context.Context,
```
starius, line 155: godoc

staticaddr/deposit/manager.go
```diff
@@ -451,6 +465,92 @@ func (m *Manager) updateDepositConfirmations(ctx context.Context,
...
+			continue
+		}
+
+		delete(m.activeDeposits, outpoint)
+		toDeactivate = append(toDeactivate, deactivatedDeposit{
+			outpoint: outpoint,
+			fsm:      fsm,
+		})
+	}
+	m.mu.Unlock()
```
starius: I propose to move to a function/closure to make Unlock call in `defer`.

staticaddr/deposit/manager_reconcile_test.go
```diff
@@ -232,6 +243,102 @@ func TestUpdateDepositConfirmationsRecomputesPositiveHeight(t *testing.T) {
 	mockStore.AssertExpectations(t)
 }
 
+// TestReconcileDepositsDeactivatesVanishedUnconfirmedDeposit verifies that a
+// missing wallet outpoint is removed from the live active set without mutating
+// its historical DB state.
+func TestReconcileDepositsDeactivatesVanishedUnconfirmedDeposit(t *testing.T) {
+	ctx := context.Background()
```
starius: t.Context()

loopdb/sqlc/migrations/000021_static_loopin_risk_decision.up.sql
```diff
@@ -0,0 +1,9 @@
+-- confirmation_risk_decision records the server's confirmation-risk decision
+-- for a static address loop-in. The empty string means no decision has been
+-- received yet.
```
starius: Let's also add two other possible values:
  - 'accepted': Loop server accepted waiting for the low-confirmation deposits; this starts/reconstructs the payment deadline.
  - 'rejected': Loop server rejected/stopped waiting for the low-confirmation deposits before paying the invoice.

Important semantic detail: once the DB has 'rejected', a later 'accepted' update is ignored.

loopdb/sqlc/migrations/000021_static_loopin_risk_decision.up.sql
```diff
@@ -0,0 +1,9 @@
...
+ALTER TABLE static_address_swaps ADD COLUMN confirmation_risk_decision TEXT NOT NULL DEFAULT '';
+
+-- confirmation_risk_decision_time records when loopd received and persisted
+-- the server's decision, so payment deadlines can be reconstructed after
+-- restart.
```
starius: Let's add this as a note:

> Same-decision replays preserve the original confirmation_risk_decision_time; changing from accepted to rejected updates the timestamp.

staticaddr/loopin/actions.go
```diff
@@ -389,6 +390,119 @@ func (f *FSM) handleInvoiceUpdate(update lndclient.InvoiceUpdate) (
 	}
 }
 
+// selectedDepositConfirmationHeights returns current confirmation heights for
+// the original deposit outpoints selected by this loop-in.
+func selectedDepositConfirmationHeights(
+	loopIn *StaticAddressLoopIn) map[string]int64 {
+
+	confirmations := make(map[string]int64, len(loopIn.Deposits))
+	outpoints := make(map[string]struct{}, len(loopIn.DepositOutpoints))
```
starius, line 405: I think it is confusing that `loopIn.Deposits` is now not the same set as `loopIn.DepositOutpoints`. IIUC, their meaning is different:
  - `DepositOutpoints`: original selected swap inputs, sent to the Loop server.
  - `Deposits`: current/recovered deposit rows, which can theoretically diverge. Still existing deposits? It is not clear from the name.

I think we should keep a stronger invariant around selected deposits.

Would it be cleaner to keep `Deposits` always aligned with `DepositOutpoints`? That is: same outpoints, same cardinality/order, and if one original outpoint is no longer visible, mark that original deposit as currently unavailable rather than letting the swap operate on a different deposit row.

Concretely, I'd expect:
- `DepositOutpoints` remains the immutable selected set.
- `Deposits` contains only deposits matching that selected set.
- current wallet availability is represented separately, e.g. transient `exists` status from a wallet/chain view.
- confirmation remains derived from `ConfirmationHeight > 0`, unless a helper accessor makes that clearer.

This would make the safety invariant easier to reason about: a swap never silently changes its funding set; it only learns that one of its original inputs is confirmed, unconfirmed, currently unavailable, or spent.

staticaddr/loopin/actions.go
```diff
@@ -556,7 +564,7 @@ func (f *FSM) SignHtlcTxAction(ctx context.Context,
 	if outpointUnavailable {
 		err = errors.New("original deposit outpoint no longer available")
 		f.Warnf("%v, canceling swap invoice", err)
-		f.cancelSwapInvoice()
+		_ = f.cancelSwapInvoice()
```
starius:
```go
if err := f.cancelSwapInvoice(); err != nil {
  f.Warnf
```

staticaddr/loopin/actions.go
```diff
@@ -782,6 +790,28 @@ func (f *FSM) cleanUpSessions(ctx context.Context,
...
+		retryTimer := time.NewTimer(monitorRetryDelay)
+		defer retryTimer.Stop()
+
+		select {
+		case <-retryTimer.C:
+			return OnRecover
+
+		case <-ctx.Done():
+			return fsm.NoOp
+		}
```
starius: Simplify:
```go
  select {
  case <-time.After(monitorRetryDelay):
        return OnRecover

  case <-ctx.Done():
        return fsm.NoOp
  }
```

staticaddr/loopin/actions.go
```diff
@@ -1425,7 +1435,7 @@ func (f *FSM) PaymentReceivedAction(ctx context.Context,
 func (f *FSM) UnlockDepositsAction(ctx context.Context,
 	_ fsm.EventContext) fsm.EventType {
 
-	f.cancelSwapInvoice()
+	_ = f.cancelSwapInvoice()
```
starius: Should we process the error here?

staticaddr/loopin/actions_test.go
```diff
@@ -103,6 +101,298 @@ func TestHandleInvoiceUpdate(t *testing.T) {
...
+		t.Fatalf("settled invoice was canceled: %v", hash)
+
+	default:
+	}
+}
+
+// TestMonitorInvoiceStreamErrorRecoversSettlement verifies that a dead invoice
+// stream checks the latest invoice state before entering recovery.
+func TestMonitorInvoiceStreamErrorRecoversSettlement(t *testing.T) {
+	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Second)
```
starius: Make a file level const for 5 seconds, so we can bump it when CI becomes slow.

staticaddr/loopin/actions_test.go
```diff
@@ -2139,6 +2589,134 @@ func waitForMonitorSubscriptions(t *testing.T, ctx context.Context,
...
+		ChainParams:    mockLnd.ChainParams,
+	}
+
+	f, err := NewFSM(ctx, loopIn, cfg, true)
+	require.NoError(t, err)
+
+	return f, depositMgr
+}
+
+type failingCancelInvoices struct {
```
starius, line 2870: add godocs here and below

staticaddr/loopin/risk_watcher.go
```diff
@@ -160,23 +162,35 @@ func (w *confirmationRiskWatcher) decisionTime(ctx context.Context,
...
+	return storedLoopIn.ConfirmationRiskDecisionTime, true
+}
+
+// decisionTime retains the best-effort behavior used for server notifications.
+func (w *confirmationRiskWatcher) decisionTime(ctx context.Context,
+	decision ConfirmationRiskDecision) time.Time {
+
+	decisionTime, ok := w.durableDecisionTime(ctx, decision)
+	if !ok {
+		return time.Now()
```
starius, line 192: For a future cleanup, I think we should either use `lnd/clock` or (better) cover all such time.Now() calls with `testing/synctest`.

starius (commented):
I propose to change the `loopIn` struct so `DepositOutpoints` and `Deposits` are the same set (always the originally selected deposits). See my full comment below.

staticaddr/deposit/manager_reconcile_test.go
hieblmi: Addressed: this test now uses t.Context().

staticaddr/deposit/manager_test.go
hieblmi: Addressed: added a Godoc comment for listUnspentOverride.ListUnspent.

loopdb/sqlc/migrations/000021_static_loopin_risk_decision.up.sql
hieblmi: Addressed: the migration comment now documents the empty, accepted, and rejected values, including that a later acceptance cannot override rejection.

staticaddr/deposit/manager.go
hieblmi: Addressed: the deposit lock scope now uses an immediately invoked closure with defer deposit.Unlock(), while preserving the confirmation-height rollback and error behavior.

staticaddr/loopin/actions.go
hieblmi: Addressed: the invoice-cancellation error is now checked and logged before returning the original unavailable-outpoint error.

loopdb/sqlc/migrations/000021_static_loopin_risk_decision.up.sql
hieblmi: Addressed: the migration comment now documents that same-decision replays preserve the timestamp, while accepted-to-rejected changes update it.

staticaddr/loopin/actions.go
hieblmi: Addressed: the retry select now uses time.After(monitorRetryDelay) directly.

staticaddr/deposit/manager.go
hieblmi: Addressed: the manager critical section now uses a closure with defer m.mu.Unlock().

staticaddr/deposit/manager_height_test.go
hieblmi: Addressed: added a Godoc comment for TestConfirmationHeightForUtxo.

staticaddr/loopin/actions_test.go
hieblmi: Addressed: added a file-level testTimeout constant and replaced the five-second test deadlines and wait with it.

staticaddr/loopin/actions_test.go
hieblmi: Addressed: added Godoc comments for failingCancelInvoices, flakySubscribeInvoices, and firstLookupBarrier.

hieblmi:
Also addressed the earlier history-cleanup request: the static loop-in replay fixture update is now squashed into the cmd/loop low-confirmation warning commit.

staticaddr/loopin/actions.go
hieblmi: done

staticaddr/loopin/actions.go
hieblmi: > Concretely, I'd expect:
> - `DepositOutpoints` remains the immutable selected set.
> - `Deposits` contains only deposits matching that selected set.

This was the expectation until unconfirmed deposits became acceptable.

But agreed—this is a valid invariant concern. `DepositOutpoints` is the immutable input set sent to the server, while `Deposits` is reconstructed through mutable database/manager associations, so recovery can lose cardinality or ordering (for example, after deposit reuse).

I think this can be addressed in a follow-up rather than blocking this PR: the current pre-signing path checks exact outpoint/cardinality equality and fails closed, and the RBF path creates a new deposit record instead of retargeting an in-flight selected deposit. The remaining risk is recovery/timeout-sweep robustness, which calls for the broader persistence and reconstruction changes tracked in [#1172](https://github.com/lightninglabs/loop/issues/1172).

staticaddr/loopin/actions.go
```diff
@@ -695,51 +898,201 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
...
+			ctx, f.loopIn.Deposits,
+			deposit.OnSweepingHtlcTimeout,
+			deposit.SweepHtlcTimeout,
+		)
+		if err != nil {
+			f.Errorf("unable to transition deposits to the htlc "+
+				"timeout sweeping state after %s: %v",
+				reason, err)
+
+			return
```
starius: Monitor can advance after the deposit transition it depends on failed.

**Problem:** `transitionDepositsToHtlcTimeout` logs and returns on `DepositManager.TransitionDeposits` failure, but the caller does not know it failed; at HTLC timeout the monitor still returns `OnSweepHtlcTimeout`, moving the loop-in FSM to `SweepHtlcTimeout` while selected deposits may still be `LoopingIn`/`Deposited`/partially transitioned. The same pattern exists on the no-HTLC path: `unlockDeposits` failure is logged but the monitor can still return `OnSwapTimedOut` and mark the swap failed while deposits remain locked or only partially unlocked.

**Impact:** a transient DB/observer/context error in `TransitionDeposits` can make the swap FSM and deposit FSM disagree about who owns the next recovery action. In the confirmed-HTLC case the next `MonitorHtlcTimeoutSweepAction` later expects deposits to be in `SweepHtlcTimeout`; if they were not moved there, timeout-sweep finalization can fail or require manual recovery. This is not #1171 armed-HTLC monitoring and not #1172 selected-deposit identity; it is current local error handling after a transition attempt fails.

**Fix direction:** make the transition/unlock helper return an error/bool and do not emit `OnSweepHtlcTimeout`/`OnSwapTimedOut` unless all selected deposits reached the required state. Retry the monitor with `OnRecover` or a bounded retry timer, while still returning `fsm.NoOp` on shutdown; do not hold extra deposit/manager locks around the retry.

**Reproduction:** reproduced with a focused action-level test that injects a deposit transition failure; current PR returns `OnSweepHtlcTimeout` anyway.

<details><summary>Failing regression test for GH comment</summary>

```go
// TestMonitorInvoiceAndHtlcTxDoesNotAdvanceWhenTimeoutDepositTransitionFails
// reproduces the local state skew where the loop-in FSM advances to the HTLC
// timeout sweep path even though the deposit manager failed to move the selected
// deposits into SweepHtlcTimeout.
func TestMonitorInvoiceAndHtlcTxDoesNotAdvanceWhenTimeoutDepositTransitionFails(
	t *testing.T) {

	ctx, cancel := context.WithTimeout(t.Context(), testTimeout)
	defer cancel()

	mockLnd := test.NewMockLnd()

	clientKey, err := btcec.NewPrivateKey()
	require.NoError(t, err)
	serverKey, err := btcec.NewPrivateKey()
	require.NoError(t, err)

	swapHash := lntypes.Hash{20, 21, 22}
	depositOutpoint := wire.OutPoint{
		Hash:  chainhash.Hash{20},
		Index: 0,
	}

	loopIn := &StaticAddressLoopIn{
		SwapHash:              swapHash,
		HtlcCltvExpiry:        mockLnd.Height,
		InitiationHeight:      uint32(mockLnd.Height),
		InitiationTime:        time.Now(),
		ProtocolVersion:       version.ProtocolVersion_V0,
		ClientPubkey:          clientKey.PubKey(),
		ServerPubkey:          serverKey.PubKey(),
		PaymentTimeoutSeconds: 3_600,
		DepositOutpoints: []string{
			depositOutpoint.String(),
		},
		Deposits: []*deposit.Deposit{{
			OutPoint: depositOutpoint,
		}},
	}
	loopIn.SetState(MonitorInvoiceAndHtlcTx)

	mockLnd.SetInvoice(&lndclient.Invoice{
		Hash:  swapHash,
		State: invoices.ContractCanceled,
	})

	depositMgr := &recordingDepositManager{
		err:            errors.New("transition failed"),
		transitionChan: make(chan depositTransition, 2),
	}
	cfg := &Config{
		AddressManager: &mockAddressManager{
			params: &script.Parameters{
				ClientPubkey:    clientKey.PubKey(),
				ServerPubkey:    serverKey.PubKey(),
				ProtocolVersion: version.ProtocolVersion_V0,
			},
		},
		ChainNotifier:  mockLnd.ChainNotifier,
		DepositManager: depositMgr,
		InvoicesClient: mockLnd.LndServices.Invoices,
		LndClient:      mockLnd.Client,
		ChainParams:    mockLnd.ChainParams,
	}

	f, err := NewFSM(ctx, loopIn, cfg, false)
	require.NoError(t, err)

	resultChan := make(chan fsm.EventType, 1)
	go func() {
		resultChan <- f.MonitorInvoiceAndHtlcTxAction(ctx, nil)
	}()

	select {
	case <-mockLnd.SingleInvoiceSubcribeChannel:
	case <-ctx.Done():
		t.Fatalf("invoice subscription not registered: %v", ctx.Err())
	}

	var confRegistration *test.ConfRegistration
	select {
	case confRegistration = <-mockLnd.RegisterConfChannel:
	case <-ctx.Done():
		t.Fatalf("htlc conf registration not received: %v", ctx.Err())
	}

	confRegistration.ConfChan <- nil

	select {
	case transition := <-depositMgr.transitionChan:
		require.Equal(t, deposit.OnSweepingHtlcTimeout, transition.event)
		require.Equal(t, deposit.SweepHtlcTimeout, transition.state)

	case <-ctx.Done():
		t.Fatalf("deposit timeout transition was not attempted: %v", ctx.Err())
	}

	require.NoError(t, mockLnd.NotifyHeight(mockLnd.Height+1))

	select {
	case event := <-resultChan:
		require.NotEqual(t, OnSweepHtlcTimeout, event,
			"monitor advanced even though deposits were not locked for timeout sweeping")

	case <-ctx.Done():
		t.Fatalf("monitor action did not exit: %v", ctx.Err())
	}
}
```

Command:

```bash
go test -run TestMonitorInvoiceAndHtlcTxDoesNotAdvanceWhenTimeoutDepositTransitionFails -count=1 -parallel=1 ./staticaddr/loopin
```

Current failure:

```text
--- FAIL: TestMonitorInvoiceAndHtlcTxDoesNotAdvanceWhenTimeoutDepositTransitionFails (0.00s)
    actions_test.go:3379:
        	Error Trace:	staticaddr/loopin/actions_test.go:3379
        	Error:      	Should not be: "OnSweepHtlcTimeout"
        	Test:       	TestMonitorInvoiceAndHtlcTxDoesNotAdvanceWhenTimeoutDepositTransitionFails
        	Messages:   	monitor advanced even though deposits were not locked for timeout sweeping
FAIL
FAIL	github.com/lightninglabs/loop/staticaddr/loopin	0.033s
FAIL
```

</details>

cmd/loop/testdata/sessions/static-loop-in/23_loop-static-in-max-swap-fee-both.json
```diff
@@ -127,6 +127,37 @@
...
+          "value_unconfirmed_satoshis": "0",
+          "value_deposited_satoshis": "2546150",
+          "value_expired_satoshis": "0",
+          "value_withdrawn_satoshis": "0",
+          "value_looped_in_satoshis": "0",
+          "value_htlc_timeout_sweeps_satoshis": "0",
+          "value_channels_opened": "0"
+        }
+      }
+    },
```
starius, line 160: I propose to record and add more CLI sessions to cover the new logic in CLI. File cmd/loop/staticaddr.go had many changes - cover them with a couple of scenarios, please.

starius (approved):
**LGTM!** Great job! 🚀

I propose to fix the AI finding in `transitionDepositsToHtlcTimeout` and to add CLI sessions covering new logic in cmd/loop/staticaddr.go.

Follow-ups for the future, not blockers:
- `time.Now()` cleanup PR [thread](https://github.com/lightninglabs/loop/pull/1141#discussion_r3562565746)
- armed HTLC recovery issue #1171
- immutable selected-deposit invariant issue #1172
- [dedup](https://github.com/lightninglabs/loop/pull/1141#issuecomment-4779540672) deposit selection logic in CLI and SelectDeposits

cmd/loop/testdata/sessions/static-loop-in/23_loop-static-in-max-swap-fee-both.json
hieblmi: Added two recorded CLI scenarios:

  - Explicit --utxo selection showing a five-confirmation warning.
  - Automatic selection showing an unconfirmed-deposit warning.

staticaddr/loopin/actions.go
hieblmi: Fixed in c718e9a9fbb55327e143e553153727f353707a2b.

`transitionDepositsToHtlcTimeout` now returns an error, and every caller retries the monitor instead of emitting `OnSweepHtlcTimeout` unless all selected deposits reached `SweepHtlcTimeout`. A retry after a partial multi-deposit transition filters out deposits that already reached the target state and only transitions the remaining deposits, then rechecks the complete selected set before advancing.

The no-HTLC path now handles unlock failures the same way, so `OnSwapTimedOut` is not emitted while deposits remain locked. Shutdown cancellation is also checked before treating an all-target transition as success; `retryMonitor` returns `fsm.NoOp` immediately for a canceled context.

Added regression coverage for a failed timeout transition, a partial multi-deposit transition followed by recovery, and a failed timeout unlock. The focused race tests and the full `staticaddr/loopin` and `staticaddr/deposit` test suites pass.

starius:
Commit "staticaddr/loopin: retry failed deposit state transitions" has long lines in description. Strange that CI didn't catch this.

PR #1154, opened 2026-06-19
confrisk: preparatory changes

hieblmi:
_extracted commits from https://github.com/lightninglabs/loop/pull/1141_
This PR pulls out general correctness and notification-delivery fixes ahead of the dynamic confirmation work.

  It fixes a few independent issues in static address loop-in/deposit handling and hardens notification fanout so slow subscribers cannot
  block unrelated notification delivery.

  ## What

  - Cancel static loop-in invoices when initialization fails before the swap is stored.
  - Finalize withdrawn deposits asynchronously so final-state cleanup cannot block the FSM.
  - Fix static loop-in state queries so failed swaps are included in final-state results.
  - Prevent slow notification subscribers from blocking manager fanout.
  - Queue must-deliver notification types while keeping optional reservation notifications best-effort.

  ## Why

  These changes reduce failure modes that are not specific to dynamic confirmations, but become more important as static loop-ins rely more
  heavily on notification-driven flows.

  Moving them first keeps the dynamic confirmation PR focused on the actual feature behavior while landing independent fixes separately.

### DIFF
--- a/staticaddr/loopin/actions.go
+++ b/staticaddr/loopin/actions.go
@@ -36,6 +36,8 @@ const (
 	defaultConfTarget = 3
 
 	DefaultPaymentTimeoutSeconds = 60
+
+	defaultInvoiceCleanupTimeout = 5 * time.Second
 )
 
 var (
@@ -57,6 +59,24 @@ var (
 func (f *FSM) InitHtlcAction(ctx context.Context,
 	_ fsm.EventContext) fsm.EventType {
 
+	var event fsm.EventType
+	invoiceNeedsCleanup := false
+	defer func() {
+		// If we created the private invoice but failed before persisting the
+		// swap, cancel it so retries do not accumulate orphan invoices.
+		if !invoiceNeedsCleanup || event != fsm.OnError {
+			return
+		}
+
+		f.cancelSwapInvoice()
+	}()
+
+	returnError := func(err error) fsm.EventType {
+		event = f.HandleError(err)
+
+		return event
+	}
+
 	// Lock the deposits and transition them to the LoopingIn state.
 	err := f.cfg.DepositManager.TransitionDeposits(
 		ctx, f.loopIn.Deposits, deposit.OnLoopInInitiated,
@@ -65,7 +85,7 @@ func (f *FSM) InitHtlcAction(ctx context.Context,
 	if err != nil {
 		err = fmt.Errorf("unable to loop-in deposits: %w", err)
 
-		return f.HandleError(err)
+		return returnError(err)
 	}
 
 	// Calculate the swap invoice amount. The server needs to pay us the
@@ -88,7 +108,7 @@ func (f *FSM) InitHtlcAction(ctx context.Context,
 		err = fmt.Errorf("unable to create random swap preimage: %w",
 			err)
 
-		return f.HandleError(err)
+		return returnError(err)
 	}
 	f.loopIn.SwapPreimage = swapPreimage
 	f.loopIn.SwapHash = swapPreimage.Hash()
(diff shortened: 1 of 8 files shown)


hieblmi:
@claude please review

hieblmi:
@claude please re-review after I added queued notification handling of reservations as well.

staticaddr/loopin/sql_store.go
```diff
@@ -203,7 +203,7 @@ func (s *SqlStore) GetStaticAddressLoopInSwapsByStates(ctx context.Context,
 }
 
 func toJointStringStates(states []fsm.StateType) string {
-	return "{" + strings.Join(toStrings(states), ",") + "}"
+	return strings.Join(toStrings(states), ",")
```
starius, line 206: Good fix! Can you adjust the commit message to explain that now it covers the first and the last states as well, and add a test for first state, please?

staticaddr/loopin/actions.go
```diff
@@ -57,6 +59,24 @@ var (
...
+	var event fsm.EventType
+	invoiceNeedsCleanup := false
+	defer func() {
+		// If we created the private invoice but failed before persisting the
+		// swap, cancel it so retries do not accumulate orphan invoices.
+		if !invoiceNeedsCleanup || event != fsm.OnError {
+			return
+		}
+
+		f.cancelSwapInvoice(ctx)
```
starius: This fix is correct for pre-storage situations.

What happens with an invoice in SignHtlcTx/MonitorInvoiceAndHtlcTx → OnError → UnlockDeposits paths?

staticaddr/loopin/actions.go
starius, line 663: This guard is dead code. `UnlockDepositsAction` returns `fsm.OnError` on **both** paths — the success path returns `fsm.OnError` (line 879), and the failure path returns `f.HandleError(err)`, which is also `fsm.OnError`. So `event != fsm.OnError` is never true and the `"unable to unlock deposits after payment deadline"` error is never logged, including when the unlock genuinely fails.

Concretely: when the payment deadline fires we cancel the invoice, try to unlock the deposits, and keep monitoring the htlc. If `TransitionDeposits(..., deposit.Deposited)` fails here we get no signal for it — the contextual log can't fire — and the FSM keeps running as though the deposits were freed while they actually stay locked until the swap resolves another way (e.g. the htlc timeout sweep). The deposits are the user's funds, so this is a "don't silently fail an operation" issue: low severity, but strictly worse than what's trivially available, since we can just check the real error.

I confirmed the premise on a live regtest run: on a failed init the loop-in FSM entered `UnlockDeposits`, the deposit transitioned `LoopingIn -> Deposited` (a *successful* unlock), and the FSM still emitted `OnError -> Failed` — i.e. `UnlockDepositsAction` returns `fsm.OnError` even on success, which is exactly why this guard can never fire.

Suggested fix: https://gist.github.com/starius/915dee5b117c07f2042170d992274ec6

starius (approved):
**LGTM!** 💾
A couple of comments left.

staticaddr/loopin/sql_store.go
hieblmi: Thanks, I added context to the commit message and added tests for it!

staticaddr/loopin/actions.go
hieblmi: good catch, I made `UnlockDepositsAction` call `cancelSwapInvoice()` so we always cancel the invoice on failed swaps.

staticaddr/loopin/actions.go
hieblmi: Thank you, this is now fixed  by

  - extract the actual deposit transition into unlockDeposits(ctx) error,
  - stop calling UnlockDepositsAction from the payment-deadline path,
  - call unlockDeposits(ctx) directly there so we can inspect the real error.

hieblmi:
@claude re-review this please

PR #1161, opened 2026-07-01
 staticaddr: harden deposit and loop-in lifecycle handling

hieblmi:
This PR prepares the static address deposit and loop-in code for later
  confirmation-tracking changes.

  It hardens deposit lifecycle handling by keeping deposits locked for the
  loop-in HTLC timeout path, ignoring expiry handling in final states, rejecting
  duplicate or invalid deposit transitions, stopping removed deposit FSMs, and
  documenting/enforcing deposit manager lock ordering.

  It also improves loop-in recovery behavior by preserving selected deposit
  outpoints, handling closed invoice update channels, adding an lnd-backed txout
  checker, and defaulting recovered payment timeout values.

  Finally, it factors active deposit block notifications, serializes deposit
  reconciliation, adds locked confirmation-height accessors, and increases the
  lint timeout for the expanded static-address test suite.


hieblmi:
@claude please review

starius:
Some commits are not signed.

staticaddr/loopin/actions.go
```diff
@@ -731,10 +786,14 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
...
 			cancelInvoice()
 
-			err = f.unlockDeposits(ctx)
-			if err != nil {
-				f.Errorf("unable to unlock deposits after "+
-					"payment deadline: %v", err)
+			if htlcConfirmed {
+				transitionDepositsToHtlcTimeout("payment deadline")
+			} else {
+				err = f.unlockDeposits(ctx)
```
starius: _AI finding._

This fixes the case where the HTLC is already confirmed at the payment deadline, but I think there is still a reuse window when the deadline fires before HTLC confirmation.

In that case we still unlock the deposits here, even though the server already has our HTLC signatures and can still publish/confirm the HTLC before CLTV. If the user reuses the deposit before that late confirmation arrives, `transitionDepositsToHtlcTimeout` will later try `OnSweepingHtlcTimeout` from whatever state the new flow put the deposit in.

That works for another loop-in because `LoopingIn` has the transition, but not for withdraw/open-channel states. A concrete sequence is:

1. payment deadline expires, `htlcConfirmed == false`, deposits are unlocked;
2. user starts a withdrawal or channel open;
3. server HTLC confirms later;
4. this monitor calls `OnSweepingHtlcTimeout`, but the deposit is now `Withdrawing`/`OpeningChannel`, where that event is invalid.

I think either the deposits need to remain unavailable until the HTLC can no longer confirm, or all reuse states that are allowed during this window need explicit handling/tests for losing the race to the HTLC.

staticaddr/deposit/manager.go
```diff
@@ -495,8 +521,15 @@ func (m *Manager) TransitionDeposits(ctx context.Context, deposits []*Deposit,
 
 	outpoints := make([]wire.OutPoint, len(deposits))
 	for i, d := range deposits {
+		if d == nil {
+			return fmt.Errorf("nil deposit at index %d", i)
+		}
+
 		outpoints[i] = d.OutPoint
 	}
+	if outpoint.HasDuplicates(outpoints) {
```
starius: This rejects duplicate outpoints within one transition request, but overlapping multi-deposit requests can still deadlock if they arrive in different orders.

For example, one goroutine can transition `[A, B]` while another transitions `[B, A]`. `lockDeposits` still locks in caller order, so G1 can hold A and wait for B while G2 holds B and waits for A. The documented manager/deposit lock ordering does not cover ordering among multiple deposit locks.

I think `lockDeposits` should canonicalize by outpoint before acquiring locks, or callers should be required to pass deposits in a canonical order. A reversed-order concurrency test would cover this.

staticaddr/loopin/sql_store_test.go
```diff
@@ -377,3 +377,76 @@ func TestCreateLoopIn(t *testing.T) {
...
+		Deposits:                []*deposit.Deposit{d},
+		ClientPubkey:            clientPubKey,
+		ServerPubkey:            serverPubKey,
+		HtlcTimeoutSweepAddress: addr,
+	}
+	swap.SetState(SignHtlcTx)
+
+	require.NoError(t, swapStore.CreateLoopIn(ctxb, &swap))
+
+	d.OutPoint = currentOutpoint
```
starius, line 523: Can we clarify the intended invariant this test is documenting?

This updates the deposit row to a different current outpoint while expecting the loop-in's stored `DepositOutpoints` to remain the original selected outpoint. That makes sense if future replacement handling treats a deposit as a stable logical record whose current outpoint can change.

If that is the intended model, I think recovery needs a follow-up: `GetLoopInByHash` can reconstruct deposits through the stable swap_hash/deposit-id mapping, but `recoverLoopIns` overwrites `loopIn.Deposits` by looking up `loopIn.DepositOutpoints` in the active-deposit map. If `DepositOutpoints` is the original-input snapshot and the deposit row now has a replacement/current outpoint, recovery can miss the deposit and continue with an incomplete deposit set.

If deposits are meant to be immutable by outpoint and replacement should cancel the current swap/create a new deposit, then this test may be stronger than the production invariant. In that case it would help to document that `DepositOutpoints` remains both the swap input snapshot and a valid recovery lookup key.

staticaddr/loopin/actions.go
hieblmi: Addressed in `7cecfc2c`.

The payment deadline now only cancels the invoice and keeps the deposits locked while the HTLC can still confirm. Deposits are only unlocked once the HTLC timeout has opened and no HTLC confirmation was seen; if the HTLC did confirm, the deposits transition to `SweepHtlcTimeout` instead. Added `TestMonitorInvoiceAndHtlcTxKeepsDepositsLockedUntilHtlcTimeout` for the race window.

staticaddr/deposit/manager.go
hieblmi: Addressed in `4fadd74b`.

`lockDeposits` now locks a canonical copy sorted by outpoint and unlocks that same copy in reverse order, so overlapping `[A, B]` / `[B, A]` requests cannot acquire deposit locks in conflicting orders. Added `TestLockDepositsCanonicalizesOutpoints` and `TestLockDepositsAllowsReversedConcurrentRequests`.

staticaddr/loopin/sql_store_test.go
hieblmi: Addressed in `37c5cf85`.

The invariant is that `DepositOutpoints` remains the original swap input snapshot. Recovery/listing now relies on the store swap_hash/deposit-id reconstruction for the current deposit rows; `recoverLoopIns` only uses current reconstructed outpoints when checking active in-memory deposits. Added coverage in `TestActiveDepositsForLoopInUsesCurrentDepositOutpoints` and `TestGetAllSwapsPreservesStoreDeposits`.

staticaddr/loopin/actions.go
starius: I can't find test TestMonitorInvoiceAndHtlcTxKeepsDepositsLockedUntilHtlcTimeout in the code. Was that test dropped in the force-push?

I think the payment-timeout reuse race is still present in current head. The earlier thread said this was addressed by keeping deposits locked until the HTLC can no longer confirm, but this branch still calls `unlockDeposits` immediately when the payment deadline fires.

After this point the monitor keeps watching HTLC confirmation, but the deposits are back in `Deposited` and can be reused for withdrawal/channel open. If the server confirms the HTLC before CLTV, the monitor later tries `OnSweepingHtlcTimeout`; that transition is valid from `Deposited`/`LoopingIn`, but not from `Withdrawing` or `OpeningChannel`.

So the same sequence still seems possible:
1. deadline fires, invoice is canceled, deposits are unlocked;
2. user starts withdraw/open-channel;
3. server HTLC confirms before CLTV;
4. timeout monitor cannot move the reused deposits into `SweepHtlcTimeout`.

Here is the test sketch which reproduces this edge case: https://gist.github.com/starius/7fe6001933000ff0f8894fd547abcf9f

It models:

1. deposit is `LoopingIn`
2. HTLC confirmation arrives
3. payment deadline unlocks the deposit back to `Deposited`
4. another realistic flow starts withdrawal via `OnWithdrawInitiated`
5. HTLC timeout opens
6. monitor returns `OnSweepHtlcTimeout`, but deposit stays `Withdrawing`

Failure:

```text
=== RUN   TestMonitorInvoiceAndHtlcTxDoesNotReleaseConfirmedHtlcDeposits
    actions_test.go:431:
        	Error:      	Not equal:
        	            	expected: "SweepHtlcTimeout"
        	            	actual  : "Withdrawing"
        	Messages:   	confirmed HTLC timeout recovery must keep deposits locked until the HTLC timeout path is swept
--- FAIL: TestMonitorInvoiceAndHtlcTxDoesNotReleaseConfirmedHtlcDeposits (1.00s)
FAIL
FAIL	github.com/lightninglabs/loop/staticaddr/loopin	1.048s
```

staticaddr/loopin/actions.go
```diff
@@ -497,6 +531,29 @@ func (f *FSM) SignHtlcTxAction(ctx context.Context,
...
 }
 
+func (f *FSM) checkDepositsAvailable(ctx context.Context) error {
+	if f.cfg.TxOutChecker == nil {
+		return nil
+	}
+
+	for _, d := range f.loopIn.Deposits {
+		txOut, err := f.cfg.TxOutChecker.GetTxOut(
+			ctx, d.OutPoint, true,
```
starius: Can we clarify the outpoint source of truth before signing?

`checkDepositsAvailable` checks each current deposit row's `d.OutPoint`, and the client HTLC construction also uses `loopIn.Deposits`/`Outpoints()`. But `DepositOutpoints` is now the original swap input snapshot, and the server locks/builds/verifies the HTLC using the original `DepositOutpoints` sent during initiation.

After recovery, `activeDepositsForLoopIn` intentionally resolves active deposits via the current deposit outpoints reconstructed by the store.

The fallback in `activeDepositsForLoopIn` makes this a little easy to miss: `DepositOutpoints` is only used when `loopIn.Deposits` is empty. Once the store reconstructs deposits, recovery switches to `Deposits[].OutPoint`, so the loop-in can continue with current deposit rows while `DepositOutpoints` still contains the server-side original snapshot.

If a deposit row's current outpoint differs from the original swap input snapshot, the client can check/sign a transaction spending the current outpoint while the server is still verifying/building the transaction for the original outpoint.

Should signing either assert that `DepositOutpoints` still matches `Deposits[].OutPoint`, or should availability checking and HTLC construction use the original `DepositOutpoints` snapshot? If divergence is impossible before signing, it would be useful to document/enforce that invariant here.

The test is here: https://gist.github.com/starius/fd1b8e9b62003f313765af8ff5b475d6

It models a recovered loop-in where:

- `DepositOutpoints` is still the original server snapshot: `...88:0`
- `loopIn.Deposits[0].OutPoint` is now the current deposit row: `...89:1`
- the current outpoint is available
- the original outpoint is not provided as available

Current behavior:

- `checkDepositsAvailable` checks and accepts `...89:1`
- `createHtlcTx` builds the HTLC input with `...89:1`
- stored `DepositOutpoints` still says server-side snapshot is `...88:0`

It is red now:

```text
=== RUN   TestSignHtlcTxPreparationRejectsDivergentDepositOutpoints
    actions_test.go:675: expected signing prep to reject divergent deposit outpoints; availability checked [0000000000000000000000000000000000000000000000000000000000000089:1] and accepted it, htlc tx input is 0000000000000000000000000000000000000000000000000000000000000089:1, stored DepositOutpoints snapshot is [0000000000000000000000000000000000000000000000000000000000000088:0]
--- FAIL: TestSignHtlcTxPreparationRejectsDivergentDepositOutpoints (0.00s)
FAIL
FAIL	github.com/lightninglabs/loop/staticaddr/loopin	0.045s
FAIL
```

staticaddr/deposit/fsm_test.go
```diff
@@ -11,6 +13,65 @@ import (
...
+				quitChan:             make(chan struct{}),
+				finalizedDepositChan: make(chan wire.OutPoint, 1),
+			}
+			depositFSM.StateMachine = fsm.NewStateMachineWithState(
+				depositFSM.DepositStatesV0(), state,
+				DefaultObserverSize,
+			)
+			depositFSM.ActionEntryFunc = depositFSM.updateDeposit
+
+			depositFSM.handleBlockNotification(context.Background(), 3)
```
starius: Use `t.Context()`

staticaddr/outpoint/duplicates.go
```diff
@@ -0,0 +1,24 @@
+package outpoint
```
starius: Do we really need a separate package for this? This might to go to `staticaddr/deposit/` or `staticaddr/staticutil/`

staticaddr/outpoint/duplicates.go
```diff
@@ -0,0 +1,24 @@
...
+
+import "github.com/btcsuite/btcd/wire"
+
+// FirstDuplicate returns the first repeated outpoint in the given list.
+func FirstDuplicate(outpoints []wire.OutPoint) (wire.OutPoint, bool) {
```
starius: The first output is used only in error messages.

Can we replace both FirstDuplicate and HasDuplicates with:
```go
func CheckDuplicates(outpoints []wire.OutPoint) error
```
?

staticaddr/loopin/actions_test.go
```diff
@@ -24,6 +25,82 @@ import (
 	"google.golang.org/grpc"
 )
 
+func TestHandleInvoiceUpdate(t *testing.T) {
```
starius, line 30: godoc

staticaddr/loopin/actions.go
```diff
@@ -359,6 +359,35 @@ func (f *FSM) cancelSwapInvoice() {
 	}
 }
 
+// handleInvoiceUpdate applies the monitor state's invoice-update semantics and
+// reports whether the update produced a terminal event.
+func (f *FSM) handleInvoiceUpdate(update lndclient.InvoiceUpdate) (
```
starius, line 365: I propose to split this commit to two:
 - refactoring (creation of `handleInvoiceUpdate` and `TestHandleInvoiceUpdate` with existing test cases)
 - handle closed invoice updates (targeted changes in `handleInvoiceUpdate` and adding new test cases)

staticaddr/deposit/fsm.go
```diff
@@ -160,12 +161,18 @@ type FSM struct {
...
+
 	// quitChan stops after the FSM stops consuming blockNtfnChan.
 	quitChan chan struct{}
 
 	// finalizedDepositChan is used to signal that the deposit has been
 	// finalized and the FSM can be removed from the manager's memory.
 	finalizedDepositChan chan wire.OutPoint
+
+	// stopOnce ensures Stop is idempotent.
+	stopOnce sync.Once
```
starius: Put it behind `stopChan`, so they go together.

staticaddr/loopin/txout_checker.go
```diff
@@ -0,0 +1,72 @@
...
+	endHeight := int32(0)
+	if includeMempool {
+		endHeight = -1
+	}
+
+	// We need lnd's wallet transaction view rather than only the funding
+	// transaction: a matching previous outpoint tells us the deposit has
+	// already been spent by a wallet-known transaction. When mempool spends
+	// matter, lnd exposes them through ListTransactions with endHeight=-1.
+	txs, err := c.client.ListTransactions(ctx, 0, endHeight)
```
starius: GetTxOut is called per outpoint and loads the whole tx history. So `checkDepositsAvailable` is O(selected_deposits * wallet_tx_history), including unmarshalling raw txs and scanning previous outpoints.

I propose to optimize it:
 - load the heavy thing once, not per deposit
 - switch to a lighter API, e.g. `WalletKit.ListUnspent` if it works here

One more subtlety: current `includeMempool=false` is probably not doing what the comment implies. In lnd, GetTransactions with `end_height=0` defaults to including unconfirmed txs. The current caller passes `true`, so not harmful here, but the abstraction is misleading.

Commit "staticaddr/loopin: scan wallet txouts once" claims to keep the availability check cheap, but it does not reduce the number of expensive `ListTransactions` calls.

staticaddr/deposit/deposit.go
```diff
@@ -73,6 +73,10 @@ func (d *Deposit) IsInFinalState() bool {
 	d.Lock()
 	defer d.Unlock()
 
+	return d.isInFinalStateNoLock()
+}
+
+func (d *Deposit) isInFinalStateNoLock() bool {
```
starius, line 84: godoc

starius, line 84: BTW should we make the follwoing methods private?
GetStateNoLock
SetStateNoLock
IsInStateNoLock

staticaddr/loopin/actions.go
```diff
@@ -565,6 +565,10 @@ func (f *FSM) MonitorInvoiceAndHtlcTxAction(ctx context.Context,
 			subscribeCtx, f.loopIn.SwapHash,
 		)
 	if err != nil {
+		if ctx.Err() != nil {
```
starius, line 636: wrap lines in this commit "staticaddr/loopin: keep htlc monitor resumable"

staticaddr/loopin/loopin_test.go
```diff
@@ -147,6 +147,21 @@ func TestCreateHtlcSweepTxSweepValue(t *testing.T) {
 			"the HTLC output")
 }
 
+func TestPaymentTimeoutDuration(t *testing.T) {
```
starius, line 152: godoc

staticaddr/loopin/loopin_test.go
```diff
@@ -147,6 +147,21 @@ func TestCreateHtlcSweepTxSweepValue(t *testing.T) {
...
+	require.Equal(
+		t, time.Duration(DefaultPaymentTimeoutSeconds)*time.Second,
+		(&StaticAddressLoopIn{}).PaymentTimeoutDuration(),
+	)
+	require.Equal(
+		t, 42*time.Second,
+		(&StaticAddressLoopIn{
+			PaymentTimeoutSeconds: 42,
+		}).PaymentTimeoutDuration(),
+	)
```
starius: fmt

staticaddr/loopin/actions.go
```diff
@@ -526,6 +531,29 @@ func (f *FSM) SignHtlcTxAction(ctx context.Context,
 	return OnHtlcTxSigned
 }
 
+func (f *FSM) checkDepositsAvailable(ctx context.Context) error {
```
starius, line 537: godoc

staticaddr/deposit/fsm_test.go
hieblmi: fixed

staticaddr/outpoint/duplicates.go
hieblmi: i removed it

staticaddr/outpoint/duplicates.go
hieblmi: Yes, I replaced it with `CheckDuplicates`.

staticaddr/loopin/actions_test.go
hieblmi: added

staticaddr/deposit/fsm.go
hieblmi: done

staticaddr/deposit/deposit.go
hieblmi: done

hieblmi: done

staticaddr/loopin/actions.go
hieblmi: done

staticaddr/loopin/loopin_test.go
hieblmi: done

staticaddr/loopin/actions.go
hieblmi: done

staticaddr/loopin/actions.go
hieblmi: done

staticaddr/loopin/actions.go
hieblmi: > So the same sequence still seems possible:
> 
> 1. deadline fires, invoice is canceled, deposits are unlocked;
> 2. user starts withdraw/open-channel;
> 3. server HTLC confirms before CLTV;
> 4. timeout monitor cannot move the reused deposits into `SweepHtlcTimeout`.
> 
> Here is the test sketch which reproduces this edge case: https://gist.github.com/starius/7fe6001933000ff0f8894fd547abcf9f
> 
> It models:
> 
> 1. deposit is `LoopingIn`
> 2. HTLC confirmation arrives
> 3. payment deadline unlocks the deposit back to `Deposited`
> 4. another realistic flow starts withdrawal via `OnWithdrawInitiated`
> 5. HTLC timeout opens
> 6. monitor returns `OnSweepHtlcTimeout`, but deposit stays `Withdrawing`
> 

I've removed the commit around this comment since the scenario here is unrealistic.

Once the client cancels the invoice the server can't sweep htlc tx.

staticaddr/loopin/actions.go
hieblmi: Addressed in `8f998340` and follow-up `090fa561`.

Before signing, the client now asserts that the stored `DepositOutpoints` snapshot exactly matches the recovered/current `Deposits[].OutPoint` order, so we fail before signing if the client and server would be using different inputs. The follow-up keeps recovered deposits ordered by the stored snapshot when the current outpoints still match it, so restart recovery preserves the signing order shared with the server.

staticaddr/loopin/txout_checker.go
hieblmi: Addressed in `b5289f00`.

`TxOutChecker` now batches the request as `GetTxOuts(ctx, outpoints)`, and the lnd-backed checker calls `ListTransactions(ctx, 0, -1)` once for the full selected set. It returns the available/unspent subset and treats missing entries as unavailable or spent. The `includeMempool` argument is gone from the interface; mempool/unconfirmed spends are included internally because the signing guard always needs them.

hieblmi:
Added one extra follow-up commit: `090fa561 staticaddr/loopin: order recovered deposits by snapshot`.

This keeps restart recovery aligned with the server/client signing order for multi-deposit swaps. When recovered deposit rows still match the stored `DepositOutpoints` snapshot, the store now orders `loopIn.Deposits` by that snapshot before signing. If the current deposit outpoints no longer match the snapshot, recovery leaves the reconstructed deposits unchanged and the signing-time invariant check rejects the mismatch explicitly.

staticaddr/loopin/actions.go
starius, line 842: Can we make the canceled-invoice recovery path retry the deposit unlock instead of assuming it already happened?

The deadline path cancels the invoice and then unlocks deposits:

    cancelInvoice()
    err = f.unlockDeposits(ctx)

but those are two separate side effects. If loopd shuts down, crashes, or the deposit transition fails after the invoice cancellation succeeds but before the deposits are moved back to Deposited, recovery lands back in this action with `LookupInvoice` returning `ContractCanceled`.

In that recovered state we do not create `deadlineChan`, so we never run the deadline unlock path again. Then, if the HTLC never confirmed and the CLTV height arrives, this branch returns `OnSwapTimedOut` directly and the loop-in becomes `Failed`, while the deposits can remain stuck in `LoopingIn`.

So the invariant in the comment below ("don't need to unlock the deposits because that happened when the payment deadline was reached") is not durable. It would be safer to attempt `unlockDeposits` here too when `!htlcConfirmed`, or route `OnSwapTimedOut` through an unlock action that can be retried/recovered.

staticaddr/deposit/fsm.go
starius, line 35: I don't think `fsm.OnError` is safe to classify as always caller-locked.

Some `fsm.OnError` transitions do come from `TransitionDeposits`, where the caller is holding the deposit locks. But deposit expiry actions can also produce `fsm.OnError` internally. For example, `PublishDepositExpirySweepAction` returns `fsm.OnError` on publish failure, and `WaitForExpirySweepAction` returns `fsm.OnError` on notifier/context errors. Those paths originate from the deposit FSM's own block handling, not from `TransitionDeposits`, so no external deposit lock is held.

Because `updateDeposit` skips locking for every event in `lockedEvents`, an internally-produced `fsm.OnError` can call `setStateNoLock` and `Store.UpdateDeposit`, whose new contract says the deposit lock must already be held.

So the event name alone is not enough to decide whether the deposit is locked. Could we remove `fsm.OnError` from this set, or otherwise distinguish the caller-held unlock path from internal expiry error transitions?

starius (approved):
**LGTM**, pending two new comments! 🥇

Also I think these commits are fixups. Could you streamline Git history, please?

- b5289f00 `staticaddr/loopin: scan wallet txouts once`
  PR-internal follow-up. It refines the `TxOutChecker` added earlier in this same PR and the signing check that uses it.

- 617f3fa8 `staticaddr/loopin: update payment timeout comment`
  PR-internal fixup/comment cleanup around the monitor/deadline logic touched by this PR.

- 090fa561 `staticaddr/loopin: order recovered deposits by snapshot`
  PR-internal follow-up to the preserved `DepositOutpoints` / recovery ordering changes.

staticaddr/loopin/actions.go
hieblmi: Agreed this should be fixed, but it predates this PR and is better handled separately from the current hardening branch. I opened #1167 to track routing this timeout path through a recoverable unlock state before entering `Failed`: https://github.com/lightninglabs/loop/issues/1167

staticaddr/deposit/fsm.go
hieblmi: Agreed. This `fsm.OnError`/`lockedEvents` coupling also predates this PR, so I opened #1168 to track fixing it separately by separating generic `fsm.OnError` from caller-held rollback events: https://github.com/lightninglabs/loop/issues/1168

PR #1162, opened 2026-07-02
cmd/loop: normalize numeric replay time zones

starius:
Recorded session replay already normalizes timestamps before comparing CLI output, but the `time.String` timestamp regex only matched alphabetic zone names like UTC and EST. Some systems render the same offset with a numeric zone name, for example -0300 -03, which left the publication deadline unnormalized and caused fixture comparisons to fail.

Match numeric zone names as well, and parse the timestamp using the explicit numeric offset instead of the display zone name. Add a regression test covering RFC3339, alphabetic `time.String` zones, and numeric `time.String` zones.

#### Pull Request Checklist


hieblmi (approved):
LGTM!

PR #1165, opened 2026-07-06
multi: add static address loop in monitor

GustavoStingelin:
Implements monitor support for static address loop-in swaps.

### DIFF
--- a/cmd/loop/testdata/sessions/basic-swaps/04_loop-monitor-static-loop-in.json
+++ b/cmd/loop/testdata/sessions/basic-swaps/04_loop-monitor-static-loop-in.json
@@ -1,0 +1,267 @@
+{
+  "metadata": {
+    "args": [
+      "loop",
+      "monitor",
+      "--network",
+      "regtest"
+    ],
+    "env": {},
+    "version": "0.33.3-beta commit= commit_hash=",
+    "run_error": "recv: rpc error: code = Canceled desc = context canceled",
+    "duration": 25407490958,
+    "clock_start_unix": 1784150305
+  },
+  "events": [
+    {
+      "time_ms": 13,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "send",
+        "message_type": "looprpc.MonitorRequest",
+        "payload": {}
+      }
+    },
+    {
+      "time_ms": 13,
+      "kind": "stdout",
+      "data": {
+        "lines": [
+          "Note: offchain cost may report as 0 after loopd restart during swap\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 16479,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "recv",
+        "message_type": "looprpc.SwapStatus",
+        "payload": {
+          "amt": "500000",
+          "id": "bec3777d45337f652d96136e5714ed90a5d4f7cda87b5842110ecc38c9a65826",
+          "id_bytes": "vsN3fUUzf2UtlhNuVxTtkKXU982oe1hCEQ7MOMmmWCY=",
+          "type": "STATIC_LOOP_IN",
+          "state": "INITIATED",
+          "static_loop_in_state": "INIT_HTLC",
+          "failure_reason": "FAILURE_REASON_NONE",
+          "initiation_time": "1784150320647961000",
+          "last_update_time": "1784150321494895000",
+          "htlc_address": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2wsh": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2tr": "",
+          "cost_server": "0",
+          "cost_onchain": "0",
+          "cost_offchain": "0",
+          "last_hop": "",
+          "outgoing_chan_set": [],
+          "label": "",
+          "asset_info": null
+        }
+      }
+    },
+    {
+      "time_ms": 16479,
+      "kind": "stdout",
+      "data": {
+        "lines": [
+          "2026-07-15T18:18:41-03:00 STATIC_LOOP_IN INIT_HTLC 0.00500000 BTC - P2WSH: bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 16482,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "recv",
+        "message_type": "looprpc.SwapStatus",
+        "payload": {
+          "amt": "500000",
+          "id": "bec3777d45337f652d96136e5714ed90a5d4f7cda87b5842110ecc38c9a65826",
+          "id_bytes": "vsN3fUUzf2UtlhNuVxTtkKXU982oe1hCEQ7MOMmmWCY=",
+          "type": "STATIC_LOOP_IN",
+          "state": "INITIATED",
+          "static_loop_in_state": "SIGN_HTLC_TX",
+          "failure_reason": "FAILURE_REASON_NONE",
+          "initiation_time": "1784150320647961000",
+          "last_update_time": "1784150321499639000",
+          "htlc_address": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2wsh": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2tr": "",
+          "cost_server": "0",
+          "cost_onchain": "0",
+          "cost_offchain": "0",
+          "last_hop": "",
+          "outgoing_chan_set": [],
+          "label": "",
+          "asset_info": null
+        }
+      }
+    },
+    {
+      "time_ms": 16482,
+      "kind": "stdout",
+      "data": {
+        "lines": [
+          "2026-07-15T18:18:41-03:00 STATIC_LOOP_IN SIGN_HTLC_TX 0.00500000 BTC - P2WSH: bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 16530,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "recv",
+        "message_type": "looprpc.SwapStatus",
+        "payload": {
+          "amt": "500000",
+          "id": "bec3777d45337f652d96136e5714ed90a5d4f7cda87b5842110ecc38c9a65826",
+          "id_bytes": "vsN3fUUzf2UtlhNuVxTtkKXU982oe1hCEQ7MOMmmWCY=",
+          "type": "STATIC_LOOP_IN",
+          "state": "INITIATED",
+          "static_loop_in_state": "MONITOR_INVOICE_HTLC_TX",
+          "failure_reason": "FAILURE_REASON_NONE",
+          "initiation_time": "1784150320647961000",
+          "last_update_time": "1784150321548450000",
+          "htlc_address": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2wsh": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2tr": "",
+          "cost_server": "0",
+          "cost_onchain": "0",
+          "cost_offchain": "0",
+          "last_hop": "",
+          "outgoing_chan_set": [],
+          "label": "",
+          "asset_info": null
+        }
+      }
+    },
+    {
+      "time_ms": 16530,
+      "kind": "stdout",
+      "data": {
+        "lines": [
+          "2026-07-15T18:18:41-03:00 STATIC_LOOP_IN MONITOR_INVOICE_HTLC_TX 0.00500000 BTC - P2WSH: bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 17027,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "recv",
+        "message_type": "looprpc.SwapStatus",
+        "payload": {
+          "amt": "500000",
+          "id": "bec3777d45337f652d96136e5714ed90a5d4f7cda87b5842110ecc38c9a65826",
+          "id_bytes": "vsN3fUUzf2UtlhNuVxTtkKXU982oe1hCEQ7MOMmmWCY=",
+          "type": "STATIC_LOOP_IN",
+          "state": "INITIATED",
+          "static_loop_in_state": "PAYMENT_RECEIVED",
+          "failure_reason": "FAILURE_REASON_NONE",
+          "initiation_time": "1784150320647961000",
+          "last_update_time": "1784150322044521000",
+          "htlc_address": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2wsh": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2tr": "",
+          "cost_server": "1828",
+          "cost_onchain": "0",
+          "cost_offchain": "0",
+          "last_hop": "",
+          "outgoing_chan_set": [],
+          "label": "",
+          "asset_info": null
+        }
+      }
+    },
+    {
+      "time_ms": 17027,
+      "kind": "stdout",
+      "data": {
+        "lines": [
+          "2026-07-15T18:18:42-03:00 STATIC_LOOP_IN PAYMENT_RECEIVED 0.00500000 BTC - P2WSH: bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv (cost: server 1828, onchain 0, offchain 0)\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 17036,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "recv",
+        "message_type": "looprpc.SwapStatus",
+        "payload": {
+          "amt": "500000",
+          "id": "bec3777d45337f652d96136e5714ed90a5d4f7cda87b5842110ecc38c9a65826",
+          "id_bytes": "vsN3fUUzf2UtlhNuVxTtkKXU982oe1hCEQ7MOMmmWCY=",
+          "type": "STATIC_LOOP_IN",
+          "state": "INITIATED",
+          "static_loop_in_state": "SUCCEEDED",
+          "failure_reason": "FAILURE_REASON_NONE",
+          "initiation_time": "1784150320647961000",
+          "last_update_time": "1784150322052572000",
+          "htlc_address": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2wsh": "bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv",
+          "htlc_address_p2tr": "",
+          "cost_server": "1828",
+          "cost_onchain": "0",
+          "cost_offchain": "0",
+          "last_hop": "",
+          "outgoing_chan_set": [],
+          "label": "",
+          "asset_info": null
+        }
+      }
+    },
+    {
+      "time_ms": 17036,
+      "kind": "stdout",
+      "data": {
+        "lines": [
+          "2026-07-15T18:18:42-03:00 STATIC_LOOP_IN SUCCEEDED 0.00500000 BTC - P2WSH: bcrt1qjsl6tmv2vf3p5wwv28jh6xwettds934k02c3l5w6zyjsyrsd3lfskgdjmv (cost: server 1828, onchain 0, offchain 0)\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 25406,
+      "kind": "signal",
+      "data": {
+        "signal": "interrupt"
+      }
+    },
+    {
+      "time_ms": 25407,
+      "kind": "grpc",
+      "data": {
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "error",
+        "error": "rpc error: code = Canceled desc = context canceled",
+        "status": {
+          "code": 1,
+          "message": "context canceled"
+        }
+      }
+    },
+    {
+      "time_ms": 25407,
+      "kind": "stderr",
+      "data": {
+        "lines": [
+          "[loop] recv: rpc error: code = Canceled desc = context canceled\n"
+        ]
+      }
+    },
+    {
+      "time_ms": 25407,
+      "kind": "exit",
+      "data": {
+        "run_error": "recv: rpc error: code = Canceled desc = context canceled"
+      }
+    }
+  ]
+}
--- a/loopd/daemon.go
+++ b/loopd/daemon.go
@@ -110,6 +110,10 @@ type Daemon struct {
 	macaroonService *lndclient.MacaroonService
 }
 
+// staticLoopInStatusChanBuffer keeps the shared swap-status fanout from
+// stalling static loop-in FSM progress during transient subscriber gaps.
+const staticLoopInStatusChanBuffer = 20
+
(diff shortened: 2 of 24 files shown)


GustavoStingelin:
Looking for ConceptACK/suggestions. 

Still need to test.

loopd/swapclient_server.go
```diff
@@ -412,6 +412,21 @@ func (s *swapClientServer) marshallSwap(ctx context.Context,
...
 		}
 
+	case swap.TypeStaticAddressLoopIn:
+		swapType = looprpc.SwapType_STATIC_ADDRESS_LOOP_IN
+
+		if loopSwap.HtlcAddressP2TR != nil {
+			htlcAddressP2TR = loopSwap.HtlcAddressP2TR.EncodeAddress()
+			htlcAddress = htlcAddressP2TR
+		} else {
+			htlcAddressP2WSH = loopSwap.HtlcAddressP2WSH.EncodeAddress()
```
starius: Static address is always a P2TR, never a P2WSH, so we can just assert that loopSwap.HtlcAddressP2TR is set and return an error otherwise.

loopd/swapclient_server.go
```diff
@@ -693,7 +713,9 @@ func filterSwap(swapInfo *loop.SwapInfo, filter *looprpc.ListSwapsFilter) bool {
 
 	// If the swap is of type loop in and the last hop filter is set, we
 	// only return swaps that match the filter.
-	if swapInfo.SwapType == swap.TypeIn && filter.LoopInLastHop != nil {
+	if (swapInfo.SwapType == swap.TypeIn ||
+		swapInfo.SwapType == swap.TypeStaticAddressLoopIn) &&
```
starius: We can use `!swapType.IsOut()` or even `swapType.IsIn()` instead of `(swapInfo.SwapType == swap.TypeIn || swapInfo.SwapType == swap.TypeStaticAddressLoopIn)`

So we can keep two distinct SwapType instances for loop-in and static loop-in, but use `IsOut/IsIn` in situations where both loop-in and static loop-in work the same way.

looprpc/client.proto
```diff
@@ -734,6 +737,9 @@ message ListSwapsFilter {
 
         // LOOP_IN indicates a loop in swap (on-chain to off-chain).
         LOOP_IN = 2;
+
+        // STATIC_ADDRESS_LOOP_IN indicates a static address loop in swap.
+        STATIC_ADDRESS_LOOP_IN = 3;
```
starius: This change seems unrelated to the `loop monitor` integration of static loop-ins. If we need it for completeness, I propose to do it in a separate commit or in a follow-up PR.

starius (commented):
Concept ACK for API shape.

One _implementation_ note. `Monitor` snapshots and subscribes only to `s.swaps`, populated from `loop.Client.FetchSwaps` and live `statusChan` updates. Static address loop-ins live under `staticLoopInManager.GetAllSwaps()` and (`ListStaticAddressSwaps`), so the current changes alone will not make `loop monitor` see them.

Some bridging is needed to deliver static loop-in updates into monitor's SwapStatus stream: initial snapshot plus live state updates. That also requires an explicit state mapping from static loop-in FSM states to generic SwapState, and filling the generic SwapInfo fields safely, especially HTLC address, amount, label, last hop, timestamps, and costs.

loopd/swapclient_server.go
hieblmi: +1

hieblmi:
Thanks @GustavoStingelin for putting up this draft PR!

I am looking a bit ahead of where we'd update the monitor with the static state, and it should probably be `updateLoopIn` which is triggered by the state machine after every fsm action. Given that this is called
frequently we should probably not do too much work for the monitor here and slow down the state
transitions.

Keeping future swap types in mind maybe it would make sense to send lightweight notifications
with minimum swap info to the monitor instance, and then let the monitor instance retrieve and assemble
the full swap info.

Maybe we could have a small monitor publisher in each existing and new swap config that does something like:
```
f.cfg.MonitorPublisher.Notify(MonitorRef{
      Type: MonitorStaticLoopIn,
      ID:   f.loopIn.SwapHash,
  })
```

What do you think @GustavoStingelin @starius ?

GustavoStingelin:
@hieblmi

> Maybe we could have a small monitor publisher in each existing and new swap config that does something like

I like the idea, but I have two questions:

1. Is this expected to be included in this PR?

2. IIUC could this create a race between the notification and the monitor full swap info assemble? For example:

```
state changes, notification is sent
state changes again, another notification is sent
monitor is still assembling the first notification
monitor reads the latest state, so the first event is shown with the second state
monitor then processes the second notification and shows the same state again
```

In that case, we could show the second state twice in the monitor and lose the first state transition from the monitor log.

looprpc/client.proto
GustavoStingelin: The thing here is, how can we tell the monitor that this is a `static loop in` swap and that we are sending a new notification type, not the classic `swap in` or `swap out` notification?

`SwapStatus` is currently a flat struct returned in the `rpc Monitor`, and `SwapType` is the field that carries this information. Adding this option here was the best design I found so far.

Are you thinking about another option? How would that look?

loopd/swapclient_server.go
GustavoStingelin: I kind of agree. My only concern is that the difference between `TypeStaticAddressLoopIn`, `TypeIn`, and `type.IsIn()` could become a bit brittle.

Internally, is it common in the codebase to classify `TypeStaticAddressLoopIn` as a generic `In` swap? If yes, I am okay with that.

If not, I think we should establish a clear separation in how we name and classify these swap types, so we do not introduce ambiguity later.

GustavoStingelin:
I am thinking that, in the future, we could have something like this:

```proto
message MonitorEvent {
  string id = 1;
  bytes id_bytes = 2;
  SwapKind kind = 3;
  int64 amount = 4;
  int64 initiation_time = 5;
  int64 last_update_time = 6;
  string label = 7;

  // Human friendly normalized lifecycle.
  MonitorPhase phase = 8;

  // Machine specific exact state.
  string state = 9;

  oneof details {
    LoopInMonitorDetails loop_in = 20;
    LoopOutMonitorDetails loop_out = 21;
    StaticLoopInMonitorDetails static_loop_in = 22;
    // Future swap types can be added here.
  }
}
```

Where:

`kind` answers, what is this?

```text
LOOP_IN
LOOP_OUT
STATIC_LOOP_IN
future: INSTANT_OUT, ASSET_LOOP_OUT, etc.
```

`phase` answers, where is it broadly in the lifecycle?

```text
INITIATING
WAITING_FOR_PAYMENT
WAITING_FOR_CHAIN
SWEEPING
SUCCESS
FAILED
```

`state` answers, what is the exact implementation state?

```text
SIGN_HTLC_TX
MONITOR_INVOICE_HTLC_TX
PREIMAGE_REVEALED
etc.
```

The idea is to separate the high level swap category, the normalized monitor lifecycle, and the exact internal state. That should make the monitor API easier to extend without overloading one field with multiple meanings.

WDYT @hieblmi @starius?

hieblmi:
> I am thinking that, in the future, we could have something like this:
> 
> ```proto
> message MonitorEvent {
>   string id = 1;
>   bytes id_bytes = 2;
>   SwapKind kind = 3;
>   int64 amount = 4;
>   int64 initiation_time = 5;
>   int64 last_update_time = 6;
>   string label = 7;

 from `SwapKind` to `label` all fields should be available in the swap details below. Do we need them here?

> 
>   // Human friendly normalized lifecycle.
>   MonitorPhase phase = 8;
> 
>   // Machine specific exact state.
>   string state = 9;

`state` here too is part of the swap details.

> 
>   oneof details {
>     LoopInMonitorDetails loop_in = 20;
>     LoopOutMonitorDetails loop_out = 21;
>     StaticLoopInMonitorDetails static_loop_in = 22;
>     // Future swap types can be added here.
>   }
> }
> ```

hieblmi:
> @hieblmi
> 
> > Maybe we could have a small monitor publisher in each existing and new swap config that does something like
> 
> I like the idea, but I have two questions:
> 
> 1. Is this expected to be included in this PR?


I tend to not assemble info in the static address state machine and then send it to the monitor, rather just send a lightweight reference which the monitor can assemble from.

> 2. IIUC could this create a race between the notification and the monitor full swap info assemble? For example:
> 
> ```
> state changes, notification is sent
> state changes again, another notification is sent
> monitor is still assembling the first notification
> monitor reads the latest state, so the first event is shown with the second state
> monitor then processes the second notification and shows the same state again
> ```
> 
> In that case, we could show the second state twice in the monitor and lose the first state transition from the monitor log.

This is a good point! We might cache the swap state at that point in time to not miss anything, but that might defeat
the purpose of being performant.

I think as long as the data we are returning to the monitor is already in memory we can think about this optimization later.

GustavoStingelin:
@hieblmi 
> from SwapKind to label all fields should be available in the swap details below. Do we need them here?

Cool, I totally agree.

I think doing that broader refactor in this PR would be too much, but I’m also not very happy with my current local code. It has some ugly workarounds to support both state models.

So what do you think about shifting to the new monitor approach now? Or would you prefer to take a look at my current code first and decide from there?

Also, just to confirm, since we release the client and loopd together, a breaking protobuf API change should not be a problem, right?

cmd/loop/testdata/sessions/basic-swaps/04_loop-monitor-static-loop-in.json
```diff
@@ -0,0 +1,111 @@
...
+        "method": "/looprpc.SwapClient/Monitor",
+        "event": "recv",
+        "message_type": "looprpc.SwapStatus",
+        "payload": {
+          "amt": "50000",
+          "id": "3f4ec343ba4087eec3cdf501b4e19733d9685cc151db360bb67d2c924677ce71",
+          "id_bytes": "P07DQ7pAh+7DzfUBtOGXM9loXMFR2zYLtn0skkZ3znE=",
+          "type": "STATIC_LOOP_IN",
+          "state": "INITIATED",
+          "static_loop_in_state": "SIGN_HTLC_TX",
```
GustavoStingelin, line 87: @hieblmi It is the thing that I'm mostly concerned about

loopd/swapclient_server.go
```diff
@@ -2059,6 +2169,166 @@ func staticAddressLoopInSwapServerCost(swp *loopin.StaticAddressLoopIn) int64 {
 	}
 }
 
+func (s *swapClientServer) staticAddressLoopInSwapInfos(
```
starius, line 2266: Please add godoc for each new symbol, including private methods. The goal (sometimes not achieved) in this code base is that everything has a godoc, including test files, struct members, consts, etc, i.e. all symbols.

Also please comment non-trivial logic in the code, prepending code comments, explaining what is going on, before come code sections.

The purpose of this is to explain to code readers (both human and LLMs) what this thing does or what is about to happen.

loopd/swapclient_server.go
```diff
@@ -2059,6 +2169,166 @@ func staticAddressLoopInSwapServerCost(swp *loopin.StaticAddressLoopIn) int64 {
...
+			Label:           swp.Label,
+			ProtocolVersion: loopdb.ProtocolVersion(
+				swp.ProtocolVersion,
+			),
+		},
+		LastUpdate:               lastUpdate,
+		SwapHash:                 swp.SwapHash,
+		SwapType:                 swap.TypeStaticAddressLoopIn,
+		StaticAddressLoopInState: swp.GetState(),
+		HtlcAddressP2TR:          staticAddress,
```
starius: HTLC address is not the same as staticAddress. These are two different addresses and it is very important to separate them. We got issues previously when people sent funds to HTLC addresses by mistake.

HTLC address also exists and is a property of a static loop-in (as well as of other swap types). It should be loaded/computed and put here.

Static address is a long-living address which is reused by (potentially) many deposits and many static loop-ins. Probably we need a new field to pass it here. The new field for the static address will only be filled if the swap is a static loop-in.

starius:
@claude review

loopd/swapclient_server.go
```diff
@@ -2059,6 +2169,166 @@ func staticAddressLoopInSwapServerCost(swp *loopin.StaticAddressLoopIn) int64 {
...
+		return loopdb.StateHtlcPublished
+
+	case loopin.Succeeded:
+		return loopdb.StateSuccess
+
+	case loopin.HtlcTimeoutSwept:
+		return loopdb.StateFailTimeout
+
+	case loopin.Failed, loopin.SucceededTransitioningFailed:
+		return loopdb.StateFailAbandoned
```
starius: I think `loopdb.StateFailAbandoned` is a wrong state for Failed and SucceededTransitioningFailed.

@hieblmi What do you think?

loopd/swapclient_server.go
```diff
@@ -575,6 +617,71 @@ func (s *swapClientServer) Monitor(in *looprpc.MonitorRequest,
...
+
+	if swp.SwapType != swap.TypeStaticAddressLoopIn {
+		return false
+	}
+	lastUpdate, ok := highWater[swp.SwapHash]
+	if !ok {
+		return false
+	}
+
+	return !swp.LastUpdate.After(lastUpdate)
```
starius: I think we should differentiate by swap state as well, not only by time. If two updates share the same time, but have different states, they should not be deduplicated.

loopd/daemon.go
```diff
@@ -661,6 +661,7 @@ func (d *Daemon) initialize(withMacaroonService bool) error {
 		LightningClient:   d.lnd.Client,
 	}
 	openChannelManager = openchannel.NewManager(openChannelCfg)
+	statusChan := make(chan loop.SwapInfo)
```
starius: The channel is not buffered. There is a write to it in `staticLoopInStatusUpdater.sendUpdate` which can potentially block. `sendUpdate` is called from a swap's `SendUpdate`, so this may block the swap's progress, which is not great.

I propose to make this channel buffered to mitigate this risk.

loopd/swapclient_server.go
```diff
@@ -2059,6 +2090,166 @@ func staticAddressLoopInSwapServerCost(swp *loopin.StaticAddressLoopIn) int64 {
...
+	if lastUpdate.IsZero() {
+		lastUpdate = swp.InitiationTime
+	}
+
+	return &loop.SwapInfo{
+		SwapStateData: loopdb.SwapStateData{
+			State: staticAddressLoopInGenericState(swp.GetState()),
+			Cost: loopdb.SwapCost{
+				Server: btcutil.Amount(
+					staticAddressLoopInSwapServerCost(swp),
```
starius, line 2344: I would maybe add/keep a comment saying this mirrors `ListStaticAddressSwaps`: only actual persisted client-visible server cost is reported; onchain/offchain remain zero until static loop-ins persist real fee data.

loopd/swapclient_server.go
```diff
@@ -2510,7 +2780,9 @@ func (s *swapClientServer) processStatusUpdates(mainCtx context.Context) {
 		// subscribers about the changes.
 		case swp := <-s.statusChan:
 			s.swapsLock.Lock()
-			s.swaps[swp.SwapHash] = swp
+			if swp.SwapType != swap.TypeStaticAddressLoopIn {
```
starius, line 2862: Claude raised an interesting point. We do now have an inconsistency of `loop monitor` and `loop listswaps` / `loop swapinfo`.

Users might see a swap hash in `loop monitor` and try to check its information in `loop listswaps` or `loop swapinfo` and be confused that the swap is missing (because it is a static loop-in).

I propose to add a note on CLI side to `loop listswaps` output and its docs that it covers only traditional loop-ins and loop-outs, not static loop-ins. We have command `loop static listswaps` which should be used instead - tell this in the message.

For `loop swapinfo <swap>` we don't have a direct replacement in the static world. @hieblmi would it make sense to extend `loop swapinfo <swap>` so it works for static loop-ins as well? Or is it better to add `loop static swapinfo <swap>` command and in `loop swapinfo` tell the user to use the new command instead if we detect that the hash is of a static loop-in?

starius (commented):
**Great progress!** :rocket: 

I found a couple of weak points in this PR, also we need to sync with @hieblmi about a couple of changes (see below).

Could you also go through Claude's feedback and address the real issues, please?

loopd/daemon.go
GustavoStingelin: yeah, it was previously blocking in the old loopin/out, will change it

loopd/swapclient_server.go
hieblmi: Imo we shouldn't show the htlc address in the static world as this is a backup transaction by the server that is rarely ever shown as we try to avoid publishing it and instead sweep directly.

 If we still think this is useful information we could recreate like it is done in `loopin.go`:
```
func (l *StaticAddressLoopIn) getHtlc(chainParams *chaincfg.Params)
```

loopd/swapclient_server.go
hieblmi: Technically not correct as abandoned is reserved in the old loop world for manual swap abandonment, if a swap is borked e.g., similar to how `lnd` allows abandoning channels.

@GustavoStingelin do we need the mapping of the static states into the old states?
I think we can just rely on the static swap state and leave the old state empty? So just use
```
	StaticAddressLoopInState: swp.GetState(),
```

loopd/swapclient_server.go
hieblmi: > 
> For `loop swapinfo <swap>` we don't have a direct replacement in the static world. @hieblmi would it make sense to extend `loop swapinfo <swap>` so it works for static loop-ins as well? Or is it better to add `loop static swapinfo <swap>` command and in `loop swapinfo` tell the user to use the new command instead if we detect that the hash is of a static loop-in?

I think we should either add `loop static swapinfo <swap>` or extend existing `loop static listswaps` with a `--swap_hash` param. In the future the goal is to get rid of `loop swapinfo` respectively replace it with `loop static [swapinfo|listswaps]`

loopd/swapclient_server.go
```diff
@@ -729,10 +836,13 @@ func (s *swapClientServer) SwapInfo(ctx context.Context,
 
 	// Just return the server's in-memory cache here too as we also want to
 	// return temporary failures to the client.
+	s.swapsLock.Lock()
+	defer s.swapsLock.Unlock()
```
hieblmi: the unlock can be done right after `s.swaps` has been read. This is to avoid holding it if `marsahllSwap` calls the asset client.

staticaddr/loopin/fsm.go
```diff
@@ -273,6 +273,19 @@ func (f *FSM) updateLoopIn(ctx context.Context, notification fsm.Notification) {
 
 		return
 	}
+
+	err = f.sendUpdate(ctx)
```
hieblmi, line 277: Here comes back my initial performance concern, since we don't want to slow further down the state transitions of the state machine. 

We should add godocs pointing out that `sendUpdate` is not supposed to do heavy lifting. If the updates become to heavy(like calculating htlc address from given parameters or something else) we could queue the monitor updates. But for now lgtm.

hieblmi (commented):
Thanks for the changes so far @GustavoStingelin 🚀 I've left some comments.

Overall I wonder if we could simplify the monitoring approach without having to modify all too much of the existing flows and also be extensible when future swap types need to be monitored.

What about if we seed a `staticSwaps map[lntypes.Hash]loop.SwapInfo` once from the database during daemon startup, then update it in processStatusUpdates. Monitor could register and snapshot both legacy and static maps under the same lock.

That would eliminate:
  - Full database reads on every monitor connection.
  - Snapshot/live timestamp deduplication.
  - Most of the race-specific logic.
  - The need to put static swaps into the legacy s.swaps map or expose them through
    ListSwaps.

hieblmi:
Just tested the current set of changes in regtest by submitting a static and legacy swap, looks good:
```
loop --network=regtest monitor               ✔  20:14:42 
Note: offchain cost may report as 0 after loopd restart during swap
2026-07-14T20:15:12+02:00 STATIC_LOOP_IN INIT_HTLC 0.00250000 BTC - P2TR: bcrt1phnvwg3wg70e4asnn5wty05v57vcd6xvrk8cek98muw0etzkgvw3q6xmxm5
2026-07-14T20:15:12+02:00 STATIC_LOOP_IN SIGN_HTLC_TX 0.00250000 BTC - P2TR: bcrt1phnvwg3wg70e4asnn5wty05v57vcd6xvrk8cek98muw0etzkgvw3q6xmxm5
2026-07-14T20:15:13+02:00 STATIC_LOOP_IN MONITOR_INVOICE_HTLC_TX 0.00250000 BTC - P2TR: bcrt1phnvwg3wg70e4asnn5wty05v57vcd6xvrk8cek98muw0etzkgvw3q6xmxm5
2026-07-14T20:15:13+02:00 STATIC_LOOP_IN PAYMENT_RECEIVED 0.00250000 BTC - P2TR: bcrt1phnvwg3wg70e4asnn5wty05v57vcd6xvrk8cek98muw0etzkgvw3q6xmxm5 (cost: server 1825, onchain 0, offchain 0)
2026-07-14T20:15:13+02:00 STATIC_LOOP_IN SUCCEEDED 0.00250000 BTC - P2TR: bcrt1phnvwg3wg70e4asnn5wty05v57vcd6xvrk8cek98muw0etzkgvw3q6xmxm5 (cost: server 1825, onchain 0, offchain 0)
2026-07-14T20:15:42+02:00 LOOP_IN INITIATED 0.00250000 BTC - P2TR: bcrt1p0u95wwwnxd00w4qyj6yaf2latsnsnmfg2lxsckpqqpunu5fgfk7smlgpgq
2026-07-14T20:15:42+02:00 LOOP_IN HTLC_PUBLISHED 0.00250000 BTC - P2TR: bcrt1p0u95wwwnxd00w4qyj6yaf2latsnsnmfg2lxsckpqqpunu5fgfk7smlgpgq
2026-07-14T20:16:00+02:00 LOOP_IN INVOICE_SETTLED 0.00250000 BTC - P2TR: bcrt1p0u95wwwnxd00w4qyj6yaf2latsnsnmfg2lxsckpqqpunu5fgfk7smlgpgq (cost: server 0, onchain 3850, offchain 0)
```

Noticed that the on-chain fee of a static swap is currently `0` but should be the sum of the deposit fees of a swap. Opened an issue for it https://github.com/lightninglabs/loop/issues/1174

loopd/swapclient_server.go
```diff
@@ -412,6 +412,21 @@ func (s *swapClientServer) marshallSwap(ctx context.Context,
...
+	case swap.TypeStaticAddressLoopIn:
+		swapType = looprpc.SwapType_STATIC_ADDRESS_LOOP_IN
+
+		if loopSwap.HtlcAddressP2TR != nil {
+			htlcAddressP2TR = loopSwap.HtlcAddressP2TR.EncodeAddress()
+			htlcAddress = htlcAddressP2TR
+		} else {
+			htlcAddressP2WSH = loopSwap.HtlcAddressP2WSH.EncodeAddress()
+			htlcAddress = htlcAddressP2WSH
+		}
```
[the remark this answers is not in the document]
GustavoStingelin: now we r setting only htlcAddressP2WSH, static loop in only use P2WSH in htlc

loopd/swapclient_server.go
GustavoStingelin: fixed by using only P2WSH bc it's the HTLC address, not the static one used for deposit

loopd/swapclient_server.go
GustavoStingelin: the code changed a bit, lmk if u have more thoughts on this

loopd/swapclient_server.go
GustavoStingelin: added, lmk if there is any specific one that u think we should better explain

loopd/swapclient_server.go
GustavoStingelin: changed to not map like that, now all use the default enum zero value (`StateInitiated`for old school loop) when it is static

loopd/swapclient_server.go
GustavoStingelin: ok, did that way

loopd/daemon.go
GustavoStingelin: buffered now

loopd/swapclient_server.go
GustavoStingelin: added the CLI observation, I guess we can create some good follow-up issues from this PR, right?

loopd/swapclient_server.go
GustavoStingelin: cool, I think it's safe for the current execution; otherwise, we would need to do a deep copy of the returned value first

GustavoStingelin: applied

staticaddr/loopin/fsm.go
GustavoStingelin: added godocs in the sendUpdate function definition

loopd/swapclient_server.go
GustavoStingelin: We are no longer putting the reusable static address into the HTLC field. Now, we only set `HtlcAddressP2WSH` to the actual HTLC address.

@hieblmi Yeah, if it is not needed, I agree that we should remove it and simplify this even further. However, I would also prefer not to expose `HtlcAddressP2WSH` here if it should not be populated. That would bring us back to the broader monitor refactor, though.

If you all agree that it should be left empty for now, I can change it. Otherwise, we can open a follow up issue for a broader refactor. It is up to you.

GustavoStingelin:
The other RPC call changes are deferred for a future follow-up issue/pr.

GustavoStingelin:
@starius @hieblmi Thanks for the reviews! All comments addressed.

hieblmi (changes requested):
There seems to be an issue with the autoloop liquidity rules enabled in this PR:
```
STATIC_LOOP_IN liquidity rules silently become Loop Out rules. Adding STATIC_LOOP_IN to the shared looprpc/client.proto:612 enum also exposes it through looprpc/client.proto:1431. However,
    liquidity/parameters.go:402 treats only LOOP_IN specially and defaults every other value to swap.TypeOut. An API client selecting the newly valid STATIC_LOOP_IN value would therefore configure automation in the opposite direction and potentially incur unintended swaps and fees.
```

I think we should reject `STATIC_LOOP_IN` in `liquidity/parameters.go:402`.

GustavoStingelin:
@hieblmi cool, fixed in the first commit ([looprpc: expose static loop-in monitor state](https://github.com/lightninglabs/loop/pull/1165/changes/4c03e469288430e592fecad4e889de2d7d42ebb1)) which introduces the new enum type. 

Also did a rebase.

hieblmi:
Thanks for the fix @GustavoStingelin, I will test this again and aim for a tAck then.

hieblmi (approved):
LGTM 💾

PR #1175, opened 2026-07-16
build: fix dependency security alerts

starius:
This change resolves the current Go dependency security alerts by migrating the remaining PostgreSQL error handling to pgx v5, temporarily selecting the official lnd sqldb pgx v5 migration until it is included in a release, updating the affected crypto, networking, container runtime, database, and supporting dependencies, and raising the application module Go versions to 1.25.12 for the applicable standard-library security fixes.

The module files were tidied after the migration, removing the legacy pgx v4, pgconn, and pgproto3 dependency chain as well as other dependencies that are no longer reachable. The temporary sqldb replacement includes a comment explaining why it is required and when it can be removed.

The static-address open-channel `--memo` help text is normalized to remove embedded indentation and line breaks, so terminal, Markdown, and man-page output use ordinary spacing.

## Testing

- `make docs` was run and the generated man page is included.
- Regenerating the documentation from the final commit produces no changes.
- Standard loop-out completed successfully on regtest.
- Standard loop-in completed successfully on regtest.
- Static-address loop-in completed successfully on regtest.
- Static-address cooperative withdrawal completed and confirmed on regtest.
- Static-address channel opening completed and produced an active channel on regtest.
- An expired static-address deposit was swept successfully by the server on regtest.
- Instant loop-out completed through the sweepless path on regtest.

#### Pull Request Checklist


assets/client.go
```diff
@@ -49,13 +49,16 @@ type TapdConfig struct {
...
+
 	return &TapdConfig{
-		Activate:     false,
-		Host:         "localhost:10029",
-		MacaroonPath: defaultConf.RpcConf.MacaroonPath,
-		TLSPath:      defaultConf.RpcConf.TLSCertPath,
-		RFQtimeout:   defaultRfqTimeout,
+		Activate: false,
+		Host:     "localhost:10029",
+		MacaroonPath: filepath.Join(
```
hieblmi, line 65: the new tapd macaroon path omits data. Tapd stores it at `<tapdDir>/data/<network>/admin.macaroon`, but Loop now looks under `<tapdDir>/mainnet/admin.macaroon`. Enabling tapd therefore fails immediately while reading
     the macaroon.

hieblmi (approved):
LGTM!

assets/client.go
starius: Good catch! I fixed it in a separate commit, since it is quite large and the feature is not critical to have it green in every commit.

PR #1176, opened 2026-07-17
staticaddr/withdraw: prepare deposits before publish

hieblmi:
Neutrino can remove a spent input from its wallet view while transaction
publication is still blocked. If the deposit remains Deposited,
reconciliation can remove it before the withdrawal manager records the
withdrawal intent.

Attach the finalized transaction, transition the deposits to Withdrawing,
and set up local tracking before publishing. Keep the prepared state on
publication errors because the wallet RPC outcome can be ambiguous and
recovery or fee bumping must be able to retry.

### DIFF
--- a/staticaddr/withdraw/manager.go
+++ b/staticaddr/withdraw/manager.go
@@ -439,19 +439,37 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
 		return "", "", err
 	}
 
-	published, err := m.publishFinalizedWithdrawalTx(ctx, finalizedTx)
-	if err != nil {
-		return "", "", err
-	}
-
-	if !published {
-		return "", "", nil
-	}
-
 	withdrawalPkScript, err := txscript.PayToAddrScript(withdrawalAddress)
 	if err != nil {
 		return "", "", fmt.Errorf("could not get withdrawal "+
 			"pkscript: %w", err)
+	}
+
+	// Attach the finalized withdrawal transaction before transitioning the
+	// deposits. The state transition persists it, allowing recovery to
+	// republish the transaction if publishing blocks or the client restarts.
+	previousWithdrawalTxns := make([]*wire.MsgTx, len(deposits))
+	for i, d := range deposits {
+		d.Lock()
+		previousWithdrawalTxns[i] = d.FinalizedWithdrawalTx
+		d.FinalizedWithdrawalTx = finalizedTx
+		d.Unlock()
+	}
+
+	// Transition before publishing so wallet reconciliation can't remove a
+	// spent deposit from the active set while publication is in progress.
+	err = m.cfg.DepositManager.TransitionDeposits(
+		ctx, deposits, deposit.OnWithdrawInitiated, deposit.Withdrawing,
+	)
+	if err != nil {
+		for i, d := range deposits {
+			d.Lock()
+			d.FinalizedWithdrawalTx = previousWithdrawalTxns[i]
+			d.Unlock()
+		}
+
+		return "", "", fmt.Errorf("failed to transition deposits %w",
+			err)
 	}
 
 	// If this is the first time this cluster of deposits is withdrawn, we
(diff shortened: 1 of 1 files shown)


staticaddr/withdraw/manager.go
```diff
@@ -479,49 +497,43 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
 	// If a previous withdrawal existed across the selected deposits, and
 	// it isn't the same as the new withdrawal, we remove it from the
 	// finalized withdrawals to stop republishing it on block arrivals.
-	deposits[0].Lock()
-	prevTx := deposits[0].FinalizedWithdrawalTx
-	deposits[0].Unlock()
+	previousWithdrawalTx := previousWithdrawalTxns[0]
```
starius, line 500: This assumption is pre-existing, but could we fix it while touching this logic? The fee-bump path verifies that all `Withdrawing` deposits reference the same previous transaction, but that invariant is not enforced for `Deposited` deposits or inconsistent recovery states.

Since we now retain every previous transaction, consider deleting every distinct non-nil previous hash other than `finalizedTx`, or explicitly validating that all entries match. Otherwise, a stale transaction associated with a deposit other than index 0 could remain in `finalizedWithdrawalTxns` and continue being republished. A multi-deposit regression test would be useful here.

staticaddr/withdraw/manager.go
```diff
@@ -439,19 +439,37 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
+		d.Lock()
+		previousWithdrawalTxns[i] = d.FinalizedWithdrawalTx
+		d.FinalizedWithdrawalTx = finalizedTx
+		d.Unlock()
 	}
 
-	withdrawalPkScript, err := txscript.PayToAddrScript(withdrawalAddress)
+	// Transition before publishing so wallet reconciliation can't remove a
+	// spent deposit from the active set while publication is in progress.
+	err = m.cfg.DepositManager.TransitionDeposits(
```
starius, line 461: Can we ensure persistence errors are observable here? `TransitionDeposits` returns success even when the FSM’s `Store.UpdateDeposit` fails, because that error is only logged. Since the explicit update now only runs for fee bumps, an initial withdrawal can reach publication while the database still says `Deposited` and has no finalized transaction. That recreates the restart/reconciliation race this change is intended to fix.

staticaddr/withdraw/manager.go
```diff
@@ -439,19 +439,37 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
+		ctx, deposits, deposit.OnWithdrawInitiated, deposit.Withdrawing,
+	)
 	if err != nil {
-		return "", "", fmt.Errorf("could not get withdrawal "+
-			"pkscript: %w", err)
+		for i, d := range deposits {
```
starius, line 465: Is this rollback safe for multiple deposits? `TransitionDeposits` processes FSMs sequentially, so an earlier deposit may already be `Withdrawing` and persisted when a later transition fails. Restoring only `FinalizedWithdrawalTx` would leave that deposit in `Withdrawing` with its old or nil transaction, while the cluster may have mixed states. We likely need atomic persistence/transition or a rollback that also restores state and database records.

staticaddr/withdraw/manager.go
starius, line 492: At this point the deposits are already `Withdrawing`. If `handleWithdrawal` fails, we return before publishing or adding local tracking. A retry is then classified as a fee bump and skips `handleWithdrawal`, so the replacement can publish without any spend/confirmation watcher and remain `Withdrawing` until restart. Could this path retry notifier setup or restore the prepared state before returning?

staticaddr/withdraw/manager.go
```diff
@@ -479,49 +497,43 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
...
-	// Deposits with the same withdrawal tx are part of the same withdrawal.
-	for _, d := range deposits {
-		d.Lock()
-		d.FinalizedWithdrawalTx = finalizedTx
-		d.Unlock()
-	}
-
 	// Add the new withdrawal tx to the finalized withdrawals to republish
 	// it on block arrivals.
 	m.mu.Lock()
```
starius, line 511: Could we add the replacement to this map only after all deposit updates succeed? If an `UpdateDeposit` below fails, the method returns but the old transaction has already been removed and the replacement remains eligible for block-triggered republication. That can broadcast a transaction whose database records are old or only partially updated.

staticaddr/withdraw/manager.go
starius, line 483: Should creating the withdrawal record be part of the durable preparation and return its error? The deposits are already persisted as `Withdrawing`, so a crash or ignored error here allows recovery to publish the transaction without a corresponding withdrawal record. Confirmation then cannot update that record, leaving the withdrawal permanently absent from history.

starius (commented):
Found some potentially fragile parts in this branch. I haven't validated them empirically, so some of them may be false-positives.

Also I propose to add **regression coverage** in a separate commit. If the main fix commit is then reverted, the new test must build and run, but fail with a symptomatic error, reproducing the failure you are fixing. So it is easy to demo that the fix is sound.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #1177, opened 2026-07-17
fix dead error variable and silent invoice cancel in setStateAbandoned

chandachewe10:
## Summary

This change fixes two issues in `setStateAbandoned` in `loopin.go`.

### 1. Removed a misleading `<nil>` from the returned error

After `persistAndAnnounceState` succeeds, execution can only reach the final return if `err` is `nil`, since any non-nil error returns earlier. The previous implementation still included `err` in the formatted error:

```go
return fmt.Errorf("swap hash abandoned by client, swap ID: %v, %v", s.hash, err)
```

As a result, the returned message always ended with `, <nil>`, adding unnecessary noise to client responses and logs. The error message now only includes the relevant swap information.

### 2. Handle `CancelInvoice` errors consistently

Previously, the result of `CancelInvoice` was ignored, causing unexpected failures to go unnoticed.

The abandon path now follows the same behavior as the timeout handling:

- Ignore `ErrInvoiceAlreadySettled`, since the invoice may already have been settled before the swap was abandoned.
- Log any other unexpected error at `Warn` level so operators can diagnose the issue without preventing the abandon process from completing.

## Testing

- `go build ./...`
- `go vet ./...`
- `go test -run TestLoopIn ./...`

### DIFF
--- a/loopin.go
+++ b/loopin.go
@@ -1174,13 +1174,17 @@ func (s *loopInSwap) setStateAbandoned(ctx context.Context) error {
 		return err
 	}
 
-	// If the invoice is already settled or canceled, this is a nop.
-	_ = s.lnd.Invoices.CancelInvoice(ctx, s.hash)
+	// Cancel the invoice so the server can no longer settle it. If the
+	// invoice is already settled we ignore the error, matching the
+	// behaviour of the timeout path. Any other unexpected error is logged
+	// but does not prevent the abandon from completing.
+	err = s.lnd.Invoices.CancelInvoice(ctx, s.hash)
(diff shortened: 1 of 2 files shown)


hieblmi (approved):
Hi @chandachewe10, thank you for the fix!

This LGTM! 💾

starius:
I rebased it and fix error classification in another commit.

lnd returns ErrInvoiceAlreadySettled from its RPC handler as an uncoded gRPC error. On the client it is reconstructed as an Unknown status, so comparing it directly with the Go sentinel never matches.

I verified empirically that previous version still produced a warning in this situation, while the fixed one correctly filters out this error.

PR #1178, opened 2026-07-17
multi: add labels for static addresses

GustavoStingelin:
Adds persistent, locally managed labels for static addresses.
- Stores labels in LoopDB
- Exposes labels through RPC responses and CLI commands
- Supports updating and clearing labels
- Associates labels with the address and UTXOs


staticaddr/address/manager.go
```diff
@@ -119,38 +126,41 @@ func (m *Manager) NewAddress(ctx context.Context) (*btcutil.AddressTaproot,
...
-		expiry := int64(addresses[0].Expiry)
-
 		defer m.Unlock()
 
+		// TODO: When multi-address support lands, resolve an existing address
+		// by label or create a new one when no matching label exists.
+		params := addresses[0]
+		clientPubKey := params.ClientPubkey
+		serverPubKey := params.ServerPubkey
+		expiry := int64(params.Expiry)
```
GustavoStingelin, line 139: Note: reviewers, pls, pay attention here.

starius:
I propose to include the current state of https://github.com/lightninglabs/loop/pull/1139 into this PR, so own label commits go after @hieblmi's commits. Add a note to the PR description not to review the multi-address commits.

I attempted to do a rebase using AI, here is my attempt: https://github.com/starius/loop/commits/staticaddr-combined/ It is not functional.

Some semantic rebase is needed. For example:
- PR 1178 adds `000022_static_address_label`, while PR 1139 already uses migrations 22–24 for multi-address persistence.
- PR 1178 assigns protobuf field 2 in `NewStaticAddressRequest` to `label`; PR 1139 assigns field 2 to `send_coins`.
- PR 1178 assigns protobuf field 3 in `NewStaticAddressResponse` to `label`; PR 1139 assigns field 3 to `send_coins_response`.

Found some issues in that multi-address + labels branch that I got:

 - the multi-address manager returns an internal `*Parameters` pointer after releasing its lock, while label update mutates that same object. `go test -race` reports a real concurrent read/write race. [repro](https://gist.github.com/starius/985175cde58a369f853f4de4819b3a54)
 - after multi-address is applied, `static new` returns a derived receive address, but `static summary` still reads the legacy/root seed. The two commands no longer describe the same address. [repro](https://gist.github.com/starius/1f1865f279f06c4be705bb42783fa676)
 - every generated change address is stored with an empty label, regardless of the labels on the consumed deposits. I think labels should be inherited by change - that is a common practice among wallets
 - testing gap: the SQL tests read the label back through the same open database/store instance. They do not close and reopen the database or rebuild the manager cache.
 - CLI: the shared `--label` help text says "label for this swap" and lacks a space, while `static deposit --label` labels the lnd funding transaction and `static new --label` labels the address.
 - `ListUnspentDeposits` includes the label, but historical deposit/list responses carry the address without the corresponding label - needs unification

Some of them will be resolved by a proper rebase, but some seem genuine.

GustavoStingelin:
Okay, if we agree to rebase this on top of #1139, I’d prefer to wait until @hieblmi has finished it and its CI is green. This should avoid premature conflicts and duplicated work from having to address #1139’s current race failures in the combined branch.

GustavoStingelin:
just rebased from master

GustavoStingelin:
@hieblmi rebased this on multi-address-core (thanks!). The rebase raised two design questions I'd like to settle before we go further:

1. How should address labels relate to funding transaction labels?
2. Should we add a ListStaticAddresses RPC?

PR #1180, opened 2026-07-21
swapserverrpc: add multi-address proofs

hieblmi:
Carry each deposit client derivation key and an optional generated change descriptor in static-address loop-in and withdrawal requests. The server can use these fields to validate scripts it does not store directly.

### DIFF
--- a/swapserverrpc/staticaddr.proto
+++ b/swapserverrpc/staticaddr.proto
@@ -124,6 +124,11 @@ message ServerPsbtWithdrawRequest {
 
     // The map of deposit txid:idx to the nonce used by the client.
     map<string, bytes> deposit_to_nonces = 2;
+
+    // The map of deposit txid:idx to the static address descriptor that was
+    // used to derive the deposit output. The server combines the client key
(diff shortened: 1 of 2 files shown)


swapserverrpc/staticaddr.proto
```diff
@@ -143,6 +148,22 @@ message ServerPsbtWithdrawSigningInfo {
     bytes sig = 2;
 }
 
+message StaticAddressDescriptor {
+    // The serialized client public key.
```
starius: Please specify what is the format of the key? Compressed (33 bytes) or x-only (32 bytes). And also specify its expected length in bytes. Also it makes sense to add that this is a Schnorr BIP340 pubkey , just in case. So IIUC, this is an x-only 32 bytes Schnorr BIP340 pubkey - please add this, if this is correct.

starius (approved):
**LGTM!** 🚀 
One note about the pubkey doc string.

swapserverrpc/staticaddr.proto
hieblmi: done

PR #1186, opened 2026-07-24
cmd/loop: skip expiring deposits in static loop-in --all

GustavoStingelin:
loop static in --all now skips confirmed deposits too close to their CSV expiry, so it won't quote swaps that could fail to complete in time.

- Filters --all selection against a minimum-remaining-blocks floor (default = shared Loop-in safety policy)
- Adds --min_expiry_blocks flag to raise/lower the floor; below-default values allowed with a warning
- Unconfirmed deposits stay eligible (relative timeout hasn't started)
- Reports skipped deposits back to the user in deterministic order
- Shares the eligibility check with the existing low-confirmation warning path


starius (commented):
Doing this in CLI is not ideal. This does not cover LiT and other direct gRPC callers.

I think this should be implemented on the `loopd` side, and we should remove the `--min_expiry_blocks` option. The minimum is derived from the swap parameters rather than being a user preference:

```
minimum runway = quote.CltvDelta + DepositHtlcDelta
```

Keeping `DepositHtlcDelta = 50` as a client-side constant in `staticaddr/loopin` is fine.

The ideal flow would be:

1. The CLI passes the `all` selection intent to loopd instead of listing deposits, calculating expiry eligibility, and expanding `--all` into explicit outpoints itself.
2. loopd refreshes the deposits and obtains a quote. The quote must expose the CLTV delta that will actually be used for a static Loop-in. Currently, the static-deposit quote path drops `quote.CltvDelta`, and `GetLoopInQuote` does not populate `InQuoteResponse.CltvDelta`; those should be fixed.
3. loopd computes each deposit's absolute expiry using its confirmation height and CSV delay, then compares its remaining runway with `quote.CltvDelta + 50`. Unconfirmed deposits should be evaluated using the same next-block-confirmation assumption as the server.
4. For an `all` selection, loopd omits deposits that do not satisfy the requirement and returns the selected and skipped deposits so the CLI can report them. For explicitly supplied outpoints which are too old, loopd should return an error rather than silently changing the user's selection.
5. Because removing deposits changes both the amount and the per-input fee, loopd recomputes the amount and obtains a final quote for the filtered deposit count.

GustavoStingelin:
Thanks for reviewing it, @starius! Some validations already live on the CLI side, so I kept them there to follow the existing layout. I agree that they would be better placed in `loopd`.

As u said, moving them will require a few additional changes and a small refactor. I am happy to do that, it just was not clear to me that this was the intended direction. I am working on this version now.

litbot-9000:
@hieblmi: review reminder
@gustavostingelin, remember to re-request review from reviewers when ready

GustavoStingelin:
paused while deciding the next directions

PR #1189, opened 2026-07-31
assets: harden tapd client lifecycle and RFQ handling

hieblmi:
## Summary

  Improve the tapd client’s lifecycle, timeout handling, concurrency, and validation.

  - Close tapd gRPC connections during daemon shutdown, initialization failures, and `view` completion.
  - Validate RFQ timeout durations before creating the client.
  - Round positive fractional timeouts up to tapd’s whole-second representation.
  - Allow cached asset-name reads while another universe lookup is in progress.
  - Reject nil, malformed, non-positive, and invalid-scale RFQ asset rates instead of risking a panic.
  - Add regression tests for timeout boundaries, concurrent cache access, and malformed rates.

### DIFF
--- a/assets/client.go
+++ b/assets/client.go
@@ -81,7 +81,7 @@ type TapdClient struct {
 
 	rfqTimeoutSeconds uint32
 	assetNameCache    map[string]string
-	assetNameMutex    sync.Mutex
+	assetNameMutex    sync.RWMutex
 	cc                *grpc.ClientConn
 }
 
@@ -169,10 +169,8 @@ func (c *TapdClient) GetRfqForAsset(ctx context.Context,
 func (c *TapdClient) GetAssetName(ctx context.Context,
 	assetId []byte) (string, error) {
 
-	c.assetNameMutex.Lock()
-	defer c.assetNameMutex.Unlock()
 	assetIdStr := hex.EncodeToString(assetId)
-	if name, ok := c.assetNameCache[assetIdStr]; ok {
+	if name, ok := c.getCachedAssetName(assetIdStr); ok {
 		return name, nil
 	}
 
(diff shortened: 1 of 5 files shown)


loopd/daemon.go
```diff
@@ -166,13 +166,23 @@ func (d *Daemon) Start() error {
...
 		// likely LiT is already running and blocking the DB
 		return fmt.Errorf("%v: make sure no other loop daemon process "+
 			"(standalone or embedded in lightning-terminal) is"+
 			"running", err)
 	}
 	if err != nil {
+		if d.assetClient != nil {
+			d.assetClient.Close()
+			d.assetClient = nil
+		}
```
starius: I propose to add `defer` inside `if d.cfg.Tapd.Activate` block above and close it there.

assets/client.go
```diff
@@ -288,6 +294,26 @@ func getPaymentMaxAmount(satAmount btcutil.Amount, feeLimitMultiplier float64) (
 	)
 }
 
+// getRfqTimeoutSeconds converts the configured RFQ timeout to the whole
+// seconds accepted by tapd. Sub-second precision is rounded up so a positive
+// timeout can never become tapd's invalid zero value.
```
starius: Isn't there a stronger reason to do it so we don't get a timeout error?

If we account on 1.5s timeout (and actively using it, e.g. use it at time 1.4s, i.e. 0.1s before the timeout), but tapd enforces the timeout of just 1s (rounded down, as before), we get a timeout error. So we round up to request enough time, right?

assets/client.go
```diff
@@ -254,6 +264,19 @@ func (c *TapdClient) GetAssetPrice(ctx context.Context, assetID string,
...
+	}
+
+	coefficient, ok := new(big.Int).SetString(assetRate.Coefficient, 10)
+	if !ok {
+		return 0, fmt.Errorf("invalid asset rate coefficient: %q",
+			assetRate.Coefficient)
+	}
+	if coefficient.Sign() <= 0 {
+		return 0, fmt.Errorf("asset rate coefficient must be positive")
+	}
```
starius, line 316: We need a similar hardening for GetRfqForAsset. Here is an example how to fix it: https://gist.github.com/starius/c23dd24330c3d36c9e852d2f4596313f

assets/client.go
```diff
@@ -163,12 +170,13 @@ func (c *TapdClient) GetRfqForAsset(ctx context.Context,
...
 
-	c.assetNameMutex.Lock()
-	defer c.assetNameMutex.Unlock()
 	assetIdStr := hex.EncodeToString(assetId)
+	c.assetNameMutex.RLock()
 	if name, ok := c.assetNameCache[assetIdStr]; ok {
+		c.assetNameMutex.RUnlock()
 		return name, nil
 	}
+	c.assetNameMutex.RUnlock()
```
starius, line 172: Let's preserve the `defer` pattern somehow. Maybe make private accessor methods for this map? To encapsulate map+mutex work there.

starius (approved):
**LGTM!** 💾

Left come comments. The main one is hardening for GetRfqForAsset.

loopd/daemon.go
hieblmi: done

assets/client.go
hieblmi: done

assets/client.go
hieblmi: extracted the checks and re-used it for GetFrqForAsset.

assets/client.go
hieblmi: done

PR #1190, opened 2026-08-02
staticaddr: coordinate concurrent deposit use

starius:
Withdrawals obtain signatures before deposits enter their durable state, leaving a window where a loop-in can select the same outputs.

Coordinate in-flight client operations with an in-memory registry. Persist withdrawal data before publication. Existing recovery resumes it after restart.

#### Pull Request Checklist
- [ ] Update `docs/release-notes/release-notes-next.md` if your PR contains
  major features, breaking changes or bugfixes


litbot-9000:
@hieblmi: review reminder

PR #1193, opened 2026-08-10
loopd: account for channel reserves in loop out validation

starius:
Use each channel's local balance minus its reserve when simulating Loop Out payment bandwidth, clamping the result at zero. This prevents swaps from starting when the gross balance covers the amount and routing fee but the spendable balance does not.

Add one-sat boundary tests and a bug-fix release note. Leave a TODO for server-fee and prepay capacity because enforcing their maximum fee caps could reject viable swaps whose actual routing fees are lower.

#### Pull Request Checklist

### DIFF
--- a/docs/release-notes/release-notes-next.md
+++ b/docs/release-notes/release-notes-next.md
@@ -6,6 +6,10 @@
 
 #### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


hieblmi (approved):
Thanks for the improvement, LGTM!

PR #1197, opened 2026-08-11
test: avoid chain notifier mock deadlock

starius:
Release the mock LND lock before sending the initial block height. A concurrent height notification can fill the buffered channel. Holding the lock while sending would then block invoice lookups and make the static loop-in test time out.

#### Pull Request Checklist

### DIFF
--- a/test/chainnotifier_mock.go
+++ b/test/chainnotifier_mock.go
@@ -129,13 +129,18 @@ func (c *mockChainNotifier) RegisterBlockEpochNtfn(ctx context.Context) (
 			}
 		}()
 
-		// Send initial block height
+		// Snapshot the initial block height while holding the lock, but
+		// release it before sending. The buffered channel may already contain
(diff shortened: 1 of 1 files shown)


GustavoStingelin (commented):
Why this PR? Is there any flaky test that we are trying to fix?

starius:
Yes, this was identified when looking though a red CI log.

GustavoStingelin:
Cool, I think we should reference the failed CIs here as documentation

starius:
Agreed. I can't find it now, but I'll take into account for the future.

GustavoStingelin:
Yeah, because there is a race between two goroutines trying to send to the same blocked channel. There is no guarantee which goroutine will reach the send first, so the messages can be received out of order. That means we could receive an older height even though we have already reached the tip. Since this is only a test mock, though, we can leave it for now and see if it actually causes problems.

GustavoStingelin (approved):

PR #1199, opened 2026-08-12
multi: migrate to btcd v2 modules

starius:
Update LND, Aperture, and Taproot Assets to revisions using the btcd v2 modules, and update lndclient to [v0.21.0-3](https://github.com/lightninglabs/lndclient/pull/285). Migrate Loop chain, transaction, and address types to their corresponding v2 packages.

The lndclient release includes the migration from: https://github.com/lightninglabs/lndclient/pull/280

Taproot Assets is temporarily replaced with its btcd v2 revision because the v0.8 release branch has not adopted the new modules.

This raises the minimum Go version to 1.26 and changes exported address types.

#### Pull Request Checklist


hieblmi (commented):
### Taproot Assets compatibility

Please bump `github.com/lightninglabs/taproot-assets` to `v0.8.1`. This PR already handles its updated `rfqmath.UnitsToMilliSatoshi` signature, and the bump keeps Loop compatible with litd versions using tapd v0.8.1.

Since v0.8.1 still uses the legacy btcd packages, Loop needs the temporary btcd-v2 replacement. Go does not propagate `replace` directives to downstream modules, so LiT needs the same replacement until a btcd-v2-compatible Taproot Assets release is available. Please re-run the LiT compatibility jobs with that setup.

PR #1202, opened 2026-08-17
staticaddr/loopin: reject expiring deposits

starius:
Manual static-address loop-ins can name explicit deposit outpoints. Those paths only checked that deposits were available. An old deposit could get a quote, then fail server init.

Apply the same swappability and expiry runway check used by automatic selection to manual quotes and initiation. The client now rejects before contacting the server.

#### Pull Request Checklist

### DIFF
--- a/loopd/swapclient_server.go
+++ b/loopd/swapclient_server.go
@@ -92,6 +92,7 @@ type swapClientServer struct {
 	config               *Config
 	network              lndclient.Network
 	impl                 *loop.Client
+	loopInQuoter         loopInQuoter
 	liquidityMgr         *liquidity.Manager
 	lnd                  *lndclient.LndServices
 	reservationManager   *reservation.Manager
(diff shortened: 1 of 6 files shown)


loopd/swapclient_server.go
```diff
@@ -1182,6 +1182,27 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
...
 		}
 
+		params, err := s.staticAddressManager.
+			GetStaticAddressParameters(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("unable to retrieve static "+
+				"address parameters: %w", err)
+		}
+
+		info, err := s.lnd.Client.GetInfo(ctx)
```
hieblmi: we also need this call in the `if` branch above, so we might as well only call it once before the `if`, then reuse it here.

loopd/swapclient_server.go
```diff
@@ -1182,6 +1182,27 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 			)
 		}
 
+		params, err := s.staticAddressManager.
```
hieblmi: we also need this call in the if branch above, so we might as well only call it once before the if, then reuse it here.

hieblmi (approved):
Thanks for the fix.

LGTM pending two improvements.

loopd/swapclient_server.go
starius: Fixed

loopd/swapclient_server.go
starius: Fixed

starius:
Fixed the review comments. Also added test coverage. Introduced `loopInQuoter` for that.

starius:
Rebased

PR #1203, opened 2026-08-21
loopd: support more HTLC recovery modes, stateless and cooperative

starius:
Allow `sweephtlc` to reconstruct protocol-11 Loop Out HTLCs when the local swap record is unavailable but the public swap data and preimage can be recovered.

Prefer signing by public key and use a bounded family-99 key scan only when direct signing fails or produces an invalid witness.

Also add cooperative sweep as an option applicable in both stateful and stateless modes. In stateless mode it also requires providing payment address.

So now we have a matrix.

| State | Spend | LND key path | Mode note |
|---|---|---|---|
| Stateful | Unilateral | Stored locator | Loads the preimage, contract, and locator from Loop DB; spends the Taproot script success path. |
| Stateful | Cooperative | Stored locator | Loads the contract, locator, and invoice payment-address proof from Loop DB; uses the MuSig2 Taproot key path without revealing the preimage to the server. |
| Stateless | Unilateral | Public key | Reconstructs the contract without Loop DB and asks LND to sign directly with `client_pubkey`; no key scan occurs. |
| Stateless | Unilateral | Scan fallback | Reconstructs the contract without Loop DB; an unknown public-key signing result must fail local witness verification, then family-99 scanning recovers the locator. |
| Stateless | Cooperative | Locator scan | Reconstructs the contract without Loop DB, scans family 99 for the locator required by LND's MuSig2 API, and supplies the invoice `payment_addr` as server ownership proof. |

I tested all the modes locally.

#### Pull Request Checklist

### DIFF
--- a/client.go
+++ b/client.go
@@ -112,6 +112,19 @@ type Client struct {
 	wg          sync.WaitGroup
 }
 
+// MuSig2SignSweep asks the swap server to cooperatively sign a Loop Out HTLC
+// sweep transaction.
+func (c *Client) MuSig2SignSweep(ctx context.Context,
+	protocolVersion loopdb.ProtocolVersion, swapHash lntypes.Hash,
+	paymentAddr [32]byte, nonce, sweepTxPsbt []byte) ([]byte, []byte,
+	error) {
+
+	return c.Server.MuSig2SignSweep(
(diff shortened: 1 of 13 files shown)


bhandras (approved):
LGTM 🎉

bhandras (approved):
LGTM 🎉

PR #1204, opened 2026-08-21
multi: bump gRPC to v1.83.1

bhandras:
## The problem

Loop's main, `looprpc`, and `swapserverrpc` modules pin gRPC v1.82.1.
This leaves each independently built module on an older gRPC dependency
set.

## The fix

Bump all three module graphs to gRPC v1.83.1. Running `go mod tidy` also
updates the `genproto/googleapis` and OpenTelemetry dependencies selected
by that release.

## What does not change

This does not change RPC definitions, generated code, wire behavior, or
runtime configuration.

## Tests

- `go build ./...` in the main, `looprpc`, and `swapserverrpc` modules
- `go test ./...` in the main, `looprpc`, and `swapserverrpc` modules
- `make mod-check`

## Compatibility and rollout

The update requires no migration or coordinated rollout.


hieblmi (approved):
LGTM

PR #1209, opened 2026-08-25
cmd/loop: fix route hint parsing

hieblmi:
## Summary

- preserve comma-containing JSON route hints across nested CLI commands
- return decoded route hints instead of discarding a shadowed slice
- add regression coverage and a bug-fix release note

## Testing

- go test ./...
- make lint
- make commitmsg-lint commit=HEAD


hieblmi:
closing in favor of https://github.com/lightninglabs/loop/pull/1213

PR #1212, opened 2026-08-26
loopin: Route invoices over asset channels

sputn1ck:
## Summary

- add asset-channel Loop In invoices via tapd, with a required asset ID,
  explicit edge node, and per-swap minimum output in indivisible asset units
- enforce that minimum in the RFQ request, verify the accepted output, and
  return/display the quoted amount; display the minimum before confirmation
- use the same accepted RFQ route hint for the swap and probe invoices,
  and use the selected edge for preflight and server fee quotes
- require a quote covering the full 30-day invoice lifetime before funding;
  reject short-lived quotes instead of risking expiry during confirmations
- validate CLI inputs before quote/swap RPCs and update generated RPC/CLI docs

Asset invoices remain BTC/msat-denominated BOLT 11 invoices. The accepted
edge RFQ converts that amount into asset units. The edge's price oracle must
support the requested lifetime; the built-in one-hour mock oracle is not
sufficient for this flow.

## Testing

- `GOWORK=off go test ./...`
- `GOWORK=off make lint workers=4`
- `GOWORK=off make rpc`
- `GOWORK=off make docs-check`
- `GOWORK=off make commitmsg-lint range="HEAD^..HEAD"`
- hybrid-network integration: short-quote and minimum-output rejection,
  canceled probe HTLC using the swap RFQ, successful client/server completion,
  and exact asset units in the selected channel and settled invoice HTLCs

#### Pull Request Checklist

- [x] Added an entry to `docs/release-notes/release-notes-next.md`

### DIFF
--- a/cmd/loop/loopin.go
+++ b/cmd/loop/loopin.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"fmt"
 	"strconv"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/lightninglabs/loop"
 	"github.com/lightninglabs/loop/labels"
 	"github.com/lightninglabs/loop/looprpc"
@@ -94,8 +96,13 @@ var (
 			},
 			&cli.StringFlag{
 				Name: "asset_edge_node",
-				Usage: "the optional pubkey of the asset channel " +
+				Usage: "the required pubkey of the asset channel " +
 					"peer to use for the loop in",
+			},
+			&cli.Uint64Flag{
+				Name: "min_asset_amount",
+				Usage: "minimum output in indivisible asset units; " +
+					"required with asset_id",
 			},
 		},
 		Action: loopIn,
@@ -164,9 +171,10 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	var assetInfo *looprpc.AssetLoopInRequest
-	if cmd.IsSet("asset_edge_node") && !cmd.IsSet("asset_id") {
-		return fmt.Errorf("asset_id must be set when asset_edge_node " +
-			"is set")
+	if (cmd.IsSet("asset_edge_node") || cmd.IsSet("min_asset_amount")) &&
+		!cmd.IsSet("asset_id") {
+
+		return fmt.Errorf("asset_id must be set with asset options")
 	}
 	if cmd.IsSet("asset_id") {
 		if cmd.Bool(privateFlag.Name) || len(hints) != 0 {
@@ -178,6 +186,9 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
 		if err != nil {
 			return fmt.Errorf("invalid asset id: %w", err)
 		}
+		if len(assetID) != 32 {
+			return fmt.Errorf("asset id must be a 32 byte value")
+		}
 
 		var assetEdgeNode []byte
 		if cmd.IsSet("asset_edge_node") {
@@ -188,10 +199,25 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
 				return fmt.Errorf("invalid asset edge node: %w", err)
 			}
 		}
+		if len(assetEdgeNode) != 33 {
+			return fmt.Errorf("asset_edge_node is required and must " +
+				"be a 33 byte public key")
+		}
+		if _, err := btcec.ParsePubKey(assetEdgeNode); err != nil {
+			return fmt.Errorf("invalid asset edge node: %w", err)
+		}
+		if len(lastHop) != 0 && !bytes.Equal(lastHop, assetEdgeNode) {
+			return fmt.Errorf("last_hop and asset_edge_node must match")
+		}
+		lastHop = assetEdgeNode
+		if cmd.Uint64("min_asset_amount") == 0 {
+			return fmt.Errorf("min_asset_amount must be positive")
+		}
 
 		assetInfo = &looprpc.AssetLoopInRequest{
-			AssetId:       assetID,
-			AssetEdgeNode: assetEdgeNode,
+			AssetId:        assetID,
+			AssetEdgeNode:  assetEdgeNode,
+			MinAssetAmount: cmd.Uint64("min_asset_amount"),
 		}
 	}
 
@@ -224,6 +250,10 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	limits := getInLimits(quote)
+	if assetInfo != nil {
+		fmt.Printf("Minimum asset output: %d units of %x\n",
+			assetInfo.MinAssetAmount, assetInfo.AssetId)
+	}
 
 	// Skip showing details if configured
 	if !(cmd.Bool("force") || cmd.Bool("f")) {
@@ -254,6 +284,9 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
 
 	fmt.Printf("Swap initiated\n")
 	fmt.Printf("ID:           %x\n", resp.IdBytes)
+	if assetInfo != nil {
+		fmt.Printf("Quoted asset output: %d units\n", resp.AssetAmount)
+	}
 
 	if resp.HtlcAddressP2Tr != "" {
 		fmt.Printf("HTLC address (P2TR): %v\n", resp.HtlcAddressP2Tr)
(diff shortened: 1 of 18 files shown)


looprpc/client.proto
```diff
@@ -2437,6 +2444,20 @@ message AssetLoopOutRequest {
...
+    The asset ID to receive when the swap invoice is paid. A Taproot Assets
+    client must be connected to loopd when this field is set.
+    */
+    bytes asset_id = 1;
+
+    /*
+    The optional node identity public key of the asset channel peer to use for
+    RFQ negotiation. If omitted, tapd selects from the eligible asset channels.
+    */
+    bytes asset_edge_node = 2;
```
hieblmi, line 2458: [P1] Require a minimum asset output: The request specifies only an asset ID and optional peer. It provides no minimum asset amount or rate floor, even though the edge quote determines how many asset units the client receives for its on-chain BTC. The quote is negotiated only after CLI confirmation and is never returned, displayed, or persisted. The daemon-wide tapd price-deviation check is not per-swap user consent. Add a quote phase or fields such as minimum_asset_amount or rate_limit, show the resulting asset amount, and enforce the limit in AddInvoiceRequest before committing the swap.

loopin.go
```diff
@@ -191,40 +237,106 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
-			Expiry:     3600,
-			RouteHints: request.RouteHints,
-			Private:    true,
-		},
-	)
-	if err != nil {
-		return nil, err
+	var probeInvoice string
+	if assetLoopIn {
+		assetProbe, err := cfg.assets.AddAssetInvoice(
```
hieblmi, line 301: [P1] Probe the exact RFQ used by the swap: This second AddAssetInvoice call negotiates a new RFQ. Checking only that its peer matches is insufficient: the probe and swap invoices have different RFQ IDs, virtual SCIDs, rates, limits, and policies. A probe can therefore succeed with a smaller asset requirement or different final-hop fee while the actual swap quote fails after the BTC HTLC is funded. Create the probe and swap invoices from the same accepted RFQ-backed route hint, or otherwise verify and reserve the exact rate, SCID, and capacity used by the swap invoice.

loopin.go
```diff
@@ -191,40 +237,106 @@ func newLoopInSwap(globalCtx context.Context, cfg *swapConfig,
...
+			return nil, fmt.Errorf("invalid asset invoice peer: %w", err)
+		}
+		if len(assetPeer) != 0 && !bytes.Equal(assetPeer, quotePeer[:]) {
+			return nil, fmt.Errorf("asset invoice peer does not match " +
+				"requested edge node")
+		}
+
+		// Pin the probe and the server payment to the same edge node as
+		// the swap invoice.
+		request.LastHop = &quotePeer
```
hieblmi, line 273: [P1] Do not treat the tapd reference quote as edge selection: When no edge is supplied, tapd negotiates with multiple peers and returns its most-expensive quote as the reference quote while putting multiple RFQ hints in the invoice. That response is not a selected route, but this code treats its peer as one and pins the server to it, forcing the worst-priced edge and disabling the other hints. The server fee quote was also calculated earlier with LastHop unset, so its invoice amount may not match this route fee. Require an explicit edge for now, or add a pre-swap selection phase that chooses an approved quote before calculating server fees.

loopin.go
```diff
@@ -31,6 +33,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// assetLoopInInvoiceExpiry is the invoice lifetime an asset edge must cover
+// with its RFQ quote. If an on-chain funding transaction does not confirm in
+// this window, the swap fails safely and the client can reclaim the HTLC.
+const assetLoopInInvoiceExpiry = int64(60 * 60)
```
hieblmi, line 39: [P1] Invoice lifetime is shorter than the payment lifecycle: The asset invoice and RFQ expire after one hour, but the server normally waits 3-6 confirmations and retries failed payments only after one hour. Normal block variance alone makes expiration likely, particularly for six-confirmation swaps. Once funded, failure can send the client to the default 1000-block refund path. The RFQ must cover confirmation and retry time, or RFQ and invoice creation must be deferred until the HTLC is sufficiently confirmed.

cmd/loop/loopin.go
```diff
@@ -150,6 +162,38 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
...
+				cmd.String("asset_edge_node"),
+			)
+			if err != nil {
+				return fmt.Errorf("invalid asset edge node: %w", err)
+			}
+		}
+
+		assetInfo = &looprpc.AssetLoopInRequest{
+			AssetId:       assetID,
+			AssetEdgeNode: assetEdgeNode,
```
hieblmi, line 193: [P2] Quote the explicit asset edge in the CLI: asset_edge_node is placed in AssetInfo here, but the preflight quote below uses only the separate last_hop flag. The CLI therefore displays and uses a fee cap for a different route, while Loop internally re-quotes with the asset edge and may fail with ErrSwapFeeTooHigh. Make asset_edge_node the effective LoopInLastHop and reject conflicting values.

cmd/loop/loopin.go
```diff
@@ -150,6 +162,38 @@ func loopIn(ctx context.Context, cmd *cli.Command) error {
...
+
+		var assetEdgeNode []byte
+		if cmd.IsSet("asset_edge_node") {
+			assetEdgeNode, err = hex.DecodeString(
+				cmd.String("asset_edge_node"),
+			)
+			if err != nil {
+				return fmt.Errorf("invalid asset edge node: %w", err)
+			}
+		}
```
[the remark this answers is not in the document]
sputn1ck: Addressed in 3caa4bf: the CLI validates the decoded 32-byte asset ID and 33-byte compressed edge public key before quote/swap RPCs. TestAssetLoopInCLI covers malformed/short IDs, missing/short edges, missing minimum output, conflicting last hops, and the edge passed to the quote.

looprpc/client.proto
sputn1ck: Addressed in 3caa4bf with a mandatory per-swap min_asset_amount (indivisible units). The CLI displays the minimum before confirmation; AddAssetInvoice converts it into an upward-rounded AssetRateLimit before requesting the RFQ and independently checks the accepted output before swap admission/funding. SwapResponse.asset_amount returns the accepted output and the CLI displays it. Unit tests cover an accepted exact minimum and rejection below it; the hybrid integration test also rejects an above-market minimum before any HTLC funding.

loopin.go
sputn1ck: Addressed in 3caa4bf: there is now one tapd AddInvoice/RFQ negotiation. The returned BOLT 11 hint is checked against the accepted peer and virtual SCID, then reused unchanged in lnd AddHoldInvoice for the probe, with the same msat amount. The unit test compares the hints and asserts only one tapd call. The integration test observes a canceled probe HTLC on the selected physical channel, compares probe/swap RFQ hints, and subsequently settles the swap using that RFQ.

loopin.go
sputn1ck: Addressed in 3caa4bf by requiring an explicit asset_edge_node. The client sets LastHop before its server fee quote and requires exactly one RFQ hint matching the requested peer and accepted virtual SCID. The multi-peer reference quote is no longer treated as route selection.

loopin.go
sputn1ck: Addressed in 3caa4bf: the asset invoice now requests a 30-day lifetime, and the client verifies that both the invoice and accepted quote cover that lifetime before swap admission/funding. This adds substantial confirmation/retry margin over the current maximum accepted 1500-block HTLC delta; an edge that cannot offer that lifetime is rejected rather than funding a one-hour quote. The built-in one-hour mock oracle therefore cannot serve this flow. Unit and hybrid-network tests cover short-lived quote rejection, and the hybrid test uses a controllable long-lived oracle for the successful swap.

cmd/loop/loopin.go
sputn1ck: Addressed in 3caa4bf: asset_edge_node becomes the effective LoopInLastHop for the CLI preflight quote and swap request, and conflicting last_hop values fail before either RPC. TestAssetLoopInCLI verifies both the conflict rejection and the exact peer passed to GetLoopInQuote.

litbot-9000:
@hieblmi: review reminder
@sputn1ck, remember to re-request review from reviewers when ready

litbot-9000:
@hieblmi: review reminder
@sputn1ck, remember to re-request review from reviewers when ready

PR #1213, opened 2026-08-27
cmd/loop: fix route hint array parsing

starius:
_Another approach for https://github.com/lightninglabs/loop/pull/1209. We can have only a single `--route_hints` with complete JSON, so the type should be `cli.StringFlag`. And then we don't need `DisableSliceFlagSeparator` and touching other flags (`--utxo` and `--deposit_outpoint`)._

Treat `--route_hints` as a single JSON array instead of a repeatable string slice. This preserves commas in the JSON without disabling slice separators for unrelated UTXO and deposit flags.

Decode the outer array into raw JSON elements, then use protojson for each `RouteHint` protobuf. This follows protobuf JSON field naming and quoted uint64 semantics, and rejects unknown fields instead of silently accepting misspelled hints.

Reject empty RouteHint messages at the RPC boundary. An empty route would be encoded as an empty BOLT 11 `r` field, which zpay32 rejects when decoding the invoice.

Add recorded CLI sessions covering valid single, multiple, and multi-hop hints, malformed input, RPC-boundary validation, and comma-separated UTXO and deposit selections.

#### Pull Request Checklist


loopd/swapclient_server.go
```diff
@@ -1264,6 +1278,21 @@ func unmarshallRouteHints(rpcRouteHints []*swapserverrpc.RouteHint) (
 
 // unmarshallHopHint unmarshalls a single hop hint.
 func unmarshallHopHint(rpcHint *swapserverrpc.HopHint) (zpay32.HopHint, error) {
+	if rpcHint == nil {
+		return zpay32.HopHint{}, fmt.Errorf("hop hint is nil")
+	}
+	if rpcHint.ChanId == 0 {
+		return zpay32.HopHint{}, fmt.Errorf("channel ID is zero")
+	}
+	if rpcHint.CltvExpiryDelta == 0 {
```
hieblmi: Zero CLTV deltas are protocol-permitted. The new check in [swapclient_server.go](https://github.com/lightninglabs/loop/blob/328f30281de756a2400f03415da51162b9028615/loopd/swapclient_server.go#L1287-L1289) rejects every hint with cltv_expiry_delta == 0. BOLT 7 defines this as an unrestricted u16 and only requires it to equal the blocks subtracted; BOLT 11 carries that value into route hints without adding a nonzero constraint. This can reject a standards-permitted private-channel policy and breaks previously accepted gRPC requests. Keep the > MaxUint16 guard, but remove the zero check unless Loop intentionally documents a stricter API requirement. [BOLT 7](https://github.com/lightning/bolts/blob/master/07-routing-gossip.md#the-channel_update-message), [BOLT 11](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md).

hieblmi (approved):
LGTM, pending one finding.

loopd/swapclient_server.go
starius: Fixed. Removed the zero check.

PR #1215, opened 2026-08-27
staticaddr/withdraw: harden replacement monitoring

hieblmi:
## Summary

This is PR 3 of 3 in the static-address multi-address stack. It depends on
#1214 and is based on `multi-address-address-hardening` so this PR only shows
withdrawal replacement and monitor hardening.

This follow-up:

- validates that a confirmed replacement transaction actually spends the
  selected deposits;
- identifies static-address change by script instead of assuming a fixed
  output index;
- reconciles partial conflicting spends without marking untouched deposits as
  withdrawn;
- follows the output from the transaction that really replaced the original
  withdrawal; and
- joins withdrawal monitor goroutines during shutdown and adds multi-address
  replacement coverage.

## Stack

1. #1139 — multi-address core
2. #1214 — address startup hardening
3. This PR — withdrawal replacement hardening

## Testing

- `go test -race ./staticaddr/...`
- `go vet ./...`
- `git diff --check`

#### Pull Request Checklist

- [x] Added an entry to `docs/release-notes/release-notes-next.md`.


staticaddr/withdraw/manager.go
```diff
@@ -468,9 +469,7 @@ func (m *Manager) WithdrawDeposits(ctx context.Context,
 				"withdrawal: %v", err)
 		}
```
starius, line 470: Please fail or reconcile when `CreateWithdrawal` fails. Continuing can report success without the association required to persist confirmation or recover history.

staticaddr/withdraw/manager.go
starius, line 667: Please return `ErrInsufficientFee`. Suppressing it makes a rejected withdrawal return success with an empty transaction hash and address.

PR #1216, opened 2026-08-28
docker: fix linux/arm64 images shipping amd64 binaries

### LINKED ISSUE #1211: Published Docker images for linux/arm64 contain amd64 binaries (exec format error)
## Summary

The multi-arch Docker images published to Docker Hub are broken for
`linux/arm64`: the arm64 manifest entry contains an amd64 filesystem.
Running `loopd` from that image on aarch64 gives `exec format error`.

The published **release binaries** (`loop-linux-arm64-*.tar.gz`) are correct —
this affects only the Docker images.

## Root cause

`Dockerfile` pins **both** build stages to the build machine's platform:

    Dockerfile:1   FROM --platform=${BUILDPLATFORM} golang:1.26-alpine as builder
    Dockerfile:21  FROM --platform=${BUILDPLATFORM} alpine as final

while `.github/workflows/docker.yml:39` requests:

    platforms: linux/amd64,linux/arm64

Because the *final* stage is also pinned to `BUILDPLATFORM`, every platform
entry is built on the runner's architecture (amd64). buildx then publishes
that same filesystem under both manifest entries, changing only the
`architecture` label.

Both lines are present at master (`341d844`) as of 2026-08-25.

## Evidence

The two per-arch entries of `lightninglabs/loop:v0.34.0-beta-v0.8.1-tapd`
share **byte-identical layer digests**, and their config blobs differ in
exactly one string:

    -  "architecture": "amd64",
    +  "architecture": "arm64",

Identical layers means identical bytes, so at most one of those two
architecture labels can be true. No layer download is needed to confirm this —
comparing the two manifests is enough.

## Reproduction

On any aarch64 host:

    docker run --rm --platform linux/arm64 \
      lightninglabs/loop:v0.34.0-beta-v0.8.1-tapd loopd --version

Expected: a version string. Actual: `exec format error`.

## Suggested fix

Remove the pin from the final stage — it should resolve per target platform:

    -FROM --platform=${BUILDPLATFORM} alpine as final
    +FROM alpine as final

`--platform=$TARGETPLATFORM` is the default for every stage, so no flag is
needed. Keeping `--platform=${BUILDPLATFORM}` on the *builder* stage is fine
and desirable (it avoids emulating the Go toolchain), but the build then needs
to cross-compile — e.g. passing `GOARCH=$TARGETARCH` through to `make install`
— rather than producing a native binary for the build host.

For comparison: `lightning-terminal`'s Dockerfile has no `--platform` pin on
any stage, and its arm64 image works correctly.

## Impact

This has been silently shipping broken arm64 images. Anyone deploying loopd in
a container on ARM hardware (Raspberry Pi, Apple Silicon, Graviton) hits it,
and the failure mode is opaque — the manifest advertises arm64 support, the
pull succeeds, and the error appears only at exec time.

## Environment

- Host: Raspberry Pi 5, aarch64, Debian
- Docker 28.5.0
- Affected: `lightninglabs/loop` Docker images, all published multi-arch tags
- Not affected: `loop-linux-arm64-*.tar.gz` release binaries (verified
  `CGO_ENABLED=0`, correct `e_machine`, runs fine)

starius:
Fixes https://github.com/lightninglabs/loop/issues/1211 and other related issues.

The published `linux/arm64` images have never held arm64 code. The Dockerfile pinned both of its stages to `${BUILDPLATFORM}`, so every platform buildx was asked for got built on the release runner's amd64 architecture and then labelled arm64, which buildx reports as a perfectly successful build. Running `loopd` from one of those images on ARM hardware fails with `exec format error`. The final stage now resolves to the target platform, and the builder stage, which stays native so the Go toolchain is never emulated, cross-compiles to it.

Nothing in the release workflow ever looked at what it was publishing, which is why this went unnoticed across every release since the workflow was added. The build now emits a local OCI layout alongside an untagged push, the layout is checked for binaries whose ELF headers match the architecture each manifest entry advertises, the staged image is confirmed to consist of exactly the manifests that were checked, and only then does that digest get the release tag. Note that a smoke test could not have caught this, since amd64 binaries in a mislabelled arm64 image run natively on an amd64 runner.

Released images also reported no build metadata, printing `commit= commit_hash=`, because BuildKit's git context drops the `.git` directory that the Makefile reads to stamp it in. Keeping that directory restores the commit in `loopd --version`.

#### Pull Request Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,6 +13,11 @@ jobs:
   main:
     runs-on: ubuntu-latest
     steps:
+      # Needed for scripts/check_multiarch_image.py below. The image itself is
+      # built from the default git context, not from this checkout.
+      - name: Git checkout
+        uses: actions/checkout@v4
+
       - name: Set up QEMU
         uses: lightninglabs/gh-actions/setup-qemu-action@2021.01.25.00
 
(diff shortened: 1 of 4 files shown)


bhandras (approved):

hieblmi (commented):
Thanks for fixing this! There are two findings:

Ancestor whiteouts do not remove discovered binaries. In [check_multiarch_image.py](https://github.com/lightninglabs/loop/blob/75e049767d9223b0d056d1e7a98f0de5a37b70ee/scripts/check_multiarch_image.py#L188-L202), .wh.bin becomes wh.bin because lstrip("./") removes the leading dot. Additionally, deleting or replacing /bin does not clear /bin/* from found. I reproduced an image whose final layer deletes /bin; the validator still reported all three binaries and exited successfully.

Unexpected platforms are promoted without validation. The loop at [lines 394–400](https://github.com/lightninglabs/loop/blob/75e049767d9223b0d056d1e7a98f0de5a37b70ee/scripts/check_multiarch_image.py#L394-L400) checks only platforms named by --platforms, but writes every manifest to --digests-out. I reproduced an extra linux/386 manifest containing amd64 binaries; it was included in the digest list while the validator exited successfully. Either reject extra platforms or run check_entry for every non-attestation entry.

GustavoStingelin (commented):
looking good, just the same findings from heebs

starius:
> Thanks for fixing this! There are two findings:
> 
> Ancestor whiteouts do not remove discovered binaries. In [check_multiarch_image.py](https://github.com/lightninglabs/loop/blob/75e049767d9223b0d056d1e7a98f0de5a37b70ee/scripts/check_multiarch_image.py#L188-L202), .wh.bin becomes wh.bin because lstrip("./") removes the leading dot. Additionally, deleting or replacing /bin does not clear /bin/* from found. I reproduced an image whose final layer deletes /bin; the validator still reported all three binaries and exited successfully.
> 
> Unexpected platforms are promoted without validation. The loop at [lines 394–400](https://github.com/lightninglabs/loop/blob/75e049767d9223b0d056d1e7a98f0de5a37b70ee/scripts/check_multiarch_image.py#L394-L400) checks only platforms named by --platforms, but writes every manifest to --digests-out. I reproduced an extra linux/386 manifest containing amd64 binaries; it was included in the digest list while the validator exited successfully. Either reject extra platforms or run check_entry for every non-attestation entry.

Fixed both. Also added the following fix: the digest list was written even when the check failed. Now a failing run exits 1 and creates no file.

PR #1218, opened 2026-08-28
staticaddr: support multiple receive and change addresses

hieblmi:
This is PR 1 of 3 in the Static Address multi-address stack.

  Static Address previously treated one legacy/root address as the owner of every
  deposit. This PR introduces fresh receive and operation-specific change
  addresses while preserving the address parameters that own each deposit.

  Loop-ins and withdrawals can consequently spend deposits received across
  multiple derived addresses, with each input signed and proven using its actual
  address parameters.

  ## Key Changes

  - Derive fresh static receive and change addresses from dedicated key families.
  - Persist each deposit's owning static-address parameters and restore them
    across restarts.
  - Backfill pre-migration deposits and fractional loop-ins to the legacy/root
    address.
  - Create cooperative signing sessions from each deposit's actual keys.
  - Send per-deposit address proofs to the server for loop-ins and withdrawals.
  - Generate and persist operation-specific change addresses for fractional
    loop-ins and partial withdrawals.
  - Rebuild and maintain an active address index during startup.
  - Import only missing wallet scripts and narrowly recognize duplicate-import
    errors for the expected Taproot output key.
  - Resolve requested funding addresses through the active script index instead
    of reconstructing and scanning every persisted address.
  - Track expiry sweeps by the spent deposit outpoint so unrelated transactions
    paying the same destination script cannot finalize a deposit.
  - Add `loop static deposit` support for creating and optionally funding a fresh
    address through lnd `SendCoins`.
  - Include each deposit's receiving address in RPC and CLI listings.

  ## RPC and Compatibility Changes

  `NewStaticAddress` now derives a fresh receive address on every request and
  never funds it. Callers must not assume repeated requests are idempotent or
  return the same address.

  Wallet funding is handled by the separate `FundStaticAddress` RPC, exposed as
  `POST /v1/staticaddr/fund`. It creates and funds a fresh address when
  `send_coins_request.addr` is empty, or funds a known existing static address
  when it is set. The `loop static deposit` command uses this endpoint.

  Funding requires `wallet:fund` in addition to `swap:execute` and `loop:in`,
  or an explicit URI grant for `FundStaticAddress`. Existing execute/in and
  `NewStaticAddress` URI macaroons cannot authorize funding. The default local
  Loop macaroon is regenerated on startup when permissions change; distributed
  or custom funding macaroons must be explicitly updated.

  Because address creation mutates wallet and database state, the RPC permission
  changes from `swap:read` to `swap:execute`. Operators using custom scoped
  macaroons must rebake them accordingly.

  `StaticAddressSummaryResponse.static_address` remains populated with the
  legacy/root address for wire compatibility, but is deprecated and must not be
  treated as the current receive address. Call `NewStaticAddress` to derive a
  fresh address.

  ## Database and Recovery

  The database now records:

  - The static-address owner of every deposit.
  - The generated change address associated with a fractional loop-in.

  Migrations backfill existing records using the legacy/root address. Store
  reconstruction restores deposit ownership and loop-in and withdrawal change
  metadata after restart.

  ## Stacked PRs

  1. This PR — multi-address persistence, derivation, signing, RPC, and CLI core.
  2. [#1215 — withdrawal replacement hardening](https://github.com/lightninglabs/loop/
  pull/1215)
     validates and follows the transactions that actually replace multi-address
     withdrawals.
  3. [#1217 — loop-in HTLC recovery](https://github.com/lightninglabs/loop/pull/1217)
     persists the confirmed HTLC output so recovery can rebuild the exact
     server-published transaction.

  The address startup and lookup hardening previously isolated in
  [#1214](https://github.com/lightninglabs/loop/pull/1214) has been folded into
  this PR, and #1214 is now closed.

  ## Testing

  - `go test ./...`
  - `go vet ./...`
  - `go test -race ./staticaddr/address ./loopd`
  - `make docs-check`
  - Commit-message lint across the complete PR range
  - `git diff --check`

  ## Release Notes

  Release notes document the new address behavior, RPC permission change,
  deprecated summary field, migration compatibility, and address lookup
  hardening.


docs/release-notes/release-notes-next.md
```diff
@@ -5,12 +5,29 @@
 * Instant Out now validates server invoices against a caller-approved maximum
   swap fee.
 
+* Static Address now derives fresh receive and change addresses while retaining
+  per-deposit address ownership across restarts for discovery, recovery, and
+  signing. The new `loop static deposit` command can create and fund an address
+  directly from the lnd wallet, and deposit listings identify the receiving
+  address. [PR #1139](https://github.com/lightninglabs/loop/pull/1139)
```
starius: PR #1139 is closed. Need to update the link.

hieblmi: fixed

litbot-9000:
@starius: review reminder

swap/keychain_test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestStaticAddressKeyFamiliesAreDisjoint documents the key-family split used
+// by static-address HTLC, receive and change key derivation.
+func TestStaticAddressKeyFamiliesAreDisjoint(t *testing.T) {
+	families := map[int32]string{
```
starius, line 12: I would remove this test. It just tests the consts' values. The values of `families` are never used.

swap/keychain.go
```diff
@@ -5,7 +5,16 @@ var (
 	// spending of the htlc.
 	KeyFamily = int32(99)
 
-	// StaticAddressKeyFamily is the key family used to generate static
-	// address keys.
+	// StaticAddressKeyFamily is the legacy static-address key family. It is
+	// used for the V0 single static-address key and for static-address HTLC
+	// keys.
 	StaticAddressKeyFamily = int32(42060)
```
starius, line 11: Can we rename it to `StaticSingleAddressKeyFamily`?

swap/keychain.go
```diff
@@ -5,7 +5,16 @@ var (
...
+
+	// StaticMultiAddressKeyFamily is the key family used to generate
+	// externally visible multi-address static-address receive keys.
+	StaticMultiAddressKeyFamily = int32(42061)
+
+	// StaticAddressChangeKeyFamily is the key family used to generate
+	// static-address change outputs.
+	StaticAddressChangeKeyFamily = int32(42062)
```
starius, line 19: It is used as a change for both single-address and multi-address. Can we add this note explicitly to the godoc?

loopd/swapclient_server.go
```diff
@@ -1979,6 +1979,20 @@ func (s *swapClientServer) ListUnspentDeposits(ctx context.Context,
 			continue
 		}
 
+		params := s.staticAddressManager.GetParameters(u.PkScript)
+		if params == nil {
+			return nil, fmt.Errorf("missing static address "+
+				"parameters for %v", u.OutPoint)
```
starius, line 2122: Should we also include address or pkscript to the error message?

staticaddr/address/sql_store.go
```diff
@@ -64,10 +70,22 @@ func (s *SqlStore) GetAllStaticAddresses(ctx context.Context) (
...
+	staticAddress, err := s.baseDB.Queries.GetLegacyAddress(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.toAddressParameters(staticAddress)
+}
+
 // toAddressParameters transforms a database representation of a static address
 // to an AddressParameters struct.
```
starius, line 86: There is no such struct type `AddressParameters`, but I propose to actually call it that (see another comment).

staticaddr/address/interface.go
```diff
@@ -6,15 +6,26 @@ import (
 	"github.com/lightninglabs/loop/staticaddr/script"
 )
 
+// Parameters aliases the script-level static address parameters for callers
+// that interact with the address manager API.
+type Parameters = script.Parameters
```
starius, line 11: Can we rename it to `AddressParameters`? So it is clear that they are parameters of a single address, not some global parameters.

staticaddr/address/manager.go
```diff
@@ -107,54 +131,195 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
...
+		return err
+	}
+
+	return m.activateAddresses(ctx, params)
+}
+
+// activateAddresses adds persisted addresses to the runtime map. A single
+// wallet read replaces the previous one-write-RPC-per-address startup path.
+func (m *Manager) activateAddresses(ctx context.Context,
+	params []*Parameters) error {
```
starius, line 148: s/params/addrParams/

staticaddr/address/manager.go
```diff
@@ -107,54 +131,195 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
...
+					"the wrong network")
+			}
+
+			pkScript, err := txscript.PayToAddrScript(addr)
+			if err != nil {
+				return nil, fmt.Errorf("derive imported wallet "+
+					"address script: %w", err)
+			}
+
+			scripts[string(pkScript)] = struct{}{}
```
starius, line 239: It is confusing that a `string` is a pkscript, but a Bitcoin address. Normally an address is a string and pkscript is `[]byte`.

staticaddr/address/manager.go
```diff
@@ -107,54 +131,195 @@ func (m *Manager) Run(ctx context.Context, initChan chan struct{}) error {
...
+}
 
-		address, err := m.GetTaprootAddress(
-			clientPubKey, serverPubKey, expiry,
-		)
+// EnsureStaticAddressSeed loads or creates the legacy/root static address
+// parameters. The root address is the only address that requires a Nautilus
+// ServerNewAddress call; all receive/change addresses derive client keys
+// locally and reuse this server pubkey/expiry seed.
+func (m *Manager) EnsureStaticAddressSeed(ctx context.Context) (*Parameters,
```
starius, line 273: It seems that "seed" and "root" mean the same thing. Can we reconcile the wording across the PR?

loopdb/sqlc/migrations/000022_deposit_static_address_id.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE deposits ADD static_address_id INT REFERENCES static_addresses(id);
+
+UPDATE deposits
+SET static_address_id = (
+    SELECT id FROM static_addresses ORDER BY id ASC LIMIT 1
```
starius, line 5: Can this migration ensure that there is exactly one static address assigned for a deposit? Here is some vibe code which should do this:

```sql
CREATE TEMPORARY TABLE migration_22_cardinality_guard (
    deposit_count BIGINT NOT NULL,
    static_address_count BIGINT NOT NULL,
    CONSTRAINT migration_22_requires_one_legacy_address
        CHECK (deposit_count = 0 OR static_address_count = 1)
);

INSERT INTO migration_22_cardinality_guard (
    deposit_count, static_address_count
)
SELECT
    (SELECT COUNT(*) FROM deposits),
    (SELECT COUNT(*) FROM static_addresses);

ALTER TABLE deposits
    ADD static_address_id INT REFERENCES static_addresses(id);

UPDATE deposits
SET static_address_id = (
    SELECT id
    FROM static_addresses
    ORDER BY id ASC
    LIMIT 1
)
WHERE static_address_id IS NULL
  AND EXISTS (SELECT 1 FROM static_addresses);

CREATE TEMPORARY TABLE migration_22_backfill_guard (
    missing_count BIGINT NOT NULL,
    CONSTRAINT migration_22_requires_every_deposit_assigned
        CHECK (missing_count = 0)
);

INSERT INTO migration_22_backfill_guard (missing_count)
SELECT COUNT(*)
FROM deposits
WHERE static_address_id IS NULL;

DROP TABLE migration_22_backfill_guard;
DROP TABLE migration_22_cardinality_guard;
```

This gives the intended behavior:

- No deposits: migration succeeds regardless of address count.
- Deposits plus exactly one address: all deposits are assigned to it.
- Deposits plus zero or multiple addresses: migration fails.
- The second guard explicitly verifies that every deposit was assigned.
- Do not add `UNIQUE(static_address_id)`—multiple deposits must be able to share one address.

The temporary-table technique is needed because the migration is shared by SQLite and PostgreSQL; SQLite `RAISE()` and PostgreSQL `DO ... RAISE` are not portable.

One operational caveat: the schema/data changes roll back on failure, but the migration framework leaves version 22 marked dirty. Recovery then requires repairing the ambiguous data and resetting/forcing the migration version to 21. If automatic retry after repair is important, put this validation in a separate programmatic preflight migration using `ResetVersionOnError: true`, then renumber the currently unshipped migrations.

staticaddr/address/manager.go
```diff
@@ -163,78 +328,124 @@ func (m *Manager) NewAddress(ctx context.Context) (*btcutil.AddressTaproot,
...
+func (m *Manager) NewChangeAddress(ctx context.Context) (*Parameters, error) {
+	seed, err := m.EnsureStaticAddressSeed(ctx)
+	if err != nil {
+		return nil, err
 	}
 
+	return m.newDerivedAddress(ctx, seed, swap.StaticAddressChangeKeyFamily)
+}
+
+func (m *Manager) newDerivedAddress(ctx context.Context, seed *Parameters,
```
starius, line 377: add godoc for this method and below

staticaddr/address/manager.go
```diff
@@ -363,25 +622,47 @@ func (m *Manager) GetStaticAddress(ctx context.Context) (*script.StaticAddress,
...
-	return address, nil
+	return staticAddressFromParams(params)
 }
 
 // ListUnspent returns a list of utxos at the static address.
 func (m *Manager) ListUnspent(ctx context.Context, minConfs,
 	maxConfs int32) ([]*lnwallet.Utxo, error) {
 
-	_, utxos, err := m.ListUnspentRaw(ctx, minConfs, maxConfs)
+	return m.ListUnspentRaw(ctx, minConfs, maxConfs)
```
starius, line 632: Why do we need both `ListUnspentRaw` and `ListUnspent`? They are exactly the same.

staticaddr/address/manager_test.go
```diff
@@ -33,6 +42,79 @@ type mockStaticAddressClient struct {
 	mock.Mock
 }
 
+type blockingImportWallet struct {
```
starius, line 45: add godoc and comments in focal points of tests. Currently all new code in this file lacks comments and godocs.

staticaddr/address/sql_store.go
```diff
@@ -51,14 +50,14 @@ func (s *SqlStore) GetStaticAddressID(ctx context.Context,
...
-	[]*script.Parameters, error) {
+	[]*Parameters, error) {
 
 	staticAddresses, err := s.baseDB.Queries.AllStaticAddresses(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	var result []*script.Parameters
+	var result []*Parameters
```
starius, line 60: pre-allocate the slice

staticaddr/address/manager.go
```diff
@@ -272,6 +483,65 @@ func validateServerAddressParams(
...
+		if params == nil {
+			continue
+		}
+
+		if legacy == nil || params.ID < legacy.ID {
+			legacy = params
+		}
+	}
+
+	return legacy
```
starius, line 542: The legacy address is found by scanning that entire map on every receive/change-address issuance. Repeated issuance is therefore cumulatively `O(N²)`.

We should cache the legacy address, since it does not change.

staticaddr/address/manager.go
```diff
@@ -292,65 +562,54 @@ func (m *Manager) GetTaprootAddress(clientPubkey, serverPubkey *btcec.PublicKey,
...
-		return nil, nil, nil
+	maxConfs int32) ([]*lnwallet.Utxo, error) {
 
-	case len(addresses) > 1:
-		return nil, nil, fmt.Errorf("more than one address found")
+	m.Lock()
+	active := make(map[string]struct{}, len(m.activeStaticAddresses))
+	for pkScript := range m.activeStaticAddresses {
+		active[pkScript] = struct{}{}
+	}
```
starius, line 571: Here we go through all static addresses copying them into the map. This method is called quote often. I propose to avoid O(N) things in hot paths (N - the number of all addresses).

staticaddr/address/interface.go
```diff
@@ -6,15 +6,26 @@ import (
...
+	CreateStaticAddress(ctx context.Context, addrParams *Parameters) error
+
+	// GetStaticAddressID retrieves the static address row ID for the
+	// address script.
+	GetStaticAddressID(ctx context.Context, pkScript []byte) (int32, error)
 
 	// GetAllStaticAddresses retrieves all static addresses from the store.
-	GetAllStaticAddresses(ctx context.Context) ([]*script.Parameters,
-		error)
+	GetAllStaticAddresses(ctx context.Context) ([]*Parameters, error)
```
starius, line 25: I don't think such a method is a good API in multi-address world. The number of static addresses in a wallet may be huge. We should avoid loading them all at once to memory. Instead we should do pagination. And use it only rarely anyway, avoiding doing it regularly. Ideally if we do it once when the daemon starts and then maintain in memory some representation sufficient to operate.

staticaddr/address/manager.go
```diff
@@ -59,9 +70,16 @@ type ManagerConfig struct {
 type Manager struct {
 	sync.Mutex
 
-	cfg *ManagerConfig
+	cfg        *ManagerConfig
+	issuanceMu sync.Mutex
```
starius, line 74: This mutex currently covers much more than in-memory state. `EnsureStaticAddressSeed` holds it while loading addresses from the DB, calling the server, deriving an lnd key, inserting and re-reading DB records, and importing Taproot scripts into the wallet. `newDerivedAddress` similarly holds it across key derivation, DB operations, and wallet import. A slow or stuck dependency therefore blocks every receive- and change-address request, and callers waiting on `sync.Mutex.Lock` cannot honor context cancellation.

I think only root-address creation/recovery requires serialization, so concurrent first callers cannot create different roots. Once the root exists, derived address issuances are independent and should not be globally serialized. Could we replace `issuanceMu` with a retryable, context-aware single-flight operation in `EnsureStaticAddressSeed`, with waiters selecting on their own `ctx.Done()`, and remove it from `newDerivedAddress`? `sync.Once` would not be suitable because transient initialization failures must remain retryable.

Separately, now that the type has multiple synchronization concerns, the embedded `sync.Mutex` should become a named private field:

```go
activeMu sync.RWMutex // guards activeStaticAddresses
```

Embedding promotes `Lock` and `Unlock` as public `Manager` methods and no longer makes it clear which state the mutex protects. Most map access paths (`GetParameters`, `legacyParameters`, and address filtering in `ListUnspentRaw`) are reads, so `RWMutex` also avoids unnecessarily excluding independent readers. The O(N) scan/copy in those latter methods should ideally be reduced separately, but at minimum the lock ownership and scope should be explicit.

looprpc/perms.go
```diff
@@ -82,7 +82,7 @@ var RequiredPermissions = map[string][]bakery.Op{
 	}},
 	"/looprpc.SwapClient/NewStaticAddress": {{
 		Entity: "swap",
-		Action: "read",
+		Action: "execute",
```
starius, line 85: This grants every existing `swap:execute` + `loop:in` macaroon access to the expanded `NewStaticAddress` endpoint. A bearer can submit `SendCoinsRequest{SendAll: true}` to move the eligible lnd wallet balance into a static address; the same permissions authorize `WithdrawDeposits` with an arbitrary `dest_addr`, allowing those funds to be withdrawn externally after confirmation. Please split funding into a separately permissioned RPC, or require a dedicated funding permission that existing execute macaroons do not carry. The release notes should also explicitly describe this authority expansion.

loopd/swapclient_server.go
```diff
@@ -2703,35 +2868,66 @@ func (s *swapClientServer) StaticOpenChannel(ctx context.Context,
...
+		Value:              int64(d.Value),
+		ConfirmationHeight: d.GetConfirmationHeight(),
+		SwapHash:           swapHash,
+	}
+
+	if s.staticAddressManager == nil {
+		return nil, fmt.Errorf("static address manager not configured")
+	}
+
+	staticAddress, err := s.staticAddressManager.GetTaprootAddress(
```
starius, line 2921: This reconstructs the Taproot tree and performs MuSig2 key aggregation for every returned deposit. Since `rpcDeposit` is used by deposit, withdrawal, and swap listings, this adds elliptic-curve work per record. `AddressParams.PkScript` already contains the canonical P2TR output script, so could we derive or cache the encoded address from it instead? The same applies to `ListUnspentDeposits`.

loopd/swapclient_server.go
```diff
@@ -1194,11 +1185,16 @@ func (s *swapClientServer) GetLoopInQuote(ctx context.Context,
 			)
 		}
 
-		err = validateStaticQuoteDepositsSwappable(
-			depositList.FilteredDeposits, staticAddrExpiry,
-			currentHeight,
+		selectedDeposits, err := s.depositManager.DepositsForOutpoints(
```
starius, line 1188: This retrieves the selected deposits twice. `ListStaticAddressDeposits` first loads the complete deposit history, filters and renders the selected records, and performs another height lookup; `DepositsForOutpoints` then reloads the same selection for expiry validation. Could we use one exact-outpoint live-deposit lookup to validate presence, state, and expiry and calculate the count and value in one pass?

swap/keychain_test.go
hieblmi: Removed the key-family constant test.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

swap/keychain.go
hieblmi: Renamed this to `StaticSingleAddressKeyFamily` and updated the Loop call sites. The key-family value remains 42060. Nautilus will need the same symbol rename in its address manager, loop-in actions, and integration-test harness when it pins this update.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

swap/keychain.go
hieblmi: Updated the godoc to explicitly cover change outputs in both single-address and multi-address flows.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

loopd/swapclient_server.go
hieblmi: Added the hex-encoded `pkScript` to the error alongside the deposit outpoint.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/sql_store.go
hieblmi: The address-manager alias is now `AddressParameters`, matching this store conversion godoc. The underlying `script.Parameters` type is unchanged.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/interface.go
hieblmi: Renamed the address-manager API alias to `AddressParameters` and updated its consumers, interfaces, and tests. It still aliases `script.Parameters`, so the underlying representation is unchanged.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/manager.go
hieblmi: Renamed the address parameter variables to `addrParams` throughout the manager.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/manager.go
hieblmi: Renamed the wallet script-key map to `scriptKeys` and documented both script indexes: their keys contain raw PkScript bytes converted to strings for map lookup, not encoded Bitcoin addresses. Retained this representation because byte slices cannot be Go map keys.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/manager.go
hieblmi: Standardized on root terminology, including `EnsureStaticAddressRoot`, local variables, the recovery test, CLI help, and RPC descriptions. Updated the corresponding generated documentation as well.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/manager.go
hieblmi: Added godocs for `newDerivedAddress`, `createAddressFromKey`, and the remaining undocumented helpers, including the lock requirement for `legacyParameters`.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/manager.go
hieblmi: Consolidated the implementation into `ListUnspent`, removed `ListUnspentRaw`, and updated the RPC caller. Wallet filtering and confirmation bounds are unchanged.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/manager_test.go
hieblmi: Added godocs for the new test fixtures, helper methods, tests, and benchmark, plus comments at the concurrency, import-recovery, wallet-reconciliation, and restart assertions.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

staticaddr/address/sql_store.go
hieblmi: Preallocated the result slice using the number of rows returned by `AllStaticAddresses`.

Addressed in 114bf22dcbd6389d3252dd2644eec60f76274e14.

looprpc/perms.go
hieblmi: Addressed in f0ded84abae2550255455af664d6554ef65b5d36 by splitting wallet funding into `FundStaticAddress` (`POST /v1/staticaddr/fund`). `NewStaticAddress` is now address-only, and `loop static deposit` calls the new endpoint.

The new RPC requires `wallet:fund` alongside `swap:execute` and `loop:in`, or an explicit URI grant for `FundStaticAddress`. Existing execute/in macaroons and URI grants for `NewStaticAddress` cannot authorize funding. Release notes and the PR description document the new authority and macaroon upgrade behavior.

Added tests using the production interceptor and signed macaroons, including old URI grants, incomplete permissions, and expired credentials. Handler tests cover new/existing address funding, invalid requests, broadcast failure recovery, and address-only requests carrying the removed funding field on the wire. Full `go test ./...`, `go vet ./...`, lint, targeted race tests, docs-check, WASM compilation, and reproducible RPC generation passed.

loopdb/sqlc/migrations/000022_deposit_static_address_id.up.sql
hieblmi: Addressed in 42370f99df8c6e17d75bb4ba70a46f462f6b91a4 with a single cardinality guard before the backfill.

A temporary table with the named `migration_22_requires_one_legacy_address` CHECK constraint requires exactly one static address whenever deposits exist. Databases without deposits are allowed through regardless of address count. This uses the same SQL on SQLite and PostgreSQL, adds no permanent constraint, and still permits multiple deposits to share an address.

We omitted the second backfill guard: after the cardinality check succeeds, the existing UPDATE assigns the sole address ID to every legacy deposit. We also kept the existing migration failure handling rather than adding a programmatic preflight or renumbering migrations.

Tests pass on both SQLite and PostgreSQL for empty databases, unused addresses, multiple deposits sharing one address, and rejection of missing/ambiguous ownership. They also verify that rejection leaves the old schema and records intact, recovery works after test-data repair and resetting to version 21, and later upgrades still allow multiple addresses. Lint and pinned sqlc generation checks pass.

The operational caveat remains: a rejected migration leaves version 22 dirty. Actual recovery requires backing up and investigating the address ownership, repairing the data, and only then resetting the migration version; users should not delete address records merely to satisfy the check.

staticaddr/address/manager.go
hieblmi: Implemented a narrower change in 205077e4c0a7543756029df15a6c296565d60710: issuance now uses a context-aware gate, and the embedded mutex is replaced by a private `activeMu` guarding the script index and activated root.

Waiters can return when their own context is canceled without canceling the current issuer or releasing its gate. The gate also checks cancellation after acquisition before starting side effects. Root creation/recovery and derived-address issuance remain serialized and retryable; failed root imports still recover from persisted address parameters.

We deliberately retained serialization instead of adding single-flight initialization or parallel derived issuance. This fixes uncancellable waiting while keeping the ordering of server, database, key-derivation, and wallet-import side effects straightforward. A slow issuer can still delay other non-canceled issuers; parallelism can be evaluated separately if that becomes a measured bottleneck.

The map lock remains a plain mutex. Separate commits cache the root and remove full-map copies, so the protected reads are now small lookups or filtering the returned wallet UTXOs, with no dependency RPC under this lock.

Race tests cover canceled root/derived waiters, continued progress of the current issuer, subsequent retries, concurrent first callers sharing one root, failed-import recovery, and address reads remaining responsive during wallet RPCs. The address package and targeted deposit/RPC race tests, vet, and lint pass.

staticaddr/address/manager.go
hieblmi: Addressed in a18fd0d77a6ebf486bdc607a0a9bdf7ec51dda9d. The manager now caches the lowest-ID activated address alongside the script index, so receive/change issuance gets the root in O(1) without scanning the map.

The root is published under the same lock as the index and only after successful wallet activation. Restart and repeated import-failure tests cover recovery without exposing an unimported root.

staticaddr/address/manager.go
hieblmi: Addressed in 8d6cc4005567d71b95f8401dc73ee5cf82f4add9. ListUnspent now calls the wallet outside the map lock, then checks the returned UTXO scripts directly against the active index under the lock. It no longer allocates and copies the entire address map on each call; filtering scales with the returned wallet UTXOs.

Tests cover multiple matching addresses, unknown scripts, confirmation bounds, and address activation while the wallet RPC is blocked.

staticaddr/address/interface.go
hieblmi: Addressed the manager loading path in 0a80c1b9e3ac9c019103d29908be9f16aab6ff49. Startup and root recovery now use ListStaticAddresses with an ascending database-ID cursor and pages of 256 records. Wallet watches are read once per load, and the completed index/root are published only after every page and required import succeeds. Normal issuance updates the in-memory index directly.

This bounds each temporary DB result, not total runtime memory: the manager still maintains its script-to-parameters index and reads the wallet watch set for reconciliation. The existing bulk-read store method remains available for compatibility, but these manager paths no longer use it.

Tests cover sparse IDs, page boundaries, a failed later page leaving the prior index intact, retries and restart recovery; the SQL pagination tests pass on SQLite and PostgreSQL.

loopd/swapclient_server.go
hieblmi: Addressed in 9fa334de39939ad3cdf13a97bd4c1292a7816a3d. Both rpcDeposit and ListUnspentDeposits now encode the address directly from the canonical P2TR PkScript, after checking its script shape and using the configured network. No rendering cache is needed, and listing records no longer rebuilds the Taproot tree or aggregates keys.

Tests compare this encoding with full reconstruction on mainnet, testnet and regtest, and reject malformed/non-P2TR scripts. Address creation and signing retain their cryptographic construction paths.

loopd/swapclient_server.go
hieblmi: Addressed in 8abe2b3eebc28d23334149f25bdee0f949de294d. After refreshing deposits, a manual quote uses one exact-outpoint lookup in the live Deposited set. The same selected records supply value, count and expiry validation; there is no historical listing/rendering pass or second selection lookup. Duplicate, malformed, missing and unavailable outpoints are rejected.

Tests check the single live lookup and retained validation. This commit also fixes the stale ListUnspentDeposits snapshot: refresh precedes the response wallet read, and results must also be in the active Deposited set, so historical rows cannot revive removed deposits.

litbot-9000:
@starius: review reminder

PR #1219, opened 2026-08-31
staticaddr/deposit: preserve confirmation heights

hieblmi:
Static Address deposit reconciliation previously combined queued block epochs with wallet confirmation counts from a different lnd sync point. During rapid lnd catch-up this repeatedly rewrote the persisted first-confirmation height and could make expiry appear eligible one block early.

This change brackets each wallet UTXO query with matching, chain-synced lnd tip observations before deriving absolute confirmation heights. It also:

- compares tip height and block hash to reject reorg races;
- defers confirmed-height, expiry, and spending decisions while no authoritative snapshot is available;
- caps queued expiry notifications at the reconciled tip; and
- continues tracking genuinely unconfirmed deposits without storing confirmed deposits as unconfirmed.

Stack order (bottom to top / merge order):

1. #1219 (`9c145058`) — confirmation-height preservation (this PR), based on `master`.
2. #1218 (`c74ac6cd`) — multi-address core, stacked on this PR.
3. #1215 (`8ab455f9`) — withdrawal hardening, stacked on #1218.

Testing:

- go test ./staticaddr/...
- go test -race ./staticaddr/deposit
- go test ./loopd -run ^$
- go vet ./staticaddr/deposit ./loopd
- git diff --check

### DIFF
--- a/staticaddr/deposit/manager.go
+++ b/staticaddr/deposit/manager.go
@@ -35,8 +35,20 @@ const (
 	PollInterval = 10 * time.Second
 )
 
+var (
+	// ErrConfirmationSnapshotUnavailable is returned to spending paths when
+	// lnd's wallet view cannot be tied to a stable, chain-synced tip.
+	ErrConfirmationSnapshotUnavailable = errors.New(
+		"authoritative deposit confirmation snapshot unavailable",
+	)
+)
+
 // ManagerConfig holds the configuration for the address manager.
 type ManagerConfig struct {
+	// LightningClient is used to obtain a chain-synced wallet height for
+	// confirmation-height reconciliation.
+	LightningClient LightningClient
+
 	// AddressManager is the address manager that is used to fetch static
 	// address parameters.
 	AddressManager AddressManager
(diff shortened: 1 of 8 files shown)


staticaddr/deposit/manager.go
```diff
@@ -271,7 +296,7 @@ func (m *Manager) pollDeposits(ctx context.Context) {
 		for {
 			select {
 			case <-ticker.C:
-				err := m.reconcileDeposits(ctx)
+				_, err := m.reconcileDeposits(ctx)
```
starius, line 299: Non-blocking: if a block-driven reconciliation returns no stable snapshot, that block's expiry notification is skipped. When this polling path later obtains a stable snapshot, it discards the reconciled tip and does not retry expiry processing, so an already-expired deposit waits until the next block. That is safe and conservative, but could delay a sweep. Would it be worth adding a test for the unsynced → synced recovery path, and either documenting the next-block delay or replaying the latest authoritative height once?

staticaddr/deposit/manager.go
```diff
@@ -290,69 +315,137 @@ func (m *Manager) pollDeposits(ctx context.Context) {
 // an unconfirmed funding transaction is replaced, a confirmed deposit is
 // reorged out, or the output was spent outside the active manager path.
 func (m *Manager) EnsureDepositsFresh(ctx context.Context) error {
-	return m.reconcileDeposits(ctx)
+	confirmationTipHeight, err := m.reconcileDeposits(ctx)
+	if err != nil {
+		return err
+	}
+	if confirmationTipHeight == 0 {
```
starius: Nit. Could we also reject a reconciled tip below the latest block epoch already consumed by the manager? For example, if `currentHeight` is 200 but both `GetInfo` calls return a stable, synced tip at 199, this method currently succeeds and spending paths may use a wallet view known to disagree with the higher queued height. That view could still expose an output spent in block 200. Even if the higher epoch is stale because of a reorg, the disagreement means we do not have an authoritative snapshot and should fail closed.

```go
knownHeight := m.currentHeight.Load()
if confirmationTipHeight == 0 ||
	confirmationTipHeight < knownHeight {

	return ErrConfirmationSnapshotUnavailable
}
```

Please also add a test with `currentHeight = 200` and a stable lnd tip at 199 that verifies `EnsureDepositsFresh` returns `ErrConfirmationSnapshotUnavailable`. The expiry path already handles this mismatch conservatively through `expiryNotificationHeight`, but spending freshness currently does not.

starius (approved):
**LGTM!** :tada: 
Everything is good. Added two nits.

PR #1223, opened 2026-09-02
instantout: invoice hardening

starius:
Reject Instant Out RPC requests that omit the caller-approved swap fee cap and require the cap in the internal Go API. Also reject invoice fee values that cannot be safely converted from millisatoshis. This is a follow-up for https://github.com/lightninglabs/loop/pull/1194

Also make a best-effort cancellation request whenever the server initialization request may have created a swap, including when its response may have been lost. Disarm cleanup only after the client persists the swap successfully.

#### Pull Request Checklist

### DIFF
--- a/docs/release-notes/release-notes-next.md
+++ b/docs/release-notes/release-notes-next.md
@@ -13,6 +13,10 @@
 
 #### Bug Fixes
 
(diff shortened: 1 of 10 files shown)


hieblmi (approved):
This LGTM!

PR #1224, opened 2026-09-02
reservation: handle fast funding confirmation

hieblmi:
## Summary

- register an initialization observer before starting the reservation FSM
- wait for the observed WaitForConfirmation transition instead of polling the current state
- cover a funding confirmation that arrives during initialization

## Motivation

A quickly confirmed reservation can enter and leave WaitForConfirmation before the manager polling loop observes that intermediate state. The manager then waits until its timeout even though the reservation has already reached Confirmed. Recording the transition with a pre-registered observer removes that race.

## Testing

- go test ./instantout/...
- go test -race ./instantout/reservation
- make fmt
- make mod-check

#### Pull Request Checklist

- [x] Added an entry to docs/release-notes/release-notes-next.md

### DIFF
--- a/instantout/reservation/manager.go
+++ b/instantout/reservation/manager.go
@@ -14,10 +14,7 @@ import (
 	reservationrpc "github.com/lightninglabs/loop/swapserverrpc"
 )
 
-var (
-	reservationStateWaitTimeout = 5 * time.Second
-	reservationStatePollDelay   = time.Second
-)
+var reservationStateWaitTimeout = 5 * time.Second
 
 // Manager manages the reservation state machines.
 type Manager struct {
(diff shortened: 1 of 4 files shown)


instantout/reservation/manager.go
```diff
@@ -192,12 +215,21 @@ func (m *Manager) newReservation(ctx context.Context, currentHeight uint32,
...
-	err = reservationFSM.DefaultObserver.WaitForState(
-		ctx, reservationStateWaitTimeout, WaitForConfirmation,
-		fsm.WithWaitForStateOption(reservationStatePollDelay),
-	)
+	// Wait until initialization reaches the confirmation monitor. The
+	// observer was registered before the initialization event, so this also
+	// succeeds if the reservation confirms before this select starts.
+	timeout := time.NewTimer(reservationStateWaitTimeout)
+	defer timeout.Stop()
+	err = nil
```
starius: `err := nil`
Currently the closure reuses the outer `err`.

starius (approved):
**LGTM!** 💾

A couple of AI findings.

**The failure path still burns the full 5s.** If `InitAction` fails, the FSM goes `Init → Failed` (`fsm.go:176-178`), `reached` never closes, and `newReservation` waits out the entire `reservationStateWaitTimeout`. Since it's called inline from `Manager.Run`'s event loop (`manager.go:125`), the manager stops processing blocks and new server notifications for 5s per failing reservation. That's pre-existing, but this commit is holding exactly the right tool: have the observer also fire on `isFinalState(notification.NextState)` and let the caller distinguish. Cheap to do now, awkward to retrofit later.

## Nits

- `fsm.WithWaitForStateOption`, `InitialWaitOption`, and `fsmOptions.initialWait` are now dead — this was the only caller repo-wide. Drop them, or say explicitly they're kept as `fsm` public API.
- `reservationFSM.LastActionError` (`manager.go:234-237`) is read without synchronization while the FSM goroutine may be writing it in `HandleError`, and it's formatted with `%v` rather than `%w`. Pre-existing, unchanged by this commit.
- `manager.go:17-19` is now a `var` block wrapping one variable.

PR #1225, opened 2026-09-02
assets: Add shared asset tooling

hieblmi:
## Summary

- make Loop the single home for the shared asset HTLC scripts, immutable
  `SwapKit`, legacy deposit kit, and generic OP_TRUE virtual-packet sweep
  helper
- preserve the existing deposit contract as the explicit
  `LegacyDepositV0` policy, pinned by golden script, key, anchor, witness, and
  virtual-packet vectors
- expose narrow boundaries that both the client and server can import
- use Taproot Assets v0.8.3, taprpc v1.3.3, and LND v0.21.3-beta,
  with Go 1.25.13 and classic btcd; this prototype does not add tap-sdk

## Security and correctness boundaries

The shared kit verifies proofs before using them and binds every Bitcoin
spend to the proof's exact anchor outpoint, output value, script, commitment
root, and unique PSBT input. It supplies every prevout to the signer, applies
the required CSV sequence to the matched input without mutating the caller's
PSBT on failure, rejects malformed signer responses, verifies the returned
Schnorr signature, and returns the matched input index to the caller.

The OP_TRUE helper rejects empty, nil, invalid, non-OP_TRUE, mixed-asset,
overflowing, and amount-mismatched proof sets. It derives the network from the
destination address and validates the prepared output and split-root witness
before attaching the asset witness.

Network validation distinguishes shared testnet HRPs by Bitcoin network
magic. Simnet explicitly accepts both btcd's native BIP-0044 coin type 115 and
lnd's testnet-compatible coin type 1 without relying on mutable global state.

Feature state machines remain responsible for trusted proof import,
canonical-chain and confirmation tracking, reorg handling, destination
validation, quote and fee limits, and durable recovery. This PR adds no Asset
Loop Out RPC, funding flow, persistence, or state machine, and does not alter
conventional Loop In or Asset Loop In.

This is prototype infrastructure, not a release or rollout change.

## Verification

- `GOTOOLCHAIN=go1.25.13 go test ./... -count=1 -timeout=10m`
- `GOTOOLCHAIN=go1.25.13 go test -race ./assets/... -count=1`
- `GOTOOLCHAIN=go1.25.13 go vet ./assets/...`
- `GOTOOLCHAIN=go1.25.13 CGO_ENABLED=0 go build -tags=dev ./cmd/loop ./cmd/loopd`
- `GOTOOLCHAIN=go1.25.13 go mod verify` in the root and `looprpc` modules
- `GOTOOLCHAIN=go1.25.13 go mod tidy -diff` in the root, `looprpc`, and
  `swapserverrpc` modules
- `GOTOOLCHAIN=go1.25.13 go test ./... -count=1` in the `looprpc` and
  `swapserverrpc` modules
- `make commitmsg-lint range=origin/master..HEAD`
- `git diff --check origin/master..HEAD`


assets/tapkit.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		seenOutpoints[outpoint] = struct{}{}
+		total += assetProof.Asset.Amount
+	}
+	if total != addr.Amount {
+		return nil, fmt.Errorf("total proof amount does not match address")
+	}
+
+	sweepVpkt, err := tappsbt.FromProofs(
+		proofs, addr.ChainParams, tappsbt.V1,
```
starius, line 100: Above we accept both V0 and V1, but map both to V1. V0 addresses accepted, but built as a V1/interactive send.

1. Upstream's own mapping (`tappsbt/create.go:39-45`) is V0→`V0`, V1→`V1`, and `tappsbt.CommitmentVersion` (`tappsbt/interface.go:198-210`) turns `V1` into a mandatory `TapCommitmentV2` while `V0` leaves it open. A V0 address therefore gets an output commitment it doesn't describe.
2. The other half is worse: upstream builds V0/V1 address sends as **non-interactive with a split-root tombstone**, with an explicit comment that V0/V1 receivers *need* that to predict the anchor output key. Loop builds a single `Interactive: true` full-value output. So for V0 *and* V1 addresses the interactivity is wrong too.
`address.V0 == 0`, so an unset `Version` silently takes this path — and that's exactly what the fixtures do (`assets/tapkit_test.go:64,113` construct `address.Tap` with no `Version`, paired with `TapCommitmentV2` proofs). The only success-path test that sets a version uses `address.V1`. **The V0 path has no coverage and fails silently.**

Fix: reject non-V1.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,894 @@
...
+}
+
+// GenTaprootAssetRootFromProof verifies the inclusion, exclusion, and optional
+// split-root proofs before returning the committed Taproot Asset root.
+func GenTaprootAssetRootFromProof(assetProof *proof.Proof) ([]byte, error) {
+	if assetProof == nil {
+		return nil, fmt.Errorf("asset proof is required")
+	}
+
+	assetCommitment, err := assetProof.VerifyProofs()
```
starius: `VerifyProofs` is a commitment check, not a validity check.

`VerifyProofs` runs inclusion + split-root + exclusion only. It skips everything `VerifyProofIntegrity` adds — proof version, `Asset.Validate`, `TxSpendsPrevOut`, block header + merkle inclusion, genesis/group reveal — and everything `Verify` adds: prior witnesses and amount conservation. No confirmation check, no unspent check.

However in *this* branch every caller uses the result to **self-derive** a pkScript or tap root (`GetPkScriptFromProof`, `GenTaprootAssetRootFromProof`, `deposit.VerifyProof`, `CreateOpTrueSweepVpkt`). A forged proof yields a script that doesn't exist on chain — the sweep just fails; no value is released. Still worth fixing now: this is about to become a shared boundary, and `VerifyProof` is a name that invites the wrong reading.

Also fix the same problem in `assets/deposit/kit.go:353`, and `assets/tapkit.go:66`.

assets/tapkit.go
```diff
@@ -0,0 +1,180 @@
...
+
+	var (
+		total         uint64
+		seenOutpoints = make(map[wire.OutPoint]struct{}, len(proofs))
+	)
+	for idx, assetProof := range proofs {
+		if assetProof == nil {
+			return nil, fmt.Errorf("asset proof %d is nil", idx)
+		}
+		if _, err := assetProof.VerifyProofs(); err != nil {
```
starius, line 66: **malformed proofs panic inside `VerifyProofs`**

Loop's exposure by call ordering: `assets/tapkit.go:66` calls `VerifyProofs` **before** its script-key nil check at `:73` (exposed to both); exported `GenTaprootAssetRootFromProof` (`swapkit.go:489`) validates nothing (exposed to both); `deposit/kit.go:353` and `swapkit.go:614` guard the script key first but check the internal key only afterwards at `:357` — or never (exposed to the internal-key panic).

assets/deposit/kit.go
```diff
@@ -0,0 +1,639 @@
...
+
+func (d *Kit) validateAsset(depositAsset *asset.Asset) error {
+	if depositAsset == nil {
+		return fmt.Errorf("deposit asset is required")
+	}
+	proofID := depositAsset.Genesis.ID()
+	if !bytes.Equal(d.assetID[:], proofID[:]) {
+		return fmt.Errorf("asset proof ID does not match deposit")
+	}
+	if depositAsset.Amount == 0 {
```
starius, line 314: `validateAsset` only rejects `Amount == 0`. `Kit` stores no expected amount even though `NewAddr` takes one, so `VerifyProof` succeeds for *any* positive amount of the right asset at the right script key. A caller reading it as "the deposit is funded" would accept an underfunded deposit. (`GroupKey` is also unchecked, unlike the swapkit path.)

assets/deposit/kit.go
```diff
@@ -0,0 +1,639 @@
...
+		return nil, fmt.Errorf("deposit kit is required")
+	}
+	if depositProof == nil {
+		return nil, fmt.Errorf("deposit proof is required")
+	}
+	if err := d.validateAsset(&depositProof.Asset); err != nil {
+		return nil, err
+	}
+
+	tapCommitment, err := depositProof.VerifyProofs()
```
starius, line 353: `VerifyProofs()` is a commitment-structure check, not a proof-validity check, and `VerifyProof` exposes its result as a standalone verdict.

`Proof.VerifyProofs()` verifies only the inclusion proof, the split-root proof (when the asset has a split commitment witness), the exclusion proofs, and that the inclusion/exclusion commitment versions are consistent. It deliberately stops there — compare it with `VerifyProofIntegrity`, which additionally checks the proof version, `Asset.Validate()`, `TxSpendsPrevOut`, and runs `HeaderVerifier` + `MerkleVerifier` plus the genesis/group reveals; and with `Proof.Verify`, which on top of that runs `verifyAssetStateTransition` (previous witnesses and amount conservation through the asset VM).

So what we establish here is: *"some Taproot Asset commitment containing this asset leaf is committed to by this anchor output, and the sibling and internal key are the ones this deposit expects."* What we do **not** establish:

- that the anchor transaction is in a block (no header or merkle verification);
- that the asset has valid provenance back to genesis, or a valid group key;
- amount conservation from the previous witnesses;
- that the anchor outpoint is confirmed, or still unspent.

A single `*proof.Proof` cannot give us provenance at all — the chain back to genesis lives in the `proof.File`.

### Why this call site specifically

The other `VerifyProofs()` callers in this branch are fine as-is. `validateSweep` (`kit.go:467`) and `htlc.SwapKit.validateSweep` / `GetPkScriptFromProof` use the returned commitment to *reconstruct* a pkScript that is then compared against the prevout being spent (`kit.go:514-536`). A forged proof there just yields a script that does not exist on chain, and the sweep fails — no trust decision is taken.

`VerifyProof` (`kit.go:389`) is different: it returns the anchor root and nothing else, and its name and doc invite a caller to read it as *"this deposit is real."* It is the one place where a commitment-only result escapes as a verdict someone can act on. Once Nautilus or Asset Loop In consumes this package, that is the line where a counterfeit or unconfirmed asset state would be accepted.

### Suggested fix

Split the two concerns so the trust decision cannot be reached by accident:

1. Keep today's cheap reconstruction, renamed to say what it does (e.g. `AnchorRootFromProofCommitment`), with a doc note that it establishes nothing about the asset's validity or on-chain existence.
2. Add one explicitly-named verification entry point that takes a `*proof.File` plus an expected amount. Options, roughly in order of fit for a client:
   - **Delegate to tapd.** `taprpc.TaprootAssets.VerifyProof(ProofFile) → {Valid, DecodedProof}` already exists in our pinned `taprpc v1.2.0`, and tapd has a fully wired `VerifierCtx` against its own chain backend. `AddressProofClient` (`kit.go:33-38`) already narrows tapd to `NewAddr`/`ExportProof`; adding `VerifyProof` is a small extension in the existing style and needs no dependency bump.
   - **Verify in-process** via `proof.Verify` with a real `VerifierCtx`, backing `HeaderVerifier`/`MerkleVerifier` with `lndclient`'s `ChainKit.GetBlockHeader` / `GetBlock` (available in our pinned lndclient; loop does not use ChainKit today, so this is new plumbing).
   - **For the unconfirmed case** — which is what Asset Loop In actually needs, since HTLC funding has to be validated before the anchor confirms — `proof.VerifyProofSuffix`. It runs proof integrity, inclusion/exclusion, anchor input spending checks and the asset VM against the supplied input files, skipping only the suffix's own header, merkle proof and timelock checks. Note it does **not** exist in v0.8.1; it is new on taproot-assets main, so it would require bumping the pin.

Whichever we pick, these remain our job at the kit level, because no verifier does them for us:

- proof outpoint equals the outpoint we expect, and its pkScript equals the one we derived ourselves from this `Kit` (we already do this in `validateSweep`; `VerifyProof` does not);
- amount equals the expected amount, and asset ID / group key match;
- confirmation depth, from our own chain source;
- the anchor outpoint is still unspent — no proof can ever tell us this.

Given the package has no callers yet, this does not have to block landing the foundation, but it should be settled before anything treats these helpers as a validation boundary.

assets/deposit/kit.go
```diff
@@ -0,0 +1,639 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("invalid co-signer public key: %w", err)
+	}
+
+	sortKeys := true
+	muSig2Key, err := input.MuSig2CombineKeys(
+		input.MuSig2Version100RC2,
+		[]*btcec.PublicKey{funderKeyCopy, coSignerKeyCopy}, sortKeys,
+		&input.MuSig2Tweaks{TaprootBIP0086Tweak: true},
```
starius, line 98: `TaprootBIP0086Tweak: true` here is inert, and it is a footgun to leave in place.

Only `PreTweakedKey` is ever read from `muSig2Key` (`kit.go:198`, `:359`, `:419`, `:429`, `:529`), and btcd snapshots that value *before* applying any tweak (`schnorr/musig2/keys.go:395-399`; the taproot branch at `:403` normalizes a local copy and mutates only `finalKeyJ` → `FinalKey`). So this flag cannot change any key this kit produces, parity included.

I think this is residue from adapting `staticaddr/script/script.go:62-99`, which is the same shape — two keys plus a CSV timeout leaf — but computes `rootHash` from the script tree up front and keeps both `PreTweakedKey` and `FinalKey`. That does not work for an asset anchor: the root is `tapCommitment.TapscriptRoot(siblingHash)` and is unknown until the deposit lands and we hold a proof, which is exactly why `VerifyProof` returns it. `TaprootBIP0086Tweak` is the only tweak variant that needs no root hash, so it satisfies the struct without data we do not have yet.

Worth removing rather than just tidying, because `MuSig2Tweaks` is also the signing-time config (`ToContextOptions` / `MuSig2CreateContext`) and `TaprootBIP0086Tweak` **takes precedence over** `TaprootTweak` — lnd ignores the script root when the BIP86 flag is set (`input/musig2.go:220-229`, `:358-363`). If someone later stores this struct on the `Kit` and adds the real anchor root for the MuSig2 key-path spend without spotting the flag, the root is silently discarded and they get BIP86 key-spend-only signatures for an output that has both a CSV leaf and an asset commitment.

Suggest matching `htlc.GetAggregateKey` (`swapkit.go:289`), which passes an empty tweak set and documents the intent:

```go
	&input.MuSig2Tweaks{},
```

plus a comment noting that the BIP341 tweak is applied later at signing time, using the anchor root from `VerifyProof`.

assets/deposit/kit.go
```diff
@@ -0,0 +1,639 @@
...
+
+	if signer == nil {
+		return nil, fmt.Errorf("signer is required")
+	}
+	sweep, err := d.validateSweep(depositProof, sweepPacket)
+	if err != nil {
+		return nil, err
+	}
+	sweepTx := sweepPacket.UnsignedTx.Copy()
+	sweepTx.TxIn[sweep.assetInputIndex].Sequence = d.csvExpiry
```
starius, line 591: Divergent contracts: `deposit.CreateTimeoutWitness` sets the CSV sequence itself and writes it back into the caller's PSBT (`kit.go:591,631`); `htlc.CreateTimeoutWitness` requires the caller to have set it and rejects any mismatch. Easy to get wrong.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,894 @@
...
+
+// AddressParams returns a copy of the contract's address parameters.
+func (s *SwapKit) AddressParams() address.ChainParams {
+	return *cloneAddressParams(s.addressParams)
+}
+
+func cloneAddressParams(params address.ChainParams) *address.ChainParams {
+	paramsCopy := params
+	if params.Params != nil {
+		bitcoinParams := *params.Params
```
starius, line 207: This derefs `params.Params` unguarded; the `htlc` twin guards it. Safe via `NewKit` today, a landmine if reused.

starius (commented):
**Style — ~18 unexported funcs and ~14 test helpers have no godoc**, against the convention. Non-test: `canonicalAddressParams`, `cloneAddressParams` (both copies), `validateScripts`, `genBtcControlBlock`, `validateAsset` (both), `findUniqueInput` (both), `validateSweep` (both), `validateSequence`, `verifyTapscriptSignature` (both), `timeoutPathSibling`, `encodedTimeoutPathSibling`, `newHtlcSwapKit`, `validateProof`.

assets/deposit/kit.go
hieblmi: Addressed without changing the existing method signature: AnchorRootFromProofCommitment now requires and checks the expected amount, while VerifyProof remains as a deprecated compatibility wrapper with its limited guarantees documented. VerifyProofFile also binds the fully verified terminal proof to the expected amount and outpoint.

assets/deposit/kit.go
hieblmi: Addressed in 818876e. VerifyProofFile delegates the complete proof file to tapd verification, then binds the terminal proof to the expected deposit outpoint, amount, asset ID, script key, internal key, and timeout sibling. The commitment-only helpers now state their limits explicitly, and the ambiguous VerifyProof entry point is retained only as a deprecated compatibility wrapper. Confirmation depth and unspent status remain explicit caller checks.

assets/tapkit.go
hieblmi: The v0.8.3 dependency update now rejects both malformed asset/script keys and missing Taproot proof internal keys inside VerifyProofs. I added regression coverage at the exported helper boundary so both cases return errors rather than panic.

assets/deposit/kit.go
hieblmi: Fixed in 818876e. The inert BIP86 flag is removed and the aggregation now uses an empty tweak set, with a comment explaining that the real BIP341 tweak is applied once the asset and timeout roots are known. The legacy aggregate-key vector remains unchanged.

assets/tapkit.go
hieblmi: Fixed in 818876e. CreateOpTrueSweepVpkt now uses tappsbt.FromAddresses for the destination output template, preserving V0→V0 and V1→V1 while using the required non-interactive split-root layout. The success test now covers both address versions and validates every input witness.

assets/htlc/swapkit.go
hieblmi: The thread is attached to the HTLC helper, which already has the guard. I added the same nil guard to the deposit clone helper so both copies are defensive even if reused outside their currently validated constructors.

assets/deposit/kit.go
hieblmi: To avoid changing existing behavior, I kept both contracts and made them explicit: deposit signs a copy and writes the CSV sequence back only after signature verification; HTLC requires the caller to set SuccessSequence/CsvExpiry and never mutates the PSBT. The existing mismatch tests remain in place, and Nautilus #1712 already uses AssetInputIndex before signing.

hieblmi:
Follow-up commit 818876e also adds documentation to the unexported production and test helpers listed in the general review. Local verification passed: go test ./..., go test -race ./assets/..., go vet ./..., golangci-lint (0 issues), plus a Nautilus #1712 itest-package compile with integration tags.

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+		return nil, fmt.Errorf(
+			"deposit asset amount mismatch: expected %d, got %d",
+			expectedAmount, depositProof.Asset.Amount,
+		)
+	}
+
+	return d.anchorRootFromProofCommitment(depositProof)
+}
+
+// VerifyProofFile asks tapd to fully verify a proof file, binds its terminal
```
starius, line 461: Could we make this method enforce full provenance rather than leaving that distinction to callers?

`VerifyProofResponse.Valid` does not necessarily mean that a proof history is rooted at issuance. In Taproot Assets v0.8.3:

- tapd's `VerifyProof` calls `File.Verify` and does not bind `GenesisPoint` (`rpcserver.go:2159`).
- `File.Verify` starts the first proof with `prev == nil` (`verifier.go:1565`).
- In that state, `Proof.Verify` accepts a non-nil `ChallengeWitness` as an ownership proof (`verifier.go:1227`). This is intentional behavior (`proof_test.go:975`).

Because the deposited asset uses an `OP_TRUE` script key, satisfying that ownership challenge is trivial. An attacker could anchor a leaf with the target genesis identity and claimed amount under the expected public Bitcoin contract. The current chain-inclusion, commitment, outpoint, amount, internal-key and sibling checks could then pass without proving legitimate issuance or transition history.

I think `VerifyProofFile` should provide the safe abstraction promised by its name:

1. Snapshot `RawProofFile` once and verify and decode that same immutable content.
2. Reject empty files and non-nil `ChallengeWitness` values.
3. Require the first asset to satisfy `IsGenesisAsset()`.
4. Apply those requirements recursively to every proof file in `AdditionalInputs`.
5. Call tapd to verify the genesis reveal, group keys, chain inclusion, commitments and state transitions, then perform the existing terminal contract bindings.

`IsGenesisAsset()` is only a structural precondition; tapd's cryptographic verification is still required. Conversely, tapd verification alone permits the ownership-proof shortcut. Please add regressions for a standalone ownership proof and an ownership-rooted `AdditionalInputs` file.

Confirmation depth and current unspent status are separate live-chain properties. If callers should have one safe go/no-go operation, a higher-level `VerifyDeposit` can encapsulate those checks while keeping `VerifyProofFile` focused on provenance and exact contract binding.

I would also consider removing or unexporting the commitment-only `VerifyProof` before merge (`kit.go:515`). Keeping an attractively named unsafe alternative makes it easy for Loop server to bypass the safe path accidentally.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,917 @@
...
+type Policy uint8
+
+const (
+	// PolicyUnknown reserves the zero value so every caller must choose the
+	// protocol whose compatibility rules it expects.
+	PolicyUnknown Policy = iota
+
+	// LegacyDepositV0 preserves the asset deposit HTLC contract that predates
+	// this shared package.
+	LegacyDepositV0
```
starius, line 49: What policy should asset loop-in use?

Do we keep LegacyDepositV0 for drafts written before?

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+	}
+	verifyResponse, err := verifier.VerifyProof(ctx, proofFileCopy)
+	if err != nil {
+		return nil, fmt.Errorf("unable to verify deposit proof file: %w", err)
+	}
+	if verifyResponse == nil || !verifyResponse.Valid {
+		return nil, fmt.Errorf("invalid deposit proof file")
+	}
+
+	decodedFile, err := proof.DecodeFile(proofFile.RawProofFile)
```
starius, line 497: The method creates a private copy and sends that to tapd, but afterward decodes the caller-owned original at line 497. Mutation while the RPC is in flight can therefore make Loop consume bytes tapd did not verify.

Make one immutable snapshot and both verify and decode that exact snapshot. Add a verifier mock that mutates the caller’s original buffer to prevent regression.

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+		...grpc.CallOption) (*taprpc.VerifyProofResponse, error)
+}
+
+var _ ProofVerifier = (taprpc.TaprootAssetsClient)(nil)
+
+// Kit contains the immutable information needed to create and operate a
+// two-party MuSig2 asset deposit.
+type Kit struct {
+	funderKey   *btcec.PublicKey
+	coSignerKey *btcec.PublicKey
```
starius, line 53: **The deposit kit collapses four independent key roles into two.**

`Kit` stores only `funderKey` and `coSignerKey`. The same pair is then:

- aggregated into the deposit’s MuSig2 internal key at `assets/deposit/kit.go:103-110`;
- used for the deposit refund script at `assets/deposit/kit.go:138-150`;
- passed to the HTLC as its sender and receiver keys at `assets/deposit/kit.go:273-285`;
- aggregated again as the HTLC internal key at `assets/htlc/swapkit.go:285-301`;
- used directly in the HTLC success and timeout leaves at `assets/htlc/swapkit.go:261-283`.

This cannot represent the intended four direction-neutral roles:

- `SenderInternalPubKey` and `ReceiverInternalPubKey` form the cooperative MuSig2 internal key.
- `SenderScriptKey` authorizes the unilateral deposit refund and HTLC timeout paths.
- `ReceiverScriptKey` authorizes the HTLC preimage-success path.

“Sender” and “receiver” are defined by the on-chain asset direction: for Asset Loop In, Loop is the sender and Loop server is the receiver; for Asset Loop Out, the mapping is reversed. This is also the model already used by Bitcoin HTLCs in `loopdb/loop.go:13-35` and `swap/htlc.go:535-615`.

There is no correct way for callers to map four keys into the current two-key constructor. Passing the internal keys makes the unilateral scripts use the wrong keys; passing the script keys makes the cooperative aggregate use the wrong keys. This changes the deposit address, HTLC output, control blocks, signing responsibilities, and recovery paths.

Please preserve all four roles as separate immutable constructor inputs. Aggregate only the sender and receiver internal keys, use the sender script key in refund/timeout leaves, and use the receiver script key in the preimage-success leaf. The shared `SwapKit` policy should derive and validate that same mapping for both Loop In and Loop Out.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,917 @@
...
+		return 0, fmt.Errorf("asset input does not spend proof outpoint")
+	}
+
+	return inputIndex, nil
+}
+
+// validateSweep binds a PSBT input and witness UTXO to the HTLC proof and
+// reconstructs the exact Taproot output being spent.
+func (s *SwapKit) validateSweep(assetProof *proof.Proof,
+	sweepPacket *psbt.Packet) (*validatedSweep, error) {
```
starius, line 621: **Witness methods will sign asset-burning outputs.**

Both `SwapKit.validateSweep` and `deposit.Kit.validateSweep` verify only that the selected input is the expected asset anchor. `assets/htlc/swapkit.go:618-708` and `assets/deposit/kit.go:583-665` return without inspecting any transaction output or binding the PSBT to a validated Taproot Asset transition.

The PSBT is then signed by `CreatePreimageWitness` at `assets/htlc/swapkit.go:812-815`, `CreateTimeoutWitness` at `assets/htlc/swapkit.go:883-886`, and deposit `CreateTimeoutWitness` at `assets/deposit/kit.go:730-733`. `SIGHASH_DEFAULT` commits to all outputs, but that does not make them asset-safe—the helper signs whatever outputs the caller supplied.

Bitcoin consensus validates the anchor spend, not Taproot Asset conservation. A transaction can therefore spend the expected P2TR anchor while omitting or replacing the destination asset commitment. Once confirmed, this destroys the active asset’s valid transition history and can also burn passive assets co-anchored in the same UTXO.

The tests currently demonstrate this gap: the HTLC fixture’s only output is a bare `OP_TRUE` output at `assets/htlc/swapkit_test.go:744-747`, and the deposit fixture does the same at `assets/deposit/kit_test.go:353-355`; both are nevertheless signed successfully.

Because this kit is intended to hide Taproot Asset internals from callers, output correctness should not be an implicit caller obligation. Before releasing a signature or witness, the shared API must bind the exact PSBT to the exact validated virtual transition and verify:

- the active asset ID and amount;
- the intended receiver asset script key;
- the anchor output index, value, internal key, sibling, commitment root, and resulting pkScript;
- all required change and passive-asset outputs;
- that the Bitcoin outputs being signed are exactly those produced by the validated Taproot Asset transition.

Tests should construct a real asset-preserving transition, then verify that mutations to the destination output, amount, keys, sibling, commitment, or passive-asset carry-forward are rejected before the signer is invoked.

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+
+	htlcAddr, err := client.NewAddr(ctx, &taprpc.NewAddrRequest{
+		AssetId:   d.assetID[:],
+		Amt:       amount,
+		ScriptKey: rpcutils.MarshalScriptKey(tapScriptKey),
+		InternalKey: &taprpc.KeyDescriptor{
+			RawKeyBytes: btcInternalKey.SerializeCompressed(),
+		},
+		TapscriptSibling: siblingBytes,
+	})
```
starius, line 265: **Medium — `NewHtlcAddr` and `CreateHtlcVpkt` describe different asset commitments.**

This request leaves both version fields at their protobuf zero values. With Taproot Assets v0.8.3, that selects Asset V0 and leaves the address version unspecified; tapd currently resolves the latter to Address V1. In contrast, `CreateHtlcVpkt` explicitly creates VPacket V1 and Asset V1 outputs at `assets/htlc/swapkit.go:373-397`.

The asset version is serialized into the asset leaf, so changing it changes the Taproot Asset commitment root and ultimately the Bitcoin output key. The address returned here and the VPacket returned by the same `SwapKit` therefore do not construct the same HTLC output. Relying on tapd's unspecified address-version default also leaves part of the contract outside the immutable policy.

Please freeze every relevant version in the policy—at minimum the address, asset, VPacket, and TapCommitment versions—and use those values consistently in both address requests and VPacket construction. Proof-file and transition-proof versions accepted by validation should be frozen there as well. Validate the versions returned by tapd, and add a test showing that the decoded address and the funded VPacket derive the same output key.

(also applies to assets/deposit/kit.go:212)

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+
+	return client.NewAddr(ctx, &taprpc.NewAddrRequest{
+		AssetId:   d.assetID[:],
+		Amt:       amount,
+		ScriptKey: rpcutils.MarshalScriptKey(tapScriptKey),
+		InternalKey: &taprpc.KeyDescriptor{
+			RawKeyBytes: d.muSig2Key.PreTweakedKey.SerializeCompressed(),
+		},
+		TapscriptSibling: siblingBytes,
+	})
```
starius, line 220: `NewAddr` returns tapd's response directly, and `NewHtlcAddr` does the same at `assets/deposit/kit.go:257-270`. There is no shared validation path, so a Loop or Loop server caller cannot independently reconstruct and validate an address without duplicating contract logic outside the shared kit.

Please expose a shared validator and run every locally generated address through it before returning. The same validator should accept an address received from a peer. It should treat the encoded address as canonical, decode it for the expected network, and verify the frozen address and asset versions, exact asset ID and amount, contract-specific asset script key, Bitcoin internal key, tapscript sibling, and resulting Taproot output key. For an RPC response, it should also reject any disagreement between the encoded address and the expanded response fields.

Recomputing the output key requires trusted asset genesis/type data. If the current `assetID`-only kit does not contain enough information, extend its immutable inputs rather than accepting tapd's returned `TaprootOutputKey` as the expected value. The purpose of this API should be to give callers one simple, authoritative contract check without requiring them to understand these internals.

(also applies to assets/deposit/kit.go:257)

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+
+	controlBlock, err := d.GenTimeoutBtcControlBlock(sweep.assetRoot)
+	if err != nil {
+		return nil, err
+	}
+	controlBlockBytes, err := controlBlock.ToBytes()
+	if err != nil {
+		return nil, err
+	}
+	sweepPacket.UnsignedTx.TxIn[sweep.assetInputIndex].Sequence = d.csvExpiry
```
starius, line 755: `CreateTimeoutWitness` signs a transaction copy after setting the asset input's sequence at `assets/deposit/kit.go:714-733`, then mutates the caller's unsigned transaction here. Any existing `SIGHASH_DEFAULT` or `SIGHASH_ALL` signature on a fee input commits to all input sequences, so a signature made before this call becomes invalid even though its own input was not otherwise changed. Updating the sequence also changes the transaction ID.

Please match the HTLC witness API at `assets/htlc/swapkit.go:725-738`: require the proof-selected input to already contain the exact policy sequence, sign that exact transaction, and never mutate the caller's PSBT. Add a regression test with an already signed fee input, and assert that both successful and rejected calls leave the packet unchanged.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,917 @@
...
+
+	return &paramsCopy
+}
+
+// Validate verifies every field required to create an HTLC virtual packet.
+func (s *SwapKit) Validate() error {
+	if err := s.validateScripts(); err != nil {
+		return err
+	}
+	if s.amount == 0 {
```
starius, line 223: **Low — `SwapKit` accepts amounts that its VPacket cannot encode.**

`Validate` rejects zero but accepts values above `math.MaxInt64`. `CreateHtlcVpkt` then places that value in a `VOutput` at `assets/htlc/swapkit.go:390-391`, but Taproot Assets rejects such outputs during encoding because the virtual transaction output amount must fit in a signed `wire.TxOut.Value`.

Please reject `amount > math.MaxInt64` during `NewSwapKit` validation so a successfully constructed kit cannot later fail solely because its immutable amount is unencodable. Boundary tests for `math.MaxInt64` and `math.MaxInt64 + 1` would make the contract explicit.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,917 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid sender public key: %w", err)
+	}
+	receiverKey, err := btcec.ParsePubKey(
+		params.ReceiverPubKey.SerializeCompressed(),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("invalid receiver public key: %w", err)
+	}
+	if senderKey.IsEqual(receiverKey) {
```
starius, line 108: **Low — distinct-key validation compares full points even though the contracts use x-only keys.**

`btcec.PublicKey.IsEqual` distinguishes `P` from `-P`, but BIP340/Taproot serializes both points to the same x-only public key. A caller can therefore provide opposite-parity encodings of the same key and bypass the "keys must differ" check, even though one underlying signing key controls both nominal roles in the resulting contract.

Please compare `schnorr.SerializePubKey` encodings after parsing and apply the same x-only distinctness rule to every policy role that must be independently controlled. Add a regression test using `P` and `-P`.

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+		return nil, fmt.Errorf("expected deposit outpoint is required")
+	}
+	if expectedAmount == 0 {
+		return nil, fmt.Errorf("expected deposit amount must be positive")
+	}
+
+	proofFileCopy := &taprpc.ProofFile{
+		RawProofFile: append([]byte(nil), proofFile.RawProofFile...),
+		GenesisPoint: proofFile.GenesisPoint,
+	}
```
starius, line 488: **Low — `VerifyProofFile` duplicates oversized proofs before enforcing Taproot Assets' size limit.**

The first operation on the raw proof is an unbounded `append` copy. Tapd's `VerifyProof` RPC applies `proof.CheckMaxFileSize`, but that happens only after this allocation and after the buffer crosses the pluggable verifier boundary. An oversized untrusted proof can therefore force a second allocation of the full payload before it is rejected.

Please capture the input slice, call `proof.CheckMaxFileSize` before allocating the immutable snapshot, and then use that one snapshot for both verification and decoding. Oversized input should be rejected before invoking the verifier.

assets/deposit/kit.go
```diff
@@ -0,0 +1,763 @@
...
+	if err != nil {
+		return nil, err
+	}
+	tapScriptKey, _, _, _, err := htlc.CreateOpTrueLeaf()
+	if err != nil {
+		return nil, err
+	}
+
+	return client.NewAddr(ctx, &taprpc.NewAddrRequest{
+		AssetId:   d.assetID[:],
```
starius, line 213: **Low — the RPC request exposes mutable storage from the supposedly immutable deposit kit.**

`d.assetID[:]` is a slice backed directly by the array stored in `Kit`. Because `AddressProofClient` is a pluggable interface, an implementation or interceptor can retain and mutate `request.AssetId`, changing `d.assetID` after construction. On the HTLC path this can also make the address response disagree with the `SwapKit` that was created before the RPC call.

Please copy `AssetId` into request-owned storage before crossing the interface boundary in both address methods. A mock that mutates and retains the request slice should demonstrate that subsequent kit operations remain unchanged.

assets/htlc/swapkit.go
```diff
@@ -0,0 +1,917 @@
...
+	}
+
+	branch := txscript.NewTapBranch(timeoutLeaf, successLeaf)
+
+	return commitment.NewPreimageFromBranch(branch), nil
+}
+
+// CreateHtlcVpkt creates the version-one virtual packet for the HTLC. The
+// split-root and HTLC output indices and their interactive flags are consensus
+// with the existing server implementation.
```
starius, line 352: **Non-blocking documentation request — clarify that `CreateHtlcVpkt` returns a funding template.**

The zero-valued interactive split-root placeholder at `assets/htlc/swapkit.go:381-388` is correct for tapd's funding pipeline, but it is not valid as a final VPacket: `PrepareOutputAssets` rejects interactive zero-valued outputs. `FundVirtualPsbt` first converts the template to allocations, replaces the placeholder amount with actual change, and omits the local output entirely for a full-value interactive transfer.

Could the method documentation state that its result must pass through `FundVirtualPsbt` and must not be validated or committed as a final VPacket? Renaming it to something like `CreateHtlcFundingTemplate` would make the contract even clearer, but documentation is sufficient here. A funded-template test covering both exact-value and change cases would preserve this subtle dependency.

assets/htlc/script.go
```diff
@@ -0,0 +1,125 @@
...
+	return builder.Script()
+}
+
+// GetOpTrueScript returns a script that always evaluates to true.
+func GetOpTrueScript() ([]byte, error) {
+	return txscript.NewScriptBuilder().AddOp(txscript.OP_TRUE).Script()
+}
+
+// CreateOpTrueLeaf creates the legacy Taproot Asset script key whose only
+// script path is OP_TRUE beneath the public Taproot Assets NUMS key.
```
starius, line 71: **Low — the policy permanently fixes a shared OP_TRUE asset script key.**

`CreateOpTrueLeaf` has no contract-specific input, so all deposits and HTLCs use the same asset script key. This matches the historical prototype and is valid for individual swaps, but it prevents multiple outputs for the same asset from being constructed in one transaction and precludes script-key-addressed proof delivery without another disambiguation mechanism.

Please document whether those limitations are intentional for this immutable policy. If same-transaction batching or script-key-addressed proof delivery is required later, a new policy will need a deterministic contract-specific discriminator while retaining OP_TRUE as its spendable script path.

litbot-9000:
@bhandras: review reminder
@sputn1ck: review reminder

litbot-9000:

PR #1228, opened 2026-09-09
staticaddr/loopin: prevent sweepless signing timeouts

starius:
For a sweepless Static Address loop-in, the server can ask the client to presign many fee variants of the sweep transaction. The client previously handled those notifications one at a time, and Aperture's L402 interceptor also held its token-payment mutex across each authenticated unary RPC. On a high-latency connection this made every signature response wait for another full round trip, so later variants could miss the server's signing deadline even though the client was otherwise able to sign them.

This change processes sweep requests through a bounded 128-worker pool and lets unary RPCs that already have a paid L402 token run concurrently. Token acquisition, payment resumption, and payment-required challenges still use Aperture's serialized path, preventing concurrent calls from initiating duplicate payments, while streaming RPC authentication remains unchanged. Shutdown cancels and drains the workers before manager-owned resources are released, and the fixed pool prevents a burst of server notifications from creating unbounded signing work.

  - [ ] TODO: change the Aperture client itself not to serialize the requests
  - [ ] TODO: when https://github.com/lightninglabs/aperture/pull/281 is merged, we could remove most of commit "loop: parallelize authenticated L402 calls"

Since this path signs server-provided transactions, the change also tightens the validation boundary before creating any MuSig2 session. The client verifies that the PSBT contains an unsigned transaction, that the server supplied an exact and unique prevout for every input, and that every deposit it is asked to sign is locally owned with the expected value and script. Malformed or inconsistent requests are rejected before signing instead of producing a signature over an incorrect digest or reaching a nil prevout.

#### Pull Request Checklist
- [x] Add an entry to `docs/release-notes/release-notes-next.md`, or apply the

### DIFF
--- a/docs/release-notes/release-notes-next.md
+++ b/docs/release-notes/release-notes-next.md
@@ -13,9 +13,10 @@
 
 #### Bug Fixes
 
-* Static Address loop-in sweep signing requests are now processed concurrently,
-  preventing slow connections from delaying later signing responses until the
-  server times them out.
+* Static Address loop-in sweep signing requests and their authenticated
+  responses are now processed concurrently, preventing slow connections from
+  delaying later responses until the server times them out. Sweep work is
+  bounded, and server-provided prevouts are validated before signing.
 
 * Instant Out now attempts to cancel server-side swaps when client
   initialization fails, allowing locked reservations to be released without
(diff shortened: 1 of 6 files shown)


litbot-9000:
@hieblmi: review reminder

PR #1233, opened 2026-09-18
sweepbatcher: filter batches by signing mode

starius:
Standard clients can leave regular Loop In HTLC sweeps and presigned static-address sweeps pending at the same time. Fee-only selection offers sweeps to incompatible batches, producing warnings before retrying a compatible batch.

Filter incompatible signing modes in both greedy selection and its fallback. Keep the admission warnings as safeguards.

Add regression coverage for both mismatch directions through both selection paths. All four cases emit the original warnings without the fix. Manual regtest swaps reproduce both warnings on the old server and complete without either warning with this change.

#### Pull Request Checklist


sweepbatcher/sweep_batch.go
```diff
@@ -789,6 +789,23 @@ func (b *batch) addSweeps(ctx context.Context, sweeps []*sweep) (bool, error) {
...
+func (b *batch) matchesSweepMode(presigned bool) bool {
+	done, err := b.scheduleNextCall()
+	defer done()
+	if err != nil {
+		return false
+	}
+
+	// Admission keeps every sweep in a batch in the same signing mode.
+	for _, sweep := range b.sweeps {
+		return sweep.presigned == presigned
```
hieblmi, line 803: ah ok, this expects every sweep in the sweeps to have the same mode.

hieblmi (approved):
Thank you for the fix, this looks good to me!

bhandras (approved):

